PR #2, opened 2016-08-22
Add k8s.io Nginx configuration.

rmmh:
README.md sketches one way to do testing. If we don't want to require a separate cluster for testing, the Makefile should be modified to use namespaces.

### DIFF
--- a/k8s.io/TODO
+++ b/k8s.io/TODO
@@ -1,0 +1,5 @@
+canary process
+README
+Makefile to sed-in the www/... and certs, and to capture updates and tests
+where to check it all in?
+ - SSL certs
--- a/k8s.io/TODO
+++ b/k8s.io/TODO
@@ -1,5 +1,0 @@
-canary process
-README
-Makefile to sed-in the www/... and certs, and to capture updates and tests
-where to check it all in?
- - SSL certs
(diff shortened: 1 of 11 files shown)


k8s.io/TODO
```diff
@@ -0,0 +1,5 @@
+canary process
```
thockin: don't check this one in :)

k8s.io/test.py
```diff
@@ -0,0 +1,225 @@
...
+                'issues?q=is%3Aopen+is%3Aissue+label%3ABounty')
+            self.assert_redirect(base + 'help-wanted',
+                'https://github.com/kubernetes/kubernetes/labels/help-wanted')
+            self.assert_redirect(base + 'start',
+                'http://kubernetes.io/docs/getting-started-guides/')
+
+    # TODO: external go-get ???
+
+    def test_ci_test(self):
+        # FIXME: https://ci-test.kubernetes.io is not on the
```
thockin: not on the ...   the suspense is killing me!

thockin:
Mostly just nits.

Can you add a redacted form of the SSL secret?  Like https://github.com/kubernetes/k8s.io/blob/master/reviewable.k8s.io/secret-ssl.yaml (but without namespace specified)

thockin:
LGTM

PR #3, opened 2016-08-23
Remove some over-precise leftovers

thockin:
Don't need namespace in every file - assume it is in the kubectl context.


rmmh:
LGTM

PR #14, opened 2016-09-27
add vanity domains for os packages

mikedanese:
This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/k8s.io/14)


thockin:
I see you already canaried and pushed this.  For future, please drop me a
note when you do, so I don't almost repeat what you did.

mikedanese:
Ok, will do.

PR #15, opened 2016-09-29
Redirect dl.k8s.io to the kubernetes-release GCS bucket

ixdy:
Idea is that scripts can use https://releases.k8s.io/downloads/release/v1.4.0/kubernetes.tar.gz rather than needing to know the GCS bucket name (kubernetes-release) or use the github URL (which doesn't have anything besides kubernetes.tar.gz).

More concrete example: I want to update https://github.com/kubernetes/kubernetes/blob/master/cluster/centos/config-build.sh#L44 to use `https://releases.k8s.io/downloads/release/v${K8S_VERSION}/kubernetes-server-linux-amd64.tar.gz`.

Open to bikeshedding if you have a better suggestion.

---
This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/k8s.io/15)

### DIFF
--- a/k8s.io/configmap-nginx.yaml
+++ b/k8s.io/configmap-nginx.yaml
@@ -90,6 +90,19 @@ data:
         rewrite ^/(.*)?$    https://github.com/kubernetes/kubernetes/tree/master/$1 redirect;
       }
       server {
+        server_name dl.k8s.io dl.kubernetes.io;
+        listen 80;
+        listen 443 ssl;
+
+        location / {
+          # Don't require /release/ if you want to get at the Kubernetes release artifacts, the common case.
+          rewrite ^/(v[0-9]+\.[0-9]+\.[0-9]+(-(alpha|beta)\.[0-9]+)?/.*)$   https://storage.googleapis.com/kubernetes-release/release/$1 redirect;
+          # Provide a convenient redirect for CI (continuous integration) artifacts as well, which live in a different bucket.
+          rewrite ^/ci/?(.*)$              https://storage.googleapis.com/kubernetes-release-dev/ci/$1 redirect;
+          rewrite ^/(.*)$                  https://storage.googleapis.com/kubernetes-release/$1 redirect;
+        }
+      }
+      server {
         server_name docs.k8s.io docs.kubernetes.io;
         listen 80;
         listen 443 ssl;
--- a/k8s.io/test.py
+++ b/k8s.io/test.py
@@ -120,6 +120,38 @@ class RedirTest(unittest.TestCase):
                 'https://github.com/kubernetes/kubernetes/tree/master/$path',
                 path=path)
 
+    def test_dl(self):
+        # FIXME: https://dl.{k8s,kubernetes}.io is not on the cert
+        for base in ('http://dl.k8s.io', 'http://dl.kubernetes.io'):
+            # Valid release version numbers
+            for extra in ('', '-alpha.$rc_ver', '-beta.$rc_ver'):
+                self.assert_redirect(
+                    base + '/v$major_ver.$minor_ver.$patch_ver' + extra + '/$path',
+                    'https://storage.googleapis.com/kubernetes-release/release/v$major_ver.$minor_ver.$patch_ver' + extra + '/$path',
+                    major_ver=rand_num(), minor_ver=rand_num(), patch_ver=rand_num(), rc_ver=rand_num(), path=rand_num())
+            # Not a release version
+            self.assert_redirect(
+                base + '/v8/engine',
+                'https://storage.googleapis.com/kubernetes-release/v8/engine')
+            # Not a valid release version (gamma)
+            self.assert_redirect(
+                base + '/v1.2.3-gamma.4/kubernetes.tar.gz',
+                'https://storage.googleapis.com/kubernetes-release/v1.2.3-gamma.4/kubernetes.tar.gz')
+            # A few /ci/ tests
+            self.assert_redirect(
+                base + '/ci/v$ver/$path',
+                'https://storage.googleapis.com/kubernetes-release-dev/ci/v$ver/$path',
+                ver=rand_num(), path=rand_num())
+            self.assert_redirect(
+                base + '/ci/latest-$ver.txt',
+                'https://storage.googleapis.com/kubernetes-release-dev/ci/latest-$ver.txt',
+                ver=rand_num())
+            # Base case
+            self.assert_redirect(
+                base + '/$path',
+                'https://storage.googleapis.com/kubernetes-release/$path',
+                path=rand_num())
+
     def test_docs(self):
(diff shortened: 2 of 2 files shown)


thockin:
why not `dl.k8s.io/v1.4.0/kubernetes-1.4.0.tgz`?  less characters overall.  `dl` is easy to remember.  we can argue about the path below it and the actual filename.  I think the filename has to have the version in it.

ixdy:
I like that! Done.

`https://dl.k8s.io/v[major].[minor]` will automatically add the `/release/` to the path, so you can do something like https://dl.k8s.io/v1.4.0/kubernetes.tar.gz, but you can still get to all the random crap we mirror, like https://dl.k8s.io/docker/docker-engine_1.9.1-0~wheezy_amd64.deb.

I also added a `/ci/` redirect, so you could do
```shell
latest=$(curl -L https://dl.k8s.io/ci/latest.txt)
curl -O "https://dl.k8s.io/ci/${latest}/kubernetes.tar.gz"
curl -O "https://dl.k8s.io/ci/${latest}/bin/bin/linux/amd64/kubecfg"
```
etc etc and have it do the right thing.

I'm guessing you'll need add DNS separately out-of-band?

k8s.io/test.py
```diff
@@ -120,6 +120,19 @@ def test_code(self):
                 'https://github.com/kubernetes/kubernetes/tree/master/$path',
                 path=path)
 
+    def test_dl(self):
+        # FIXME: https://dl.kubernetes.io is not on the cert
```
thockin: is dl.k8s.io on the cert?  Why would that one be but not kubernetes?

k8s.io/configmap-nginx.yaml
```diff
@@ -90,6 +90,18 @@ data:
         rewrite ^/(.*)?$    https://github.com/kubernetes/kubernetes/tree/master/$1 redirect;
       }
       server {
+        server_name dl.k8s.io dl.kubernetes.io;
+        listen 80;
+        listen 443 ssl;
+
+        location / {
+          # Don't require /release/ if you want to get at the Kubernetes release artifacts, the common case.
+          rewrite ^/(v[0-9]+\.[0-9]+.*)$   https://storage.googleapis.com/kubernetes-release/release/$1 redirect;
```
thockin: This URL doesn't work.  Is it supposed to?

k8s.io/configmap-nginx.yaml
```diff
@@ -90,6 +90,18 @@ data:
...
+          rewrite ^/ci/?(.*)$              https://storage.googleapis.com/kubernetes-release-dev/ci/$1 redirect;
```
thockin: What does `ci` mean here?

k8s.io/configmap-nginx.yaml
thockin: should this be `^/v([0-9]+\.[0-9]+)(/.*)?$` and `https://storage.googleapis.com/kubernetes-release/release/$1$2` ?

see the examples domain further down

k8s.io/test.py
ixdy: dl.k8s.io isn't on the cert. I guess we'd need to get a new cert with that added somehow? not sure how to do that.

I just cargo-culted the comment from other test cases. things like https://get.k8s.io are also broken.

k8s.io/configmap-nginx.yaml
ixdy: "continuous integration" - i.e. release artifacts generated by Jenkins.

thockin: That's what I assumed, but I didn't know what sort of stuff would live there.

k8s.io/test.py
thockin: get.k8s.io works just fine

k8s.io/configmap-nginx.yaml
ixdy: fair point, added a comment.

k8s.io/test.py
ixdy: er. sorry, mistyped. I meant https://get.kubernetes.io.

k8s.io/configmap-nginx.yaml
ixdy: in particular, the `v` is a part of the path. For examples we remove it, since the git branch name is `release-1.4`, not `release-v1.4`.

thockin: yeah, i was treating GCS as a file server again, and it just isn't.  I need GCSweb.

k8s.io/test.py
thockin: yes, tha''s true

k8s.io/configmap-nginx.yaml
thockin: re: examples, my point was to catch `/v1.4/foo` but not `v1.4.foo` - but now I am unclear.  Is the path supposed to have `v1.4` or `v.1.4.0`?  The example above show 3 digits, but your test shows 2 digits

k8s.io/configmap-nginx.yaml
```diff
@@ -90,6 +90,19 @@ data:
         rewrite ^/(.*)?$    https://github.com/kubernetes/kubernetes/tree/master/$1 redirect;
       }
       server {
+        server_name dl.k8s.io dl.kubernetes.io;
+        listen 80;
+        listen 443 ssl;
+
+        location / {
+          # Don't require /release/ if you want to get at the Kubernetes release artifacts, the common case.
+          rewrite ^/(v[0-9]+\.[0-9]+.*)$   https://storage.googleapis.com/kubernetes-release/release/$1 redirect;
```
ixdy: yeah, this was me mostly being lazy. I wanted to not accidentally catch something like /v1/, though nothing in `gs://kubernetes-release/` matches anything like that right now, so just `/v[0-9]+` would work.

it looks like everything in `gs://kubernetes-release/release` is of the form `vX.X.X(-(alpha|beta).X)?`. if you want me to make the regex more specific, I can do that.

thockin:
I want to be as precise as we can afford to be.  If we expect it to be 3
digits, let's say that.  If we want to allow 2 or 3 digits, either fix up
the regex or make 2 rules.  Thanks.

ixdy:
Done, now with even more unit tests.

k8s.io/test.py
```diff
@@ -120,6 +120,39 @@ def test_code(self):
...
+        for base in ('http://dl.k8s.io', 'https://dl.k8s.io', 'http://dl.kubernetes.io'):
```
thockin: still need to take https out of this?

ixdy:
I haven't pushed this anywhere besides my own GKE cluster yet.

thockin:
LGTM

thockin:
it's live

ixdy:
thanks!

PR #19, opened 2016-10-04
Fix HTTPS in k8s.io unit test

jbeda:
I'm not sure how this ever worked before.  Note that this is me running this on my mac with python 2.7.12 installed via homebrew.

---

This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/k8s.io/19)

### DIFF
--- a/k8s.io/test.py
+++ b/k8s.io/test.py
@@ -10,6 +10,7 @@ except ImportError:
     import urllib.parse as urlparse
 import os
 import random
+import ssl
 import subprocess
 import unittest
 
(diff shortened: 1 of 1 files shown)


jbeda:
@thockin @ixdy @rmmh

rmmh:
I'd rather declare outright that this test requires Python 3 than add hacks to support Python 2.

jbeda:
Hmm. Some indication of that would have saved me a bunch of time this
afternoon. ;)

rmmh:
I'll have to check my mac's .bash_history, but I think I installed a bunch of ssl modules from pip that magically made the SNI work properly.

ixdy:
So on the ancient python version we have installed on workstations at Google (2.7.6), it seems like Python doesn't do any verification of TLS certificates. This PR would probably verify certs correctly, except that our Python is so old we don't even have `ssl.create_default_context()`.

In true k8s spirit, I instead just used docker to run the test, with this change included.
```
docker run -it --rm --name k8s-io-test -e TARGET_IP=23.236.58.218 -v "$PWD":/usr/src/k8s.io -w /usr/src/k8s.io python:2 bash -c "pip install pyyaml && python test.py"
```

ixdy:
need to fix the merge conflict I just introduced.

LGTM, with the caveat that Googlers now need to use `make docker-test` instead of `make test`. :\

thockin:
Is this alive?

jbeda:
Sorry -- I suck.  So behind on stuff.  If y'all still want it I can rebase/fix.  Or we can close and just live with it.  Let me know either way.

thockin:
I can leave it.  Just pinging.

rmmh:
Closing for now.

ixdy:
I think we might need this PR again. This test has mysteriously started failing on my workstation.

ixdy:
(I think I had some old version of the python image in https://github.com/kubernetes/k8s.io/pull/19#issuecomment-265556199. 2.7.8 works, 2.7.9 fails. python is up to 2.7.14 now :\ )

ixdy:
rebased this into #104.

PR #28, opened 2016-10-13
Integrate changes to kubernetes/cluster/get-kube.sh into get-kube-secure.sh

ixdy:
The script on https://get.k8s.io is pretty out-of-date at this point.

---
This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/k8s.io/28)

### DIFF
--- a/k8s.io/configmap-www-get.yaml
+++ b/k8s.io/configmap-www-get.yaml
@@ -4,10 +4,11 @@ metadata:
   name: www-get
 data:
   # Adding new entries here will make them appear as files in the deployment.
+  # TODO: is there a better way to keep this up-to-date with cluster/get-kube.sh in kubernetes/kubernetes?
   get-kube-secure.sh: |
     #!/bin/bash
 
-    # Copyright 2014 The Kubernetes Authors All rights reserved.
+    # Copyright 2014 The Kubernetes Authors.
     #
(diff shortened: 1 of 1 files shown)


ixdy:
ping @thockin @rmmh ?

k8s.io/configmap-www-get.yaml
```diff
@@ -4,10 +4,11 @@ metadata:
   name: www-get
 data:
   # Adding new entries here will make them appear as files in the deployment.
+  # TODO: is there a better way to keep this up-to-date with cluster/get-kube.sh in kubernetes/kubernetes?
```
thockin: we could maybe redirect directly to that file?

thockin:
lgtm

k8s.io/configmap-www-get.yaml
ixdy: that'd probably work, though we'd need to update our documentation (such as http://kubernetes.io/docs/getting-started-guides/binary_release/#download-kubernetes-and-automatically-set-up-a-default-cluster) to pass `-L` to `curl`, since by default it'll helpfully just print out "302 Found".

thockin:
Or we could proxy that file through, so no redirect needed.

Issue #18, opened 2016-10-04
Update TLS cert for most of the aliases


jbeda:
A wildcard cert would work but I'm assuming that it is hard to get one of those.  Right now the cert for k8s.io has a set of SANs:

```
k8s.io
changelog.k8s.io
code.k8s.io
docs.k8s.io
examples.k8s.io
get.k8s.io
issue.k8s.io
issues.k8s.io
jenkins.k8s.io
pr.k8s.io
prs.k8s.io
rel.k8s.io
releases.k8s.io
slack.k8s.io
submit-queue.k8s.io
www.k8s.io
```

This is missing:

```
yum.k8s.io
apt.k8s.io
dl.k8s.io
features.k8s.io
feature.k8s.io
reviewable.k8s.io
testgrid.k8s.io
```

mikedanese:
@ixdy

ixdy:
There are a few others missing, too:
* `ci-test.k8s.io`
* `gcsweb.k8s.io`
* `go.k8s.io`
* `pr-test.k8s.io`

and also all of the `*.kubernetes.io` variants of all of these. And we'll undoubtedly have more as we add additional vanity subdomains.

I think we want to move to LetsEncrypts certs instead of the Google-issued one we have right now. This would let us add additional vanity subdomains fairly easily, but someone needs to setup the automation around this. (LE does not do wildcard certs, either.)

We have about 2 months to get this done.

techtonik:
https://github.com/AnalogJ/lexicon with https://github.com/AnalogJ/lexicon/blob/master/examples/letsencrypt.default.sh should do the trick.

How do you maintain DNS records? Is there any .yaml with mapping?

thockin:
I have seen about such client libs.  I don't have time to try them all.  I
need a recommendation of one that is trustworthy and easy and can pull a
cert with an arbitrary number of SANs.

techtonik:
Let's encrypt allows to `--expand` existing certificate with more names - http://stackoverflow.com/questions/35777157/how-to-add-change-certificates-issued-by-letsencrypt, but `k8s.io` domains use Google Internet authority, which is not accessible to public and hence doesn't have public API or client libs. So a decision to switch to let's encrypt should be taken first.

thockin:
Yeah, if we're going to switch we have to switch it all.

ixdy:
[...]

PR #30, opened 2016-10-16
Machine readable domain inventory

techtonik:
to maintain a list of all domains programmatically

---

This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/k8s.io/30)

### DIFF
--- a/check.py
+++ b/check.py
@@ -56,13 +56,14 @@ def sitecheck(site):
     return check
 
 
-errors = 0
-for site in inv['domains']:
-    if not sitecheck(site):
-        errors += 1
+if __name__ == '__main__':
+    errors = 0
+    for site in inv['domains']:
+        if not sitecheck(site):
+            errors += 1
 
-print('\n{} domains, {} errors'.format(len(inv['domains']), errors))
-sys.exit(errors)
+    print('\n{} domains, {} errors'.format(len(inv['domains']), errors))
+    sys.exit(errors)
 
 """
 
--- a/check.py
+++ b/check.py
@@ -40,7 +40,8 @@ for site in inv['domains']:
         if type(reason.message) != str:
             raise
         if ("[Errno 11001] getaddrinfo failed" in reason.message or     # Windows
-            "[Errno -2] Name or service not known" in reason.message):  # Linux
+            "[Errno -2] Name or service not known" in reason.message or # Linux
+            "[Errno 8] nodename nor servname " in reason.message):      # OS X
             message = 'DNSLookupError'
         else:
             raise
--- a/check.py
+++ b/check.py
@@ -9,14 +9,26 @@ import sys
 import yaml
 import requests
 
+statusmap = {
+    '200': 'OK',
+    '404': 'Not Found',
+    '302': 'Redirect',
+}
+
+
 inv = yaml.load(open('inventory.yaml'))
+# calc names width for pretty printing
+colwidth = max([len(item) for item in inv['domains']])+1
 
 
 errors = 0
 
 for site in inv['domains']:
+    status = None
+    message = ''
     try:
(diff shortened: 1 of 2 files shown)


ixdy:
how was this list compiled? "programmatically" implies a script, which I don't see here.

techtonik:
I compiled it manually and got stuck to create a link checking and reporting script. I thought I could reuse link checking code from Sphinx, but got carried away with refactoring, which now seems pointless. I'll try to commit something in the next couple of days.

ixdy:
OK. I just worry that having a list separate from the configmaps isn't very helpful, since it'll almost certainly get out of sync.

techtonik:
I've not yet learned about http://kubernetes.io/docs/user-guide/configmap/ so I started with an oldschool way.

thockin:
What's the objective?

techtonik:
@thockin automate maintenance of domains. Right now implement a human readable status check for config, like in #18 to catch things like expired domain names or missing entries (through DNS API).

techtonik:
Ready for review.
```
k8s.io               302   Redirect
changelog.k8s.io     302   Redirect
code.k8s.io          302   Redirect
docs.k8s.io          302   Redirect
examples.k8s.io      302   Redirect
get.k8s.io           200   OK
issue.k8s.io         302   Redirect
issues.k8s.io        302   Redirect
jenkins.k8s.io       None  DNSLookupError
pr.k8s.io            302   Redirect
prs.k8s.io           302   Redirect
rel.k8s.io           302   Redirect
releases.k8s.io      302   Redirect
slack.k8s.io         200   OK
submit-queue.k8s.io  200   OK
www.k8s.io           302   Redirect
yum.k8s.io           302   Redirect
apt.k8s.io           302   Redirect
dl.k8s.io            302   Redirect
features.k8s.io      302   Redirect
feature.k8s.io       302   Redirect
reviewable.k8s.io    302   Redirect
testgrid.k8s.io      302   Redirect
ci-test.k8s.io       302   Redirect
gcsweb.k8s.io        404   Not Found
go.k8s.io            404   Not Found
pr-test.k8s.io       302   Redirect

27 domains, 3 errors
```

thockin:
I'm afraid I still don't understand - we already have a test that tests these more completely and correctly.  2 of your 3 errors (404) are not errors, just that the / URL doesn't have meaning.  The third appears to be a domain we don't actually use.

Other than being yet another place to update when adding a domain - what value does this have?

techtonik:
@thockin can you point me to that test? I'd like to see where it gets its list of domains. If I knew it exists, I would spend my free time on writing domain management config instead.

The point is to have a human-friendly, but also machine readable format for various services that project maintains. For the newcomers like myself it is very hard to understand what is available - where is bugtracker, binaries, and interesting dashboard etc. All these services are probably documented somewhere, but most interesting ones for development are usually known only to existing developers (like `gcsweb` with convenient `kubectl` downloads for Windows).

"domain" for the web is the same as "entrypoint" for binary executable on operating system. It is where the process starts. Or maybe domain is the executable of the web. So being able to "dir" the new project is an awesome feature as for me. =)

techtonik:
It will help me to redirect my efforts into something usable for the project if you could explain how the current domain configuration is managed. Do you manually setup 302 redirects and add new third level domains? If so which interface is used?

thockin:
https://github.com/kubernetes/k8s.io/blob/master/k8s.io/test.py

thockin:
Is this still something you're thinking about?  I'm not sure what the goal is..

techtonik:
@thockin yes, but 1. more generic declarative format for the start, 2. available for 3rd party developers to implement, 3. less verbose license terms

PR #31, opened 2016-10-19
Proxy https://get.k8s.io to GitHub instead of including a copy

ixdy:
Follow-up to suggestion on #28 from @thockin.

Tested with `make test` on my private cluster.

---
This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/k8s.io/31)


thockin:
LGTM, but I don't have time to test it right now.  will hold it in pattern until I get a break.  Thanks.

I should just give you access to that cluster so you can admin it :)

ixdy:
might be handy, especially if I end up owning #18...

thockin:
deployed

PR #37, opened 2016-12-07
Test https on all subdomains

ixdy:
Refactored test.py a bit so that we automatically test http and https in all test cases.

I've also added a Dockerfile and make rule so we can run test.py with a less-ancient version of python that actually validates the cert. This doesn't actually work without @jbeda's fix in #19, but #19 will fail with ancient python because `ssl.create_default_context()` doesn't exist. It's a mess.

I've pushed the updated nginx configmap to canary but not yet prod.

### DIFF
--- a/k8s.io/Dockerfile-test
+++ b/k8s.io/Dockerfile-test
@@ -1,0 +1,10 @@
+FROM python:2
+MAINTAINER Jeff Grafton <jgrafton@google.com>
+
+WORKDIR /workspace
+
+RUN pip install pyyaml
+
+COPY test.py configmap-*.yaml /workspace/
+
+ENTRYPOINT /workspace/test.py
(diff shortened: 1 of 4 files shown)


rmmh:
Why not make the dockerfile based on Python 3? It has much better SSL support.

thockin:
lgtm

thockin:
Deployed?

ixdy:
@rmmh `urllib.urlopen` doesn't exist in python3, and I'm failing at trying to fix (while also juggling with #19). Let's merge this, merge #19, and then we can fix for python3?
@thockin deployed to canary, will deploy to prod and then merge.

rmmh:
there's no use of urllib.urlopen here? I ran the test script in python3 before

ixdy:
@rmmh https://github.com/kubernetes/k8s.io/blob/master/k8s.io/test.py#L247 - I added this since you first wrote the script. I did not test it in python3, obviously. :)

PR #41, opened 2017-01-10
add api machinery

deads2k:
Add entry for `k8s.io/apimachinery` for https://github.com/kubernetes/gengo/pull/29

@thockin is there any way for me to test this?


thockin:
Not really.  We have to deploy it.  We should probably create a siteadmin SIG or something to share the load.

thockin:
pushed to prod

deads2k:
> pushed to prod

That did the trick, thanks.

PR #43, opened 2017-01-10
Switch main site host header to real backend

thockin:
pushed and activated now that the switchover is done

### DIFF
--- a/k8s.io/configmap-nginx.yaml
+++ b/k8s.io/configmap-nginx.yaml
@@ -41,7 +41,7 @@ data:
 
         # Proxy to the real site.
         location / {
-          proxy_set_header Host kubernetes.io;
+          proxy_set_header Host kubernetes.github.io;
           proxy_set_header X-Real-IP $remote_addr;
           proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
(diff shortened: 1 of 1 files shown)


ixdy:
Not a huge deal, but it seems like some self-generated redirects now go to kubernetes.github.io - e.g. if you go to https://kubernetes.io/docs instead of https://kubernetes.io/docs/ (trailing slash).

thockin:
crap, how do I fix that?  I could re-enable github to think it owns the
custom domain (would have to disable 'enforce ssl') and have it generate
redirects to  the custom domain.  That's probably the only solution, eh?

thockin:
OK, I fixed it by basically rolling this back.  Good catch

thockin:
ugh, it only does HTTP, not https.

ixdy:
ugh, that might be worse than redirecting to kubernetes.github.io.

thockin:
I think I fixed it.  PR coming once I fix tests.  If you are up - poke
around the site?

ixdy:
yeah, seems to be working. sweet!

(and now I notice that blog.kubernetes.io is HTTP-only, but this time it's Blogger that doesn't support HTTPS for custom domains.)

PR #44, opened 2017-01-11
Fix redirects for the main site

thockin:
Fixes wonky redirects, adds protocl upgrades, and adds test cases

### DIFF
--- a/k8s.io/test.py
+++ b/k8s.io/test.py
@@ -39,86 +39,112 @@ def do_get(url):
 
 
 class RedirTest(unittest.TestCase):
-    def assert_redirect(self, partial_url, expected_loc, **kwargs):
+    def assert_code(self, url, expected_code):
+        print('GET: %s => %s' % (url, expected_code))
+        resp, body = do_get(url)
+        self.assertEqual(resp.status, expected_code)
+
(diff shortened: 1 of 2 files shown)


thockin:
already pushed

k8s.io/test.py
```diff
@@ -39,194 +39,220 @@ def do_get(url):
...
+            k = '$%s' % k
+            v = '%s' % v
+            url = url.replace(k, v)
+            expected_loc = expected_loc.replace(k, v)
+        print('REDIR: %s => %s' % (url, expected_loc))
+        resp, body = do_get(url)
+        self.assertEqual(resp.status, expected_code)
+        self.assertEqual(resp.getheader('location'), expected_loc)
+
+    def assert_redirect(self, partial_url, expected_loc, expected_code, **kwargs):
```
rmmh: since `assert_perm_redirect` is only used twice, it would be cleaner to just make expected_code a default arg:

    def assert_redirect(self, partial_url, expected_loc, expected_code=302, **kwargs)

thockin:
I actually like the name being very clear about perm vs temp..  Anyway, if
we want to argue this, can we do in a followup.  This is already pushed,
I'd like it to get off my plate soon :)

PR #45, opened 2017-01-13
add k8s.io/apiserver

deads2k:
Add entry for godeping k8s.io/apiserver.


deads2k:
@lavalamp I recall needing this for godeping https://github.com/kubernetes/k8s.io/pull/41

lavalamp:
LGTM

lavalamp:
I probably need to remember how to push this before it will take effect.

deads2k:
> I probably need to remember how to push this before it will take effect.

https://github.com/kubernetes/apiserver is synced and ready when this rolls out.  Mention me when its pushed?

thockin:
pushed - should work now

PR #47, opened 2017-01-23
www-golang: add ingress

euank:
Someone ran into this not working https://github.com/kubernetes/ingress/issues/147


thockin:
I wish I could make it work generically...

euank:
Using `github.com/kubernetes/...` for the import path works generically and is easier for a beginner to go from import path -> source code.

No clue why the vanity urls are being used in the first place.

thockin:
There was some thought that we might move off of github, and a shorter URL
was attractive.

vharsh:
> There was some thought that we might move off of github, and a shorter URL was attractive.
Looks like this was possible just because `go get` is not just `git clone` outright as most beginners begin to think.
view-source:https://k8s.io/kubectl?go-get=1

PR #49, opened 2017-02-04
PROOF OF CONCEPT: Make go-get fetch client-go release-1.5 branch

sttts:
/cc @deads2k @lavalamp @caesarxuchao

### DIFF
--- a/k8s.io/configmap-www-golang.yaml
+++ b/k8s.io/configmap-www-golang.yaml
@@ -72,8 +72,8 @@ data:
     </head></html>
   client-go.html: |
     <html><head>
-      <meta name="go-import" content="k8s.io/client-go git https://github.com/kubernetes/client-go">
-      <meta name="go-source" content="k8s.io/client-go     https://github.com/kubernetes/client-go https://github.com/kubernetes/client-go/tree/master{/dir} https://github.com/kubernetes/client-go/blob/master{/dir}/{file}#L{line}">
+      <meta name="go-import" content="k8s.io/client-go git https://gopkg.in/kubernetes/client-go.v1">
+      <meta name="go-source" content="k8s.io/client-go     https://github.com/kubernetes/client-go https://github.com/kubernetes/client-go/tree/release-1.5{/dir} https://github.com/kubernetes/client-go/blob/release-1.5{/dir}/{file}#L{line}">
     </head></html>
   apimachinery.html: |
     <html><head>


sttts:
Not sure this is nice.

lavalamp:
@caesarxuchao there was a reason we didn't do this, right? Can you remember what it was?

thockin:
Fascinating trick.  I am strangely OK with this...  I think..

thockin:
The alternative to depending on a random third-party FREE website would be to run our own instance of gopkg.

lavalamp:
Any suggestion on how to test this not in production?

sttts:
@lavalamp just create /etc/hosts alias for k8s.io pointing to 127.0.0.1 and serve the `client-go` file with the changed contents using some webserver:
```
<html><head>
  <meta name="go-import" content="k8s.io/client-go git https://gopkg.in/kubernetes/client-go.v1">
  <meta name="go-source" content="k8s.io/client-go     https://github.com/kubernetes/client-go https://github.com/kubernetes/client-go/tree/release-1.5{/dir} https://github.com/kubernetes/client-go/blob/release-1.5{/dir}/{file}#L{line}">
</head></html>
```

caesarxuchao:
@sttts @lavalamp Is the code downloaded to gopkg.in/kubernetes/client-go.v1 or k8s.io/client-go? I remember it's the former, that means we need to rewrite all the import paths.

sttts:
@caesarxuchao the code goes through `gopkg.in`'s https proxy. But go will use the `k8s.io/client-go` name, like it does not with calling it `k8s.io/client-go` and not `github.com/kubernetes/client-go`.

sttts:
I want to mention one downside of this approach: `go get` will checkout the code using `gopkg.in`, i.e. the git remote looks like this:
```
[remote "origin"]
	url = https://gopkg.in/kubernetes/client-go.v1
	fetch = +refs/heads/*:refs/remotes/origin/*
```

In other words, a `git pull --rebase` or probably also `go get -u` will stay on the v1 branch. If you delete the repo and `go get k8s.io/client-go` again, you get the new major release, if there is any.

caesarxuchao:
> the code goes through gopkg.in's https proxy. But go will use the k8s.io/client-go name

That's nice! @lavalamp I think we had tried changing this file to redirect to a stable release, we also had tried using gopkg.in, but we hadn't tried to combine two approaches together like what this PR does.

thockin:
This is a bit of a wart...

An alternative again might be to export to YET ANOTHER repo that is just
the import-rewritten tip of the stable branch...

PR #50, opened 2017-02-14
expose kube-aggregator and sample-apiserver

deads2k:
expose kube-aggregator and sample-apiserver so they are `go get`able.

@lavalamp


lavalamp:
LGTM

lavalamp:
@thockin is there someone besides you I can bug for a push?

pwittrock:
@thockin Let me know if this is something it is possible for me to help out with in a way that lessens your load.

deads2k:
Bump

deads2k:
@lavalamp I think this breaking https://github.com/kubernetes/kubernetes/issues/42821

deads2k:
@lavalamp @thockin hit this again.  Can we move forward with exposing it?

thockin:
2 things to make this easier on me:

1) someone figure out how to make nginx server a templated literal response, rather than a file per repo.  E.g. if I ask for k8s.io/frobber, it would expand this same HTML block and sub "frobber" in as repo name.

2) a few more people hold keys to this cluster.  Already @ixdy has keys and a couple others in engprod.

I'll push this now.

thockin:
pushed live - should self-update within a minute, and go-get will start
working.

PR #51, opened 2017-02-16
Add redirect for go.k8s.io/oncall

ixdy:
Easier to tell folks to use https://go.k8s.io/oncall rather than trying to remember the GCS path.

Already pushed and tested on canary.


ixdy:
Thanks. Bouncing k8s.io prod now.

PR #55, opened 2017-03-02
Add triage and history go links

fejta:
@ixdy 

Make go.k8s.io/test-history and go.k8s.io/triage work!

### DIFF
--- a/k8s.io/configmap-nginx.yaml
+++ b/k8s.io/configmap-nginx.yaml
@@ -187,9 +187,11 @@ data:
         location / {
           rewrite ^/bounty$          https://github.com/kubernetes/kubernetes.github.io/issues?q=is%3Aopen+is%3Aissue+label%3ABounty redirect;
           rewrite ^/help-wanted$     https://github.com/kubernetes/kubernetes/labels/help-wanted redirect;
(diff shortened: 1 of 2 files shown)


k8s.io/configmap-nginx.yaml
```diff
@@ -190,6 +190,8 @@ data:
           rewrite ^/oncall$          https://storage.googleapis.com/kubernetes-jenkins/oncall.html;
           rewrite ^/partner-request$ https://docs.google.com/forms/d/e/1FAIpQLSdN1KtSKX2VAOPGABFlShkSd6CajQynoL4QCVtY0dj76MNDKg/viewform redirect;
           rewrite ^/start$           http://kubernetes.io/docs/getting-started-guides/ redirect;
+          rewrite ^/test-history$    https://storage.googleapis.com/kubernetes-test-history/static/index.html
```
ixdy: these are missing ` redirect;`.

(also, curiously the `/oncall` one is missing `redirect` but it works?)

ixdy: looks like ` redirect` is optional, since these all start with `https`.
The semicolon is not optional. :)

fejta: Done

PR #57, opened 2017-04-21
Add kubernetes/autoscaler to configmap-www-golang.yaml

mwielgus:
cc: @MaciekPytel @fgrzadkowski


thockin:
pushed, should be live momentarily.

PR #70, opened 2017-07-28
Configure nginx for netlify cutover

thockin:
Netlify will handle the kubernetes.io domain directly.  We will still
handle k8s.io (including go-get) and all the vanity URLs, and redirect
to kubernetes.io as needed.

This required a bunch of rejiggering the nginx config, but I think it is
clearer now.

I also beefed up the tests some and covered some cases that we missed
before.

Tests pass against canary, and I pushed the nginx config to prod.

### DIFF
--- a/k8s.io/configmap-nginx.yaml
+++ b/k8s.io/configmap-nginx.yaml
@@ -21,13 +21,21 @@ data:
       #
       # Upgrade HTTP to HTTPS.
       server {
-        server_name kubernetes.io;
-        listen 80;
-        return 301 https://$server_name$request_uri;
+        server_name k8s.io;
+        listen 80 default_server;
+
+        location /_healthz {
+          add_header Content-Type text/plain;
+          return 200 'ok';
+        }
+
+        location / {
+          return 301 https://$host$request_uri;
+        }
       }
       # Handle HTTPS.
       server {
-        server_name kubernetes.io;
+        server_name k8s.io;
         listen 443 ssl;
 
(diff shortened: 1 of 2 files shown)


k8s.io/configmap-nginx.yaml
```diff
@@ -21,13 +21,21 @@ data:
       #
       # Upgrade HTTP to HTTPS.
       server {
-        server_name kubernetes.io;
-        listen 80;
-        return 301 https://$server_name$request_uri;
+        server_name k8s.io;
+        listen 80 default_server;
+
+        location /_healthz {
```
ixdy: what is using this? is there another config change missing from this PR?

k8s.io/configmap-nginx.yaml
```diff
@@ -21,13 +21,21 @@ data:
...
+          add_header Content-Type text/plain;
+          return 200 'ok';
+        }
+
+        location / {
+          return 301 https://$host$request_uri;
```
ixdy: should this go directly to https://kubernetes.io to skip a hop though https://k8s.io?

k8s.io/configmap-nginx.yaml
ixdy: oh, I understand what this is doing now. never mind.

k8s.io/configmap-nginx.yaml
thockin: No.  I want go-get to go through HTTPS and that has to stay local.

PR #77, opened 2017-08-16
k8s.io/configmap-www-golang.yaml: add code-generator repo

sttts:


thockin:
I don't even know what this repo is for?

thockin:
@ixdy to merge and push

ixdy:
this seems fine, but maybe we should actually have something in https://github.com/kubernetes/code-generator first?

sttts:
> this seems fine, but maybe we should actually have something in https://github.com/kubernetes/code-generator first?

On its way. Will ping here when the publishing bot is updated.

cblecker:
@ixdy @sttts Would this hurt to merge/deploy before there is actual code there? I'm toying with `dep` again and it's dying because of this :(

sttts:
@cblecker I don't think it hurts. We just have an empty repo behind it.

ixdy:
passes tests on canary. merging.

PR #79, opened 2017-08-24
Remove kubernetes.io proxy passthru to GitHub pages

ixdy:
The https://kubernetes.io test started failing, but we've completely switched to netlify, so we don't need it anymore anyway.

Already deployed and tested on canary.


thockin:
LGTM, please deploy.

ixdy:
deployed along with #77.

PR #80, opened 2017-08-28
Add actual list of redirectors currently active

dims:
We don't have a list of these url(s) anywhere on a single page which
makes it harder to discover

### DIFF
--- a/k8s.io/README.md
+++ b/k8s.io/README.md
@@ -3,6 +3,22 @@ Overview
 This contains the Nginx configuration for k8s.io and the associated subdomain
 redirectors.
 
+Redirections
+====
+- https://go.k8s.io/bounty
+- https://go.k8s.io/help-wanted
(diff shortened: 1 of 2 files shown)


k8s.io/README.md
```diff
@@ -3,6 +3,20 @@ Overview
 This contains the Nginx configuration for k8s.io and the associated subdomain
 redirectors.
 
+Redirections
```
ixdy: maybe add a comment in the ngnix configmap to keep this list up-to-date?

fejta: Also link to the configmap so someone can find the source of truth?

thockin: We should also document the vanity URLs we support (there are a lot of them, eg. pr.k8s.io)

dims: Ack @thockin i'll figure that out

jdumars:
Thanks @dims !

PR #90, opened 2017-11-13
Create go-import meta for k8s.io/kubectl

apelisse:
Add redirection for `k8s.io/kubectl`. We will start moving some code from `k8s.io/kubernetes` into `k8s.io/kubectl`, and vendor that.


thockin:
pushed live

apelisse:
thank you

PR #93, opened 2017-11-28
Add a redirect (short link) for the cluster-api project.

roberthbailey:
I also added documentation for another redirect that was in the nginx config but not in the README. 

/cc @pipejakob

### DIFF
--- a/k8s.io/configmap-nginx.yaml
+++ b/k8s.io/configmap-nginx.yaml
@@ -208,6 +208,7 @@ data:
         location / {
           rewrite ^/bounty$          https://github.com/kubernetes/kubernetes.github.io/issues?q=is%3Aopen+is%3Aissue+label%3ABounty redirect;
           rewrite ^/bot-commands$    https://github.com/kubernetes/test-infra/blob/master/commands.md redirect;
+          rewrite ^/cluster-api$     https://github.com/kubernetes/kube-deploy/blob/master/cluster-api/README.md redirect;
           rewrite ^/help-wanted$     https://github.com/kubernetes/kubernetes/labels/help-wanted redirect;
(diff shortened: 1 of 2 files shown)


k8s.io/configmap-nginx.yaml
```diff
@@ -208,6 +208,7 @@ data:
         location / {
           rewrite ^/bounty$          https://github.com/kubernetes/kubernetes.github.io/issues?q=is%3Aopen+is%3Aissue+label%3ABounty redirect;
           rewrite ^/bot-commands$    https://github.com/kubernetes/test-infra/blob/master/commands.md redirect;
+          rewrite ^/cluster-api$     https://github.com/kubernetes/kube-deploy/blob/master/cluster-api/README.md
```
mindprince: Missing ` redirect;` at the end.

roberthbailey: good catch!. fixed and re-pushed.

thockin:
This is the first sub-project related shortlink, AFAIK.  We don't have rules for this, yet.  Where is this to be used?  What I want to avoid is hundred of these, pointing to sub-dirs of github repos that perhaps should just be their own github repos...

roberthbailey:
Jacob suggested using a friendly short-link in my kubecon presentation and Ryan said that we had such a thing and pointed me at these files. But I think it'll be useful to have in general. What is the benefit of keeping this to a slim set? Why not have hundreds of human friendly short links?

thockin:
Because it is not nearly self-service enough nor namespaced enough?

roberthbailey:
Self-service could be fixed pretty easily by a new community member. Namespacing is a more difficult question.

thockin:
I pushed this live.  I'm fine with it, and we'll deal with it when it
becomes a problem, I guess.

PR #101, opened 2018-01-02
Make go-get content be automatic

### LINKED ISSUE #72: `k8s.io/dns` (and others) missing from golang responses
We should probably (a) have this auto generated from some list of repos someplace and (b) make sure that we keep it up to date.  But I can at least say that `k8s.io/dns` is missing.

thockin:
Downside is it always assumes the go-get name and the repo name are the same, and that it comes from the kubernetes org.

I am not sure we should make that tradeoff.  Want to hear thoughts.

Fixes #72

### DIFF
--- a/k8s.io/configmap-nginx.yaml
+++ b/k8s.io/configmap-nginx.yaml
@@ -204,7 +204,7 @@ data:
 
         location / {
           rewrite ^/v([0-9]+\.[0-9]+)(/.*)?$    https://github.com/kubernetes/kubernetes/tree/release-$1/examples$2  redirect;
-          rewrite ^/(.*)$                       https://github.com/kubernetes/kubernetes/tree/master/examples/$1     redirect;
+          rewrite ^/(.*)$                       https://github.com/kubernetes/examples/tree/master/$1                redirect;
         }
       }
       server {
--- a/k8s.io/test.py
+++ b/k8s.io/test.py
@@ -273,7 +273,8 @@ class RedirTest(unittest.TestCase):
 
     def test_examples(self):
         for base in ('examples.k8s.io', 'examples.kubernetes.io'):
-            self.assert_temp_redirect(base, 'https://github.com/kubernetes/kubernetes/tree/master/examples/')
+            self.assert_temp_redirect(base, 'https://github.com/kubernetes/examples/tree/master/')
+            self.assert_temp_redirect(base + '/', 'https://github.com/kubernetes/examples/tree/master/')
 
             ver = '%d.%d' % (rand_num(), rand_num())
             self.assert_temp_redirect(base + '/v$ver',


k8s.io/configmap-nginx.yaml
```diff
@@ -39,23 +39,28 @@ data:
...
+                              content="go.clustermaster.net/$1
+                                       https://github.com/kubernetes/$1
+                                       https://github.com/kubernetes/$1/tree/master{/dir}
+                                       https://github.com/kubernetes/$1/blob/master{/dir}/{file}#L{line}">
+                  </head></html>
+              ';
+          }
+
           # Default to redirecting to the "real" site.
           return 301 https://kubernetes.io$request_uri;
```
rmmh: Does this interfere with the config map?

k8s.io/configmap-nginx.yaml
```diff
@@ -39,23 +39,28 @@ data:
...
         }
 
-        location / {
+        location ~* ^/([^_][^/]*)(/.*)?$ {
+          if ($arg_go-get = "1") {
+              # This is a go-get operation.
+              return 200 '
+                  <html><head>
+                        <meta name="go-import"
+                              content="go.clustermaster.net/$1
```
rmmh: throughout: should be k8s.io/$1

k8s.io/configmap-nginx.yaml
```diff
@@ -39,23 +39,28 @@ data:
...
-          alias /www/golang;
-        }
-
         location /_healthz {
           add_header Content-Type text/plain;
           return 200 'ok';
         }
 
-        location / {
+        location ~* ^/([^_][^/]*)(/.*)?$ {
```
rmmh: Does this interfere with the normal redirect?

ixdy:
Do we expect there to be any cases where `k8s.io/foo` is *not* `github.com/kubernetes/foo`? This looks fine to me (besides @rmmh's comments).

spiffxp:
@thockin re: our recent discussion of having sigs host code during steering committee, would this interfere with the idea of having `github.com/kubernetes-sig-foo/repo` host `foo.k8s.io/repo`?  Asking for a friend over in https://github.com/kubernetes/community/pull/1524#issuecomment-355439096

totherme:
I think I'm spiffxp's friend over in kubernetes/community#1524 :)

We've created a git repo to host testing frameworks. It lives at https://github.com/kubernetes-sig-testing/frameworks. Note the new org name.

If we're automatically mapping `k8s.io/reponame` to `github.com/kubernetes/reponame` then can we also automatically map `testing.k8s.io/reponame` to `github.com/kubernetes-sig-testing/reponame`? 

Or more generally, could we map `signame.k8s.io/reponame` to `github.com/kubernetes-sig-signame/reponame` ?

hoegaarden:
Would this do what @totherme asked for?

```
  server {
    server_name k8s.io ~^(?<sig>.+)\.k8s\.io$;
    listen 443 ssl;

    location /_healthz {
      add_header Content-Type text/plain;
      return 200 'ok';
    }

    location ~* ^/([^_][^/]*)(/.*)?$ {
      if ($sig != "") {
        set $sig "kubernetes-sig-$sig";
      }
      if ($sig = "") {
        set $sig "kubernetes";
      }
      if ($arg_go-get = "1") {
          # This is a go-get operation.
          return 200 '
              <html><head>
                    <meta name="go-import"
                          content="go.clustermaster.net/$1
                                   git https://github.com/$sig/$1">
                    <meta name="go-source"
                          content="go.clustermaster.net/$1
                                   https://github.com/$sig/$1
                                   https://github.com/$sig/$1/tree/master{/dir}
                                   https://github.com/$sig/$1/blob/master{/dir}/{file}#L{line}">
              </head></html>
          ';
      }

      # Default to redirecting to the "real" site.
      return 301 https://kubernetes.io$request_uri;
    }
  }
```

hoegaarden:
@thockin @spiffxp 
Is there any movement on this PR? Is there anything we can help to get this setup?

thockin:
I just have not had time to think about it further.  I'm not sure if I want to implement the testing.k8s.io redirects or not.  Will circle back soon.

k8s.io/configmap-nginx.yaml
thockin: oops.  Fixed in next push.

thockin:
WRT adding other targets for go-get that redirect to other github orgs:  I am not against it, per se, but I'd want to think about the convention (since we have a bunch of domains already).  Can we make that a followup PR, rather than overloading this?

thockin:
PTAL - I think we can proceed with this.

k8s.io/configmap-nginx.yaml
rmmh: `~*` is unnecessary, since this isn't case sensitive. Why not `^/([^/]*)(/.*)?$`, which would match all URLs?

k8s.io/test.py
```diff
@@ -344,6 +344,7 @@ def assert_body_url(self, url, expected_content_url):
         self.assertMultiLineEqual(body, expected_body)
 
     def test_go_get(self):
+        # FIXME: need to fix this if we keep this idea.
```
ixdy: does this need to be fixed?

dims: i can help with this

dims: @thockin here's a fix for the test.py - https://github.com/dims/k8s.io/commit/e09f0cb1f71dd8dfbcc10e97f71dc128387480b1#diff-58c10258360cbb1470635b52e0887e93

spiffxp:
@thockin needs-rebase, if you're still interested in this

thockin:
Since the sigs redirect has been working, time to do this one?  I canaried it and it looks good.  I messed up and made this on my master branch, will open a new PR.

Thanks to http://nginx.viraptor.info/ for helping find the location-match issue with healthz

PR #102, opened 2018-01-23
Update gcsweb deployment to v1.0.6

ixdy:
The main change this gives us is https://github.com/kubernetes/test-infra/pull/5942, which serves the CSS directly, rather than relying on yahooapis.com.

/assign @thockin 
cc @BenTheElder

### DIFF
--- a/gcsweb.k8s.io/deployment.yaml
+++ b/gcsweb.k8s.io/deployment.yaml
@@ -20,7 +20,7 @@ spec:
       terminationGracePeriodSeconds: 30
       containers:
         - name: gcsweb
-          image: k8s.gcr.io/gcsweb-amd64:v1.0.4
+          image: k8s.gcr.io/gcsweb-amd64:v1.0.6
           args:
             - -b=crreleases
             - -b=istio-prow


BenTheElder:
/hold
I think this is supposed to be at a new URI now. staging.k8s.io or something like that..

ixdy:
no, `staging-k8s.gcr.io` is where you push, but `k8s.gcr.io` is from where you pull.

ixdy:
(also my branch was out of date. this is now only changing a `4` to a `5`.)

BenTheElder:
ah, thanks. FWIW that was also changed for a different image. https://github.com/kubernetes/test-infra/pull/6319
/hold cancel

ixdy:
hm, I deployed this, but http://gcsweb.k8s.io/styles/style.css is 404ing. probably gonna revert.

ixdy:
I found the issue; see https://github.com/kubernetes/test-infra/pull/6403.

stevekuznetsov:
Whoopsie, so I guess this should bump two releases then

ixdy:
yep, updated to v1.0.6 now.

ixdy:
http://gcsweb.k8s.io/styles/style.css

that's better!

PR #105, opened 2018-02-13
Add k8s.io/federation to k8s.io golang metadata

irfanurrehman:
Fixes https://github.com/kubernetes/federation/issues/230.


ixdy:
deployed.

PR #109, opened 2018-03-22
Add k8s.io URL for testing_frameworks repo

totherme:
The testing_frameworks repo currently has no canonical k8s.io URL. To
reduce the pain of potential migrations if the repo moves in the future,
we would like a k8s.io URL to be the canonical import path.

This might eventually end up being covered in the general case by #101 , but there has been no movement on that PR for over a month and in that time the testing_frameworks repo has already moved once.

CCing @timothysc just in case there's anything else we should do before merging this.

### DIFF
--- a/k8s.io/configmap-www-golang.yaml
+++ b/k8s.io/configmap-www-golang.yaml
@@ -142,6 +142,11 @@ data:
       <meta name="go-import" content="k8s.io/test-infra git https://github.com/kubernetes/test-infra">
       <meta name="go-source" content="k8s.io/test-infra     https://github.com/kubernetes/test-infra https://github.com/kubernetes/test-infra/tree/master{/dir} https://github.com/kubernetes/test-infra/blob/master{/dir}/{file}#L{line}">
     </head></html>
+  testing_frameworks.html: |
+    <html><head>
+      <meta name="go-import" content="k8s.io/testing_frameworks git https://github.com/kubernetes-sigs/testing_frameworks">
+      <meta name="go-source" content="k8s.io/testing_frameworks     https://github.com/kubernetes-sigs/testing_frameworks https://github.com/kubernetes-sigs/testing_frameworks/tree/master{/dir} https://github.com/kubernetes-sigs/testing_frameworks/blob/master{/dir}/{file}#L{line}">
+    </head></html>
   metrics.html: |
     <html><head>
       <meta name="go-import" content="k8s.io/metrics git https://github.com/kubernetes/metrics">


hoegaarden:
Is there currently a mapping between a repo in `kubernetes-sigs` and the SIG it is attached to, except the github topics? If not, I am not sure how we can map from `sig-testing.k8s.io/some_package` to `github.com/kubernetes-sigs/some_package` without some sort of config file.
I could also imagine [the automatic mapping](https://github.com/kubernetes/k8s.io/pull/101/files#diff-72968a5008e3bd97c21262c9ba0a0753R63) only for `k8s.io` for all the core repos and having a special config file per `<sig>.k8s.io` subdomain (or one config file with the mappings for all other sigs/subdomains).

thockin:
The config files are tedious, which is why I want to nix them.

`sig,k8s,io/testing_frameworks` ->
github.com/kubernetes-sigs/testing_frameworks ?

hoegaarden:
I am happy with `sigs.k8s.io/xxx` -> `github.com/kubernetes-sigs/xxx`.

With the current repos in the kubernetes-sigs that would mean:

```
sigs.k8s.io/cluster-api              ->  github.com/kubernetes-sigs/cluster-api
sigs.k8s.io/testing_frameworks       ->  github.com/kubernetes-sigs/testing_frameworks
sigs.k8s.io/application              ->  github.com/kubernetes-sigs/application
sigs.k8s.io/poseidon                 ->  github.com/kubernetes-sigs/poseidon
sigs.k8s.io/kubebuilder              ->  github.com/kubernetes-sigs/kubebuilder
sigs.k8s.io/aws-encryption-provider  ->  github.com/kubernetes-sigs/aws-encryption-provider
```

However, the reason why we want a canonical URL actually is, that it makes it easier to move repos without breaking the import path. Without allowing (additional) config, the mapping approach in #101 wont help in this regard.

thockin:
If you move within github, it will auto-redirect for you.

If you move off of github, well, how likely do you think that really is?

hoegaarden:
Ok then! Is there anything we can help on to get #101 merged?
We can, as soon as #101 is out the door, PR a follow up -- based on the idea in https://github.com/kubernetes/k8s.io/pull/101#issuecomment-360095519.

thockin:
I could use some help getting the test case done in #101.  My python is rusty and I am buried in PRs.  If we fix the testcase, we could trial the idea with a whole new vanity domain for sigs, before the main domain.

Volunteers?  Seems like it should be easy - take a template, sub out a var for the correct input string, iterate for 8 random inputs.

dims:
@thockin ack. i can take care of that

dims:
oops. assigned myself the wrong one :)

/unassign @dims

dims:
@thockin left a link to my patch for test.py in #101 PTAL

totherme:
Looks like this might be sorted out as part of #114 ?

krousey:
@totherme Yep. We're just waiting on the certificate to update to include sigs.k8s.io in the SAN. You could use go get -insecure for now.

totherme:
Looks like this is all sorted. Thanks!

Issue #110, opened 2018-03-29
Tests fail with SSL verification error


hoegaarden:
The tests fail with the following error

> ssl.SSLError: [SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed (_ssl.c:777)

(both `make test` and `make docker-test`)

As far as we understand, this is because nginx uses a self signed certificate and the https client doesn't have this self signed certificate in its certificate store.

PR #111, opened 2018-03-29
Disable SSL certificate verification

totherme:
This tests' purpose is not to test the certificates in use. Therefore we
can disable the certificate verification so that tests are easier
(possible? See issue #110) to run locally with self-signed certificates.

The comment
https://github.com/kubernetes/k8s.io/pull/19#issuecomment-265556199
suggests that when these tests were first written, certifiacte
verification was disabled by default. So we're confident that the intent
of the tests doesn't rely on working certificate verification.

Closes: #110 

CC: @hoegaarden


ixdy:
I do use this to verify the certificates on the production instance of k8s.io. Adding an option to disable the checks would be fine with me, though.

krousey:
Close in favor of #115 which has been merged?

PR #112, opened 2018-04-02
Add go.k8s.io/github-labels redirect

spiffxp:
It's an easier URI to remember than https://github.com/kubernetes/test-infra/blob/master/label_sync/labels.md and I would like to point people to it

ref: kubernetes/test-infra#7349
ref: kubernetes/kubernetes#34255


ixdy:
maybe add to the list in `k8s.io/README.md` too?

spiffxp:
oh right, you mean that comment that says "Please update README.md" that I completely overlooked? 🤦‍♂️

ixdy:
deployed

PR #114, opened 2018-04-03
Redirect sigs.k8s.io to github.com/kubernetes-sigs

krousey:
This provides vanity urls for the kubernetes-sigs repos that function as a "go get" redirect, a repo redirect, and a file redirect.

Based this off of #101. I've tested the cases I implemented with `curl --insecure --header "Host: sigs.k8s.io" <urls>`. I wrote python tests, but I keep getting SSL errors on a lot of them. Open to advice on how best to run the tests.

cc @rmmh @thockin @ixdy

### DIFF
--- a/k8s.io/configmap-nginx.yaml
+++ b/k8s.io/configmap-nginx.yaml
@@ -60,6 +60,51 @@ data:
           return 301 https://kubernetes.io$request_uri;
         }
       }
+
+      # Vanity redirects for the new kubernetes-sigs repos
+      #
+      # Upgrade HTTP to HTTPS.
+      server {
+        server_name sigs.k8s.io;
+        listen 80;
+
+        location / {
+          return 301 https://$host$request_uri;
+        }
+      }
+      # Handle HTTPS
+      server {
+        server_name sigs.k8s.io;
+        listen 443 ssl;
+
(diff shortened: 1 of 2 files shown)


dims:
@krousey see the `PYTHONHTTPSVERIFY=0` i used in my attempt to avoid the SSL error

krousey:
@dims locally patching in the change from #111 allowed me to run them.

Also, comparing HTML is a pain with spacing. So I just stubbed a basic HTML parser to only care about the go meta tags.

thockin:
merging

thockin:
@ixdy - you have been managing certs - can we add sigs.k8s.io and sigs.kubernetes.io ?

thockin:
canaried, tests pass with #111 manually patched, deployed to prod, tests pass.

Thanks @krousey - if we like this, we can flip the main URL too.

ixdy:
I learned this weekend with kates.io that Let's Encrypt supports wildcard certificates now, so I can replace the k8s.io one with a wildcard cert. 

(Still need to work on getting automation around this, though.)

ixdy:
done; not yet using wildcard certs because the stable version of lego hasn't been updated to ACMEv2 yet.

PR #115, opened 2018-04-03
Disable SSL verification is TARGET_IP is specified

krousey:
This allows us to still run SSL verification against the production
endpoint, but disables it for all dev and canary tests.

This is a conditional version of #111 

cc @thockin


thockin:
LGTM

PR #116, opened 2018-04-05
Promote new search engine for kubernetes main github repos

dims:
Etsy/Hound based Search engine maintained by @dims

### DIFF
--- a/k8s.io/configmap-nginx.yaml
+++ b/k8s.io/configmap-nginx.yaml
@@ -123,6 +123,13 @@ data:
         rewrite ^/(.*)?$    https://github.com/kubernetes/kubernetes/tree/master/$1 redirect;
       }
       server {
+        server_name cs.kubernetes.io cs.k8s.io;
+        listen 80;
+        listen 443 ssl;
+
+        rewrite ^/(.*)?$    https://k8s-code.appspot.com/?q=$1 redirect;
+      }
+      server {
         server_name dl.k8s.io dl.kubernetes.io;
         listen 80;
         listen 443 ssl;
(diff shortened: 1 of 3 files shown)


thockin:
blocks on #96 ?

dims:
@thockin not really since this just lists the existing instance on GAE. so we can merge this as-is.

ixdy:
looking again at #96, do we want to support `codesearch.` as well as `cs.`?

dims:
@ixdy i'll rework #96 to use `cs` later. let's go with `cs.` (and not add `codesearch.`)

dims:
@ixdy i've closed #96 and am maintaining the code here - https://github.com/dims/k8s-code.appspot.com

So @thockin @ixdy i'll leave it upto you all on whether this PR should be merged as-is.

Thanks,
Dims

k8s.io/configmap-nginx.yaml
```diff
@@ -122,6 +122,13 @@ data:
 
         rewrite ^/(.*)?$    https://github.com/kubernetes/kubernetes/tree/master/$1 redirect;
       }
+      server {
+        server_name cs.kubernetes.io cs.k8s.io;
+        listen 80;
+        listen 443 ssl;
+
+        rewrite ^/$       https://k8s-code.appspot.com redirect;
```
ixdy: would it be worth passing along any the rest of the request URI as an argument to the search?

e.g. translate https://cs.k8s.io/foo -> https://k8s-code.appspot.com/?q=foo ?

dims: Done! thanks @ixdy

ixdy:
/lgtm

canary tests pass

ixdy:
[https://cs.k8s.io/it%20works!](https://cs.k8s.io/it%20works!)

dims:
Thanks a ton @ixdy !

PR #118, opened 2018-05-08
Add config for discuss.k8s.io prototype

castrojo:
See https://github.com/kubernetes/community/blob/master/keps/sig-contributor-experience/0007-20180403-community-forum.md for more information.

### DIFF
--- a/k8s.io/configmap-nginx.yaml
+++ b/k8s.io/configmap-nginx.yaml
@@ -190,10 +190,12 @@ data:
       server {
         server_name discuss.kubernetes.io discuss.k8s.io;
         listen 80;
+        return 301 https://$server_name$request_uri;
+      }
+      server {
+        server_name discuss.kubernetes.io discuss.k8s.io;
         listen 443 ssl;
-
         rewrite ^/(.*)?$    https://kubernetes.hosted-by-discourse.com/$1 redirect;
-        }
       }
       server {
         server_name docs.k8s.io docs.kubernetes.io;
--- a/k8s.io/configmap-nginx.yaml
+++ b/k8s.io/configmap-nginx.yaml
@@ -188,6 +188,14 @@ data:
         }
       }
       server {
+        server_name discuss.kubernetes.io discuss.k8s.io;
+        listen 80;
+        listen 443 ssl;
+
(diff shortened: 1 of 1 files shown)


thockin:
Do we still need this, now that we have DNS in place?

k8s.io/configmap-nginx.yaml
```diff
@@ -187,6 +187,14 @@ data:
           rewrite ^/(.*)$                  https://storage.googleapis.com/kubernetes-release/$1 redirect;
         }
       }
+      server {
+        server_name discuss.kubernetes.io discuss.k8s.io;
+        listen 80;
+        listen 443 ssl;
```
zparnold: In order for SSL to work in nginx without browsers complaining you'll need to specify where the cert files are

zparnold: Something like:
```
listen   443;

ssl    on;
ssl_certificate    /etc/ssl/your_domain_name.pem; (or bundle.crt)
ssl_certificate_key    /etc/ssl/your_domain_name.key;
```

zparnold: Ignore me. I didn't read further down the file. 😄 ssl is configured elsewhere similarly

zparnold:
@thockin I would argue that we do, so we have control over upgrading the connection to SSL before forwarding to discourse. But additionally we could simply ask the lovely folks at discourse whether or not there is a way to configure SSL upgrades on their side to retain the DNS masking

ixdy:
I think what @thockin is trying to say (and I agree) is that none of this will have any effect, since discuss.k8s.io (and discuss.kubernetes.io) resolve to kubernetes.hosted-by-discourse.com, not the nginx server we run.

ixdy:
It looks like it might be possible to configure discourse to force https, though: https://docs.discourse.org/#tag/Site-Settings-Security

castrojo:
OK there doesn't appear to be a setting that's exposed to me in the UI, so I'll ask them to see if they can force https on their end. Thanks for the feedback everyone!

thockin:
If they can't, we can intermediate but it's better if we don't need to.

castrojo:
They've enabled forced https for us, so we're good to go!

PR #119, opened 2018-05-10
Make go-get content be automatic

thockin:
Replaces #101 

This obsoletes the explicit list of go-get redirects which were all identity-mapped anyway.  As a result `go get k8s.io/anything` will always pull `github.com/kubernetes/anything`.

Since the sigs redirect has been working, time to do this one, I think. I canaried it and it looks good. 

Thanks to http://nginx.viraptor.info/ for helping find the location-match issue with healthz


thockin:
deployed, test passes

PR #120, opened 2018-05-15
Remove redirection for code search

dims:
Instead of redirect to GAE, we will use a DNS entry

/assign @thockin


thockin:
You know what would be great is a a declarative GCE DNS config system.  Check config into git, a bot applies it to DNS.  Then I could do this whole transition by PR :)

Oh well, later.

PR #123, opened 2018-05-25
Update the go.k8s.io/cluster-api redirect

krousey:
We moved the repo a while ago. This should update as well.

/cc @roberthbailey


roberthbailey:
Thanks Kris, this was on my todo list but I hadn't gotten to it. :)

/lgtm

PR #128, opened 2018-07-19
Remove specific redirection for cluster-api

spangenberg:
Cluster API exists under kubernetes-sigs and follows the default
redirection rules. This redirection rule is not needed anymore.

Fixes https://github.com/kubernetes-sigs/cluster-api/issues/118

### DIFF
--- a/k8s.io/configmap-nginx.yaml
+++ b/k8s.io/configmap-nginx.yaml
@@ -82,7 +82,8 @@ data:
         server_name sigs.k8s.io;
         listen 443 ssl;
 
-        location ~ ^/(?<sig_repo>[^/]*)(?<repo_subpath>/.*)?$ {
+        # The ?! block is negative-lookahead to prevent `/repo/` from grouping into (`repo`, `/`) while `/repo/path` will still group as (`repo`, `/path`).
+        location ~ ^/(?<sig_repo>.*?)(?!/+$)(?<repo_subpath>/.*)?$ {
           if ($arg_go-get = "1") {
             # This is a go-get operation.
             return 200 '
--- a/k8s.io/configmap-nginx.yaml
+++ b/k8s.io/configmap-nginx.yaml
@@ -258,7 +258,6 @@ data:
         location / {
           rewrite ^/bounty$          https://github.com/kubernetes/kubernetes.github.io/issues?q=is%3Aopen+is%3Aissue+label%3ABounty redirect;
           rewrite ^/bot-commands$    https://prow.k8s.io/command-help.html redirect;
-          rewrite ^/cluster-api$     https://github.com/kubernetes-sigs/cluster-api/blob/master/README.md redirect;
           rewrite ^/github-labels$   https://github.com/kubernetes/test-infra/blob/master/label_sync/labels.md redirect;
           rewrite ^/help-wanted$     https://github.com/kubernetes/kubernetes/labels/help%20wanted redirect;
           rewrite ^/needs-ok-to-test$ https://github.com/pulls?utf8=%E2%9C%93&q=is%3Aopen+is%3Apr+user%3Akubernetes+label%3Aneeds-ok-to-test+-label%3Aneeds-rebase redirect;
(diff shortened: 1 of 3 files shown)


k8s.io/configmap-nginx.yaml
```diff
@@ -258,7 +258,6 @@ data:
         location / {
           rewrite ^/bounty$          https://github.com/kubernetes/kubernetes.github.io/issues?q=is%3Aopen+is%3Aissue+label%3ABounty redirect;
           rewrite ^/bot-commands$    https://prow.k8s.io/command-help.html redirect;
-          rewrite ^/cluster-api$     https://github.com/kubernetes-sigs/cluster-api/blob/master/README.md redirect;
```
ixdy: this only affects https://go.k8s.io/cluster-api, so it won't fix the original issue.

the section to fix so that https://sigs.k8s.io/cluster-api/ does the right thing would be 
https://github.com/kubernetes/k8s.io/blob/04adf1d3ff64805e23dafbf80898b5c60d050b91/k8s.io/configmap-nginx.yaml#L80-L111

ixdy: it'd also be good to extend test.py with a test case for this.

spangenberg:
Here also the regex changes visualised and how it reacts to different inputs.
https://regex101.com/r/8LFNx3/1

k8s.io/configmap-nginx.yaml
thockin: Is this PR intending to both fix the go-get and the go.k8s.io URL?  In this case "go" doesn't mean "golang", just "go to".  I don't think we want this part - it's sort of a breaking "API" change.

k8s.io/configmap-nginx.yaml
```diff
@@ -82,7 +82,7 @@ data:
         server_name sigs.k8s.io;
         listen 443 ssl;
 
-        location ~ ^/(?<sig_repo>[^/]*)(?<repo_subpath>/.*)?$ {
+        location ~ ^/(?<sig_repo>.*?)(?!/+$)(?<repo_subpath>/.*)?$ {
```
thockin: This is a complicated enough regex that I don't want it.  Can you pklease explain to me what it is doing and why it is needed?

spangenberg: Sure, I was also hesitant first to touch it.

thockin: I don't know what `(?!/+$) ` represents.

Can you show me a match that was wrong before and is right after this change?

spangenberg: Sure, thats the current implementation:
https://regex101.com/r/jRLEd8/1/

You see that the first match, matches the group `repo_subpath` to `/`, which fails the redirect.

The change I propose:
https://regex101.com/r/8LFNx3/1

This group `repo_subpath` doesn't match at all, only if there is anything behind it.

`(?!/+$)` drops any '/' if there is nothing else until the end of line.

thockin: I feel like a dunce,.  Thank you for trying to educate me.  I see now that the old form forces the / to be in group which would be wrong considering how it is used.

A more obvious (to me, not a regex wizard) might be: `^\/(?<sig_repo>[^\/]*)(\/(?<repo_subpath>.*)?)$`

This sets up a middle capture group (unnamed) for the / between repo and path, changing the usage below to require adding a / (which actually makes it more obvious to my eyes). 

We could make it a non-capture group, but I don't know that it is warranted - less is more?

You know what?  After staring at yours, I get it now.  OK.  Can you just add a comment like:

```
# The ?! block is negative-lookahead to prevent `/repo/` from grouping into (`repo`, `/`) while `/repo/path` will still group as (`repo`, `/path`).
```

One day I will have to come back and look at this :)

spangenberg: Perfect, sounds good to me since it's not so obvious.
I added the comment, guess thats a good first approach until we maybe come up with a better solution.

thockin:
running canary now.

thockin:
Deployed

PR #137, opened 2018-09-23
Document blog.k8s.io and sigs.k8s.io

nikhita:
This PR has the following changes:

- ~~https://changelog.k8s.io currently leads to a 404 github page. This PR redirects it to https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG.md.~~

- Add https://blog.k8s.io and https://sigs.k8s.io in the README. Should we also have blog.kubernetes.io and sigs.kubernetes.io?

/cc @BenTheElder @dims @cblecker 
/assign @ixdy

### DIFF
--- a/k8s.io/README.md
+++ b/k8s.io/README.md
@@ -9,6 +9,7 @@ Vanity URL(s)
 |  | k8s.io | kubernetes.io |
 | --- | --- | --- |
 | APT downloads| https://apt.k8s.io | https://apt.kubernetes.io |
+| Blog | https://blog.k8s.io | |
 | Changelog | https://changelog.k8s.io | https://changelog.kubernetes.io |
 | CI logs | https://ci-test.k8s.io | https://ci-test.kubernetes.io |
 | Git repo | https://code.k8s.io | https://code.kubernetes.io |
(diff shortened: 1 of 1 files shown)


k8s.io/README.md
```diff
@@ -25,6 +26,7 @@ Vanity URL(s)
 | PR Dashboard | https://pr-test.k8s.io | https://pr-test.kubernetes.io |
 | Pull requests | https://pr.k8s.io <br> https://prs.k8s.io | https://pr.kubernetes.io <br> https://prs.kubernetes.io |
 | Downloads | https://releases.k8s.io <br> https://rel.k8s.io | https://releases.kubernetes.io <br> https://rel.kubernetes.io |
+| Kubernetes SIGs | https://sigs.k8s.io | |
```
BenTheElder: I think we might also be lacking docs about the go get handling for k8s.io / sigs.k8s.io

nikhita: +1, I was working on a repo in kubernetes-sigs recently and found out that the import path is `sigs.k8s.io` by looking at the list of imports in a `.go` file. :/

k8s.io/configmap-nginx.yaml
```diff
@@ -144,7 +144,7 @@ data:
         listen 80;
         listen 443 ssl;
 
-        rewrite ^/(.*)?$    https://github.com/kubernetes/kubernetes/releases/tag/$1 redirect;
+        rewrite ^/(.*)?$    https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG.md redirect;
```
ixdy: I don't think this is quite right... something like https://changelog.k8s.io/v1.11.2 now doesn't redirect as usefully.

I agree that with an empty request URI this is better. maybe have multiple cases?

cblecker: You could do both though, as @ixdy is suggesting. If you have the rewrite case of `^/$` then if there is no path, it will redirect to a different location. Just put that ahead of the per-version rewrite rule

k8s.io/README.md
```diff
@@ -9,6 +9,7 @@ Vanity URL(s)
 |  | k8s.io | kubernetes.io |
 | --- | --- | --- |
 | APT downloads| https://apt.k8s.io | https://apt.kubernetes.io |
+| Blog | https://blog.k8s.io | |
```
ixdy: I'm not sure why this table has both `k8s.io` and `kubernetes.io` for each, but we should probably include `blog.kubernetes.io` and `sigs.kubernetes.io` as well.

ixdy:
oops

nikhita:
> oops

merging is so fast these days :fire: :upside_down_face: 

Sorry for the delay here, I'll address the comments in a separate PR. :+1:

PR #135, opened 2018-09-21
Add go.k8s.io/good-first-issue redirect

nikhita:
This would be helpful to point for new contributors.

/cc @dims @justaugustus @BenTheElder @cblecker 
/assign @ixdy

PR #142, opened 2018-10-06
Add Travis

nikhita:
So that we can catch test failures early: https://github.com/kubernetes/k8s.io/pull/135#issuecomment-424085461

Created https://github.com/kubernetes/test-infra/pull/9731 to make sure that required contexts are enforced.

/hold
an admin needs to enable travis for this repo

/cc @ixdy @cblecker @spiffxp @dims @BenTheElder 
/assign @ixdy

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,5 @@
+language: python
+python:
+  - "2.7"
+install: pip install pyyaml
+script: make -C k8s.io all
--- a/README.md
+++ b/README.md
@@ -1,2 +1,8 @@
 # k8s.io
+
+[![BuildStatus Widget]][BuildStatus Result]
+
 Kubernetes files for various sites and infrastructure.
+
+[BuildStatus Result]: https://travis-ci.org/kubernetes/k8s.io
+[BuildStatus Widget]: https://travis-ci.org/kubernetes/k8s.io.svg?branch=master


BenTheElder:
Why not Prow?

cblecker:
+1 to prow instead of travis

.travis.yml
```diff
@@ -0,0 +1,5 @@
+language: python
+python:
+  - "2.7"
+install: pip install pyyaml
```
nikhita: > My suggestion would be looking at the community verify job and the gcloud-in-go image. It has python, so it should work.

@cblecker thought so too, but it doesn't look like pyyaml is included in it? Will it still work then? :thinking: 

https://github.com/kubernetes/test-infra/blob/master/images/gcloud/Dockerfile

BenTheElder:
That works but actually I'd kinda like to see new jobs either:
- use a standard / third party / self maintained image like `golang` + the pod utilities and not require maintenance from test-infra
- use kubekins-e2e, which despite the name is really just a standard base for all the tools we need for CI for a given kubernetes version, and requires and receives lots of active maintenance anyhow

BenTheElder:
Sorry for the delay, better docs here: https://github.com/kubernetes/test-infra/blob/master/prow/pod-utilities.md

.travis.yml
```diff
@@ -0,0 +1,5 @@
...
+script: make -C k8s.io all
```
ixdy: by itself, this won't do much - it'll run `make test`, but that just runs `test.py` against the production `k8s.io` site,

for this to be a meaningful presubmit, we'd probably need to deploy the configs to a test cluster and then run `test.py` against that.

fejta:
Nikita you can repo what the prow job will do locally with something like this:

`docker run --rm=true -v $PWD:/whatever -w /whatever IMAGE ./script-in-repo.sh`

This will be equivalent to a prowjob config of:
```yaml
presubmits:
  kubernetes/k8s.io:
  - name: my-fancy-presubmit
     always_run: true
     decorate: true
     spec:
       containers:
         image: IMAGE
         args: ./script-in-repo.sh
```

For image I would choose one of the tags here: https://hub.docker.com/_/python/ . aka `python:2-jessie`

For the script I would expect it to have all the steps you have in the travis.yaml, aka:

```bash
set -o errexit
pip install pyyaml
make -C k8s.io all
```

This will do the exact same thing as travis, but will integrate with the rest of our tooling (testgrid, triage dashboard, etc).

fejta:
IMO:

> use a standard / third party / self maintained image like golang + the pod utilities and not require maintenance from test-infra
👍 
> use kubekins-e2e, which despite the name is really just a standard base for all the tools we need for CI for a given kubernetes version, and requires and receives lots of active maintenance anyhow
👎

fejta:
Can we close this? I think you now have a prowjob

nikhita:
Yeah, let's close this.

PR #144, opened 2018-10-10
[WIP] Automation/Script Updates

hh:

PR #145, opened 2018-10-10
[WIP] Dns.org / ideation on some literate programming / automation

hh:
This should differ from #144 only by the addition of one or more [org-mode](https://orgmode.org/) as I ideate with some form of literate CI / Automation.

Human readable Artifacts (txt, markdown, etc) to be generated during the CI run.

Ie [AXFR](https://en.wikipedia.org/wiki/DNS_zone_transfer) diffs between production and canary to be verified before the merge / push to production.


thockin:
I have to admit, I am not super sold on the need to anything more than text/markdown for this.  At least not yet.  Keeping an open mind, but I really am disinclined to want to learn another meta-language

PR #146, opened 2018-10-22
Remove old DNS records, add comments

thockin:


thockin:
Manual merge on the presumption of LGTM

PR #147, opened 2018-10-22
DNS: add canary

thockin:
Add canary-push capability and a glimmer of a test.  Need help with the test - anyone have time?

@kubernetes/k8s-infra-team

### DIFF
--- a/dns/test.sh
+++ b/dns/test.sh
@@ -1,0 +1,248 @@
+#!/bin/sh
+#
+# Test for Kubernetes DNS.
+#
+# $1: "canary" or "prod"
+
+if [ "$1" != "canary" -a "$1" != "prod" ]; then
+    echo "usage: $0 <canary|prod>"
+    exit 1
+fi
+
+SERVER_CANARY_K8S=ns-cloud-c1.googledomains.com.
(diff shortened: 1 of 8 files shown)


thockin:
Waiting for approval.  @hh I think you said "merge and iterate" as an approach, so it's in your court :)

dns/test.sh
```diff
@@ -0,0 +1,248 @@
...
+    echo "  got:"
+    echo "$R"
+    echo "--------------------"
+    exit 2
+fi
+
+echo "PASS"
+
+exit 0
+### Input data for k8s.io below...
```
hh: @thockin I'm unsure how this test data below is accessed.

hh: I'm leaning towards loading the zone-configs directly for the verification.

hh:
I'm thinking a python script that iterates over zone-configs and verifies all entries with http://www.dnspython.org/.

thockin:
@hh Python or Go is fine to me - the Go test doesn't seem that hard to write, but my Go is better than my Python.  Can we merge this and get the test in as followup?

thockin:
Oh, this again.

@hh approved

PR #148, opened 2018-10-22
Use custom subdomains for gubernator and testgrid

ixdy:
These are DNS changes only, already deployed.

https://gubernator.k8s.io and https://testgrid.k8s.io seem to work, though the gubernator PR dashboard isn't working due to some configuration issues with GitHub OAuth which @rmmh is looking into.

Once that's fixed, I'll update the redirectors on k8s.io as well.

/assign @thockin


dims:
LGTM 👍

thockin:
#
/lgtm
/approve

PR #149, opened 2018-10-30
Bump Python image + requests version

cblecker:
ref [CVE-2018-18074](https://nvd.nist.gov/vuln/detail/CVE-2018-18074)


BenTheElder:
I think this probably doesn't affect us FWIW (?)

cblecker:
Got a GitHub vuln alert, so I just figured I'd take care of it.

BenTheElder:
Thanks, mostly was thinking if we needed to patch this in more places
quickly or not, I don't think we're sending http auth headers in many
places, and not with python where I know we are. Might want to do
test-infra too still though...

PR #155, opened 2018-11-28
Add DNS OWNERS

thockin:
This is the same list as the dns-admins google group, but I don't know how to sync those (new infra project needed :)


justinsb:
/lgtm

If we ever do auto-promotion we might need a double-approval policy or similar, but looks like a good place to start

Issue #160, opened 2018-12-09
https broken on docs.k8s.io and docs.kubernetes.io redirectors


ixdy:
We're currently using DNS challenges with Let's Encrypt, which basically means we create a TXT record at `_acme-challenge.SUBDOMAIN.DOMAIN.`, e.g. `_acme-challenge.docs.kubernetes.io.` for `docs.kubernetes.io`.

Unfortunately, we recently added wildcard CNAME records for `*.docs.kubernetes.io` and `*.docs.k8s.io`, which means that these TXT records aren't resolving and we can't verify these domains.

The certificate is expiring, and I don't have a great solution at the moment. We might need to use another challenge method, or we should probably investigate switching to a wildcart cert, now that Let's Encrypt supports those.

ixdy:
I think I know how to fix this - if we create dummy A records for `_acme-challenge.docs.kubernetes.io` and `_acme-challenge.docs.k8s.io`, that should prevent the wildcard from taking effect for the TXT records.

ixdy:
/close

New cert has been deployed.

Issue #38, opened 2016-12-07
Put gcsweb behind ingress and install a TLS cert for it


ixdy:
Should be pretty straightforward. It seems like using ingress has the best Let's Encrypt automation support.

ixdy:
The submit-queue has been using kube-lego for a while, and it seems to be working fine, even though it's not for "production use".

It seems like the simplest option, and is probably easily extensible to the rest of the k8s.io services.

thockin:
ACK

ixdy:
kube-lego is deprecated; cert-manager is the endorsed replacement.

munnerz:
If someone wants to point me to where the definitions for gcsweb live, and if this is still something that needs doing, I'd be happy to open a PR!

munnerz:
d'oh. just spotted `/gcsweb.k8s.io`...

Presumably the cluster that's being deployed into is not already running cert-manager, and presumably we want to use GCLB ingress (i.e. `ingress-gce`) for this? 😄

thockin:
James,

Presumably true, so far. :)

ixdy:
cc @hh @justinsb 

This is probably a good first step - if we get TLS automation working for gcsweb, then the k8s.io redirector site should be an easy extension from there.

ixdy:
Neither gcsweb nor the k8s.io redirector is running behind ingress, so that's the 0th step.

After that we can set up TLS automation.

thockin:
GCSweb should be easy since it has only one hostname.  The redirector is
the fugly one.

ixdy:
We should also put slack.k8s.io behind TLS.

thockin:
Also true.

thockin:
https://github.com/kubernetes/k8s.io/pull/180 - put gcsweb behind ingress

thockin:
As far as I can tell the ingress part of gcsweb is done.  Simple recipe to follow for slack (when ready).

Now for the TLS part :)

ixdy:
I played around with cert-manager a bit this evening, got it working on my own deployment: https://gcsweb.ixdyland.net/gcs/kubernetes-jenkins/

Should be pretty easy to deploy this for the "real" gcsweb as well as the k8s.io redirector.

Only issue with putting the k8s.io redirector behind a GCLB Ingress is that I'm not sure we can do http -> https upgrades.

ixdy:
[...]

PR #162, opened 2019-01-10
Create dummy A records for _acme-challenge.docs

ixdy:
This prevents resolvers from following the wildcard CNAME for *.docs
when trying to resolve the _acme-challenge.docs TXT record.

This should partially address https://github.com/kubernetes/k8s.io/issues/160; I still need to update the DNS records and then re-request a cert.

/assign @thockin

### DIFF
--- a/dns/zone-configs/k8s.io.yaml
+++ b/dns/zone-configs/k8s.io.yaml
@@ -46,6 +46,11 @@ docs:
 '*.docs':
   type: CNAME
   value: netlifyglobalcdn.com.
+# Create a dummy A record so that the Let's Encrypt TXT records are not caught
+# by the wildcard CNAME record. (@ixdy)
+_acme_challenge.docs:
+  type: A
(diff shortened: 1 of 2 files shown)


thockin:
/lgtm
/approve

We'll have to discuss whether cert procurement should be allowed to mutate DNS directly or go through this PR process for short-lved records like the TXT records for this.

ixdy:
Deployed these changes manually, and working on refreshing the TLS cert to fix #160.

Eventually (hopefully sooner rather than later) I'd like to move to either https://github.com/GoogleCloudPlatform/gke-managed-certs (if/when available) or cert-manager. With cert-manager we may be able to use an HTTP challenge instead of a DNS challenge, which might resolve the DNS access concerns.

ixdy:
(x-ref https://github.com/kubernetes/k8s.io/issues/38; we should setup automation on all the things, not just gcsweb)

PR #165, opened 2019-01-14
Fix bad var name in check_zone.py

thockin:
This changed late in the dev cycle, but should have been caught
pre-merge.

Also add slightly better debug string and TODO to fix retry for new
records (propagation delay).


dims:
LGTM 👍 

will let @hh apply lgtm

thockin:
@ixdy PTAL?

ixdy:
oh I can't approve. you need to self-approve I guess?

PR #169, opened 2019-01-23
Update OWNERS and SECURITY_CONTACTS

spiffxp:
This is by no means a power grab nor a voluntolding.  I'm open to all
input, these are just initial suggestions in PR form.

Reasons for root:
- drop rmmh, not on this project
- drop lavalamp, reduce surface area
- push thockin, ixdy to subdirs
- add wg-k8s-infra chairs
- add sig-contribex tech leads (its their subproject)
- keep mikedanese as backup non-crazy-overloaded google contact

Reasons for subdirs:
- ixdy and thockin are the only ones with experience
  deploying these services, and also the only ones
  who can deploy because google
- goal is to de-google these dirs, and grow the pool

### DIFF
--- a/OWNERS
+++ b/OWNERS
@@ -1,6 +1,11 @@
+# See the OWNERS docs at https://go.k8s.io/owners
+
 approvers:
-  - ixdy
-  - lavalamp
-  - mikedanese
-  - rmmh
-  - thockin
+- cblecker # sig-contribex tl
+- dims # wg-k8s-infra chair
+- mikedanese
+- nikhita # sig-contribex tl
+- spiffxp # wg-k8s-infra chair
+
+labels:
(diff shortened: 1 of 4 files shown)


spiffxp:
Doh, I forgot to put on a `/hold` for comment

Will discuss at tomorrow's wg-k8s-infra meeting and ping folks out of band to see if they're OK with this.  Totally happy to revert as needed

spiffxp:
I heard back from @lavalamp that he'd prefer to stay in

spiffxp:
Discussed at meeting, no objections.  Pinged affected individuals in person, no objections (once we clarified this was for kubernetes/k8s.io and not kuberntes/kubernetes)

PR #174, opened 2019-01-30
MVP for artifactserver / redirector service

justinsb:
Trivial artifactserver / 302 service.  I figure this can be a useful test for deploying onto our test cluster.


justinsb:
We have a bit of a chicken & egg situation here for deploying this to the dev cluster.  I guess we should also figure out (ignoring poultry biology) whether we want to use the same promotion process everywhere, even on the infra cluster itself.  I'm guessing we do, and we'll just have to do some manual pushes to break the bootstrapping problem.

artifactserver/manifest.yaml
```diff
@@ -0,0 +1,66 @@
...
+  template:
+    metadata:
+      labels:
+        app: artifactserver
+    spec:
+      terminationGracePeriodSeconds: 30
+      containers:
+        - name: artifactserver
+          #image: k8s.gcr.io/infra/artifactserver:latest
+          image: justinsb/artifactserver:latest
```
justinsb: I'm currently breaking the deadlock this way...

justinsb:
cc @brendandburns

brendandburns:
/lgtm

@justinsb I went and found all of the various IP address ranges:

[AWS](https://docs.aws.amazon.com/general/latest/gr/aws-ip-ranges.html), [Azure](https://www.microsoft.com/en-us/download/details.aspx?id=56519) and [GCP](https://cloud.google.com/compute/docs/faq#find_ip_range)

dims:
/approve 

(since @brendandburns is happy!)

PR #178, opened 2019-02-06
Add current offline page content for slack.k8s.io.

coderanger:
And a placeholder for Slackin as we think more about that.

/cc @castrojo @cblecker


cblecker:
/lgtm
/approve

Adding current content. We can iterate from here.

PR #180, opened 2019-02-06
Put GCSWeb behind Ingress

thockin:
After this rolls out through DNS, we can turn down the old Service LB.

xref: #38


ixdy:
/lgtm

I'm not sure why I didn't do this sooner

thockin:
DNS change is pushed.  Will wait A few hours before decomissioning the old IP

thockin:
Old IP is turned down.

PR #181, opened 2019-02-06
Finish gcsweb service-> ingress

thockin:


thockin:
Why the hold?

BenTheElder:
holdover from test-infra, just to allow merging when the author is ready
since we're generally auto-deploying the contents
/hold cancel

PR #182, opened 2019-02-07
Use cert-manager with Let's Encrypt to enable TLS for gcsweb

### LINKED ISSUE #38: Put gcsweb behind ingress and install a TLS cert for it

ixdy:
The first commit is basically following the [cert-manager setup guide](https://cert-manager.readthedocs.io/en/latest/getting-started/install.html#installing-with-regular-manifests) and installing `ClusterIssuers` for the staging and prod Let's Encrypt endpoints.

The second commit then uses this setup to obtain a certificate for gcsweb and install it in the new Ingress.

I haven't deployed any of this to the real cluster, but I've tested a version of it on my own deployment and cluster: http://gcsweb.ixdyland.net/gcs/kubernetes-jenkins/

Fixes #38 
(and paves the way to using this for spartakus and the k8s.io redirectors)

/assign @thockin @hh 
cc @munnerz

### DIFF
--- a/cert-manager/README.md
+++ b/cert-manager/README.md
@@ -1,0 +1,16 @@
+To bootstrap (per the [cert-manager getting started
+guide](https://cert-manager.readthedocs.io/en/release-0.6/getting-started/install.html#installing-with-regular-manifests)):
+
+```
+kubectl create clusterrolebinding cluster-admin-binding --clusterrole=cluster-admin --user=$(gcloud config get-value core/account)
+kubectl apply -f cert-manager.yaml --validate=false  # --validate=false no longer needed with k8s 1.13+
+kubectl apply -f letsencrypt-staging.yaml
+kubectl apply -f letsencrypt-prod.yaml
+```
+
+This will set up cluster-wide webhooks and issuers, you can subsequently create
+`Certificate` resources in other namespaces without repeating these steps.
+
+Note that due to https://github.com/jetstack/cert-manager/issues/1343, you
+should not add the `tls` field to the `Ingress` object until after the first
+certificate has been issued.
--- a/gcsweb.k8s.io/ingress.yaml
+++ b/gcsweb.k8s.io/ingress.yaml
@@ -7,6 +7,8 @@ metadata:
   annotations:
     kubernetes.io/ingress.global-static-ip-name: gcsweb-k8s-io
 spec:
+  tls:
+  - secretName: gcsweb-k8s-io-tls
   backend:
     serviceName: gcsweb
     servicePort: http
--- a/cert-manager/OWNERS
+++ b/cert-manager/OWNERS
@@ -1,0 +1,3 @@
+approvers:
+  - ixdy
+  - munnerz
--- a/cert-manager/cert-manager.yaml
+++ b/cert-manager/cert-manager.yaml
@@ -1,0 +1,805 @@
+# Saved from https://raw.githubusercontent.com/jetstack/cert-manager/v0.6.1/deploy/manifests/cert-manager.yaml
+apiVersion: apiextensions.k8s.io/v1beta1
+kind: CustomResourceDefinition
+metadata:
+  name: certificates.certmanager.k8s.io
+  labels:
+    app: cert-manager
+spec:
+  additionalPrinterColumns:
+  - JSONPath: .status.conditions[?(@.type=="Ready")].status
+    name: Ready
+    type: string
+  - JSONPath: .spec.secretName
+    name: Secret
+    type: string
+  - JSONPath: .spec.issuerRef.name
+    name: Issuer
+    type: string
+    priority: 1
+  - JSONPath: .status.conditions[?(@.type=="Ready")].message
+    name: Status
+    type: string
+    priority: 1
+  - JSONPath: .metadata.creationTimestamp
+    description: |-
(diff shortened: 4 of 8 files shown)


ixdy:
To bootstrap cert-manager components (per the cert-manager docs):
```console
cd cert-manager/
kubectl apply -f 00-crds.yaml
kubectl create clusterrolebinding cluster-admin-binding --clusterrole=cluster-admin --user=$(gcloud config get-value core/account)
kubectl apply -f cert-manager.yaml --validate=false
kubectl apply -f letsencrypt-staging.yaml
kubectl apply -f letsencrypt-prod.yaml
```

Then create the `Certificate` and update the `Ingress` per normal:
```console
cd gcsweb.k8s.io/
kubectl apply -f certificate.yaml
kubectl apply -f ingress.yaml
```

Subsequent migrations (spartakus, redirector, etc) only need the latter steps (creating a `Certificate` and updating the `Ingress`).

munnerz:
Awesome to see, and this lgtm 😄 

One thing to note - we've identified an issue with certificate rotation in the webhook component, which could leave the webhook solving with old certificates when renewal times comes along. More details here: https://github.com/jetstack/cert-manager/issues/1340

The certificate duration here is configured to be 1y for the serving cert, and 5y for the CA used. Just a heads up to upgrade to v0.7 once that is released some time in the next year, or otherwise {somehow remember} to restart your webhook pod in ~8mths time (we renew 4m before expiry).

If you want to keep an eye on the expiry state of your certificates, we also expose expiry time prometheus metrics. You might want to set up prometheus/alertmanager/stackdriver or the like, to automatically alert you if a certificate is nearing expiry 😄

cert-manager/README.md
```diff
@@ -0,0 +1,12 @@
+To bootstrap (per the [cert-manager getting started
+guide](https://cert-manager.readthedocs.io/en/latest/getting-started/install.html#installing-with-regular-manifests)):
```
munnerz: nit: may want to pin `latest` to `release-0.6`?

cert-manager/README.md
```diff
@@ -0,0 +1,12 @@
...
+
+```
+kubectl create clusterrolebinding cluster-admin-binding --clusterrole=cluster-admin --user=$(gcloud config get-value core/account)
+kubectl apply -f cert-manager.yaml --validate=false
```
munnerz: nit: possibly add a note mentioning that `--validate=false` can go away with k8s 1.13+

gcsweb.k8s.io/ingress.yaml
```diff
@@ -7,6 +7,8 @@ metadata:
   annotations:
     kubernetes.io/ingress.global-static-ip-name: gcsweb-k8s-io
 spec:
+  tls:
+  - secretName: gcsweb-k8s-io-tls
```
munnerz: Due to an issue with the way cert-manager obtains certificates from LE (we generate and store a private key *before* the corresponding certificate has been obtained from Let's Encrypt) (more info: https://github.com/jetstack/cert-manager/issues/1343) - if you add this field here when no Secret resource already exists, the ingress-gce ingress controller will fail to actually update the GCP API with the load balancer configuration.

Because of this, you must instead create the Certificate resource and obtain the certificate *before* updating the ingress resource to reference that secret.

We'll be building a workaround into v0.7 which is due in ~1.5wks (most likely, pre-generating a self-signed certificate which can be used in the interim whilst CM is obtaining a publicly trusted certificate).

What this means, is that in order to deploy this *the first time*, you'll need to apply these changes *after* creating all the other resources in this PR 😄

cert-manager/README.md
ixdy: done

cert-manager/README.md
ixdy: done

gcsweb.k8s.io/ingress.yaml
ixdy: I think I discovered this, but I appreciate the clarification. I added a note to our README as well.

hh:
During our meeting today approval for this PR was requested. 
/cc @munnerz @thockin

spiffxp:
/approve
/hold
I don't have enough context to full review, @munnerz can you take a look?

For followup, I feel like we should have a cert-manager/OWNERS file so this doesn't have to hit root OWNERS

ixdy:
good point re: `OWNERS`, though I'm not sure who should be in it. :)

gcsweb.k8s.io/ingress.yaml
munnerz: I've got a fix in progress that will be landing in v0.7 here: https://github.com/jetstack/cert-manager/pull/1392 😄

munnerz:
Taking another look, this all looks 👍 to me (although sods law states ...)

I'll add my lgtm, but defer a `/hold cancel` to someone in a better position to manage the roll out of this change 😄 

> good point re: `OWNERS`, though I'm not sure who should be in it. :)

I'd be happy to put my name forward to be pinged for reviews etc. on this 😄

ixdy:
added `OWNERS` with me and @munnerz. any other volunteers? :)

thockin:
I am OK with this, but defer actual review to you folks.  I will approve but not remove he hold.  At your leisure.

/approve

ixdy:
this needs one more `/lgtm` (for the OWNERS file).

ixdy:
/hold cancel
:shipit:

Issue #184, opened 2019-02-11
DNS REQUEST: kind.sigs.k8s.io subproject docs


BenTheElder:
### Type of DNS update: Create
one of: Create, Delete, Update

### Domain being modified: k8s.io
e.g. `k8s.io`

### Existing DNS Record:

```yaml
NONE
```

### New DNS Record:

```yaml
# https://github.com/kubernetes-sigs/kind docs (@bentheelder, @munnerz)
kind.sigs:
  type: CNAME
  value: kindocs.netlify.com
```

### Reason for update:

The SIG-Testing [kind](https://github.com/kubernetes-sigs/kind) subproject has been working on subproject documentation using tooling aligned with SIG-Docs, currently at https://kindocs.netlify.com/ under a Kubernetes-owned Netlify team. 

Ideally we'd like to start using a provider independent URL, `kind.sigs.k8s.io` aligns with our current API namespace, in the future if this pattern works well other subprojects could potentially follow.

The record above follows https://www.netlify.com/docs/custom-domains/#manual-no-www

> If you are planning to set your site’s custom domain to a non-www subdomain, like blog.yourdomain.com, most of the above instructions won’t be as applicable to you. The things to make sure of are:
>
> 1. You should use a CNAME DNS record for that subdomain to point to your [your-site-name].netlify.com so you can use the CDN.
> 2. You don’t need a www.subdomain.yourdomain.com DNS entry.
> 3. You probably want to stick with your external DNS host. If you use Netlify DNS, you’ll be hosting yourdomain.com with us, not subdomain.yourdomain.com. Netlify DNS can only manage entire DNS zones for root domains.

cc @munnerz @chenopis

BenTheElder:
cc @dims @cblecker ref: https://github.com/kubernetes/k8s.io/blob/master/dns/OWNERS

I briefly tested that it is possible to do something like this with my own netlify backed domain (with external DNS) and a subdomain backed by the kind netlify.

From that test I believe this will work fine as long as someone who owns both the k8s.io netlify project and the kind project does this on the netlify end (I think IE @chenopis 🙏).

I wrote the PR to do this here: https://github.com/kubernetes/k8s.io/pull/185

cblecker:
Is this the right format for docs sites like this (vs. `kind.k8s.io` or something else).

thockin:
What @cblecker said.  I don't have context - are we going to create a netlify site and domain for every sig-owned thing that writes docs?

[...]

PR #185, opened 2019-02-11
add kind.sigs.k8s.io

BenTheElder:
ref: #184


BenTheElder:
/hold
discussion ongoing in the issue https://github.com/kubernetes/k8s.io/issues/184

thockin:
I will push, there's a bug in this so watch for a followup

Issue #158, opened 2018-12-05
[Umbrella Issue] Setup a GCR Repository for projects to use


dims:
Split from #153 (see that for some context)

cc @javier-b-perez @mkumatag @listx @thockin

dims:
Related to https://github.com/kubernetes/k8s.io/issues/157

javier-b-perez:
Do we have an agreement in the structure of the storage?

dims:
@javier-b-perez we need to come up with a proposal

javier-b-perez:
You mean a proposal for the structure, like in the [doc](https://docs.google.com/document/d/1cQUAU3AeRahL8Y4R6BDiMiMV44ESPDFNyZQIlgSA5eQ/edit#)?

Option 1)
We have a single staging area

`staging-k8s.gcr.io/<TYPE>/[<PROJECT>/]<IMAGE_NAME>:<TAG>`

TYPE

- kubernetes - core kubernetes images like pause, apiserver etc.
- kubernetes-sigs - Kubernetes SIG-related work
- tests - used for tests
- examples - example images used in the documentation

PROJECT: (optional) can be e.g. dashboard.
IMAGE_NAME: name of the container image

TAG: structure `<VERSION>-<OS>-<ARCH>`

- `<TAG>`: image version
- `<OS>`: the same as GOOS, e.g. linux or windows
- `<ARCH>`: the same as GOARCH, e.g. amd64 or ppc64le


Example
k8s.gcr.io/kubernetes/pause

- k8s.gcr.io/kubernetes/pause:3.1-linux-amd64
- k8s.gcr.io/kubernetes/pause:3.1-linux-arm
- k8s.gcr.io/kubernetes/pause:3.1-linux-ppc64le
- k8s.gcr.io/kubernetes/pause:3.1-linux-s390x

Option 2)
Tim [mentioned](https://github.com/kubernetes/community/pull/2695#discussion_r231327369) an option to have per-project staging areas

`gcr.io/<PROJECT>/[<TYPE>/]<IMAGE_NAME>:<TAG>`

Example
k8s.gcr.io/kubernetes/pause

- gcr.io/k8s-kubernetes-staging/pause:3.1-linux-amd64
...


Then we promote images to the official registry `k8s.gcr.io/<TYPE>/[<PROJECT>/]<IMAGE_NAME>:<TAG>`

marquiz:
[...]

PR #186, opened 2019-02-11
Script adding a GCR staging repo

thockin:


thockin:
As promised on Friday, the script.

dims:
Related to https://github.com/kubernetes/k8s.io/issues/158

gcr/add-staging-repo.sh
```diff
@@ -0,0 +1,107 @@
...
+color 6 "Configuring billing for ${PROJECT}"
+gcloud beta billing projects link "${PROJECT}" \
+    --billing-account "${BILLING}"
+
+# Grant project viewer so the UI will work.
+color 6 "Granting project viewer to ${ADMINS}"
+gcloud \
+    projects add-iam-policy-binding "${PROJECT}" \
+    --member "group:${ADMINS}" \
+    --role roles/viewer
```
javier-b-perez: do we just want admins as 'viewers'? what about 'owners'?

gcr/add-staging-repo.sh
```diff
@@ -0,0 +1,107 @@
...
+gsutil iam ch "group:${ADMINS}:legacyBucketOwner" "gs://${BUCKET}"
+
+# Make the bucket publicly readable.
+color 6 "Making the bucket public"
+gsutil iam ch allUsers:objectViewer "gs://${BUCKET}"
+
+# Grant repo writers access to write.
+color 6 "Granting bucket objectAdmin to ${WRITERS}"
+gsutil iam ch "group:${WRITERS}:objectAdmin" "gs://${BUCKET}"
+color 6 "Granting bucket legacyBucketReader to ${WRITERS}"
```
javier-b-perez: do we need 'legacyBucketReader'?
are we going to store GCS buckets as well or just GCR?

gcr/add-staging-repo.sh
thockin: They only need project viewer for the UI to work.  They have bucket admin for the specific buckets.

We're trying to keep untra-limited perms when possible.

gcr/add-staging-repo.sh
thockin: without buckets.list it didn't work, and there doesn't seem to be a role that includes that permission that isn't "legacy" (and I didn't want to make a custom role, though we could, I guess...)

gcr/add-staging-repo.sh
javier-b-perez: who is going to manage the projects? just who create them?

gcr/add-staging-repo.sh
javier-b-perez: you are right, the other alternative is 'roles/storage.admin'

javier-b-perez:
LGTM

as side note, we should also document somewhere who is going to own the groups like 'k8s-infra-gcr-staging-${REPO}@googlegroups.com', probably another google group?

gcr/add-staging-repo.sh
thockin: the set of people who can create projects is very small and should, itself be governed by a googlegroup

gcr/add-staging-repo.sh
thockin: which is too broad

dims:
/assign @justinsb 

Justin, can you please peek?

thockin:
I am working on a quick followup commit to age-out old data

thockin:
PTAL at last commit, too

gcr/add-staging-repo.sh
```diff
@@ -0,0 +1,135 @@
...
+
+# The group that can write to this staging repo.
+WRITERS="k8s-infra-gcr-staging-${REPO}@googlegroups.com"
+
+# The GCP org stuff needed to turn it all on.
+ORG="758905017065" # kubernetes.io
+BILLING="018801-93540E-22A20E"
+
+# Make the project, if needed
+if ! gcloud projects describe "${PROJECT}" >/dev/null 2>&1; then
```
justinsb: What happens if this project name happens to be an unrelated 3rd party?

gcr/add-staging-repo.sh
```diff
@@ -0,0 +1,135 @@
...
+gcloud --project "${PROJECT}" \
+    services enable containerregistry.googleapis.com
+
+# Push an image to trigger the bucket to be created
+color 6 "Activating the registry bucket"
+docker pull k8s.gcr.io/pause
+docker tag k8s.gcr.io/pause "gcr.io/${PROJECT}/pause"
+docker push "gcr.io/${PROJECT}/pause"
+gcloud --project "${PROJECT}" \
+    container images delete --quiet "gcr.io/${PROJECT}/pause:latest"
```
justinsb: Scary!  Worth tagging `gcr.io/${PROJECT}/ceci-nest-pas-une-image` or some other thing that isn't important?

justinsb:
Two paranoid suggestions, but lgtm

/lgtm

gcr/add-staging-repo.sh
thockin: You won't be able to list those.  I don't see a way to only list projects under a specific org - do you know of one?  Assuming you have a project with this name pattern in your scope, something else will fail.  Or if you have project admin, maybe it won't fail.

We could require this create to happen, which makes the scripts not re-runnable.  We could add flags to say --project-exists or something.  Is that worth the effort?

gcr/add-staging-repo.sh
thockin: hahah, sure.  In next push

thockin:
New push is up with image name change.  Other is unresolved. Launch and iterate?

dims:
thanks for adding the age-out. +1 to launch/iterate

/approve

leaving `/lgtm` for @justinsb

thockin:
We could check the project existence with:

gcloud projects describe k8s-staging-coredns --flatten='parent[]'
--format='csv[no-heading](type, id)' | grep ^organization, | cut -f2 -d,
and match that against the

thockin:
Pushed with that added.  @justinsb should be placated now...

thockin:
ping for lgtm

PR #192, opened 2019-02-20
Add slack.k8s.io existing config

thockin:
These are imperfect in many ways, but worth capturing.  Before we
convert this to new cluster it needs at least:

* convert to deployment
* token moved to a secret
* run as non-root
* appropriate request/limit settings
* probably put behind an ingress with TLS

### DIFF
--- a/slack.k8s.io/replicationcontroller.yaml
+++ b/slack.k8s.io/replicationcontroller.yaml
@@ -1,0 +1,32 @@
+apiVersion: v1
+kind: ReplicationController
+metadata:
+  name: slackin-controller
+  annotations:
+    original-image-before-attacks: gcr.io/kubernetes-tools/slackin-kubernetes
+  labels:
+    app: slackin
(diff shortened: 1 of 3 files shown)


dims:
/lgtm

let's get this in, we can iterate on it we have a fresh cluster.

slack.k8s.io/replicationcontroller.yaml
```diff
@@ -0,0 +1,32 @@
+apiVersion: v1
+kind: ReplicationController
```
justinsb: Hello RC my old friend

justinsb:
/approve

I can't figure out what the upstream source is.  It seems to have been built on `2017-09-01`, and includes ` "powered by ", dom("a href=http://rauchg.com/slackin target=_blank",` but I think it has socket.io version 1.3.5 which suggests a _much_ older snapshot of the code.

PR #197, opened 2019-02-20
Manifests for k8s-container-image-promoter prow jobs

dims:
Placeholders for manifests for use with https://github.com/GoogleCloudPlatform/k8s-container-image-promoter

Change-Id: Ic5a0eb9be30e5bd4682faa484436fb0d543235ce


thockin:
/approve

I'll let @listx check syntax

thockin:
Maybe worth including the googlegroup as a comment for each staging repo

dims:
Already done :) see the first line of the yaml

thockin:
As per zoom, let's move the script from the current "gcr" dir to this root in the same PR.

thockin:
I have almost totally rewritten the script.  Will wait until you git-move it before sending a PR

k8s.gcr.io/k8s-staging-csi/manifest.yaml
```diff
@@ -0,0 +1,9 @@
+# google group for gcr.io/k8s-staging-csi is k8s-infra-gcr-staging-csi@googlegroups.com
+registries:
+  src: gcr.io/k8s-staging-csi
+  dest: gcr.io/k8s-staging
+service-account: service-account@kubernetes-public.iam.gserviceaccount.com
+#images:
```
listx: Can you uncomment this line? The `images` key is part of a Go struct so the marshaling logic will break if it is missing IIRC.

k8s.gcr.io/k8s-staging-cluster-api/manifest.yaml
```diff
@@ -0,0 +1,9 @@
+# google group for gcr.io/k8s-staging-cluster-api is k8s-infra-gcr-staging-cluster-api@googlegroups.com
+registries:
+  src: gcr.io/k8s-staging-cluster-api
+  dest: gcr.io/k8s-staging
+service-account: service-account@kubernetes-public.iam.gserviceaccount.com
+images:
```
listx: You should do `images: []` to denote an empty list.

k8s.gcr.io/k8s-staging-csi/manifest.yaml
listx: Err, might as well make that `images: []`.

k8s.gcr.io/k8s-staging-cluster-api/manifest.yaml
dims: ack

k8s.gcr.io/k8s-staging-csi/manifest.yaml
dims: ack

thockin:
I am impatient and need to send a PR on top of this

PR #199, opened 2019-02-21
Add redirect for api review process doc

liggitt:
To be used in the label description

cc @spiffxp


spiffxp:
/lgtm
I leave the /approve to @ixdy or @thockin, who can actually deploy this

In theory I can too, but I need to learn how, and I'm about to turn into a distracted squirrel for a few hours.  If I come back and it's still here I'll give it a shot.

thockin:
will push momentarily

PR #191, opened 2019-02-20
Update with latest status

castrojo:
Signed-off-by: Jorge O. Castro <jorgec@vmware.com>

PR #201, opened 2019-02-22
Update slack.k8s.io image to v3 tag

spiffxp:
I manually updated it in cluster as part of https://github.com/kubernetes/k8s.io/pull/191#issuecomment-466504010

Before I noticed https://github.com/kubernetes/k8s.io/pull/192

Back in sync now


spiffxp:
/approve
self-approving

PR #202, opened 2019-02-22
Update GCR scripts

thockin:
Add a script to create prod repos and convert logic into a library to as
to DRY.


thockin:
Now with regional prod repos, rather than the single one.  That is a commit of its own, just in case we want to debate it.

The theory them is that promoter pulls from k8s-staging-foo/... and pushes to all of {us,eu,asia}.gcr.io/k8s-gcr-prod/...

Once @listx has promoter responding to github commits, we can talk about how to bulk import the trillions of existing images and flopping the global name to auto-regionalize.

listx:
I guess this is blocked on https://github.com/GoogleCloudPlatform/k8s-container-image-promoter/issues/13.

OTOH, we could just have multiple promoter manifests for now (one for each of {us,eu,asia}.gcr.io/k8s-gcr-prod), and then dedup them into 1 manifest file in the future once https://github.com/GoogleCloudPlatform/k8s-container-image-promoter/issues/13 lands sometime in the future.

Do the different regional repos have different service account credentials? I don't think so, right (they all just use the same one)? If so, then the prow job could dynamically generate the regional manifests (just replacing the `registries: dest: ...` key in the promoter manifest) and run it against those manifests on the fly; this is hacky but it would be simple to implement and relieve developers of the need to update all regional manifests if they were checked in statically as separate YAMLs.

I will mention this when I edit the prow config PR to include such a script. wdyt?

thockin:
Doing it on the fly seems hacky, but contained, so OK for testing.

Same svcacct privs for all 3 repos (same project)

Also, there's an open question in here about retention policy vs GC.  Your
thoughts appreciated.

k8s.gcr.io/lib.sh
```diff
@@ -0,0 +1,279 @@
...
+function ensure_prod_lifecycle() {
+    if [ $# -lt 1 -o $# -gt 2 -o -z "$1" ]; then
+        echo "ensure_prod_lifecycle(project, [region]) requires 1 or 2 arguments" > /dev/stderr
+        return 1
+    fi
+    project="$1"
+    region="${2:-}"
+
+    # Set lifecycle policies.
+    # TODO: do we want this?  It will inhibit promoter's GC, but will protect
```
listx: I don't think the retention policy is the right safeguard here. Right off the bat it would prevent us from performing reverts against mistakenly-promoted images (I've already seen this happen on more than one occasion for k8s.gcr.io).

For pruning/deleting old images, it would be better to have a script that periodically runs and looks for old images in the promoter manifest, and make it file a PR to delete that image from the manifest. It would make everything explicit.

On a related note, I need to add a feature of the promoter that "snapshots" a GCR into YAML (create a promoter manifest that captures the existing state).

thockin: Yeah, I am really on the fence.  

When we promote a bad image, what do we want to happen?  Do we really need to delete it, or just change the tag back?

I worry that this is a vector for mistakes that retention policy (especially if locked!!) could prevent.   E.g. maybe we don't need a graveyard, just a GC and very long minimum retention.  The behavior I really want (delayed delete) isn't part of GCS.  I think I'd rather pay the 3-year or 5-year cost of hosting something that was accidentally promoted if it means I know that real images can't go away for literally years.

What we really want is immutable labels.  I don't think we have a way to specify that, either.


(and yes, we will need a dump tool to import the thousands of existing images)

listx: I guess setting a GCS retention policy does not prevent us from moving around/renaming Docker tags? If so, then I think having a retention policy is fine.

The only thing I don't like is how we are relinquishing control of image deletion to the underlying GCS layer beneath GCR. This is something the promotion/demotion process can handle, so why put this piece in?

Also, this prevents us from performing real e2e tests around promotion using these exact GCRs in the future (we won't be able to delete test data). Maybe this seems far-fetched, but in my mind I think it is useful to have real tests that run on real infrastructure --- just in case the real one isn't behaving as expected for whatever reason.

I vote to add in a retention policy later after thinking about it some more. I feel like it should be a more thought-out decision.

k8s.gcr.io/ensure-prod-repo.sh
```diff
@@ -0,0 +1,86 @@
+#!/bin/sh
```
listx: Out of curiosity, is there a reason you didn't use bash here? Like, I'm fine with `sh` but I saw bash everywhere in kuberenetes/release repo.

k8s.gcr.io/lib.sh
thockin: Why: belts and suspenders.  k8s is too important to risk.  Consider what happens if a bug nukes an image that people use.  The impact is too big to ignore.

We could e2e to alternate places, that seems ok.

I am fine to not do this for now, but I *do* want to think about it. 

I will let it sit for the weekend.

k8s.gcr.io/ensure-prod-repo.sh
thockin: I try to not say "bash" unless I need bash-isms.

k8s.gcr.io/lib.sh
listx: For people coming into the discussion from the GH side, the ML thread is: https://groups.google.com/forum/#!topic/kubernetes-wg-k8s-infra/Dz4urYxWKi4

k8s.gcr.io/ensure-prod-repo.sh
```diff
@@ -0,0 +1,86 @@
...
+set -o errexit
+set -o nounset
+set -o pipefail
+
+SCRIPT_PATH=$(readlink -f "$0")
+SCRIPT_DIR=$(dirname "$SCRIPT_PATH")
+source "${SCRIPT_DIR}/lib.sh"
+
+function usage() {
+    echo "usage: $0" > /dev/stderr
```
listx: Nit: we should use `>&2` here instead of `/dev/stderr` (https://unix.stackexchange.com/a/274044).

k8s.gcr.io/lib.sh
```diff
@@ -0,0 +1,279 @@
...
+        image="ceci-nest-pas-une-image"
+        dest="${host}/${project}/${image}"
+        docker pull k8s.gcr.io/pause
+        docker tag k8s.gcr.io/pause "${dest}"
+        docker push "${dest}"
+        gcloud --project "${project}" \
+            container images delete --quiet "${dest}:latest"
+    fi
+
+    gsutil iam ch allUsers:objectViewer $(gcs_bucket_for "${project}" "${region}")
```
listx: Since $region can be an empty string (line 138), we should only expand to a double-quoted var if it is not empty. So I think

```
${region:+"$region"}
```

would be better. Then from `gcs_bucket_for`s perspective, the region, if there, is guaranteed to be non-empty, which would simplify the validation logic on line 33.

Also I would move the call to `gcs_bucket_for` here to its own line, assign to a variable and use that instead of doing it inline like this.

k8s.gcr.io/ensure-staging-repo.sh
```diff
@@ -0,0 +1,72 @@
...
+# Each staging repo exists in its own GCP project, and is writable by a
+# dedicated googlegroup.
+
+set -o errexit
+set -o nounset
+set -o pipefail
+
+SCRIPT_PATH=$(readlink -f "$0")
+SCRIPT_DIR=$(dirname "$SCRIPT_PATH")
+source "${SCRIPT_DIR}/lib.sh"
```
listx: Isn't `source` a bashism? According to https://superuser.com/a/46146 the POSIX-y way is to just use `.` to source.

k8s.gcr.io/lib.sh
```diff
@@ -0,0 +1,279 @@
...
+        echo "ensure_project(project) requires 1 argument" > /dev/stderr
+        return 1
+    fi
+    project="$1"
+
+    if ! gcloud projects describe "${project}" >/dev/null 2>&1; then
+        gcloud projects create "${project}" \
+            --organization "${GCP_ORG}"
+    else
+        o=$(gcloud projects \
```
listx: Nit: use `org` instead of `o`

k8s.gcr.io/lib.sh
```diff
@@ -0,0 +1,279 @@
...
+# $1: The GCP project
+function ensure_repo() {
+    if [ $# -lt 1 -o $# -gt 2 -o -z "$1" ]; then
+        echo "ensure_repo(project, [region]) requires 1 or 2 arguments" > /dev/stderr
+        return 1
+    fi
+    project="$1"
+    region="${2:-}"
+
+    if ! gsutil ls $(gcs_bucket_for "${project}" "${region}") >/dev/null 2>&1; then
```
listx: Please separate the call to `gcs_bucket_for` to its own line.

k8s.gcr.io/lib.sh
```diff
@@ -0,0 +1,279 @@
...
+        host=$(gcr_host_for "${region}")
+        image="ceci-nest-pas-une-image"
```
listx: Nice image name :sunglasses:

k8s.gcr.io/lib.sh
```diff
@@ -0,0 +1,279 @@
...
+    region="${2:-}"
+
+    # Set lifecycle policies.
+    gsutil retention set 30d $(gcs_bucket_for "${project}" "${region}")
+    echo '
+        {
+          "rule": [
+            {
+              "condition": {
+                "age": 30
```
listx: If the `30` here is written to match the `30d` in line 255, I would throw this into a function and reuse the number. Also might as well do a heredoc instead to pass in variables to the multiline string.

k8s.gcr.io/lib.sh
```diff
@@ -0,0 +1,279 @@
...
+              "condition": {
+                "age": 90
+              },
+              "action": {
+                "type": "Delete"
+              }
+            }
+          ]
+        }
+        ' | gsutil lifecycle set /dev/stdin $(gcs_bucket_for "${project}" "${region}")
```
listx: Hmm neat trick with the echo + /dev/stdin there.

k8s.gcr.io/ensure-prod-repo.sh
thockin: ACK in next push

k8s.gcr.io/ensure-staging-repo.sh
thockin: interesting.

k8s.gcr.io/lib.sh
thockin: ACK

k8s.gcr.io/lib.sh
thockin: ACK

k8s.gcr.io/lib.sh
thockin: blame @justinsb

k8s.gcr.io/lib.sh
thockin: I explicitly check for "" in bucket_for, to keep this simple.  Alternatively, I can make the 2nd arg required, even if it is "", which doesn't seem much better.  Like, I had to go look that syntax up and test it to convince myself it works.  I'll do it if you think it's really better but I assume that is a bash-ism?

k8s.gcr.io/lib.sh
thockin: They are not necessarily the same value

k8s.gcr.io/lib.sh
thockin: falls victim to the same non-portability you called out above, but I couldn't get it to parse EOL properly with a HERE doc, and gave up.  I could write it to a file if it matters.

k8s.gcr.io/lib.sh
listx: It is actually a POSIX thing, as per https://unix.stackexchange.com/a/415992.

I admit that these POSIX variable expansions are not intuitive --- I'm fine with leaving this as-is for now.

k8s.gcr.io/lib.sh
listx: Ack

k8s.gcr.io/lib.sh
listx: Hmm true re: non-portability.

heredoc issues aside, is there a threat of /dev/stdin becoming corrupted due to human input while this script is running...? If so we should just write to a temporary file.

thockin: no, because the | explicitly maps /dev/stdin to that contents via a pipe

/dev/stdin is supported on linux and mac, which is all I care about for this script.  I will revisit a heredoc later if needed

k8s.gcr.io/lib.sh
thockin: Interesting.  Bash docs say "If parameter is null or unset, nothing is substituted, otherwise the expansion of word is substituted.".  Testing shows `""` counts as null or unset.

PR #203, opened 2019-02-23
add script to invoke the image promoter against multiple manifests

listx:
Instead of defining this logic inline in the upcoming Prow job
configuration [1], define it here as a script.

[1]: https://github.com/kubernetes/test-infra/pull/11414

### DIFF
--- a/k8s.gcr.io/run-cip.sh
+++ b/k8s.gcr.io/run-cip.sh
@@ -1,0 +1,36 @@
+#!/bin/sh
+
+set -o errexit
+set -o nounset
+set -o pipefail
+
+usage()
+{
+    echo >&2 "usage: $0 <path/to/cip/binary> [<path/to/manifest.yaml>,<path/to/service-account.json>, ...]"
+    echo >&2 "The 2nd argument onwards are '<manifest>,<service-account>' pairs."
+    echo >&2
+}
+
+if [ $# -lt 2 ]; then
+    usage
+    exit 1
+fi
+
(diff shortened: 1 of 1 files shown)


k8s.gcr.io/run-cip.sh
```diff
@@ -0,0 +1,36 @@
+#!/bin/sh
```
thockin: Needs license and comment

k8s.gcr.io/run-cip.sh
```diff
@@ -0,0 +1,36 @@
...
+
+set -o errexit
+set -o nounset
+set -o pipefail
+
+usage()
+{
+    echo >&2 "usage: $0 <path/to/cip/binary> [<path/to/manifest.yaml>,<path/to/service-account.json>, ...]"
```
thockin: is "cip" a thing?  I had to think about it. :)  I just say "promoter"

listx:
> I have no opinion on whether this should be part of promoter, part of prow, or a script in-between. What's weird to me is that this is checked in to a weird repo that is neither associated with the prow config nor the promoter itself.

I put it here because it's where the manifests live. Prow shouldn't care about the business logic (it should only know how to call it). I guess I could put this in the promoter directly as it adds to a feature. Will add there directly; closing.

listx:
Added an issue in the promoter's repo to track this: https://github.com/GoogleCloudPlatform/k8s-container-image-promoter/issues/21

PR #205, opened 2019-02-27
Fix bug in GCR scripts

thockin:
I burned it all down and re-ran to clear lifecycle state, and found a bug.


k8s.gcr.io/ensure-staging-repo.sh
```diff
@@ -54,10 +54,10 @@ PROJECT="k8s-staging-${REPO}"
 WRITERS="k8s-infra-gcr-staging-${REPO}@googlegroups.com"
 
 # The GCP project name.
-PROJECT="k8s-gcr-prod"
+PROJECT="k8s-staging-${REPO}"
```
listx: Looks like we have to delete this line entirely? We already have it in line 51.

thockin:
good catch, pushed

PR #210, opened 2019-03-28
Upgrade cert-manager to v0.7.0

ixdy:
It sounds like this should be a safe upgrade, just needing to apply the new `cert-manager.yaml`?

/assign @munnerz

### DIFF
--- a/cert-manager/cert-manager.yaml
+++ b/cert-manager/cert-manager.yaml
@@ -1,13 +1,14 @@
-# Saved from https://raw.githubusercontent.com/jetstack/cert-manager/v0.6.1/deploy/manifests/cert-manager.yaml
+# Saved from https://raw.githubusercontent.com/jetstack/cert-manager/v0.7.0/deploy/manifests/cert-manager.yaml
 apiVersion: apiextensions.k8s.io/v1beta1
 kind: CustomResourceDefinition
 metadata:
-  name: certificates.certmanager.k8s.io
+  creationTimestamp: null
   labels:
-    app: cert-manager
+    controller-tools.k8s.io: "1.0"
+  name: certificates.certmanager.k8s.io
 spec:
   additionalPrinterColumns:
-  - JSONPath: .status.conditions[?(@.type=="Ready")].status
+  - JSONPath: .status.conditions[?(@.type==\"Ready\")].status
     name: Ready
     type: string
   - JSONPath: .spec.secretName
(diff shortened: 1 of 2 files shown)


cert-manager/cert-manager.yaml
```diff
@@ -433,183 +1491,6 @@ spec:
...
----
-# Source: cert-manager/charts/webhook/templates/ca-sync.yaml
-## This file contains a CronJob that runs every week to automatically update the
-## caBundle set on the APIService and ValidatingWebhookConfiguration resource.
-## This allows us to store the CA bundle in a Secret resource which is
-## generated by cert-manager's 'selfsigned' Issuer.
-apiVersion: batch/v1beta1
-kind: CronJob
-metadata:
-  name: cert-manager-webhook-ca-sync
```
ixdy: how do we deal with resources like this that were deleted?

munnerz: Not sure on the best way to handle this here - I assume this is deployed with some equivalent of `kubectl apply`?

I think in this instance, these resources will need to be manually deleted 🙈

munnerz:
There are no special considerations between v0.6->v0.7.

The only significant change that's relevant here, is generating a temporary self signed certificate before issuing, which should fix up the errors we saw with GCLBs when issuing for the first time in v0.6.

Are there any alerts set up to monitor certificate expiry? The expiry date is exposed as a prometheus metric as well as in the `status.notAfter` field, so a few different ways to get the data out 😄

cert-manager/cert-manager.yaml
ixdy: yeah, we're just using `kubectl apply` here...

ixdy:
No monitoring yet, but that would be good to enable, yes.

munnerz:
The old webhook-casync cronjob resources will need manual cleanup when this is applied. It *is* safe to leave them there however, although the cainjector will supersedes it (and both will perform the same task)

/lgtm
/approve

munnerz:
Unhold at will :) 

/hold

ixdy:
I meant
/hold cancel

but I also realized that I forgot to remove a deprecated bit of the README, so this needs re-lgtm.

ixdy:
This has been deployed:
```
kubectl apply -f cert-manager.yaml --validate=false
kubectl -n cert-manager delete cronjob cert-manager-webhook-ca-sync
kubectl -n cert-manager delete job cert-manager-webhook-ca-sync
kubectl -n cert-manager delete configmap cert-manager-webhook-ca-sync
kubectl -n cert-manager delete serviceaccount cert-manager-webhook-ca-sync
kubectl -n cert-manager delete clusterrole cert-manager-webhook-ca-sync
kubectl -n cert-manager delete clusterrolebinding cert-manager-webhook-ca-sync
```

PR #211, opened 2019-03-29
Add a URL shortening service

misterikkit:
This defines manifests to create a shlink instance that will serve
redirects from go.k8s.io.

Shlink is the open source URL shortener with the best name and logo. It
is also MIT licensed and readily available on docker hub. While this app
expects to use a persistent database, we inject data from a config map
into an ephemeral SQLITE database during pod startup.

The container's only control is a REST endpoint under /rest/v1, but the
only way to generate an API key is by exec-ing into the shlink
container.

See https://shlink.io/api-docs/authentication

Note: This reloads config by restarting the service when the config map changes.

/ref https://github.com/kubernetes/community/issues/2959

### DIFF
--- a/k8s.io/deployment-short-urls.yaml
+++ b/k8s.io/deployment-short-urls.yaml
@@ -1,0 +1,58 @@
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: short-urls
+spec:
+  selector:
+    matchLabels:
+      app: short-urls
+  template:
+    metadata:
+      labels:
+        app: short-urls
+    spec:
+      containers:
+      - name: shlink
+        image: shlinkio/shlink:1.16.2
+        command: ["/bin/sh", "-c"]
+        args:
+        - |
+          # Start shlink server in the background.
+          /bin/sh /etc/shlink/docker-entrypoint.sh &
+          PID=$!
+          # Wait for service to become available.
+          while ! shlink short-url:list; do
+            sleep 10s
+          done
+          # Load URLs from the config map.
+          while read slug url; do
+            shlink short-url:generate "--customSlug=${slug}" "${url}"
+          done < /etc/config/short-urls
+          # Create healthcheck URL for our readiness probe.
+          shlink short-url:generate -c shlink-healthcheck https://k8s.io
+          # Block until the config map changes, then exit. This will restart
+          # the container and load the new config.
+          # TODO: live reload without downtime
+          inotifyd - /etc/config/short-urls:e
+        env:
+          - name: SHORT_DOMAIN_HOST
+            value: "go.k8s.io"
+          - name: NOT_FOUND_REDIRECT_TO
+            value: "https://k8s.io"
+          - name: SHORT_DOMAIN_SCHEMA
+            value: "https"
+          - name: DB_DRIVER
+            value: "sqlite"
+        ports:
+        - containerPort: 8080
+        readinessProbe:
+          exec:
+            command: ["shlink", "short-url:parse", "shlink-healthcheck"]
+        volumeMounts:
+        - name: short-urls-config
+          mountPath: /etc/config
+        # TODO: set resource requests
+      volumes:
+      - name: short-urls-config
+        configMap:
+          name: short-urls
--- a/k8s.io/configmap-short-urls.yaml
+++ b/k8s.io/configmap-short-urls.yaml
@@ -1,0 +1,21 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: short-urls
+data:
+  short-urls: |
+    api-review      https://github.com/kubernetes/community/blob/master/sig-architecture/api-review-process.md
+    bot-commands    https://prow.k8s.io/command-help
+    github-labels   https://github.com/kubernetes/test-infra/blob/master/label_sync/labels.md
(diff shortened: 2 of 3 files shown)


misterikkit:
I'm running a demo of this service at http://35.238.224.90/ e.g.
http://35.238.224.90/start
http://35.238.224.90/triage

unknown paths redirect to k8s.io, but http://35.238.224.90 will show a shlink 404 page.

k8s.io/deployment-short-urls.yaml
```diff
@@ -0,0 +1,58 @@
...
+        image: shlinkio/shlink:1.16.2
+        command: ["/bin/sh", "-c"]
+        args:
+        - |
+          # Start shlink server in the background.
+          /bin/sh /etc/shlink/docker-entrypoint.sh &
+          PID=$!
+          # Wait for service to become available.
+          while ! shlink short-url:list; do
+            sleep 10s
```
thockin: log something (stderr)?

k8s.io/deployment-short-urls.yaml
```diff
@@ -0,0 +1,58 @@
...
+          done
+          # Load URLs from the config map.
+          while read slug url; do
+            shlink short-url:generate "--customSlug=${slug}" "${url}"
+          done < /etc/config/short-urls
+          # Create healthcheck URL for our readiness probe.
+          shlink short-url:generate -c shlink-healthcheck https://k8s.io
```
thockin: I don't grok what this is doing?  is it making a redirect to k8s.io as a way to prove readiness?  Very soon we won't follow the URL at all but just accept 301/302 as succes, which still seems OK

k8s.io/deployment-short-urls.yaml
```diff
@@ -0,0 +1,58 @@
...
+          # Block until the config map changes, then exit. This will restart
+          # the container and load the new config.
+          # TODO: live reload without downtime
+          inotifyd - /etc/config/short-urls:e
```
thockin: What happens if shlink dies for some reason in the background?

k8s.io/deployment-short-urls.yaml
```diff
@@ -0,0 +1,58 @@
...
+          - name: NOT_FOUND_REDIRECT_TO
+            value: "https://k8s.io"
+          - name: SHORT_DOMAIN_SCHEMA
+            value: "https"
+          - name: DB_DRIVER
+            value: "sqlite"
+        ports:
+        - containerPort: 8080
+        readinessProbe:
+          exec:
```
thockin: why not httpGet ?

thockin:
Also, this should all live in a `go.k8s.io` subdir

k8s.io/deployment-short-urls.yaml
misterikkit: This entry is added to the database after all the "real" entries are added. This is because the shlink service is up and running, but I didn't want the readiness probe to be OK until we finish loading config.

misterikkit: In my mind this was going to help with deployment rollouts when we have more than 1 replica, so that "ready" instances have all the short URLs loaded. However, since the container is restarting inside the pod, I'm not sure if the deployment controller is capable of doing the right thing.

k8s.io/deployment-short-urls.yaml
misterikkit: the container stops serving but stays up until the next config map update. :shame:

misterikkit:
Thanks for the extensive feedback!

> Could it be simplified with a 1 container which runs a simpler loop that populates the sqlite from the configmap on a shared volume, and 1 container (sibling) that serves? I assume since it is sqlite, it could be updated atomically?

I actually tried to do it this way before sending this PR, and ran into a couple snags.
1. I didn't want to reverse engineer or couple ourselves to their database schema. So I used their CLI. I think the CLI might be able to operate directly on a sqlite file.
2. shlink (or at least the provided container image) expects its data dir in a specific place, and that dir is pre-populated with a file tree. I couldn't figure out how to mount a shared volume between two containers at that path and not hide the pre-defined file tree.

> To get this to launchable, we'd need a plan (scripted) for how to test changes before going live.

Fortunately, shlink seems to completely ignore "Host:" headers, so it's easy to serve the same set of redirects on any domain or static IP. I'll work with the WG to define and document a test/rollout proceedure.

k8s.io/deployment-short-urls.yaml
thockin: If, for example, shlink deadlocks the readiness probe will fail, and it will be removed from LBs.  You can probably use this as both a readiness and a liveness probe with different timings, so it doesn't become ready until all URLs are loaded and if it stops serving kubelet will kill it.

k8s.io/deployment-short-urls.yaml
thockin: ooooops..  https://media0.giphy.com/media/9wGjpl8z1CJ5m/giphy.gif?cid=790b76115c9ea09d716e42364db94eed

k8s.io/deployment-short-urls.yaml
thockin: yeah, but httpGet is lower impact and should have the same result, right?  If teh URL is defined, it will return 302 and if not it will return 404 I assume...

misterikkit: True if we don't use the "default redirect to" feature. I don't know if we care about mistyped short links. 404 is probably better than a default?

thockin:
I think 404 sounds appropriate.

misterikkit:
After discussion with @stp-ip it seems like txtdirect addresses a lot of the concerns raised about this implementation. And setting it up would not involve the scary hacks needed to make shlink behave like a config-driven app.

PR #214, opened 2019-04-03
 Add a cip-test GCRs for promoter testing

thockin:
I ran `ensure-staging-repo.sh cip-test` which produced a project "k8s-staging-cip-test".

I ran this new `ensure-prod-repo.sh` which produced a project "k8s-cip-test-prod".

I created a new googlegroup `k8s-infra-gcr-staging-cip-test` which includes @listx

### DIFF
--- a/k8s.gcr.io/ensure-prod-repo.sh
+++ b/k8s.gcr.io/ensure-prod-repo.sh
@@ -35,57 +35,53 @@ if [ $# != 0 ]; then
 fi
 
 # The GCP project names.
+TEST_PROJECT="k8s-cip-test-prod"
 PROD_PROJECT="k8s-gcr-prod"
 TRASH_PROJECT="k8s-gcr-graveyard"
+
+ALL_PROJECTS="${TEST_PROJECT} ${PROD_PROJECT} ${TRASH_PROJECT}"
 
 # Regions for prod.
 PROD_REGIONS=(us eu asia)
 
 # Make the projects, if needed
-color 6 "Ensuring prod project exists: ${PROD_PROJECT}"
-ensure_project "${PROD_PROJECT}"
-color 6 "Ensuring graveyard project exists: ${TRASH_PROJECT}"
-ensure_project "${TRASH_PROJECT}"
+for prj in ${ALL_PROJECTS}; do
+    color 6 "Ensuring project exists: ${prj}"
+    ensure_project "${prj}"
 
-color 6 "Configuring billing for prod"
-ensure_billing "${PROD_PROJECT}"
-color 6 "Configuring billing for graveyard"
-ensure_billing "${TRASH_PROJECT}"
+    color 6 "Configuring billing: ${prj}"
+    ensure_billing "${prj}"
 
-# Enable container registry APIs
-color 6 "Enabling the container registry API for prod"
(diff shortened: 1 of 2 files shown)


k8s.gcr.io/ensure-prod-repo.sh
```diff
@@ -35,57 +35,67 @@ if [ $# != 0 ]; then
...
+
 # Regions for prod.
 PROD_REGIONS=(us eu asia)
 
 # Make the projects, if needed
-color 6 "Ensuring prod project exists: ${PROD_PROJECT}"
-ensure_project "${PROD_PROJECT}"
-color 6 "Ensuring graveyard project exists: ${TRASH_PROJECT}"
-ensure_project "${TRASH_PROJECT}"
+for prj in ${ALL_PROJECTS}; do
```
listx: Why not just use 1 loop `for prj in ${ALL_PROJECTS}; do` and move each component into it (instead of wrapping each component with a loop)?

I think having 1 loop would be easier in case this script dies (you wouldn't have to check any of the projects that have not yet been iterated through by the for-loop).

k8s.gcr.io/lib.sh
```diff
@@ -229,7 +234,7 @@ function empower_promoter() {
     if ! gcloud --project "${project}" iam service-accounts describe "${acct}" >/dev/null 2>&1; then
         gcloud --project "${project}" \
             iam service-accounts create \
-            "${name}" \
+            "${PROMOTER_SVCACCT}" \
```
listx: It's not obvious to me why this was changed...

k8s.gcr.io/ensure-prod-repo.sh
listx: Easier as in easier to debug, that is.

k8s.gcr.io/lib.sh
thockin: This was a bug - name is not defined here.  We never hit it before because testing always stopped at the `if` and never executed this line until we had a legit new project.

k8s.gcr.io/ensure-prod-repo.sh
thockin: Indeed that would be simpler.  Myopia...

Will push with that change momentarily.

thockin:
repushed

k8s.gcr.io/ensure-prod-repo.sh
```diff
@@ -35,57 +35,53 @@ if [ $# != 0 ]; then
...
+    ensure_billing "${prj}"
+
+    color 6 "Enabling the container registry API: ${prj}"
+    enable_api "${prj}" containerregistry.googleapis.com
+
+    color 6 "Enabling the container analysis API for: ${prj}"
+    enable_api "${prj}" containeranalysis.googleapis.com
+
+    color 6 "Ensuring the registry exists and is readable: ${prj}"
+    for r in "${PROD_REGIONS[@]}"; do
```
listx: I would have used just 1 loop here as well for the regions.

thockin:
I prefer seeing that as an inner-loop while I watch the output of it.

listx:
Oh look! The `pull-k8sio-cip` check (correctly) got skipped, because you didn't touch any manifest files. Nice.

thockin:
@dims for approve

Issue #215, opened 2019-04-03
move k8s.io redirector behind Ingress and automate TLS using cert-manager


ixdy:
The TLS cert for the k8s.io redirector expires in about a week (April 11). Let's get it automated!

I've already got cert-manager set up in the cluster, so now it's just a matter of getting the redirector running behind an `Ingress` without breaking anything.

Steps to get there:
1. Teach nginx how to handle forwarded HTTPS requests from the GCP Ingress (since it won't terminate HTTPS directly in this case).
2. Create an `Ingress` to the existing `LoadBalancer` service, using a new static IP as the existing static IP is still bound to the `LoadBalancer`. Use the existing soon-to-expire TLS cert to terminate HTTPS on the `Ingress`.
3. Update DNS records to point to the new static IP and wait for propagation.
4. Switch the service to `NodePort`, releasing the old static IP.
5. Create cert-manager `Certificate` resources - use the Let's Encrypt staging endpoint for canary, and prod endpoint for prod. (Note: because of the large number of subdomains, this creates a lot of backend services at once (~50). We'll need to increase our backend service quota first.)
6. After the TLS cert has been created (may take some time), update the `Ingress` to use the new secret. (I suggest using a new name here, since we'll be using different Let's Encrypt endpoints for canary and prod.)
7. Switch off HTTPS support in nginx and delete the old certificate.
8. Hopefully never think about this again. :)

/assign
cc @thockin @munnerz @nikhita

ixdy:
(I've basically gotten all of this working on my test instance; just need to roll it out to k8s.io now.)

ixdy:
OK, looking at this a bit more, it seems like it's not possible to rename the IP address automatically allocated by the Ingress, so we need to first explicitly create one.

Unfortunately we already have `k8s-io-canary` and `k8s-io-prod` IP addresses, so I guess I'll have to use a different name.

PR #216, opened 2019-04-03
Create Ingress for k8s.io redirector

ixdy:
This is the first part of several PRs - basically steps 1 and 2 on https://github.com/kubernetes/k8s.io/issues/215.

The bulk of this PR is new logic in the nginx config to properly upgrade HTTP to HTTPS when running behind the `Ingress`, while maintaining existing behavior for clients that connect via the `LoadBalancer` directly.

/assign @thockin

### DIFF
--- a/k8s.io/ingress-prod.yaml
+++ b/k8s.io/ingress-prod.yaml
@@ -1,0 +1,15 @@
+apiVersion: extensions/v1beta1
+kind: Ingress
+metadata:
+  name: k8s-io
+  labels:
+    app: k8s-io
+  namespace: k8s-io-prod
+  annotations:
+    kubernetes.io/ingress.global-static-ip-name: k8s-io-ingress-prod
+spec:
+  tls:
+  - secretName: ssl
+  backend:
+    serviceName: k8s-io
+    servicePort: http
--- a/k8s.io/Makefile
+++ b/k8s.io/Makefile
@@ -14,6 +14,7 @@ deploy:
 	kubectl apply -f configmap-www-get.yaml
 	kubectl apply -f service-dev.yaml
 	kubectl apply -f deployment.yaml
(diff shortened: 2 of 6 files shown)


ixdy:
I created the `k8s-io-ingress-canary` static IP and deployed this to the canary namespace; both the existing `LoadBalancer` service and the `Ingress` are working as expected.

I haven't yet created an IP for prod, nor have I deployed anything to the prod namespace.

k8s.io/ingress-prod.yaml
```diff
@@ -0,0 +1,15 @@
...
+metadata:
+  name: k8s-io
+  labels:
+    app: k8s-io
+  namespace: k8s-io-prod
+  annotations:
+    kubernetes.io/ingress.global-static-ip-name: k8s-io-ingress-prod
+spec:
+  tls:
+  - secretName: ssl
```
thockin: I assume the cert manager work comes after the flip?

ixdy: right, because I can't do the http challenges until the domain is pointing at the Ingress.

... which makes me realize that I'm not sure if we'll be able to generate a cert for the canary instance, since the domains don't point at it. hm.

ixdy:
This should be pretty low-risk (especially since canary worked out fine), but I'll deploy this tomorrow when I'm awake and watching and also send an update for DNS.

k8s.io/ingress-prod.yaml
ixdy: I guess I could create a [self-signed `ClusterIssuer`](https://cert-manager.readthedocs.io/en/latest/tasks/issuers/setup-selfsigned.html), since we don't need real certs for the canary instance.

ixdy:
Changes deployed to prod; everything is looking fine so far.

PR #218, opened 2019-04-04
k8s.io redirector: remove/fix old sites

ixdy:
This is tangentially related to #215.

After looking at our DNS records, I noticed that testgrid no longer points at the redirector, so we can remove it (and not request its subdomain in our new cert).

The blog now seems to be hosted on https://kubernetes.io/blog, whereas blog.kubernetes.io points at the redirector. This makes both https://blog.kubernetes.io/ and https://blog.k8s.io/ now actually go to the blog.

I also added a handler for `sigs.kubernetes.io`, since that name points at the redirector, though most folks will probably be using sigs.k8s.io.

/assign @thockin

### DIFF
--- a/k8s.io/configmap-nginx.yaml
+++ b/k8s.io/configmap-nginx.yaml
@@ -64,7 +64,7 @@ data:
 
       # Vanity redirects for the new kubernetes-sigs repos
       server {
-        server_name sigs.k8s.io;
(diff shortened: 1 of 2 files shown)


ixdy:
/close

I'm just going to fold this into another PR.

PR #7, opened 2016-08-23
Make ci-test.k8s.io/e2e/123 redirect to Gubernator.

rmmh:
One potential problem: deploying a new configmap changes the files in the running containers, but doesn't restart them *or* force nginx to reload. The deployment should run `nginx -s reload` on the running pods to trigger a config reload.

---
This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/k8s.io/7)

PR #219, opened 2019-04-04
Enable Let's Encrypt automation for k8s.io redirector

### LINKED ISSUE #215: move k8s.io redirector behind Ingress and automate TLS using cert-manager

ixdy:
OK, the last sequence of changes to fix #215.

Commits, in order they should be applied:
1. Add a self-signing dummy `ClusterIssuer`. We'll use this for the canary instance of the redirector, since we can't use Let's Encrypt HTTP challenges since the domains don't point at the canary Ingress.
2. Remove and fix old sites on the redirector. I took this from #218. These are either subdomains that no longer point at the redirector, or are redirections we were doing wrong.
3. Create `Certificate` resources which will generate TLS certificates (saving them into a secret named `k8s-io-tls`). I manually created these lists by looking at the DNS configuration for subdomains that point at the redirector. TODO: better management of these lists.
4. After the new TLS secrets have been created (will take some time), point the `Ingress` at them.
  - Note that I already bumped up our backend services quota on the GCP project to 100.
5. Switch service to `NodePort`, since we can use the `Ingress` exclusively now.
6. Stop serving HTTPS from nginx. (I'll also delete the old secret at this point.)


thockin:
There's a lot in here, but it looks OK.  What is "dev" ?  I have no memory of that being added.  What is it?

ixdy:
it appears to be what gets deployed if you run `make deploy` - I guess if you're testing changes on your own kubernetes cluster? It was added in https://github.com/kubernetes/k8s.io/pull/7. 

Admittedly I've just been using the canary configs to test on my own cluster. I can delete `dev` if you'd prefer.

ixdy:
I'm going to go ahead and deploy these changes to canary.

ixdy:
Canary deployment is working and passing all tests.

k8s.io/Makefile
```diff
@@ -14,6 +14,7 @@ deploy:
 	kubectl apply -f configmap-www-get.yaml
 	kubectl apply -f service-dev.yaml
 	kubectl apply -f deployment.yaml
+	kubectl apply -f certificate-dev.yaml
```
thockin: I don't think anyone uses the -dev variants.  Can we just nuke them?  In fact we might want to nuke this rule - I forgot it existed

ixdy: deleted!

ixdy:
@thockin any other comments?

cblecker:
/lgtm
/approve
/hold

This looks sane to me.

ixdy:
Updated ConfigMap, Deployment, and Service have been applied; tests still pass.

Applying Certificate resource now.

ixdy:
Oops, bumped into a quota limit:
```
loadbalancer-controller  Error during sync: googleapi: Error 403: Quota 'HEALTH_CHECKS' exceeded. Limit: 50.0 globally., quotaExceeded
```

Quota bump has been requested.

ixdy:
I deleted the certificate resource in the prod namespace. Will reattempt once quota bump has been applied.

ixdy:
quota bump is in; trying to create a certificate again.

ixdy:
This has now been fully deployed. Last remaining issue I found will be addressed by https://github.com/kubernetes/k8s.io/pull/224.

PR #220, opened 2019-04-05
fixes to promoter manifests

listx:
There is no better source of clarity than real-world testing.

Functionally, we remove coredns 1.3.1 from `gcr.io/k8s-staging-coredns`'s manifest, because it was used for testing the promotion of manifest lists (verified with local testing).

/cc @thockin @dims

### DIFF
--- a/k8s.gcr.io/k8s-staging-cluster-api/manifest.yaml
+++ b/k8s.gcr.io/k8s-staging-cluster-api/manifest.yaml
@@ -1,7 +1,7 @@
 # google group for gcr.io/k8s-staging-cluster-api is k8s-infra-gcr-staging-cluster-api@googlegroups.com
-src-registry: gcr.io/k8s-staging-cluster-api
 registries:
 - name: gcr.io/k8s-staging-cluster-api
+  src: true
 - name: us.gcr.io/k8s-gcr-prod
   service-account: k8s-infra-gcr-promoter@k8s-gcr-prod.iam.gserviceaccount.com
 - name: eu.gcr.io/k8s-gcr-prod
@@ -9,3 +9,4 @@ registries:
 - name: asia.gcr.io/k8s-gcr-prod
   service-account: k8s-infra-gcr-promoter@k8s-gcr-prod.iam.gserviceaccount.com
 images: []
+renames: []
--- a/k8s.gcr.io/k8s-staging-cluster-api/manifest.yaml
+++ b/k8s.gcr.io/k8s-staging-cluster-api/manifest.yaml
@@ -2,6 +2,10 @@
 src-registry: gcr.io/k8s-staging-cluster-api
 registries:
 - name: gcr.io/k8s-staging-cluster-api
-- name: gcr.io/k8s-gcr-prod
+- name: us.gcr.io/k8s-gcr-prod
+  service-account: k8s-infra-gcr-promoter@k8s-gcr-prod.iam.gserviceaccount.com
+- name: eu.gcr.io/k8s-gcr-prod
+  service-account: k8s-infra-gcr-promoter@k8s-gcr-prod.iam.gserviceaccount.com
+- name: asia.gcr.io/k8s-gcr-prod
   service-account: k8s-infra-gcr-promoter@k8s-gcr-prod.iam.gserviceaccount.com
 images: []
--- a/k8s.gcr.io/k8s-staging-cluster-api/manifest.yaml
+++ b/k8s.gcr.io/k8s-staging-cluster-api/manifest.yaml
@@ -2,7 +2,6 @@
 src-registry: gcr.io/k8s-staging-cluster-api
 registries:
 - name: gcr.io/k8s-staging-cluster-api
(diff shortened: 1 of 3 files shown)


k8s.gcr.io/k8s-staging-cluster-api/manifest.yaml
```diff
@@ -2,7 +2,10 @@
 src-registry: gcr.io/k8s-staging-cluster-api
 registries:
 - name: gcr.io/k8s-staging-cluster-api
```
thockin: why is the src registry in this list?

k8s.gcr.io/k8s-staging-cluster-api/manifest.yaml
```diff
@@ -2,7 +2,10 @@
...
-  service-account: service-account@kubernetes-public.iam.gserviceaccount.com
-- name: gcr.io/k8s-gcr-prod
-  service-account: service-account@kubernetes-public.iam.gserviceaccount.com
+- name: us.gcr.io/k8s-gcr-prod
```
thockin: If we wanted this have subdirs, would it work?  (we don't yet, but might for things like e2e test images)

k8s.gcr.io/k8s-staging-cluster-api/manifest.yaml
listx: It's part of the promoter design. For the promoter, it gathers the state of the world by looking up info in `registries`, so src registry needs to be in that list.

I updated the docs also: https://github.com/kubernetes-sigs/k8s-container-image-promoter/commit/f2daa341f7d5969941d9a96f123fa636ab1a16be

k8s.gcr.io/k8s-staging-cluster-api/manifest.yaml
listx: Do you mean this: https://github.com/kubernetes-sigs/k8s-container-image-promoter/issues/20 ?
We need to implement that feature...

To rephrase, that issue is describing this scenario:

Push to `gcr.io/k8s-staging-cluster-api/foo-image`, but we want it to show up as either

(1) `gcr.io/k8s-gcr-prod/bar-image` (notice the image rename)
(2) `gcr.io/k8s-gcr-prod/some/different/subdir/path/foo-image` (notice how the "directory path" to the image is different)

or a mix of both (1) and (2) (image rename *and* path rename).

This is what you mean, right?

k8s.gcr.io/k8s-staging-cluster-api/manifest.yaml
listx: To clarify, having src registry's name in `registries` list does NOT make the promoter try to promote from src to src (NO self-loops allowed!).

k8s.gcr.io/k8s-staging-cluster-api/manifest.yaml
thockin: Mostly I care about (2), I think.

listx: Ok, I will start working on (2) and report progress in kubernetes-sigs/k8s-container-image-promoter#20.

listx: I created a new PR to handle subdirs (renames) https://github.com/kubernetes-sigs/k8s-container-image-promoter/pull/40. It supports the general case of renaming both the image name itself or the subdirectory (or both). Rename handling required some refactoring and a slight tweak to the promotion logic, which resulted in some delay. Anyway after that's merged, I have to

- tag the new promoter release
- create new promoter images
- update the Prowjobs to use the new promoter images
- update this PR (the redundant `src-registry` toplevel field is no longer required)

It's mostly housekeeping stuff at this point. The e2e test suite is the next priority item...

listx:
Waiting for https://github.com/kubernetes/test-infra/pull/12227 to get merged. Then we can merge this afterwards.

listx:
@dims @thockin PTAL

We should be able to handle renames now.

k8s.gcr.io/k8s-staging-cluster-api/manifest.yaml
```diff
@@ -1,8 +1,12 @@
...
-  service-account: service-account@kubernetes-public.iam.gserviceaccount.com
+  src: true
+- name: us.gcr.io/k8s-gcr-prod
+  service-account: k8s-infra-gcr-promoter@k8s-gcr-prod.iam.gserviceaccount.com
+- name: eu.gcr.io/k8s-gcr-prod
+  service-account: k8s-infra-gcr-promoter@k8s-gcr-prod.iam.gserviceaccount.com
+- name: asia.gcr.io/k8s-gcr-prod
+  service-account: k8s-infra-gcr-promoter@k8s-gcr-prod.iam.gserviceaccount.com
 images: []
+renames: []
```
thockin: Any clues on what this is?

listx: https://github.com/kubernetes-sigs/k8s-container-image-promoter#renames

PR #222, opened 2019-04-06
Add script to create staging buckets

justinsb:


k8s.gcr.io/ensure-staging-bucket.sh
```diff
@@ -0,0 +1,93 @@
...
+BUCKET="gs://${PROJECT}"
+
+# A short retention - it can always be raised, but it is hard to lower
+# We expect promotion within 30d, or for testing to "move on"
+# 30d is also short enough that people should notice occasionally,
+# and not accidentally think of the staging buckets as permanent.
+RETENTION=30d
+
+# Make the project, if needed
+color 6 "Ensuring project exists: ${PROJECT}"
```
justinsb: Should I merge into ensure-staging-repo.sh (and rename it) ?

k8s.gcr.io/ensure-staging-bucket.sh
```diff
@@ -0,0 +1,93 @@
...
+# The name of the sub-project being created, e.g. "coredns".
+REPO="$1"
+
+# The GCP project name.
+PROJECT="k8s-staging-${REPO}"
+
+# The group that can write to this staging repo.
+# TODO: Do we want a separate group for GCS vs GCR?
+#WRITERS="k8s-infra-gcr-staging-${REPO}@googlegroups.com"
+WRITERS="k8s-infra-gcs-staging-${REPO}@googlegroups.com"
```
justinsb: WYDT @thockin ... how are we creating these groups?

k8s.gcr.io/lib.sh
```diff
@@ -195,7 +200,33 @@ function empower_gcr_admins() {
...
-# Grant write privileges to a group
+# Grant full privileges to GCS admins
+# $1: The GCP project
+# $2: The bucket
+function empower_gcs_admins() {
+    if [ $# -ne 2 -o -z "$1" -o -z "$2" ]; then
+        echo "empower_gcs_admins(project, bucket) requires 2 arguments" >&2
+        return 1
+    fi
+    project="$1"
```
justinsb: Technically we don't need the project, but it feels inconsistent not to require it (IMO)

justinsb:
I'm tempted to try this from go, because my bash is terrible...

justinsb:
Tried the other half of this in go: #223

k8s.gcr.io/ensure-staging-bucket.sh
thockin: This is  my first reaction.  I think it's net simpler to fold them together.  It means all staging repos have GCR and GCS but that's probably fine.  Consistency is valuable.

That said, do you think there's any reason to differentiate ACLs for GCS and GCR?  E.g. What if we just renamed the groups from "k8s-infra-gcr-staging-${REPO}@googlegroups.com" to just "k8s-infra-staging-${REPO}@googlegroups.com" and let them control both GCS and GCR ?

k8s.gcr.io/ensure-staging-bucket.sh
thockin: We have to do it manually for now.  Don't do it until we settle on ACLs

k8s.gcr.io/ensure-staging-bucket.sh
```diff
@@ -0,0 +1,93 @@
...
+ensure_billing "${PROJECT}"
+
+# Enable GCS APIs
+color 6 "Enabling the GCS API"
+enable_api "${PROJECT}" storage-component.googleapis.com
+
+# Create the bucket
+color 6 "Ensuring the bucket exists and is readable"
+if ! gsutil ls "${BUCKET}" >/dev/null 2>&1; then
+  gsutil mb --retention ${RETENTION} -p ${PROJECT} ${BUCKET}
```
thockin: Can you split the "make bucket" step from "set retention" step?  That way, if we re-run this it will re-assert retention.  Then if we change it to 60 days, we can just re-run this script.

Reconcilers FTW.

Also retention is not deletion.  Retention is a minimum.  To auto-delete, I think you need lifecycle, something like:

```
    # Set lifecycle policies.
    gsutil retention set 30d "${bucket}"
    echo '
        {
          "rule": [
            {
              "condition": {
                "age": 31
              },
              "action": {
                "type": "Delete"
              }
            }
          ]
        }
        ' | gsutil lifecycle set /dev/stdin "${bucket}"
```

Please try this by setting a 1 or 2 day and confirming it works as documented :)

k8s.gcr.io/ensure-staging-bucket.sh
```diff
@@ -0,0 +1,93 @@
...
+color 6 "Configuring billing for ${PROJECT}"
+ensure_billing "${PROJECT}"
+
+# Enable GCS APIs
+color 6 "Enabling the GCS API"
+enable_api "${PROJECT}" storage-component.googleapis.com
+
+# Create the bucket
+color 6 "Ensuring the bucket exists and is readable"
+if ! gsutil ls "${BUCKET}" >/dev/null 2>&1; then
```
thockin: Please wrap these up into function in lib.sh, e.g. `ensure_bucket` and `ensure_bucket_lifecycle`

k8s.gcr.io/lib.sh
```diff
@@ -34,6 +34,11 @@ function color() {
 # The group that admins all GCR repos.
 GCR_ADMINS="k8s-infra-gcr-admins@googlegroups.com"
 
+# The group that admins all GCS buckets.
+# TODO: Do we want a separate group from GCR??
```
thockin: We could rename to "k8s-infra-storage-admins@googlegroups.com" - it's less obvious for the admins than it is for the staging, IMO

k8s.gcr.io/lib.sh
```diff
@@ -217,6 +248,27 @@ function empower_group() {
         "${bucket}"
 }
 
+# Grant write privileges on a bucket to a group
+# $1: The GCP project
+# $2: The googlegroups group
+# $3: The bucket
+function empower_group_to_bucket() {
```
thockin: maybe we should rename existing functions like empower_group to empower_group_for_gcr or somethign?

k8s.gcr.io/ensure-staging-bucket.sh
justinsb: I think we agreed no reason to have different groups, and that we would rename the groups when we moved to gsuite.

k8s.gcr.io/lib.sh
justinsb: Did some renames where I thought it made sense.

justinsb:
All fixed up ... please take a look @thockin

k8s.gcr.io/lib.sh
```diff
@@ -168,6 +172,23 @@ function ensure_repo() {
...
+function ensure_gcs_bucket() {
+    if [ $# -lt 2 --o -z "$1" -o -z "$2" ]; then
+        echo "ensure_gcs_bucket(project, bucket) requires 2 arguments" >&2
+        return 1
+    fi
+    project="$1"
+    bucket="$2"
+
+    if ! gsutil ls "${bucket}" >/dev/null 2>&1; then
+      gsutil mb -p ${project} ${bucket}
```
thockin: quote variables pls

k8s.gcr.io/ensure-staging-storage.sh
```diff
@@ -73,6 +87,32 @@ empower_gcr_admins "${PROJECT}"
 
 # Enable repo writers
 color 6 "Empowering ${WRITERS}"
```
thockin: ... to GCR

k8s.gcr.io/ensure-staging-storage.sh
```diff
@@ -73,6 +87,32 @@ empower_gcr_admins "${PROJECT}"
...
+# Set bucket auto-deletion
+color 6 "Ensuring the bucket has auto-deletion of ${AUTO_DELETION_DAYS} days"
+ensure_gcs_bucket_auto_deletion "${BUCKET}" "${AUTO_DELETION_DAYS}"
+
+# Enable admins on the bucket
+color 6 "Empowering GCS admins"
+empower_gcs_admins "${PROJECT}" "${BUCKET}"
+
+# Enable writers on the bucket
+color 6 "Empowering ${WRITERS}"
```
thockin: ... to GCS

k8s.gcr.io/lib.sh
```diff
@@ -247,3 +315,45 @@ function empower_promoter() {
...
+
+    gsutil retention set "${retention}" "${bucket}"
+}
+
+# Ensure the bucket auto-deletion policy is set
+# $1: The GCS bucket
+# $2: The auto-deletion policy
+function ensure_gcs_bucket_auto_deletion() {
+    if [ $# -lt 2 -o -z "$1" -o -z "$2" ]; then
+        echo "ensure_gcs_bucket_auto_deletion(bucket, auto_delettion_days) requires 2 arguments" >&2
```
thockin: typo: delettion

PR #223, opened 2019-04-06
WIP: Tool to create LoadBalancer in front of GCS backend

justinsb:
All hard-coded currently, but working towards declarative management.


justinsb:
/assign thockin

WIP, but this is IMO an easier way to manage than bash.  This could be because I don't have a good bash hammer though.

thockin:
Who write this in Go rather than shell+gcloud?

justinsb:
So I'm hoping we can:

* Have richer reconciliation than we can realistically/reliably implement in bash (we're not there yet in the go code, to be fair)
* Use this same code/mechanism for auditing (hence the separate Find function)
* Use this same code/mechanism for dry-run (again, the Find vs Reconcile split)
* Reuse and share this code in other places in k8s where we configure infrastructure
* Work towards a maintainable code base that allows other developers to contribute - bash is not great for collaboration, and our bash code doesn't have a great track record

justinsb:
We chose to do this in bash instead

/close

PR #224, opened 2019-04-11
Don't request www.kubernetes.io SAN on TLS certificate

ixdy:
www.kubernetes.io is a CNAME to the real kubernetes.io site, not our redirector, so we shouldn't request a cert for it, especially since we can't solve HTTP challenges for it.


ixdy:
(this has already been deployed)

munnerz:
🎉🎉

/lgtm

munnerz:
🎉🎉

/lgtm

PR #225, opened 2019-04-12
Migrate go.k8s.io shortener to TXTDirect

stp-ip:
Proposed canary test on deployment:
* [ ] check default redirect to kubernetes.io
* [ ] check redirect for correct redirect of go.k8s.io/oncall


cblecker:
Can you do this on a separate domain first, to test? Are there any limitations to using the hosted service, vs setting up our own?

/hold
for review

dns/zone-configs/k8s.io.yaml
```diff
@@ -232,3 +228,76 @@ velodrome:
 kind.sigs:
   type: CNAME
   value: k8s-kind.netlify.com.
+
+# TODO: Split into separate file
+# URL shortener service.
+go:
+  type: CNAME
+  value: txtd.io.
```
thockin: I am very uncomfortable taking such a dependency on a free service that has no documented SLA, no public monitoring, and no repercussions to an outage.

I'd rather run something ourselves.

If we think this is a cleaner model for config (I have no strong feeling - @misterikkit ?) it seems pretty trivial to write a simple form of this that we can make our own guarantees about.

cblecker: @thockin It's an OSS project: https://github.com/txtdirect/txtdirect

We could easily create a manifest and run it ourselves, but I think @stp-ip was just POC'ing it out.

stp-ip: Basically my team and I run the "free" service. It's doing around 2 million requests per month and has monitoring etc.
But this was suggested as a first step to test on a canary beforehand and then move to a self-hosted option.
We can provide the k8s deployment code (as that's what we also run the hosted service on) and publicly build docker images for txtdirect: https://gcr.io/v2/txtdirect-223710/txtdirect/tags/list

As @cblecker mentioned this is mostly a PoC to get the process and feeling and then be able to switch to self-hosted for prod.

stp-ip: As said. All open source and definitely not pushing the hosted option. In my mind it always was supposed to run the self-hosted after the PoC.

stp-ip:
> Can you do this on a separate domain first, to test? Are there any limitations to using the hosted service, vs setting up our own?
> 
> /hold
> for review

Let me change this to a different sub domain so we can try it out without impacting the old redirects.
No real limitations from our side for the PoC. Also happy to scale it up, if necessary.

stp-ip:
Kept old CNAME and added go2.k8s.io with the changes to make testing easier.

dns/zone-configs/k8s.io.yaml
thockin: Roger.  Sorry for being harsh.  It has nothing to do with YOUR service and everything to do with A service. :)

thockin:
As a solution this seems OK.  I would insist on the self-hosted form for now, at least.

We should do it as this suggests - a whole new zone to be delegated so all these configs are in a single file.

We would need to adapt the canary process to it, too.

dns/zone-configs/k8s.io.yaml
misterikkit: Regarding the config model, this is more complex than my POC. It puts structured data in the TXT record and that record needs to go in a DNS config file. People will likely be copy-pasting existing records when adding new ones. (that's probably fine)

a potential benefit is that our DNS config is well maintained, versus a file of custom schema in a different part of the repo.

The main reason I've come around to the side of using TXTDirect is the operational complexity is low compared to what I was building before.

stp-ip: To help with people understanding the records more easily and making copy-pasting easier. I moved all "static" config of the record to the start and the last value is the one for `to=` and hopefully can be easily grasped and changed.

stp-ip: > Roger. Sorry for being harsh. It has nothing to do with YOUR service and everything to do with A service. :)

No worries. Just wanted to give context and as mentioned never was the intention to add A service dependency.

stp-ip:
I think this is ready to test.
Uses delegation for the sub zone and the hosted service to decouple DNS testing and process from getting it running within the cluster. If this is a blocker on the other hand we can wait for the kuberneters deployment.
Only deploys to canary. Is that accessible outside? go.canary.k8s.io/oncall for example doesn't seem to work as expected.

stp-ip:
/hold

Ok I think this needs another 1on1 session before I feel like I understand the full picture.

stp-ip:
We declare a canary delegation both in canary.k8s.io and k8s.io. This feels incorrect from a DNS perspective. It keeps the files in sync so that might be the reasoning, but it's not actually deployed `dig NS canary.canary.k8s.io`. What am I missing here?

stp-ip:
/assign cblecker

Suggestion. Until we have the cluster, #227 and the canary process figured out we enable this on a test domain with the online service.
Using the canary might be possible as a testing domain, but maybe there is a totally separate one we can use. `x-k8s.io` ?
This way we could figure out the DNS workflow, once a cluster is up we can switch to self hosted before going into the final testing and deploy.

dns/zone-configs/go.canary.k8s.io.yaml
```diff
@@ -0,0 +1,57 @@
+---
+# The go.k8s.io zone
+# URL shortener service
+'':
+  type: CNAME
+  value: txtd.io.
+
+# TXTDirect data for shortener
+# Docs: https://about.txtdirect.org/docs/specification/#host-type
+_redirect:
```
thockin: I don't understand why some are `_redirect.foo` and some are `redirect.foo` ?

Can you please include comments to explain to people who are reading/editing this without being experts.

dns/zone-configs/k8s.io.yaml
```diff
@@ -32,6 +32,14 @@ canary:
   - ns-cloud-c3.googledomains.com.
   - ns-cloud-c4.googledomains.com.
 
+go.canary:
```
thockin: Why do we need this.  Why is it not sufficent to delegate `go` which implicitly does `go.canary` before going live?

dns/zone-configs/go.canary.k8s.io.yaml
stp-ip: Should always be `_redirect`. Fixing on next push.

dns/zone-configs/k8s.io.yaml
stp-ip: This was suggested to be done to be more explicit. so I followed that request.

stp-ip: Happy to remove this.

thockin: if I understand it properly, this is not necessary.  Do I understand?  When we push to canary.k8s.io, we should be able to test go.canary.k8s.io, distinct from go.k8s.io - right?  We should add tests to the DNS test to ensure that canaries are correct.

stp-ip: As far as I understand it with the explicit addition here it would enable a separate zone within google cloud dns.
But happy to drop this.
And yes we should be able to test go.canary.k8s.io distinctly from go.k8s.io. as we are already within a separate zone.

thockin:
Will a single caddy server handle both zones or do we need a canary caddy?

stp-ip:
Could handle both. Depends on what we want and how we want the general
staging process to look like.
The Caddy/TXTDirect instance doesn't care as it always looks up the record
for the incoming host url.

stp-ip:
Proposed canary test on deployment:
* [ ] check default redirect to kubernetes.io
* [ ] check redirect for correct redirect of go.k8s.io/oncall

bartsmykla:
Hi @stp-ip what is the status of that PR? Should we put it on the next k8s-infra-wg meeting's aggenda?

PR. Could you also rebase?

/assign

stp-ip:
This is ongoing in the background.
Won't make the next meeting due to travel so happy to push this to the agenda the meeting after.

A bit of context:
We are working on a full migration example of the k8s.io nginx config and using this as base for additional test coverage in TXTDirect. Additionally we created a TXT record spec validator that we can use within the canary tests to verify the TXT record is valid.
Furthermore we added additional e2e tests and a few simplification features are in the works such as aliasing to reduce DNS record duplication.

TXTDirect containers now have their own staging repo and we are working on building all dev images into the staging repo as well and get a production image into the production repo as well. 

Releases are monthly and we should be able to push for updates PRs and another review within Q2 as an estimate.

BenTheElder:
/uncc
out of bandwidth and it looks like there is other review / bikeshed

bartsmykla:
Hi @stp-ip can you give us some update over the status of this issue?

stp-ip:
Was blocked by a few things, but those cleaned up in the last few weeks. Will take another go at it once I worked through my backlog after I'm back in the office.

bartsmykla:
Thank you for update!

PR #227, opened 2019-04-18
[WIP] Add redirector service

stp-ip:
Add basic files for the redirector service.

ToDo:
* [ ] canary process
* [ ] deploy process

### DIFF
--- a/redirector.k8s.io/OWNERS
+++ b/redirector.k8s.io/OWNERS
@@ -1,0 +1,10 @@
+# See the OWNERS docs at https://go.k8s.io/owners
+
+approvers:
+- ixdy
+- thockin
+
+# limiting factor from being an approver is inability to deploy changes once PRs merge
+reviewers:
+- nikhita
+- stp-ip
--- a/redirector.k8s.io/README.md
+++ b/redirector.k8s.io/README.md
@@ -1,0 +1,34 @@
+Overview
+====
+This contains the TXTDirect configuration for redirector.k8s.io service.
+
+Used by go.k8s.io.
+
+Vanity URL(s)
+====
+
+|  | k8s.io | kubernetes.io |
+| --- | --- | --- |
+| Redirections | https://go.k8s.io | https://go.kubernetes.io |
+
+Redirections
+====
+- https://go.k8s.io/api-review
+- https://go.k8s.io/bot-commands
+- https://go.k8s.io/github-labels
+- https://go.k8s.io/good-first-issue
+- https://go.k8s.io/help-wanted
+- https://go.k8s.io/needs-ok-to-test
+- https://go.k8s.io/oncall
+- https://go.k8s.io/owners
+- https://go.k8s.io/partner-request
+- https://go.k8s.io/pr-dashboard
+- https://go.k8s.io/start
+- https://go.k8s.io/stuck-prs
+- https://go.k8s.io/test-health
+- https://go.k8s.io/test-history
+- https://go.k8s.io/triage
+
+NOTE: please see k8s.io/k8s.io/dns/go.k8s.io.yaml for redirect configuration
+
+Testing + Deploying TBD
--- a/redirector.k8s.io/configmap-txtdirect.yaml
+++ b/redirector.k8s.io/configmap-txtdirect.yaml
@@ -1,0 +1,24 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: txtdirect
+data:
+  caddyfile: |
+    :8080 {
+      tls off
+      txtdirect {
+        enable host path
+        redirect https://kubernetes.io
+        resolver 127.0.0.1:8053
+        prometheus
+      }
+      errors stdout
+    }
+    :8080/healthz {
+      tls off
+      status 200 /
+    }
+    :8080/probez {
+      tls off
+      status 200 /
+    }
--- a/redirector.k8s.io/deployment.yaml
+++ b/redirector.k8s.io/deployment.yaml
@@ -1,0 +1,107 @@
+apiVersion: extensions/v1beta1
+kind: Deployment
+metadata:
+  name: redirector-k8s-io
+  labels:
+    app: redirector-k8s-io
+    version: dirty
+spec:
+  revisionHistoryLimit: 2
+  replicas: 2
+  strategy:
+    rollingUpdate:
+      maxSurge: 4
+      maxUnavailable: 0
+  selector:
+    matchLabels:
+      app: redirector-k8s-io
+      component: txtdirect
+  template:
+    metadata:
+      labels:
+        app: redirector-k8s-io
+        component: txtdirect
+    spec:
+      securityContext:
+        fsGroup: 1337
+        runAsNonRoot: true
+      containers:
+      - name: txtdirect
+        image: gcr.io/txtdirect-223710/txtdirect:dev-7c870d0
+        securityContext:
+          runAsUser: 1337
+          readOnlyRootFilesystem: true
+          allowPrivilegeEscalation: false
+          capabilities:
+            drop: ["all"]
+        command: [ "/txtdirect", "--log=stdout", "--conf=/config/caddyfile" ]
+        resources:
+          requests:
+            memory: "16Mi"
+            cpu: "25m"
+          limits:
+            memory: "32Mi"
+            cpu: "50m"
+        readinessProbe:
+          httpGet:
+            path: /healthz
+            port: 8080
+            httpHeaders:
+              - name: X-Probez
+                value: liveness.k8s
+        livenessProbe:
+          httpGet:
+            path: /healthz
+            port: 8080
+            httpHeaders:
+              - name: X-Probez
+                value: liveness.k8s
+          initialDelaySeconds: 90
+        volumeMounts:
+        - name: config
+          mountPath: /config
+        - name: data
+          mountPath: /.caddy
+        ports:
+          - name: http
+            containerPort: 8080
+      - name: coredns
+        image: coredns/coredns:1.3.1
+        securityContext:
+          runAsUser: 1337
+          readOnlyRootFilesystem: true
+          allowPrivilegeEscalation: false
+          capabilities:
+            drop: ["all"]
+        args: [ "--conf=/config-cache/corefile" ]
+        resources:
+          requests:
+            memory: "16Mi"
+            cpu: "25m"
+          limits:
+            memory: "32Mi"
+            cpu: "50m"
+        livenessProbe:
+          httpGet:
+            path: /health
+            port: 8153
+            httpHeaders:
+              - name: X-Probez
+                value: liveness.k8s
+          initialDelaySeconds: 90
+        volumeMounts:
+        - name: config-cache
+          mountPath: /config-cache
+        ports:
+          - name: dns
+            containerPort: 8053
+      volumes:
+      - name: config-cache
+        configMap:
+          name: txtdirect-cache
+      - name: config
+        configMap:
+          name: txtdirect
+      - name: data
+        persistentVolumeClaim:
+          claimName: txtdirect
--- a/redirector.k8s.io/ingress-canary.yaml
+++ b/redirector.k8s.io/ingress-canary.yaml
@@ -1,0 +1,15 @@
+apiVersion: extensions/v1beta1
+kind: Ingress
+metadata:
+  name: redirector-k8s-io
+  labels:
+    app: redirector-k8s-io
+  namespace: k8s-io-canary
+  annotations:
+    kubernetes.io/ingress.global-static-ip-name: redirector-k8s-io-ingress-canary
+spec:
+  tls:
+  - secretName: redirector-k8s-io-tls
+  backend:
+    serviceName: redirector-k8s-io
+    servicePort: http
--- a/redirector.k8s.io/ingress-prod.yaml
+++ b/redirector.k8s.io/ingress-prod.yaml
@@ -1,0 +1,15 @@
+apiVersion: extensions/v1beta1
+kind: Ingress
+metadata:
+  name: redirector-k8s-io
+  labels:
+    app: redirector-k8s-io
+  namespace: redirector-k8s-io-prod
+  annotations:
+    kubernetes.io/ingress.global-static-ip-name: redirector-k8s-io-ingress-prod
+spec:
+  tls:
+  - secretName: redirector-k8s-io-tls
+  backend:
+    serviceName: redirector-k8s-io
+    servicePort: http
--- a/redirector.k8s.io/certificate-canary.yaml
+++ b/redirector.k8s.io/certificate-canary.yaml
@@ -1,0 +1,14 @@
+apiVersion: certmanager.k8s.io/v1alpha1
+kind: Certificate
+metadata:
+  name: k8s-io-selfsigned
+  namespace: redirector-k8s-io-canary
+spec:
+  secretName: redirector-k8s-io-tls
+  issuerRef:
+    kind: ClusterIssuer
+    name: selfsigning-issuer
+  commonName: redirector.k8s.io
+  dnsNames:
+  - go.k8s.io
+  - go.kubernetes.io
--- a/redirector.k8s.io/certificate-prod.yaml
+++ b/redirector.k8s.io/certificate-prod.yaml
@@ -1,0 +1,14 @@
+apiVersion: certmanager.k8s.io/v1alpha1
+kind: Certificate
+metadata:
+  name: redirector-k8s-io-letsencrypt-prod
+  namespace: k8s-io-prod
+spec:
+  secretName: redirector-k8s-io-tls
+  issuerRef:
+    kind: ClusterIssuer
+    name: letsencrypt-prod
+  commonName: redirector.k8s.io
+  dnsNames:
+  - go.k8s.io
+  - go.kubernetes.io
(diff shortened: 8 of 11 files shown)


redirector.k8s.io/deployment.yaml
```diff
@@ -0,0 +1,107 @@
...
+metadata:
+  name: redirector-k8s-io
+  labels:
+    app: redirector-k8s-io
+    version: dirty
+spec:
+  revisionHistoryLimit: 2
+  replicas: 2
+  strategy:
+    rollingUpdate:
```
thockin: This block doesn't seem to make sense to me

redirector.k8s.io/deployment.yaml
```diff
@@ -0,0 +1,107 @@
...
+      app: redirector-k8s-io
+      component: txtdirect
+  template:
+    metadata:
+      labels:
+        app: redirector-k8s-io
+        component: txtdirect
+    spec:
+      securityContext:
+        fsGroup: 1337
```
thockin: why?  comment

redirector.k8s.io/deployment.yaml
```diff
@@ -0,0 +1,107 @@
...
+        runAsNonRoot: true
+      containers:
+      - name: txtdirect
+        image: gcr.io/txtdirect-223710/txtdirect:dev-7c870d0
+        securityContext:
+          runAsUser: 1337
```
thockin: why?  comment

redirector.k8s.io/deployment.yaml
```diff
@@ -0,0 +1,107 @@
...
+          capabilities:
+            drop: ["all"]
+        command: [ "/txtdirect", "--log=stdout", "--conf=/config/caddyfile" ]
+        resources:
+          requests:
+            memory: "16Mi"
+            cpu: "25m"
+          limits:
+            memory: "32Mi"
+            cpu: "50m"
```
thockin: don't set CPU limit

redirector.k8s.io/deployment.yaml
```diff
@@ -0,0 +1,107 @@
...
+          allowPrivilegeEscalation: false
+          capabilities:
+            drop: ["all"]
+        command: [ "/txtdirect", "--log=stdout", "--conf=/config/caddyfile" ]
+        resources:
+          requests:
+            memory: "16Mi"
+            cpu: "25m"
+          limits:
+            memory: "32Mi"
```
thockin: don't set memory limit > request

redirector.k8s.io/deployment.yaml
```diff
@@ -0,0 +1,107 @@
...
+            cpu: "50m"
+        readinessProbe:
+          httpGet:
+            path: /healthz
+            port: 8080
+            httpHeaders:
```
thockin: comment on this

redirector.k8s.io/deployment.yaml
```diff
@@ -0,0 +1,107 @@
...
+              - name: X-Probez
+                value: liveness.k8s
+        livenessProbe:
+          httpGet:
+            path: /healthz
+            port: 8080
+            httpHeaders:
```
thockin: comment on this

redirector.k8s.io/deployment.yaml
```diff
@@ -0,0 +1,107 @@
...
+          mountPath: /config
+        - name: data
+          mountPath: /.caddy
+        ports:
+          - name: http
+            containerPort: 8080
+      - name: coredns
+        image: coredns/coredns:1.3.1
+        securityContext:
+          runAsUser: 1337
```
thockin: why?  comment

redirector.k8s.io/deployment.yaml
```diff
@@ -0,0 +1,107 @@
...
+          readOnlyRootFilesystem: true
+          allowPrivilegeEscalation: false
+          capabilities:
+            drop: ["all"]
+        args: [ "--conf=/config-cache/corefile" ]
+        resources:
+          requests:
```
thockin: same comments as above

redirector.k8s.io/deployment.yaml
```diff
@@ -0,0 +1,107 @@
...
+            memory: "16Mi"
+            cpu: "25m"
+          limits:
+            memory: "32Mi"
+            cpu: "50m"
+        livenessProbe:
+          httpGet:
+            path: /health
+            port: 8153
+            httpHeaders:
```
thockin: same comments as above

redirector.k8s.io/deployment.yaml
stp-ip: Happy to go with the default of 25% maxUnavailable and 25% maxSurge aka we remove this block.

redirector.k8s.io/deployment.yaml
stp-ip: That got juggled I think. Doesn't make sense as it is here. Will fix it.

redirector.k8s.io/deployment.yaml
stp-ip: httpHeaders helpful for debugging and to distinguish, what might be kept within the logs.
Ok to remove this.
Also copy+paste error as the header should be `readiness.k8s`

redirector.k8s.io/deployment.yaml
stp-ip: Generally used to make sure files are owned correctly. Used in other deployments that we have.
If this feels unnecessary, happy to remove.

redirector.k8s.io/deployment.yaml
stp-ip: Run as non root and as specified user to overwrite any specific user configuration set in dockerfiles. Being explicit here was the goal. Happy for thoughts.

redirector.k8s.io/deployment.yaml
stp-ip: Same as above.

redirector.k8s.io/deployment.yaml
stp-ip: Be specific on what user runs the container instead of relying on dockerfile settings etc.

redirector.k8s.io/deployment.yaml
stp-ip: ack

redirector.k8s.io/deployment.yaml
stp-ip: ack

stp-ip:
It's the backend service for go.k8s.io. The specific config is all DNS based. It could also be used as a backend for redirects such as blog.k8s.io later on etc. So that was the reason to call it redirector.k8s.io and only add the ingress for go.k8s.io etc.

redirector.k8s.io/deployment.yaml
thockin: I meant why this number -- I know what the feature does :)

redirector.k8s.io/deployment.yaml
thockin: I just want a comment - even "# Pick any non-root UID"

redirector.k8s.io/deployment.yaml
thockin: I see what it is for - the paste error makde it hard.  Don't care, but we don't do it else where

redirector.k8s.io/ingress-canary.yaml
```diff
@@ -0,0 +1,15 @@
+apiVersion: extensions/v1beta1
+kind: Ingress
+metadata:
+  name: redirector-k8s-io
+  labels:
+    app: redirector-k8s-io
+  namespace: k8s-io-canary
```
cblecker: wrong namespace

redirector.k8s.io/ingress-prod.yaml
```diff
@@ -0,0 +1,15 @@
+apiVersion: extensions/v1beta1
+kind: Ingress
+metadata:
+  name: redirector-k8s-io
+  labels:
+    app: redirector-k8s-io
+  namespace: redirector-k8s-io-prod
```
cblecker: doesn't match namespace above?

redirector.k8s.io/deployment.yaml
stp-ip: Thought you wanted to get the reasoning behind why the feature was used ;)
Basically just a random number that wasn't used for something specific yet and was easy to remember.

redirector.k8s.io/configmap-txtdirect.yaml
```diff
@@ -0,0 +1,24 @@
...
+kind: ConfigMap
+metadata:
+  name: txtdirect
+data:
+  caddyfile: |
+    :8080 {
+      tls off
+      txtdirect {
+        enable host path
+        redirect https://kubernetes.io
```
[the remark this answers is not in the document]
stp-ip: Yes. Anything that doesn't match get's redirected to kubernetes.io as default for the instance. This includes any request that can't be matched to a hostname.
Another option is to 404.

For go.k8s.io there is a default redirect configurable within the DNS records, which should take precedence as long as the hostname go.k8s.io matches. (path type record.To)

redirector.k8s.io/OWNERS
```diff
@@ -0,0 +1,10 @@
+# See the OWNERS docs at https://go.k8s.io/owners
+
+approvers:
+- ixdy
```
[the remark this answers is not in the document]
stp-ip: Removed for now. Happy to add back in, if requested.
Copied from the old OWNERS as I felt this is moving out currently assigned responsibilities.

thockin:
Is this ready for review?  I lost track of whether we were doing it or not...

stp-ip:
Ready for another review and still something on our backlog afaik.

This is the self-hosting part, which would run in the cluster. The second part are the DNS changes in #225.

Both changes should be in the ready to review and ready to test after.

There are still some left over things that are unclear. RBAC, e2e testing, deploy automation (both for DNS and K8s) etc., which are denoted within the PRs task list.

redirector.k8s.io/OWNERS
```diff
@@ -0,0 +1,9 @@
+# See the OWNERS docs at https://go.k8s.io/owners
+
+approvers:
+- thockin
+- cblecker
+
+# limiting factor from being an approver is inability to deploy changes once PRs merge
```
thockin: If we deploy this to existing clusters, yes.  If we assume this does not launch until new cluster is up, I don't see why you could be approver

redirector.k8s.io/README.md
```diff
@@ -0,0 +1,36 @@
...
+Vanity URL(s)
+====
+
+|  | k8s.io | kubernetes.io |
+| --- | --- | --- |
+| Redirections | https://go.k8s.io | https://go.kubernetes.io |
+
+Redirections
+====
+- https://go.k8s.io/api-review
```
thockin: I don't think we should document them here - just let the DNS config be canonical?

redirector.k8s.io/configmap-txtdirect.yaml
```diff
@@ -0,0 +1,25 @@
...
+      errors stdout
+    }
+    :8080/healthz {
+      tls off
+      status 200 /
+    }
+    :8080/probez {
+      tls off
+      status 200 /
+    }
```
thockin: missing EOL

redirector.k8s.io/deployment.yaml
```diff
@@ -0,0 +1,101 @@
+apiVersion: extensions/v1beta1
+kind: Deployment
+metadata:
+  name: redirector-k8s-io
+  namespace: redirector-k8s-io-prod
+  labels:
+    app: redirector-k8s-io
+    version: dirty
```
thockin: ?

redirector.k8s.io/deployment.yaml
```diff
@@ -0,0 +1,101 @@
+apiVersion: extensions/v1beta1
+kind: Deployment
+metadata:
+  name: redirector-k8s-io
+  namespace: redirector-k8s-io-prod
```
thockin: most of the YAML we store doesn't have a namespace specified, and depends on users setting context

redirector.k8s.io/deployment.yaml
```diff
@@ -0,0 +1,101 @@
...
+        app: redirector-k8s-io
+        component: txtdirect
+    spec:
+      securityContext:
+        # Assign files to non-root supplementary GID
+        fsGroup: 1337
+        runAsNonRoot: true
+      containers:
+      - name: txtdirect
+        image: gcr.io/txtdirect-223710/txtdirect:dev-00984e2
```
thockin: we'll have to get a binary to upload to k8s.gcr.io.  We can grant you a staging repo and use the promoter

redirector.k8s.io/deployment.yaml
```diff
@@ -0,0 +1,101 @@
...
+          # Run as any non-root user UID
+          runAsUser: 1337
+          readOnlyRootFilesystem: true
+          allowPrivilegeEscalation: false
+          capabilities:
+            drop: ["all"]
+        command: [ "/txtdirect", "--log=stdout", "--conf=/config/caddyfile" ]
+        resources:
+          requests:
+            memory: "32Mi"
```
thockin: better not to set a request for memory, just limit.  That way no mistakes can end up with them being different

redirector.k8s.io/deployment.yaml
```diff
@@ -0,0 +1,101 @@
...
+        readinessProbe:
+          httpGet:
+            path: /healthz
+            port: 8080
+        livenessProbe:
+          httpGet:
+            path: /healthz
+            port: 8080
+          initialDelaySeconds: 90
```
thockin: why so long?

redirector.k8s.io/deployment.yaml
```diff
@@ -0,0 +1,101 @@
...
+            memory: "32Mi"
+        readinessProbe:
+          httpGet:
+            path: /ready
+            port: 8181
+        livenessProbe:
+          httpGet:
+            path: /health
+            port: 8282
+          initialDelaySeconds: 90
```
thockin: why so long?

redirector.k8s.io/README.md
```diff
@@ -0,0 +1,36 @@
+Overview
+====
+This contains the TXTDirect configuration for redirector.k8s.io backend service.
```
thockin: Can you include a brief overview of how it works?  E.g. why does it need a DNS server running in the pod?

redirector.k8s.io/deployment.yaml
stp-ip: Currently using master, once we agree on feasibility will switch to the latest stable release.

redirector.k8s.io/deployment.yaml
stp-ip: +1

redirector.k8s.io/deployment.yaml
stp-ip: Was taken from a production deployment, where it mirrored in DNS data locally.
Will reduce to 10s

redirector.k8s.io/README.md
stp-ip: Will do.

Sidenote: DNS server is not strictly necessary, but helps with latency due to local cache vs Google DNS cache. Also gives us more insight into requested DNS records etc. so generally recommended.

redirector.k8s.io/deployment.yaml
stp-ip: Happy to drop this.

redirector.k8s.io/OWNERS
stp-ip: > Why would this not be "go.k8s.io" dir, since that is the name we have already?

Cause the redirector can do more than just serve go.k8s.io. Therefore making it a service distinct from the first use case aka go.k8s.io seemed like a good solution.

thockin:
such as?

stp-ip:
Example: blog.kubernetes.io
In general most of the current nginx config could be replaced in theory.
Even specific sig links and such could be done.

thockin:
ok.  Can we shorten just a bit to redirect.k8s.io, then? :)

stp-ip:
> ok. Can we shorten just a bit to redirect.k8s.io, then? :)

That would have been the first choice, but the current nginx is located/hosted at `redirect.k8s.io` and to make an iterative migration possible I opted for `redirector.k8s.io` instead.
Can also use `txtd.k8s.io`, but that felt a bit explanatory.

stp-ip:
Proposed canary test on dns change:
* [ ] check oncall TXT record is formatted correctly
* [ ] (optionally) check all TXT records to be formatted correctly (small go program doing the parsing that TXTDirect does)

stp-ip:
Additionally with terraform we could script a production monitoring job with stackdriver for the kubernetes.io defaulting, /oncall and maybe various other URLs being served by the redirector later on.

redirector.k8s.io/deployment.yaml
stp-ip: PR: https://github.com/kubernetes/k8s.io/pull/354

redirector.k8s.io/OWNERS
stp-ip: Not sure if I understood this correctly. Do you mean I should remove the reviewer role and add myself to approver for simplification assuming this will only deployed to the new cluster?
(went ahead and did this change)

bartsmykla:
@stp-ip the same question as in: https://github.com/kubernetes/k8s.io/pull/225#issuecomment-590224333 - what do we need to move this forward? :-)

stp-ip:
Responded with the full context over in #225

stp-ip:
/remove-lifecycle stale

Have this on my radar still and looking forward to being able to put the finishing touches on.

PR #233, opened 2019-04-24
Specify bash as our shell

justinsb:
In case the runner uses a different shell.


justinsb:
Thanks @cblecker  - switched to `usr/bin/env bash` for our MacOS friends :-)

thockin:
Do we use bash-isms?

PR #236, opened 2019-04-26
Enable bucket-only policy for GCS buckets

thockin:
Simpler auth is better - we do not want per-object rules.

### DIFF
--- a/k8s.gcr.io/lib.sh
+++ b/k8s.gcr.io/lib.sh
@@ -166,6 +166,7 @@ function ensure_repo() {
     fi
 
     gsutil iam ch allUsers:objectViewer "${bucket}"
+    gsutil bucketpolicyonly set on "${bucket}"
 }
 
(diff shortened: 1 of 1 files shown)


cblecker:
/lgtm
/approve

/hold
holding if @justinsb has any comments

justinsb:
/lgtm

This seems right, because:

* It means our audit reports won't have to deal with object ACLs
* We can always turn on ACLs if we ever do need it

I did read a bit about it, and it does seem that some services might have problems e.g. exporting to these buckets https://cloud.google.com/storage/docs/bucket-policy-only#restrictions  But we can cross that bridge when we come to it; I don't think we would have to apply this to every bucket, just the project artifact buckets.

/unhold

thockin:
Woops, I enabled it on the main project. Will fix.

thockin:
Fixed :)

PR #238, opened 2019-04-30
Extend prod repo script to create prod bucket also

justinsb:
Includes #222

### DIFF
--- a/k8s.gcr.io/ensure-prod-repo.sh
+++ b/k8s.gcr.io/ensure-prod-repo.sh
@@ -14,7 +14,8 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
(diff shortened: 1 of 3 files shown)


k8s.gcr.io/ensure-prod-repo.sh
```diff
@@ -40,6 +41,9 @@ TRASH_PROJECT="k8s-gcr-graveyard"
 
 ALL_PROJECTS="${TEST_PROJECT} ${PROD_PROJECT} ${TRASH_PROJECT}"
 
+# GCS bucket for prod
+PROD_BUCKET=gs://k8s-prod-artifacts
```
thockin: We may want a "trash" bucket eventually, too

PR #247, opened 2019-05-01
Storage renames

thockin:
Fix some bugs in the GCLB script.

Rename prod projects to not say "gcr" but "artifacts".

@listx - I will have to get you a new credential for the promoter, so I will not nuke the old projects until we talk.

### DIFF
--- a/k8s.gcr.io/ensure-prod-gclb.sh
+++ b/k8s.gcr.io/ensure-prod-gclb.sh
@@ -34,14 +34,14 @@ if [ $# != 0 ]; then
 fi
 
 # The GCP project name.
-PROD_PROJECT="k8s-gcr-prod"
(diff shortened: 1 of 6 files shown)


listx:
I just need to relay those new secrets to the Prow folks so that they can add the new secrets for the trusted Prow cluster.
Then I need to update the jobs to use the new secrets.

I think that's it? I don't think the timing of the above vs. this PR matters that much because we don't have any images in gcr.io/k8s-gcr-prod (the old one).

/lgtm

thockin:
OK.  Ping me side-band when you are in and I will get you a new key, and
then nuke the old projects

spiffxp:
/approve
/hold
Please /hold cancel whenever you're ready for this to merge

PR #252, opened 2019-05-08
GCS: configure index.html as default page on bucket

justinsb:
This stops us serving listings by default (which will end pretty big!)

### DIFF
--- a/k8s.gcr.io/lib.sh
+++ b/k8s.gcr.io/lib.sh
@@ -191,6 +191,18 @@ function ensure_gcs_bucket() {
     gsutil iam ch allUsers:objectViewer "${bucket}"
 }
 
+# Sets the web policy on the bucket, including a default index.html page
+# $1: The bucket
+function ensure_gcs_web_policy() {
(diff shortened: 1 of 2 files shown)


k8s.gcr.io/lib.sh
```diff
@@ -191,6 +191,20 @@ function ensure_gcs_bucket() {
...
 
+# Sets the web policy on the bucket, including a default index.html page
+# $1: The GCP project
+# $2: The bucket
+function ensure_gcs_web_policy() {
+    if [ $# -lt 2 -o -z "$1" -o -z "$2" ]; then
+        echo "ensure_gcs_web_policy(project, bucket) requires 2 arguments" >&2
+        return 1
+    fi
+    project="$1"
```
thockin: never usedf?

justinsb: So...

It is perhaps surprising that the bucket is globally named and that you don't need a project, and so in the GCS functions I started passing the project anyway for homogeneity with the GCR functions.

But, yes, it is perhaps a little silly.

thockin:
closing in favor of #253

PR #258, opened 2019-05-10
Use new kubernetes.io googlegroups

thockin:
I have started activating this is prod.  Found one mistake in group creating, will send a PR to fix that one.

PR #260, opened 2019-05-10
Staging storage loop

thockin:
this includes #258 because of a conflict

### DIFF
--- a/cert-manager/letsencrypt-prod.yaml
+++ b/cert-manager/letsencrypt-prod.yaml
@@ -4,7 +4,7 @@ metadata:
   name: letsencrypt-prod
 spec:
   acme:
-    email: k8s-infra-team-private+letsencrypt@googlegroups.com
+    email: k8s-infra-team-private+letsencrypt@kubernetes.io
     server: https://acme-v02.api.letsencrypt.org/directory
     privateKeySecretRef:
       name: letencrypt-prod-account-key
--- a/cert-manager/letsencrypt-staging.yaml
+++ b/cert-manager/letsencrypt-staging.yaml
@@ -4,7 +4,7 @@ metadata:
   name: letsencrypt-staging
 spec:
   acme:
-    email: k8s-infra-team-private+letsencrypt@googlegroups.com
+    email: k8s-infra-team-private+letsencrypt@kubernetes.io
     server: https://acme-staging-v02.api.letsencrypt.org/directory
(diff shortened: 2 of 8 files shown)


cblecker:
/lgtm
/approve
/hold

Does this need to be deployed?

thockin:
Everything except cip-test is live.  As soon as groups sync, I can do that
one.

*From: *Kubernetes Prow Robot <notifications@github.com>
*Date: *Thu, May 9, 2019 at 10:02 PM
*To: *kubernetes/k8s.io
*Cc: *Tim Hockin, Author

[APPROVALNOTIFIER] This PR is *APPROVED*

cert-manager/letsencrypt-prod.yaml
```diff
@@ -4,7 +4,7 @@ metadata:
   name: letsencrypt-prod
 spec:
   acme:
-    email: k8s-infra-team-private+letsencrypt@googlegroups.com
+    email: k8s-infra-team-private+letsencrypt@kubernetes.io
```
munnerz: This change has prompted me to take a look at how updating the email address field is handled in cert-manager.

We actually don't properly handle this case, and so your email address *will not* be updated with the ACME server (although it is possible to do this via the ACME API).

I've opened an issue to track this our side, and have added it into the v0.9 milestone (due in 1-1.5mths): https://github.com/jetstack/cert-manager/issues/1667

For the time being, I'd recommend:

1) manually making the API call to update your email (😞)
2) keeping the old email address working until https://github.com/jetstack/cert-manager/issues/1667 lands (probably best to switch it back here too, so this isn't forgotten)
3) deleting your ACME private key, which will cause a new private key & account to be automatically created with the new email address (this will also mean you'll have to re-authorize your domains, although you'll most likely need to do this anyway as *authorizations* expire after 60d).

cc @ixdy

thockin:
I have no context on this stuff -- @ixdy any bandwidth to help?

cert-manager/letsencrypt-prod.yaml
ixdy: option 3 seems like the least amount of work. do I need to do anything special to get the domains reauthorized, or would that just happen automatically?

thockin: ping @munnerz - I'd love to take this off my "thing I am worrying about" list

PR #266, opened 2019-05-10
Add a script to recreate the "main" GCP project

thockin:
This covers everything but DNS, which I will do separately.

When we have clusters, some of this might move out to a clusters script.

All these scripts need a new dir, rather than GCR :)

### DIFF
--- a/k8s.gcr.io/ensure-main-project.sh
+++ b/k8s.gcr.io/ensure-main-project.sh
@@ -1,0 +1,179 @@
+#!/usr/bin/env bash
+#
+# Copyright 2019 The Kubernetes Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# This script creates & configures the "main" GCP project for Kubernetes.
+
+set -o errexit
+set -o nounset
+set -o pipefail
+
+SCRIPT_DIR=$(dirname "${BASH_SOURCE[0]}")
+. "${SCRIPT_DIR}/lib.sh"
+
+function usage() {
+    echo "usage: $0" > /dev/stderr
+    echo > /dev/stderr
+}
+
+if [ $# != 0 ]; then
+    usage
+    exit 1
+fi
+
+# The GCP project name.
+PROJECT="kubernetes-public"
+
+# The BigQuery dataset for billing data.
+BQ_BILLING_DATASET="kubernetes_public_billing"
+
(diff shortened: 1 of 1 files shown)


thockin:
HOLD until we are really confident in this.

thockin:
I have run this and created kubernetes-public-thockin1 which seems to parallel kubernetes-public for everything I have looked at

k8s.gcr.io/ensure-main-project.sh
```diff
@@ -0,0 +1,172 @@
...
+
+color 6 "Empowering cluster admins"
+gcloud projects add-iam-policy-binding "${PROJECT}" \
+    --member "group:${CLUSTER_ADMINS_GROUP}" \
+    --role roles/compute.viewer
+gcloud projects add-iam-policy-binding "${PROJECT}" \
+    --member "group:${CLUSTER_ADMINS_GROUP}" \
+    --role roles/container.admin
+if ! gcloud --project "${PROJECT}" iam roles describe ServiceAccountLister >/dev/null 2>&1; then
+    # Don't use `yes` here, it causes SIGPIPE -> pipefail -> errexit
```
justinsb: Does `--quiet` work?

k8s.gcr.io/ensure-main-project.sh
```diff
@@ -0,0 +1,172 @@
...
+    {
+      "groupByEmail": "${ACCOUNTING_GROUP}",
+      "role": "roles/bigquery.user"
+    }
+  ]
+}
+__EOF__
+
+FINAL=$(tempfile -p k8s-infra-bq-access-new)
+jq -s '.[0].access + .[1].access | { access: . }' "${CUR}" "${ENSURE}" > "${FINAL}"
```
justinsb: A comment would be great here to explain what is happening!  We're merging the existing and new permissions?  (And meta: any reason not just to set the permissions explicitly so we do delete any unwanted permissions?)

justinsb:
I think this looks right.  With DNS excluded, the risk is primarily for the k8s cluster(s) we are running?  (I am assuming there's nothing that important in gcr or gce yet)?

Are you planning on actually deleting the project and recreating it?

/lgtm

k8s.gcr.io/ensure-main-project.sh
thockin: it does!  TIL

k8s.gcr.io/ensure-main-project.sh
thockin: comment added

thockin:
new push is up

I do not plan to nuke the project, just trying to make sure everything we rely on is recreatable and explainable.

thockin:
Removed the hold.   PTAL.  I have a followup to move this all to a new home and do some cleanups.

thockin:
ping

k8s.gcr.io/ensure-main-project.sh
dims: @thockin I don't see the comment. forgot to push?

dims:
LGTM, still missing a comment requested by @justinsb

k8s.gcr.io/ensure-main-project.sh
thockin: Weird, I swear I wrote a comment.  Added now.

Issue #270, opened 2019-06-03
DNS REQUEST: relnotes.k8s.io


jeefy:
### Type of DNS update:
Create

### Domain being modified:
k8s.io

### New DNS Record:
```yaml
# https://github.com/kubernetes-sigs/release-notes
relnotes:
  type: CNAME
  value: kubernetes-sigs-relnotes.netlify.com
  ```
### Reason for update:
Creation of a user-facing release notes website. As this will be linked to and distributed for future releases, a shorter URL would be preferred and thus should not fall under the `sigs` subdomain.

Relevant [KEP](https://github.com/kubernetes/enhancements/blob/master/keps/sig-release/20190331-release-notes.md)

cc @spiffxp @justaugustus @onyiny-ang

justaugustus:
Approved from SIG Release: https://github.com/kubernetes/org/issues/873#issuecomment-498117333

nikhita:
>   value: kubernetes-sigs-relnotes.netlify.com

The netlify link for this would be `kubernetes-sigs-release-notes.netlify.com` to stay consistent. Hope that works!

jeefy:
That works. The Netlify link won't be directly used anyhow. :) Thanks @nikhita!!

thockin:
Is the netlify up and running?

thockin:
I see it is active :)

PR #272, opened 2019-06-03
Create relnotes.k8s.io for k-sigs/release-notes

nikhita:
Fixes https://github.com/kubernetes/k8s.io/issues/270
xref https://github.com/kubernetes/org/issues/873

The site is live at http://kubernetes-sigs-release-notes.netlify.com/

/cc @dims @cblecker @jeefy @mrbobbytables

### DIFF
--- a/dns/zone-configs/k8s.io.yaml
+++ b/dns/zone-configs/k8s.io.yaml
@@ -170,6 +170,10 @@ issue:
 issues:
   type: CNAME
   value: redirect.k8s.io.
+# https://github.com/kubernetes-sigs/kind docs (@bentheelder, @munnerz)
+kind.sigs:
+  type: CNAME
+  value: k8s-kind.netlify.com.
 # Running in GKE (@dchen1107).
 node-perf-dash:
   type: A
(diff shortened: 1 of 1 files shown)


dns/zone-configs/k8s.io.yaml
```diff
@@ -240,3 +240,7 @@ velodrome:
 kind.sigs:
   type: CNAME
   value: k8s-kind.netlify.com.
+# https://github.com/kubernetes-sigs/release-notes docs (@jeefy)
+relnotes:
```
thockin: Please keep this list alphabetical.

We generally try to make the same name work as both k8s.io and kubernetes.io, but given that this is external I presume we won't want TLS for both ?

nikhita: Oh this was already added in https://github.com/kubernetes/k8s.io/pull/264/files#diff-72d1f1b649e651304c1e89efa0533680

/close

thockin:
This was a screwup on my end.  I patched it into a branch that had other changes and that branch merged, but I messed up.

It merged the wrong value too

PR #274, opened 2019-06-04
Add members to blog@kubernetes.io

nikhita:
Ref: https://github.com/kubernetes/community/issues/3763
Emails are as per what @kbarnard10 DM'ed me with.

/cc @dims @kbarnard10 @philips 
/assign @dims 

/hold
I'm not entirely sure if this is the right way:
- I cannot access https://groups.google.com/a/kubernetes.io/forum/#!forum/blog
- I am not adding myself in this list because I don't need to be there...but I'm not sure if there would be any errors since I'm the admin right now?

### DIFF
--- a/groups/groups.yaml
+++ b/groups/groups.yaml
@@ -3,6 +3,15 @@
 # group is prefixed with "k8s-infra" to avoid polluting the other existing gsuite
 # mailing lists.
 groups:
+  - email-id: blog@kubernetes.io
+    members:
+      - codyclark@google.com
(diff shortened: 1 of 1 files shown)


philips:
@nikhita OK, I made @parispittman Owner. She can do stuff from here.

I am not familiar with all of this automation though so I need someone to tell me what is happening here.

nikhita:
Closing since this automation is only for email aliases with the `k8s-infra` prefix.

Members have been added manually by @dims.

/close

PR #279, opened 2019-06-11
Clean up GCP k8s-infra scripts

thockin:
Followup to #266 

* Move all scripys to infra/gcp
* Clean up some of the logic in lib.sh (mostly renames for clarity)
* End goal: to recreate all the infrastructure, run `for x in ensure-*.sh; do ./$x; done`; dot back, enjoy the show

NB: DNS is still missing from this, but I think everything else is here

TODO: Make it even more declarative by extract things like the list of all staging projects from the scripts


infra/gcp/ensure-main-project.sh
```diff
@@ -0,0 +1,173 @@
...
+    },
+    {
+      "groupByEmail": "${ACCOUNTING_GROUP}",
+      "role": "roles/bigquery.user"
+    }
+  ]
+}
+__EOF__
+
+FINAL=$(tempfile -p k8s-infra-bq-access-new)
```
ameukam: I'm not sure `tempfile` is a common program on many Linux distributions and i can't find it for Mac OS. Maybe a warning in a README as prerequisite ?

thockin:
How about `mktemp -p /tmp` - is that available?

thockin:
Ping.  Prereq merged.

cblecker:
Haven't reviewed this myself, but @thockin says it's just moving scripts around and @justinsb has reviewed it.

/rubber-stamp
/approve

PR #283, opened 2019-06-17
dns: remove devstats CNAME entries

tpepper:
The official preferred url is https://k8s.devstats.cncf.io and after
changes recently using the Kubernetes org CNAMEs results in a
certificate error.  No docs appear to reference the vanity url, and since
it now errors and is not a supported config on the cncf/devstats side,
the CNAMEs should be removed.

Signed-off-by: Tim Pepper <tpepper@vmware.com>

### DIFF
--- a/dns/zone-configs/k8s.io.yaml
+++ b/dns/zone-configs/k8s.io.yaml
@@ -116,10 +116,6 @@ cs:
   type: A
   value: 147.75.97.58
 # sig-contributor-experience
-devstats:
-  type: CNAME
-  value: k8s.devstats.cncf.io.
-# sig-contributor-experience
 discuss:
   type: CNAME
   value: kubernetes.hosted-by-discourse.com.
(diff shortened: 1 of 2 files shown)


cblecker:
I think this still used in a lot of places.. do we know why it broke?

tpepper:
> I think this still used in a lot of places.. do we know why it broke?

I grepped a number of repos and didn't find meaningful references to this url.   There's a few old docs, eg: https://github.com/search?q=org%3Akubernetes+org%3Akubernetes+%22devstats.k8s.io%22&type=Code
but I don't think we update those type of old doc?

Devstats/CNCF folks say https://k8s.devstats.cncf.io is the only url they support and want used for quite some time.  Since almost a year and a half ago whenever I used the k8s.io vanity url they told me not to. 
 And they run with a wildcard cert of *.devstats.cncf.io.

My guess is they previously had in their web server config an URL rewrite rule or a redirect to move the vanity url over to their preferred url?  They say that couldn't have been and it would have been on a k8s.io web server.  But, given that it's CNAME and the associated A record is under their control and points to their server IP, I don't think a client would ever interact with a k8s web server for this url.

/shrug

cblecker:
/approve
/hold

I'll deploy this tomorrow then.

PR #290, opened 2019-07-08
Set bucketpolicyonly on non-GCR buckets

thockin:
I noticed this was not set.  Seems like we want it for consistency and simplicity?


thockin:
I will re-reconcile now.

PR #313, opened 2019-07-12
Don't allow invitations or adds from a group

justinsb:
We will remove additional members on the next reconcile anyway, so no point
allowing other membership flows.


justinsb:
Builds on #313 - split for readability (security critical!)

dims:
/approve

@cblecker can you please `/lgtm`

PR #331, opened 2019-08-09
Allow promoter test to access staging

thockin:
It's a bit odd, but I think it's better than yet another service account and credential to track.

I have NOT applied this yet, will wait for @listx to be ready to confirm.

PTAL, but I will hold it until confirmed.

/hold


thockin:
Applied and confirmed

PR #334, opened 2019-08-14
GCS: Create disaster recovery script

justinsb:
We create an independent copying program that will copy binary
artifacts without overwriting (so that we won't overwrite our
backups).

### DIFF
--- a/artifacts/dev/disaster-recovery/copy-buckets.py
+++ b/artifacts/dev/disaster-recovery/copy-buckets.py
@@ -1,0 +1,66 @@
+#!/usr/bin/env python3
+
+import os, sys
+import google.auth
+import google.cloud.storage
+import google.oauth2.credentials
+
+if len(sys.argv) != 3:
+  print("syntax: %s <src> <dest>" % (sys.argv[0]))
+  sys.exit(1)
+
+storage_client = google.cloud.storage.Client()
+
+# Build buckets, stripping gs:// prefix
+src_bucket = sys.argv[1]
+if src_bucket.startswith("gs://"):
+  src_bucket = src_bucket[5:]
+src_bucket = storage_client.get_bucket(src_bucket)
+
+dest_bucket = sys.argv[2]
+if dest_bucket.startswith("gs://"):
+  dest_bucket = dest_bucket[5:]
+dest_bucket = storage_client.get_bucket(dest_bucket)
+
+# Build map of all blobs in both buckets
+src_blobs = {}
+for blob in storage_client.list_blobs(src_bucket):
+  src_blobs[blob.name] = blob
+
+dest_blobs = {}
+for blob in storage_client.list_blobs(dest_bucket):
+  dest_blobs[blob.name] = blob
+
+# Copy blobs from src to dest, if the blobs don't exist in dest
+# If the blob does exist in dest, but the md5 does not match,
+# we don't overwrite it (this is for disaster recovery);
+# we will exit with an error
+num_copies = 0
+num_matches = 0
+
+mismatches = []
+for k, src_blob in src_blobs.items():
+  src_md5 = src_blob.md5_hash
+  dest_blob = dest_blobs.get(k)
+  if dest_blob:
+    if dest_blob.md5_hash == src_md5:
+      num_matches+=1
+      continue
+    else:
+      # Don't erase our backup!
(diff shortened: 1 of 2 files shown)


artifacts/dev/disaster-recovery/copy-buckets.py
```diff
@@ -0,0 +1,66 @@
...
+src_blobs = {}
+for blob in storage_client.list_blobs(src_bucket):
+  src_blobs[blob.name] = blob
+
+dest_blobs = {}
+for blob in storage_client.list_blobs(dest_bucket):
+  dest_blobs[blob.name] = blob
+
+# Copy blobs from src to dest, if the blobs don't exist in dest
+# If the blob does exist in dest, but the md5 does not match,
```
thockin: shouldn't the receiving bucket have sufficient retention that even a bug in this script can't overwrite?

justinsb: I think we need both retention and no-replace.  Retention prevents us replacing the file, comparison stops us trying to replace the file (which I believe fails even if the contents are the same), but we also need the script to report the failure nicely.

But ... perhaps what you're saying is that if we have retention enabled, we can do a `gsutil rsync` and that will be good enough?  That is plausible....

thockin: If I set a retention policy of 5 years, I can not remove or overwrite the file until that expires (I just tried it manually with GCS).

My question then is: can we / should we rely on this in the backup-bucket to ensure that backups are additive?

Further: we don't seem to set retention on the *prod buckets.  Perhaps we should do that, also?  Can we think of any reason we wouldn't want to guarantee that community-owned artifacts are available for an extended period of time?  We can't do this with GCR tags, but we can set it on the GCS buckets for non-GCR artifacts.

 MAYBE we can do it on the bucket for GCR (retaining the image by SHA, if not the tags).  ISTR there was a problem (we used to have it and removed it, forget why).  Will test.

FURTHER YET: should we lock these retentions?  This is risky in case we ever actually legit need to remove an artifact (e.g. a cred leak or something).

thockin: And yes, at that point, a simple additive rsync may be sufficient, with a different cred controlling the 2 buckets, it sounds pretty reasonable yeah?

justinsb: Yes, obviously slightly disappointed in myself for not seeing this _before_ I wrote the script, but I believe you're right.  We should set large retention, and just use a `gsutil rsync -c`.  (The `-c` is optional when they are both gs:// urls)

We do need to set `-C` so that one changed file doesn't stop the others.  But then I think that's it (as you pointed out).  We do need to watch for errors in the output and make sure that we don't just turn off the job instead of fixing it.

As for the other topics, yes I think we should set retention on the prod buckets, and we should set retention on the backup buckets.  I actually think we should lock the retention on the backup buckets (so we know they are always there).

As for setting retention on the prod buckets, I'm not sure.  Presumably we expect these to be mirrored over time, so it's not clear we actually could delete an artifact.  Someone changing an artifact could be pretty damaging until we corrected it (which we would do automatically, but not immediately).  We should say that you should check the hashes of all artifacts (this lets us use mirrors), but likely this bucket will be the canonical source of those hashes for the immediate future.

Because of that, I think we should lock it.  But we should try swapping out the backing bucket to make sure we can do it without "too much" disruption if we have to.

With a locked retention policy, disaster in this case means somebody uploading an additioanl file.  Either the next version of k8s, maybe some credential-stealing JS (but we shouldn't authn to `artifacts.k8s.io`), or some content designed to cause embarrassment .  We would want to remove that, but we wouldn't be compromising the integrity of the artifacts themselves.  I'd imagine the biggest risk would be uploading a new binary, then sending out a CVE notification with a link to it to various mailing lists.   I think some form of offline signing is the only realistic way of protecting against that, but plenty of existing solutions there in e.g. OS packages.

cblecker: The only concern would be if an artifact was uploaded that had such a security flaw that it should never be used. I don't care if we maintain a backup of that artifact, but deleting it from the prod bucket seems like something we should maintain an option for (especially if we have backups)

PR #335, opened 2019-08-16
create gcr.io/k8s-staging-promoter

listx:


k8s.gcr.io/k8s-staging-cip/manifest.yaml
```diff
@@ -0,0 +1,11 @@
...
+- name: gcr.io/k8s-staging-cip
+  src: true
+- name: us.gcr.io/k8s-artifacts-prod
+  service-account: k8s-infra-gcr-promoter@k8s-artifacts-prod.iam.gserviceaccount.com
+- name: eu.gcr.io/k8s-artifacts-prod
+  service-account: k8s-infra-gcr-promoter@k8s-artifacts-prod.iam.gserviceaccount.com
+- name: asia.gcr.io/k8s-artifacts-prod
+  service-account: k8s-infra-gcr-promoter@k8s-artifacts-prod.iam.gserviceaccount.com
+images: []
+renames: []
```
detiber: Should this have renames similar to what we are doing in cluster-api? https://github.com/kubernetes/k8s.io/blob/master/k8s.gcr.io/k8s-staging-cluster-api/manifest.yaml#L20-L28

At least until the automatic rewriting of the subdirectory to k8s-artifacts-prod is in place.

listx:
That mysterious error from pull-k8sio-cip should be addressed with this PR: https://github.com/kubernetes/test-infra/pull/13940

k8s.gcr.io/k8s-staging-cip/manifest.yaml
listx: I think most if not all rename cases will be unnecessary (we don't need their full power; e.g. consider the test case I added here for a "rebase": https://github.com/kubernetes-sigs/k8s-container-image-promoter/commit/a4993ef8621744d08b1987cceac77ed6f8e6ccd0).

Hmm, that being said I think I should rename the dest registries to have the correct (rebased name) endpoint. Lemme update it

listx: Updated, PTAL

detiber: Nice, I like this approach much better than the renames :)

listx:
Aside: looks like the pull-k8sio-cip job succeeded but for the wrong reason. I have another [PR](https://github.com/kubernetes-sigs/k8s-container-image-promoter/pull/101) on the way to handle that (after which I need to bump the images yet again in the Prow jobs, but that's trivial).

listx:
Ok the pull-k8sio-cip job is now passing for the correct reason (nothing to promote).

PR #336, opened 2019-08-16
Add retention for prod GCS (not GCR) buckets

thockin:
I can't see why we would NOT want to guarantee this.  The value of 10 years was picked from the air.  It could be argued for 5y or for 100y.  5y seems so recent.


infra/gcp/ensure-prod-storage.sh
```diff
@@ -138,7 +142,6 @@ empower_group_to_fake_prod \
     "k8s-infra-staging-release-test@kubernetes.io"
 
 # Special case: don't use retention on cip-test buckets
-# (the retention appears to have been locked, so we set it to 1s instead)
-gsutil retention set 1s gs://k8s-cip-test-prod
+gsutil retention clear gs://k8s-cip-test-prod
```
justinsb: OK, well I guess it wasn't locked!  Thanks for fixing :-)

thockin:
too late :)

PR #339, opened 2019-08-20
promoter manifests: avoid rename feature

listx:
The "rename" feature provides the power to rename images in an arbitrary
way, but this power is not really needed in the existing use cases we
have. We can get the same effect by just renaming the destination
registry slightly differently.

No functional change.

### DIFF
--- a/k8s.gcr.io/k8s-staging-build-image/manifest.yaml
+++ b/k8s.gcr.io/k8s-staging-build-image/manifest.yaml
@@ -2,11 +2,11 @@
 registries:
 - name: gcr.io/k8s-staging-build-image
   src: true
-- name: us.gcr.io/k8s-artifacts-prod
+- name: us.gcr.io/k8s-artifacts-prod/build-image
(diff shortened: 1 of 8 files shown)


listx:
/hold

Looks like there's a small bug I need to fix. Looking at the pull-k8sio-cip job, it shouldn't be trying to promote images in the US region that are already promoted. I think I know what's going on, so let me: (1) fix the bug, (2) update the prow images, and (3) re-run the the pull-k8sio-job test.

listx:
For (1) I created https://github.com/kubernetes-sigs/k8s-container-image-promoter/pull/106 (merged), (2) is done (https://github.com/kubernetes/test-infra/pull/13993).

Retesting now for (3):

/test pull-k8sio-cip

detiber:
/lgtm
Thanks for getting this in place, been meaning to adjust the cluster-api related repos for the past few days.

justinsb:
I like this - much safer!

/lgtm

PR #341, opened 2019-08-21
promote the promoter

listx:
This will give the promoter a more official home.

/assign @thockin @justinsb @ps882


listx:
Ok so it's a good thing that the pull-k8sio-cip job failed. I know how to fix this --- please wait!

listx:
Fix in-flight: https://github.com/kubernetes-sigs/k8s-container-image-promoter/pull/112

I need to bump the images again in the prow jobs also --- only that will make the pull-k8sio-cip job pass.

thockin:
I cleared the hold.  Linus, the promoter will just pick this up now, yeah?

listx:
> I cleared the hold. Linus, the promoter will just pick this up now, yeah?

Actually no --- if the prow jobs used a static tag like `:latest` to refer to the promoter, then yes (that is, we would constantly update `latest` to a newer version as we create them in a PR like this). But currently we don't do that. I was wondering if we should so that we won't have to edit k8s/test-infra to bump the promoter images; in the case of the promoter I think it's OK to use `latest` because the promoter already emits version information in its logs.

Anyway for now we just do things manually and will have to (1) create a "promote the promoter" PR like this, followed by (2) an update to the Prow jobs that reference the newly-promoted image.

PR #346, opened 2019-08-28
Add support for non-k8s-infra (and existing) mailing lists

dims:
- Separate command to print current configuration and subscriptions of all the mailing lists 
```
go run reconcile.go --print
```
- Add current metadata (name, description) for all the gsuite mailing lists along with their owners and managers
- Support to update the metadata
- Add support to tweak individual settings for each mailing list, so we don't stomp on how existing lists have been set up
- Folks who admin mailing lists can add/delete people as they do currently do, they can choose to use groups.yaml too (reconcile script will not clean up "members" added by hand)

### DIFF
--- a/groups/reconcile.go
+++ b/groups/reconcile.go
@@ -57,7 +57,17 @@ type GroupsConfig struct {
 
 type GoogleGroup struct {
 	EmailId     string `yaml:"email-id"`
+	Name        string `yaml:"name"`
 	Description string `yaml:"description"`
+
+	// settings that govern who can do what actions
+	Settings map[string]string `yaml:"settings,omitempty"`
+
+	// List of owners of the google group
+	Owners []string `yaml:"owners,omitempty"`
+
+	// List of managers of the google group
+	Managers []string `yaml:"managers,omitempty"`
 
 	// List of members in the google group
 	Members []string `yaml:"members,omitempty"`
@@ -77,6 +87,7 @@ var groupsConfig GroupsConfig
 func main() {
 	configFilePath := flag.String("config", "config.yaml", "the config file in yaml format")
 	confirmChanges := flag.Bool("confirm", false, "false by default means that we do not push anything to google groups")
+	printConfig := flag.Bool("print", false, "print the existing group information")
 
 	flag.Usage = Usage
 	flag.Parse()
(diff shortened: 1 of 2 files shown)


nikhita:
> Folks who admin mailing lists can add/delete people as they do currently do, they can choose to use groups.yaml too (reconcile script will not clean up "members" added by hand)

@dims out of curiosity, is there a particular reason why we aren't treating groups.yaml as the source of truth for non-k8s-infra mailing lists? Doing that would eliminate periodic runs of `--print`, right? :thinking:

dims:
@nikhita the `--print` is just a clutch for now, since there are admins for gsuite who have access to the UI to make changes directly. It can help identify if things have drifted. In time, we can get rid of it. essentially used `--print` to bootstrap the groups.yaml as well (especially the settings, as i did not want to type them out - to avoid typos etc)

thockin:
> there are admins for gsuite who have access to the UI to make changes directly.

Can we just tell them NOT to do that?

groups/reconcile.go
```diff
@@ -214,19 +243,34 @@ func printGroupMembersAndSettings(srv *admin.Service, srv2 *groupssettings.Servi
 		log.Println("No groups found.")
 		return nil
 	}
+	log.Printf("groups:")
 	for _, g := range g.Groups {
-		// Don't touch existing mailing lists, we should
-		// always prefix with "k8s-infra-"
-		if !strings.HasPrefix(g.Email, "k8s-infra-") {
-			continue
+		//// Don't touch existing mailing lists, we should
```
thockin: I don't understand this comment being here

groups/reconcile.go
```diff
@@ -321,20 +408,44 @@ func updateGroupSettingsToAllowExternalMembers(srv *groupssettings.Service, grou
...
+		WhoCanModerateMembers: "NONE",
 	}
 
-	// We will remove any additional members, so there's no point in allowing invitations / adds
-	settings.WhoCanInvite = "NONE_CAN_INVITE"
-	settings.WhoCanAdd = "NONE_CAN_ADD"
-	settings.WhoCanApproveMembers = "NONE_CAN_APPROVE"
-	settings.WhoCanModifyMembers = "NONE"
-	settings.WhoCanModerateMembers = "NONE"
+	for key, value := range groupSettings {
```
thockin: wouldn't it be easier to unmarshall into a struct?

dims: can fix this in a subsequent PR

groups/reconcile.go
dims: removing the comment and commented out code

dims:
@thockin @cblecker i've sync'ed groups.yaml. Let's please merge this before we let any more updates to groups.yaml.

dims:
i still need to find why we are not seeing `gke-security-groups@kubernetes.io`

cblecker:
/lgtm
/approve

Let's get this in and iterate

cblecker:
/lgtm cancel
/approve cancel

I found a problem with this. We're not exporting full lists and I'm not quite sure why.

dims:
@justaugustus - we don't mess with subscriptions ("members") for lists (specifically those not used for ACLs) so we are fine with nesting. We do have the definitions/metadata/owners/managers for each list that we can manage using groups.yaml. 

@cblecker - can you explain a little more "exporting full lists"?

cblecker:
@dims If at L275 you put in a:
```
		for _, m := range l.Members {
			fmt.Printf("%+v\n", m)
		}
```

and then run a `--print`, you'll see there is a lot more members in that list than are being printed out. In particular the `distributors-announce` list, you can see how many more are in that list then get printed into YAML.

justaugustus:
@dims -- Any update on this one?

dims:
on my TODO list for today! 🤞

dims:
@thockin @cblecker i've verified the following using my sc1 gsuite credentials
- number of lists
- number of members in the lists (example distributor-announce)

Thanks,
Dims

cblecker:
I have a follow up, but let's merge and iterate.

/lgtm
/approve

PR #353, opened 2019-09-05
Update terraform for GKE clusters

thockin:
I went through the gcloud docs and mapped every flag to terraform, and
made sure this represents my best understanding of them.

I ran this against the test project.  I think we are just about ready to
do it for real.

Two things I wanted to resolve:

1) Where should we store the state?  It doesn't seem right to ask
everyone to sign up for an account on another service, when we could
store state on GCS.  How?

2) I'd like us to agree on how we want to do multiple clusters in a
single project, because I assume that will be pretty important.

### DIFF
--- a/clusters/k8s-infra-dev-cluster-turnup/10-clusterconfiguration.tf
+++ b/clusters/k8s-infra-dev-cluster-turnup/10-clusterconfiguration.tf
@@ -16,8 +16,8 @@ locals {
 // Create SA for nodes
 resource "google_service_account" "cluster_node_sa" {
   project      = data.google_project.project.id
-  account_id   = "sa-${local.cluster_name}"
-  display_name = "Service Account for ${local.cluster_name}"
+  account_id   = "gke-nodes-${local.cluster_name}"
+  display_name = "Nodes in GKE cluster '${local.cluster_name}'"
 }
 
 // Add roles for SA
@@ -37,10 +37,11 @@ resource "google_project_iam_member" "cluster_node_sa_monitoring_metricwriter" {
   member  = "serviceAccount:${google_service_account.cluster_node_sa.email}"
 }
 
+// BigQuery dataset for usage data
 resource "google_bigquery_dataset" "usage_metering" {
   dataset_id  = replace("usage_metering_${local.cluster_name}", "-", "_")
   project     = data.google_project.project.id
-  description = "GKE Usage Metering for ${local.cluster_name}"
+  description = "GKE Usage Metering for cluster '${local.cluster_name}'"
   location    = local.bigquery_location
 
   access {
@@ -57,20 +58,26 @@ resource "google_bigquery_dataset" "usage_metering" {
   delete_contents_on_destroy = true
 }
 
-// This retrives the available GKE versions
-data "google_container_engine_versions" "cluster" {
-  project  = data.google_project.project.id
-  location = local.cluster_location
-}
-
 // Create GKE cluster, but with no node pools. Node pools can be provisioned below
 resource "google_container_cluster" "cluster" {
   name     = local.cluster_name
   location = local.cluster_location
 
-  provider           = google-beta
-  project            = data.google_project.project.id
-  min_master_version = data.google_container_engine_versions.cluster.latest_master_version
+  provider = google-beta
+  project  = data.google_project.project.id
+
+  // GKE clusters are critical objects and should not be destroyed
+  // IMPORTANT: should be true on production cluster
+  lifecycle {
+    prevent_destroy = false
+  }
+
+  // Network config
+  network = "default"
+  ip_allocation_policy {
+    use_ip_aliases    = true
+    create_subnetwork = true
+  }
 
   // Start with a single node, because we're going to delete the default pool
   initial_node_count = 1
@@ -89,14 +96,24 @@ resource "google_container_cluster" "cluster" {
     }
   }
 
+  // Enable google-groups for RBAC
+  authenticator_groups_config {
+    security_group = "gke-security-groups@kubernetes.io"
+  }
+
+  // Enable workload identity for GCP IAM
+  workload_identity_config {
+    identity_namespace = "${data.google_project.project.id}.svc.id.goog"
+  }
+
   // Enable Stackdriver Kubernetes Monitoring
   logging_service    = "logging.googleapis.com/kubernetes"
   monitoring_service = "monitoring.googleapis.com/kubernetes"
 
-  // Set maintenance time to 03:00 PST
+  // Set maintenance time
   maintenance_policy {
     daily_maintenance_window {
-      start_time = "11:00"
+      start_time = "11:00" // (in UTC), 03:00 PST
     }
   }
 
@@ -120,14 +137,24 @@ resource "google_container_cluster" "cluster" {
 
   // Configure cluster addons
   addons_config {
+    horizontal_pod_autoscaling {
+      disabled = false
+    }
+    http_load_balancing {
+      disabled = false
+    }
     network_policy_config {
       disabled = false
     }
   }
 
-  // GKE clusters are critical objects and should not be destroyed
-  // IMPORTANT: should be true on production cluster
-  lifecycle {
-    prevent_destroy = false
+  // Enable PodSecurityPolicy enforcement
+  pod_security_policy_config {
+    enabled = false // TODO: we should turn this on
+  }
+
+  // Enable VPA
+  vertical_pod_autoscaling {
+    enabled = true
   }
 }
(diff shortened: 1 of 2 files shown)


clusters/k8s-infra-dev-cluster-turnup/10-clusterconfiguration.tf
```diff
@@ -120,14 +143,24 @@ resource "google_container_cluster" "cluster" {
...
       disabled = false
     }
   }
 
-  // GKE clusters are critical objects and should not be destroyed
-  // IMPORTANT: should be true on production cluster
-  lifecycle {
-    prevent_destroy = false
+  // Enable PodSecurityPolicy enforcement
+  pod_security_policy_config {
```
cblecker: Doesn't turning on PSP enforcement require that we already have all applications set up with a PSP?

clusters/k8s-infra-dev-cluster-turnup/10-clusterconfiguration.tf
```diff
@@ -120,14 +143,24 @@ resource "google_container_cluster" "cluster" {
 
   // Configure cluster addons
   addons_config {
+    horizontal_pod_autoscaling {
```
cblecker: These are the defaults. Do we need to set these?

clusters/k8s-infra-dev-cluster-turnup/10-clusterconfiguration.tf
```diff
@@ -89,14 +102,24 @@ resource "google_container_cluster" "cluster" {
     }
   }
 
+  // Enable google-groups for RBAC
+  authenticator_groups_config {
+    security_group = "gke-security-groups@kubernetes.io"
+  }
+
+  // Enable workload identity for GCP IAM
+  workload_identity_config {
```
cblecker: Does enabling this on the cluster mean that we *have* to use it?

clusters/k8s-infra-dev-cluster-turnup/11-pool-1-configuration.tf
```diff
@@ -30,9 +30,15 @@ resource "google_container_node_pool" "pool-1" {
     max_node_count = 20
   }
 
+  // Set a realistic limit on pods per node, including daemonsets
+  max_pods_per_node = 32
```
cblecker: What's the reason for a non-GKE default?

clusters/k8s-infra-dev-cluster-turnup/10-clusterconfiguration.tf
```diff
@@ -68,9 +69,21 @@ resource "google_container_cluster" "cluster" {
   name     = local.cluster_name
   location = local.cluster_location
 
-  provider           = google-beta
-  project            = data.google_project.project.id
-  min_master_version = data.google_container_engine_versions.cluster.latest_master_version
```
cblecker: Why are we deleting this?

clusters/k8s-infra-dev-cluster-turnup/10-clusterconfiguration.tf
thockin: Good point.  I missed the big red warning. :)  Will update.

clusters/k8s-infra-dev-cluster-turnup/10-clusterconfiguration.tf
thockin: I called them out explicitly because I went hunting for them and didn't find them.  I thought it was useful as "documentation".  Disagree?

clusters/k8s-infra-dev-cluster-turnup/10-clusterconfiguration.tf
thockin: I don't think so.  It's totally opt-in.  Additionally, I see that it is incompatible with metadata concealment, so we should probably disable that flag too.

clusters/k8s-infra-dev-cluster-turnup/11-pool-1-configuration.tf
thockin: It economizes IP usage, but that's not a problem for us, so I will remove it.

clusters/k8s-infra-dev-cluster-turnup/10-clusterconfiguration.tf
thockin: provider and project are still present, but I see no point in specifying a min_master_version - I want to use whatever GKE thinks is best and most stable (default).

Disagree?

thockin:
re-pushed with updated config

clusters/k8s-infra-dev-cluster-turnup/10-clusterconfiguration.tf
cblecker: Sometimes GKE is behind, and we might need a newer version. GKE default is still `1.12.8-gke.10`. What if we need a feature in 1.13?

thockin:
Then we can make an explicit call if/when that happens.

thockin:
BTW - I have left this cluster up and running for us to verify.  I have a few things I want to try, but no time today, probably.

cblecker:
A few changed needed. Patch:
```
diff --git a/clusters/k8s-infra-dev-cluster-turnup/10-clusterconfiguration.tf b/clusters/k8s-infra-dev-cluster-turnup/10-clusterconfiguration.tf
index 4f24863..f3d4bc1 100644
--- a/clusters/k8s-infra-dev-cluster-turnup/10-clusterconfiguration.tf
+++ b/clusters/k8s-infra-dev-cluster-turnup/10-clusterconfiguration.tf
@@ -58,12 +58,6 @@ resource "google_bigquery_dataset" "usage_metering" {
   delete_contents_on_destroy = true
 }
 
-// This retrives the available GKE versions
-data "google_container_engine_versions" "cluster" {
-  project  = data.google_project.project.id
-  location = local.cluster_location
-}
-
 // Create GKE cluster, but with no node pools. Node pools can be provisioned below
 resource "google_container_cluster" "cluster" {
   name     = local.cluster_name
@@ -81,7 +75,7 @@ resource "google_container_cluster" "cluster" {
   // Network config
   network = "default"
   ip_allocation_policy {
-    use_ip_aliases = true
+    use_ip_aliases    = true
     create_subnetwork = true
   }
 
@@ -156,7 +150,7 @@ resource "google_container_cluster" "cluster" {
 
   // Enable PodSecurityPolicy enforcement
   pod_security_policy_config {
-    enabled = false  // TODO: we should turn this on
+    enabled = false // TODO: we should turn this on
   }
 
   // Enable VPA
diff --git a/clusters/k8s-infra-dev-cluster-turnup/11-pool-1-configuration.tf b/clusters/k8s-infra-dev-cluster-turnup/11-pool-1-configuration.tf
index 41c0d31..2514f5a 100644
--- a/clusters/k8s-infra-dev-cluster-turnup/11-pool-1-configuration.tf
+++ b/clusters/k8s-infra-dev-cluster-turnup/11-pool-1-configuration.tf
@@ -32,15 +32,22 @@ resource "google_container_node_pool" "pool1" {
 
   // Set machine type, and enable all oauth scopes tied to the service account
   node_config {
-    machine_type    = "n1-standard-4"
-    disk_size_gb    = 100
-    disk_type       = "pd-standard"
+    machine_type = "n1-standard-4"
+    disk_size_gb = 100
+    disk_type    = "pd-standard"
 
     service_account = google_service_account.cluster_node_sa.email
     oauth_scopes    = ["https://www.googleapis.com/auth/cloud-platform"]
 
-    // workload_metadata_config is not needed with cluster
-    // workload_identity_config enabled.
+
+    // Needed for workload identity
+    workload_metadata_config {
+      node_metadata = "GKE_METADATA_SERVER"
+    }
+    metadata = {
+      disable-legacy-endpoints = "true"
+    }
+
   }
 
   // If we need to destroy the node pool, create the new one before destroying
```

thockin:
All make sense to me but the workload_metadata_config part.  Those are defaults.  Yet terraform really wants me to have them - it says it will replace the nodepool, but it will come out the same.  That seems broken... ?

Regardless, patch applied and pushed.

thockin:
Curious: I used the cluster name "thtest" before, but switched it back to "k8s-services-cluster", but the cluster still came up as "thtest".  I am going to burn down and repeat.

thockin:
I am a doofus and did not refresh my screen.  Sorry.  Will turn cluster back up :)

cblecker:
> All make sense to me but the workload_metadata_config part. Those are defaults.

It depends on the GKE API. Sometimes it omits fields that are unset, sometimes it forces them to a specific setting. In this case the workload identity setting on the cluster forces an explicit setting on the node pool.. we just need to tell TF that this is known and expected.

thockin:
Right, after I patched your change in, `tf apply` showed me no delta, so you are (of course) correct

PR #354, opened 2019-09-05
Add txtdirect staging repo

stp-ip:
Create staging repo to prepare redirector pulling artifact from Kubernetes controlled location (and (help with testing).

### DIFF
--- a/groups/groups.yaml
+++ b/groups/groups.yaml
@@ -699,6 +699,15 @@ groups:
       - bowei@google.com
       - antonio.ojea.garcia@gmail.com
 
+  - email-id: k8s-infra-staging-txtdirect@kubernetes.io
+    name: k8s-infra-staging-txtdirect
+    description: |-
+      ACL for staging TXTDirect
+    settings:
+      ReconcileMembers: "true"
+    members:
+      - michaelg@okkur.org
+
   - email-id: k8s-infra-team-private@kubernetes.io
     name: k8s-infra-team-private
     description: |-
--- a/infra/gcp/ensure-staging-storage.sh
+++ b/infra/gcp/ensure-staging-storage.sh
@@ -64,6 +64,7 @@ STAGING_PROJECTS=(
     publishing-bot
     release-test
     service-apis
+    txtdirect
 )
 if [ $# = 0 ]; then
     # default to all staging projects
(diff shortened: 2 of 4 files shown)


cblecker:
Why do we need this?

stp-ip:
Tim was suggesting it and it felt like a prerequisite for the redirector PR: https://github.com/kubernetes/k8s.io/pull/227#discussion_r313019427

cblecker:
@thockin Do we need to host our own binary? Why couldn't we just use upstream's directly?

cblecker:
@stp-ip Does txtdirect publish it's own official image somewhere? Or is it just a docker file, and up to the user to build it?

stp-ip:
> @stp-ip Does txtdirect publish it's own official image somewhere? Or is it just a docker file, and up to the user to build it?

txtdirect builds their own images yes. 

In detail:
txtdirect provides a vanity url for containers such as `c.txtdirect.org/txtdirect:dev-9f1280f`, which currently is backed by: https://console.cloud.google.com/gcr/images/txtdirect-223710/GLOBAL/txtdirect

Build is done by GCB into GCR. Vanity url is doing redirects to the underlying store as well as redirect non container requests to the above GCR url so one can see the list just by visiting: `c.txtdirect.org/txtdirect`

cblecker:
I'm fine then with just referencing those images, and not building our own.

@thockin Do you disagree?

stp-ip:
Rebased

stp-ip:
Had a brief discussion with Tim and Ben on this and a rough idea was a multi tier system:
1. Use staging repo + GCB build from the project itself <- opt-in
2. Use staging repo + GCB build from a build repo under k8s <- opt-in, but preferred for critical images such as CoreDNS
3. Use upstream images 

Risks/opportunities
1: Adds another process and does not prevent the code from disappearing
2: Most flexible and might be enough to replace 1 as well. We may bundle release code for each build within the repo or github releases.
3: Risk of loosing access to both builds, build code and code in general.

We don't have clear policies around this and I wanted to loop Tim back in to discuss your concerns.
There needs to be a conscious decision here, which I would love to write down into policy afterwards.
In my view. Let's go with 2 and keep 3 to a minimum.

thockin:
Any reason this needs to be help any longer?

stp-ip:
Rebase done.

thockin:
actuated now

PR #355, opened 2019-09-05
Add stp-ip to CoreDNS staging group

stp-ip:
Thought it might be helpful for testing or pushing critical CoreDNS patch releases to add myself due to the intersection of CoreDNS maintainer and helping with the infra wg.

Happy for this to be closed without merge, if this doesn't make sense.

### DIFF
--- a/groups/groups.yaml
+++ b/groups/groups.yaml
@@ -526,6 +526,7 @@ groups:
       ReconcileMembers: "true"
     members:
       - davanum@gmail.com
+      - michaelg@okkur.org
 
   - email-id: k8s-infra-staging-csi@kubernetes.io
     name: k8s-infra-staging-csi


cblecker:
@dims Who’s responsible for this image? sig-network? sig-node?

Ultimately, being involved with the wg doesn’t make a difference, it’s up to the responsible sig to decide who and how their image gets published.

dims:
i'd say sig-release ... WDYT @justaugustus ?

cblecker:
Q: Do we actually rebuild the coredns image? Or do we just pull down the image, and push it up to our repo?

If it's the former, we should put it in GCB because it's a "core" image, and reduce humans to 0.
If it's the latter, could we not use the promoter?

Either way, for core images like this, I think the goal is to get humans out of the loop (ref #367)

dims:
@stp-ip wanna rebase this please?

PR #364, opened 2019-09-11
Enable GCS and multiple clusters for terraform

thockin:
Change terraform to use GCS for state storage.

Enable multiple clusters in one project.

There are a couple of places that need to be updated to make a new cluster, which is unfortunate, but it's fairly well encapsulated.
    
Each clusters/foo dir represents a project "foo".
    
Each clusters/foo/bar dir represents a cluster "bar" in project "foo".

I tested this by turning up 2 clusters in parallel.  All seems good.  Clusters are running as of this message.


infra/gcp/ensure-main-project.sh
```diff
@@ -48,6 +48,9 @@ CLUSTER_ADMINS_GROUP="k8s-infra-cluster-admins@kubernetes.io"
 # The accounting group.
 ACCOUNTING_GROUP="k8s-infra-gcp-accounting@kubernetes.io"
 
+# The GCS bucket which hold terraform state for clusters
+CLUSTER_TERRAFORM_BUCKET="gs://${k8s-infra-clusters-terraform}"
```
cblecker: This doesn't look right.. why is wrapped like a variable, but has dashes in it?

infra/gcp/ensure-main-project.sh
```diff
@@ -67,6 +70,9 @@ enable_api "${PROJECT}" storage-component.googleapis.com
 color 6 "Enabling the OSLogin API"
 enable_api "${PROJECT}" oslogin.googleapis.com
 
+color 6 "Ensuring the cluster terraform-state bucket exists"
+ensure_gcs_bucket "${PROJECT}" "${CLUSTER_TERRAFORM_BUCKET}"
```
cblecker: Won't this make a world readable bucket?

clusters/k8s-infra-dev-cluster-turnup/test-eu/00-inputs.tf
```diff
@@ -7,16 +7,11 @@ This file defines:
...
-  backend "remote" {
-    hostname     = "app.terraform.io"
-    organization = "k8s-infra"
-
-    workspaces {
-      // should match the name of the gcp project
-      name = "k8s-infra-dev-cluster-turnup"
-    }
+  backend "gcs" {
+    bucket = "k8s-infra-dev-cluster-turnup-terraform-state"
```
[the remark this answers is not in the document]
thockin: I intended for "real" clusters to always share a bucket with different prefixes.  I figured this was easiest, though I could be talked out of it.

infra/gcp/ensure-main-project.sh
thockin: good catch

infra/gcp/ensure-main-project.sh
thockin: I didn't think that was a problem, but looking at the contents, maybe it is?

I will change it to private.  Slightly bigger change.

thockin:
new push is up with changes

clusters/k8s-infra-dev-cluster-turnup/test-eu/00-inputs.tf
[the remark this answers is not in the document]
cblecker: I'm okay with one central bucket for all projects/clusters.

cblecker:
/lgtm
/approve

/hold
I'm good with this going in if you are, @thockin

PR #365, opened 2019-09-12
Prepare for a prod cluster

thockin:
This declares a prod cluster called `aaa`.

Let's bikeshed.

I don't think we should name it `prod`, because eventually we will have to burn it down for some reason and then what - `prod2` ? 

I could be convinced to embed region or major geo or something in the name, but I don't think it is necessary.

We could use cutesy names (e.g. smurfs or greek gods or methematicians) if people REALLY care.  I don't.

I went for something abstract, short, and unambiguous.  Three letters.  In truth, probably 2 letters would suffice.  When we EOL `aaa`, we can replace it with `aab`.  We can assign blocks to continents, or not.  This also gives us freedom to repurpose clusters in-place (e.g. run test and prod together).

We may consider other names for special-purpose things like prow trusted clusters.  Or not.


infra/gcp/clusters/README.md
```diff
@@ -0,0 +1,20 @@
...
+but for now they are designed to be straight forward and verbose.
+
+Prerequisites:
+- Be a member of the k8s-infra-cluster-admins@kubernetes.io group.
+- Have Terraform installed
+  (https://releases.hashicorp.com/terraform/${TERRAFORM_VERSION}/terraform_${TERRAFORM_VERSION}_linux_amd64.zip)
+
+Instructions:
+- Ensure you are logged into your GCP account with `gcloud auth application-default login`
+- From within a cluster directory:
```
ameukam: Is it possible to add : `terraform init` before ? It helps download the terraform providers for the resources.

thockin:
Great idea, done.

ameukam:
+1 for `aaa` and `k8s-infra-dev-cluster-turnup`.

cblecker:
I’ll give this one last round of reviews today and then we can push the button :)

thockin:
Once clear, I will turn on a cluster, hopefully before we meet again.

infra/gcp/clusters/kubernetes-public/aaa/11-pool1-configuration.tf
```diff
@@ -0,0 +1,57 @@
...
+  // by 3, as the cluster will exist in three zones
+  autoscaling {
+    min_node_count = 1
+    max_node_count = 20
+  }
+
+  // Set machine type, and enable all oauth scopes tied to the service account
+  node_config {
+    // IMPORTANT: Should be n1-standard-1 on test clusters
+    machine_type = "n1-standard-2"
```
cblecker: The current clusters in kubernetes-public are n1-standard-4

thockin:
Yeah, I don't know if we need standard-4 * 3 zones yet.

cblecker:
If we need to swap it out, we can do it easily, so I'm fine then rolling forward with this.

cblecker:
/lgtm

Note, this will need to be applied after merge.

thockin:
I will apply hopefully in the AM.

CamelCaseNotation:
What's the purpose of this cluster? What is intended to be ran in the cluster?

thockin:
This cluster is intended to run a variety of services and tools for example:

* gcsweb
* k8s.io site (vanity & shortcut redirects, `go get` support, etc)
* go.k8s.io
* kube-perf
* velodrome
* publisher bot

...and probably more over time.

PR #371, opened 2019-09-20
promoter: add 20190927-v2.3.1-21-g21a4735

listx:
This new version allows tagless promotions.

/assign @ps882


listx:
/hold

Actually, I would like to make one more improvement to the promoter before pushing up the next version.

listx:
This new version brings in the new CSV-based `-snapshot` output as per https://github.com/kubernetes-sigs/k8s-container-image-promoter/pull/136. This is the version we'll be using for performing backups from Prow.

/assign @thockin

PR #376, opened 2019-09-26
Add prod projects for GCR backups

thockin:
Fixes #375

PR #377, opened 2019-09-26
add scripts for backups

listx:
This is a rough cut of some tooling for basic backups from prod to a
backup.

/assign @thockin @justinsb

### DIFF
--- a/infra/gcp/backup_tools/backup.sh
+++ b/infra/gcp/backup_tools/backup.sh
@@ -1,0 +1,81 @@
+#!/usr/bin/env bash
+#
+# Copyright 2019 The Kubernetes Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# USAGE NOTES
+#
+# Backs up prod registries. This is a thin orchestrator as all the heavy lifting
+# is done by the gcrane binary. The only required argument is the path to the
+# service account that has write access to the backup GCRs.
+
+set -o errexit
+set -o nounset
+set -o pipefail
+
+copy_with_date()
+{
+    local gcrane_img
+    local gcrane_digest
+    local timestamp
+    local source_gcr_repo
+    local backup_gcr_repo
+    local sa_key_path
+
+    # The image is pinned to an exact sha256 version because the gcrane binary's
+    # behavior must never change (otherwise, the "docker run" invocation below
+    # might explode unexpectedly).
+    gcrane_digest="560757a7b63c85e9b95d2d971a18aa2a2425899e36f1550ce3584018d9ac49ea" # version 7683b4ee5f6150cb47a791309f781c522b95a58f
+    gcrane_img="gcr.io/go-containerregistry/gcrane@sha256:${gcrane_digest}"
+
+    # We use a timestamp of the form YYYY/MM/DD/HH because this makes the backup
+    # folders more easily traversable from a human perspective.
+    timestamp="$(date -u +"%Y/%m/%d/%H")"
+
+    if (( $# != 3 )); then
+        cat << EOF >&2
+copy_with_date: usage <source_gcr_repo> <backup_gcr_repo> <sa_key_path>
+e.g. copy_with_date "us.gcr.io/k8s-artifacts-prod" "us.gcr.io/k8s-artifacts-prod-bak" /path/to/sa/key.json
+EOF
+        exit 1
+    fi
+
+    source_gcr_repo="${1}" # "us.gcr.io/k8s-artifacts-prod"
+    backup_gcr_repo="${2}" # "us.gcr.io/k8s-artifacts-prod-bak"
+    sa_key_path="${3}"
+
+    # Perform backup by copying all images recursively over.
+    docker run \
+        -v "${sa_key_path}":/auth.json \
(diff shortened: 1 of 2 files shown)


listx:
After this gets reviewed + merged, I plan on adding Prow jobs in k8s/test-infra repo to make use of these scripts at regular intervals (the more frequent, the better).

I could also port the "record_consistency.sh" logic into the promoter itself it seems like a conflict of interest, because the whole point of this PR is to add secondary checks to make sure the promoter is behaving properly.

listx:
@thockin This is much simpler than the earlier PR (1 script instead of 3). PTAL

listx:
This is blocked by #376

infra/backup_tools/README.md
```diff
@@ -0,0 +1,10 @@
+# Backup tools
+
+The tools in this folder are designed to backup the
+{asia,eu,us}.gcr.io/k8s-artifacts-prod GCRs.
```
thockin: If all 3 are supposed to be identical, do we need backups of all 3?  Or is it just easier that way?

infra/backup_tools/backup.sh
```diff
@@ -0,0 +1,69 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# USAGE NOTES
+#
+# Backs up prod registries. This is a thin orchestrator as all the heavy lifting
```
thockin: comment implies `copy_with_date.sh` and `record_consistency.sh` are separate files, but no such file exists.

infra/backup_tools/backup.sh
```diff
@@ -0,0 +1,69 @@
...
+    us.gcr.io/k8s-artifacts-prod
+)
+sa_key_path="${1}"
+
+# Activate creds.
+gcloud auth activate-service-account --key-file "${sa_key_path}"
+
+# Copy each region to its backup.
+for prod_repo in "${prod_repos[@]}"; do
+    copy_with_date "${prod_repo}" "${prod_repo}-backup" "${sa_key_path}"
```
thockin: NB I changed "-backup" to "-bak" in latest push of the ensure script (because we overflowed allowable project names :)

infra/backup_tools/backup.sh
```diff
@@ -0,0 +1,69 @@
...
+    timestamp="$(date -u +"%Y/%m/%d/%H")"
+    source_gcr_repo="${1}" # "us.gcr.io/k8s-artifacts-prod"
+    backup_gcr_repo="${2}" # "us.gcr.io/k8s-artifacts-prod-backup"
+    sa_key_path="${3}"
+
+    # Perform backup by copying all images recursively over.
+    docker run \
+        -v "${sa_key_path}":/auth.json \
+        --env GOOGLE_APPLICATION_CREDENTIALS=/auth.json \
+        gcr.io/go-containerregistry/gcrane \
```
thockin: Do we need a particular tag?  Assuming :latest is kind of scary...

What protects this backup tool from being compromised?

thockin:
Also, if pinning to a SHA, comment why :)

infra/backup_tools/README.md
listx: I only did it for symmetry (knee-jerk reaction).

Philosophically if I choose 1 region to back up out of the 3, then that 1 region becomes special and I don't like this.

infra/backup_tools/backup.sh
listx: Oops, will fix.

infra/backup_tools/backup.sh
listx: ACK

infra/backup_tools/backup.sh
listx: > Do we need a particular tag? Assuming :latest is kind of scary...

Yes, I should have referenced a sha256 digest here. Will change to *not* use `:latest`.

> What protects this backup tool from being compromised?

Good question. For now I think the sha256 is the only guarantee. If we need to update we would have to run some e2e test that ensures that the new version doesn't change behavior. Short of that, at least manual testing of some sort.

listx:
I removed the unnecessary call to gcloud (activating creds) because they are only used inside the gcrane container.

infra/gcp/backup_tools/backup.sh
```diff
@@ -0,0 +1,73 @@
...
+    # behavior must never change (otherwise, the "docker run" invocation below
+    # might explode unexpectedly).
+    gcrane_digest="560757a7b63c85e9b95d2d971a18aa2a2425899e36f1550ce3584018d9ac49ea" # version 7683b4ee5f6150cb47a791309f781c522b95a58f
+    gcrane_img="gcr.io/go-containerregistry/gcrane@sha256:${gcrane_digest}"
+
+    # We use a timestamp of the form YYYY/MM/DD/HH because this makes the backup
+    # folders more easily traversable from a human perspective.
+    timestamp="$(date -u +"%Y/%m/%d/%H")"
+
+    source_gcr_repo="${1}" # "us.gcr.io/k8s-artifacts-prod"
```
thockin: you should probably check `$#` and abort if the args are the wrong number, ok for followup

infra/gcp/backup_tools/backup.sh
```diff
@@ -0,0 +1,81 @@
...
+        exit 1
+    fi
+
+    source_gcr_repo="${1}" # "us.gcr.io/k8s-artifacts-prod"
+    backup_gcr_repo="${2}" # "us.gcr.io/k8s-artifacts-prod-bak"
+    sa_key_path="${3}"
+
+    # Perform backup by copying all images recursively over.
+    docker run \
+        -v "${sa_key_path}":/auth.json \
```
rajibmitra: will it be a good idea to use buildkit here @listx

listx: I haven't heard of that. Care to elaborate? :)

rajibmitra: Please ignore me.  all looks good.

PR #379, opened 2019-09-26
Update org admins group

cblecker:
Update per https://github.com/kubernetes/k8s.io/pull/370#issuecomment-535349146

Add @dims and @spiffxp.

/assign @dims @thockin 
/approve

PR #380, opened 2019-09-26
Rename org admins group

cblecker:
per https://github.com/kubernetes/k8s.io/pull/379#issuecomment-535620756


thockin:
holla when it's applied, so I can do org IAM :)

cblecker:
@thockin done!

thockin:
OK.  I activated this in the org.  I removed myself and dims as named IAM.  I can still refesh the IAM page, so there's that...

cblecker:
Yup, I can see it too!

PR #289, opened 2019-07-05
Add staging gcr bucket for cluster-api-aws

detiber:
Adds configuration to enable a staging gcr repo for cluster-api-provider-aws

PR #384, opened 2019-09-30
Add kube-state-metrics staging repo

brancz:
This adds a kube-state-metrics staging repo so we can stop using external registries and/or have to ping people working at google to publish images.

I'm not entirely sure I've done everything correctly, but I believe I followed these steps: https://github.com/kubernetes/k8s.io/tree/master/k8s.gcr.io#staging-repos

Let me know if this is everything that is needed. :slightly_smiling_face: 

cc @cblecker @nikhita @LiliC @tariq1890


groups/groups.yaml
```diff
@@ -209,3 +209,8 @@ groups:
       - ihor@cncf.io
       - justinsb@google.com
       - thockin@google.com
+  - email-id: k8s-infra-staging-kube-state-metrics@kubernetes.io
```
nikhita: The max number of characters can only be 30. Ref: https://github.com/kubernetes/k8s.io/pull/289#issuecomment-508806417. :grimacing:

nikhita: (side note: we should add validation for this)

tariq1890:
If I understand correctly, we use two registry repos then? Staging and Production?

nikhita:
> If I understand correctly, we use two registry repos then? Staging and Production?

Eventually, yes.

Right now the [tool](https://github.com/kubernetes-sigs/k8s-container-image-promoter) used to promote images is WIP so we just use the staging repo. When it is ready, we can start having prod ones too.

groups/groups.yaml
brancz: shortened to `k8s-infra-staging-kube-metrics`, with a comment explaining why it's shortened

brancz: shortened to k8s-infra-staging-kube-metrics, with a comment explaining why it's shortened

groups/groups.yaml
```diff
@@ -209,3 +209,10 @@ groups:
       - ihor@cncf.io
       - justinsb@google.com
       - thockin@google.com
+  # This should be kube-state-metrics, but this is limited to 36 characters, so
```
nikhita: nit: s/36/30?

brancz: whops sorry, fixed

k8s.gcr.io/k8s-staging-kube-state-metrics/manifest.yaml
```diff
@@ -0,0 +1,12 @@
+# google group for gcr.io/k8s-staging-kube-state-metrics is k8s-infra-staging-kube-state-metrics@kubernetes.io
```
nikhita: sorry for not catching this sooner...this needs to be `k8s-infra-staging-kube-metrics@kubernetes.io` too

nikhita: To be clear, I think the main problem is with the gcr repo name and not the google group name :grimacing:

brancz: hmm .. that kind of messes with well known names of these projects, is there no other way? (the `k8s-staging-` prefix seems to take a pretty significant budget of that)

nikhita: cc @dims ^

LiliC: Yes this might cause confision, kube-metrics implies something completely different to me. I would suggest if there is no way around this, we can do `k8s-infra-staging-ksm`?

brancz:
I _believe_ all comments are resolved. Please take a look. :slightly_smiling_face:

nikhita:
@dims @thockin ping :)

nikhita:
I'm _realllyyyyyy_ sorry about the back and forth dance on this, but while working on https://github.com/kubernetes/k8s.io/pull/399, I realized that using `kube-state-metrics` would be fine since we'll need to count the character length with the `k8s-staging-` prefix, not `k8s-infra-staging-` prefix. :see_no_evil: 

`k8s-staging-kube-state-metrics` is exactly 30 characters (and that's the maximum value allowed).

- `k8s-staging-` prefix is used in the manifest file, which is the one gcloud complains about.
- `k8s-infra-staging-` prefix is used for google groups, and google groups doesn't complain about the character length.

brancz:
Got it. Will update this in a bit then!

brancz:
Done.

tariq1890:
Thanks @nikhita !

One of the owners just removed their request for review :(. 
/assign @cblecker

infra/gcp/ensure-staging-storage.sh
```diff
@@ -56,6 +56,7 @@ STAGING_PROJECTS=(
     multitenancy
     publishing-bot
     release-test
+    kube-state-metrics
```
tariq1890: @brancz I think you need to maintain lexical order here. As stated in the comment above.

dims:
Apologies! i was cleaning my PR queue and lost this one.

@brancz please fix the lexical order as @tariq1890 mentions and ping me on slack and i'll get this in.

/assign

infra/gcp/ensure-staging-storage.sh
brancz: Done

brancz:
Done! Thanks! Let me know if there is anything else :slightly_smiling_face: .

PR #347, opened 2019-08-30
Bump cert-manager to v0.9.1

munnerz:
This pull request bumps cert-manager to v0.9.1.

Full details on how this should be applied can be found in the [upgrade guide](https://docs.cert-manager.io/en/latest/tasks/upgrading/index.html), however, to summarise:

1) `kubectl delete deploy --all -n cert-manager`: we have changed the `spec.selector` used on the Deployment resources for cert-manager to be in-line with advised standards. As this field is immutable, this requires the deployments to be deleted prior to applying.
2) Apply new manifests
3) Run the command advised in the upgrade guide to verify that no Certificate resources installed in your cluster still contain a now deprecated `spec.acme` field:

```
kubectl get certificate --all-namespaces \
  -o custom-columns="NAMESPACE:.metadata.namespace,NAME:.metadata.name,OWNER:.metadata.ownerReferences[0].kind,OLD FORMAT:.spec.acme"
```

If the `OLD FORMAT` column is all `<none>`, you are all good and we can go ahead and remove the `issuer.spec.http01` and `issuer.spec.dns01` fields from the Issuer resources defined in this PR.

If you have any *auto generated* Certificate resources (created by ingress-shim), these will *automatically* have the `certificate.spec.acme` field removed *after* you remove the `issuer.spec.http01` and `issuer.spec.dns01` fields from the Issuer.

If in doubt, please feel free to ping me 😄 I'm also happy to screen-share with someone that has appropriate permissions to apply these changes to get this rolled out. I've been updating our own internal clusters recently and have done so quickly without service interruption 😄

/cc @ixdy

PR #389, opened 2019-10-05
Create RBAC group for cert-manager

cblecker:
/cc @munnerz @dims @thockin 

Per #347, I want to create an RBAC group that we try out namespace scoped RBAC. This would be for the namespace we stick cert manager on "aaa". 

https://cloud.google.com/kubernetes-engine/docs/how-to/role-based-access-control#rolebinding

### DIFF
--- a/groups/groups.yaml
+++ b/groups/groups.yaml
@@ -36,6 +36,11 @@ groups:
       - justinsb@google.com
       - spiffxp@google.com
       - thockin@google.com
+  - email-id: k8s-infra-rbac-cert-manager@kubernetes.io
+    members:
+      - james@munnelly.eu
+      - thockin@google.com
+      - cblecker@gmail.com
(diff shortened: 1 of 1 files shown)


cblecker:
@dims @munnerz @thockin bump

groups/groups.yaml
```diff
@@ -36,6 +36,11 @@ groups:
       - justinsb@google.com
       - spiffxp@google.com
       - thockin@google.com
+  - email-id: k8s-infra-rbac-cert-manager@kubernetes.io
```
thockin: Really minor nit - we're eventually going to run out of characters, and this has a 15 char prefix.  

Maybe we should try to find a shorted prefix and elide some dashes like "certmanager" vs cert-manager" or event "certmgr" ?

PR #399, opened 2019-10-16
groups: validate length of emails used for staging repos

nikhita:
Ref: https://github.com/kubernetes/k8s.io/pull/289#issuecomment-508806417

For a staging repo, PROJECT_ID is equal to `k8s-staging-<project-name>`.
gcloud only allows PROJECT_ID values between lengths of 6 and 30.

After discounting the number of characters in `k8s-staging-` (12), we
are left with 18 characters for the project name. This project name is
also used in the email ID for the google group used to provide access to
the staging repo.

To ensure that the project name stays within the character limit, this
commit adds a test to validate that the project name in the email ID
doesn't have more than 18 characters.

/assign @dims @cblecker

### DIFF
--- a/groups/reconcile.go
+++ b/groups/reconcile.go
@@ -1,9 +1,12 @@
 /*
 Copyright 2019 The Kubernetes Authors.
+
 Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.
 You may obtain a copy of the License at
+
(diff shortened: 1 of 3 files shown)


nikhita:
Will add a presubmit if this looks good.

nikhita:
PR for presubmit: https://github.com/kubernetes/test-infra/pull/14794

cblecker:
Thanks @nikhita !
/approve

PR #412, opened 2019-10-23
Fix unbound variable bug

thockin:


thockin:
self-approving trivial bugfix

PR #424, opened 2019-10-25
Add Bart as cluster admin

thockin:
We will build on this to do RBAC test work, but this seems permanent (as much as any of this is permanent)


cblecker:
/lgtm
/approve

Going to manually merge to fix reconcile.go

PR #427, opened 2019-10-25
Cleanup security groups

tallclair:
Questions:

- What's the difference between `security-release-team` and `release-managers-private`?
- `security@` has `ReconcileMembers: false` - what does that do? The real list has 2 additional members, should we just add those to the manifest and set reconcile to true?

TODO: add distributors-annonuce@ members & reconcile. Is there any way to sync this with the github.com/kubernetes/security repo, so we don't need to manually keep the lists in sync?

/cc @cjcullen @liggitt @philips @lukehinds @jonpulsifer @joelsmith

### DIFF
--- a/groups/groups.yaml
+++ b/groups/groups.yaml
@@ -1007,7 +1007,6 @@ groups:
       - dmaceachern@vmware.com
       - hhorl@pivotal.io
       - idealhack@gmail.com
-      - kubernetes-release-managers-private@googlegroups.com
       - saugustus@vmware.com
       - tpepper@gmail.com
 
@@ -1038,9 +1037,9 @@ groups:
   - email-id: security-discuss-private@kubernetes.io
     name: security-discuss-private
     description: |-
-      Private discussion forum for PST members.
+      Private discussion forum for PSC members.
 
-      https://github.com/kubernetes/sig-release/blob/master/security-release-process-documentation/security-release-process.md#product-security-team-pst
+      https://github.com/kubernetes/security#product-security-committee-psc
     settings:
       AllowExternalMembers: "true"
       WhoCanJoin: "INVITED_CAN_JOIN"


groups/groups.yaml
```diff
@@ -1007,7 +1007,6 @@ groups:
       - dmaceachern@vmware.com
       - hhorl@pivotal.io
       - idealhack@gmail.com
-      - kubernetes-release-managers-private@googlegroups.com
```
tallclair: Is this needed? I think including an unmanaged group weakens the value of managing this group with a manifest.

cblecker:
@tallclair The reason I didn't reconcile the members for a couple of those, is I wasn't sure if the membership of the lists was intended to be private (like, do we really want to publish the opsgenie address, etc).

If PSC is okay with publishing addresses in git, then I'm fine with adding them and turning it on.

tallclair:
Ah, oops. Yeah, that makes sense.

It looks like we can probably filter messages to ops-genie to only accept messages coming through the google group, but maybe it's easier to just not put the address on github.

cblecker:
same thing with distributors-announce. I can easily dump the current members of that list via the API, but I'm not sure if we want to actually publish them in git.

tallclair:
> same thing with distributors-announce. I can easily dump the current members of that list via the API, but I'm not sure if we want to actually publish them in git.

We already do: https://github.com/kubernetes/security/blob/master/private-distributors-list.md#membership

That's what I was referring to about wanting to keep the lists in sync.

groups/groups.yaml
cblecker: Unsure. I think this was @justaugustus.

cblecker:
Update distributors-announce: https://github.com/kubernetes/k8s.io/pull/429

dims:
@justaugustus there's a question for you inline!

justaugustus:
@dims -- I'm actually in the midst of doing an update on security lists right now!
Once I push that PR, I'll swing back here and see what's remaining to do.

dims:
perfect! thanks @justaugustus

justaugustus:
I've picked up @tallclair's commit in https://github.com/kubernetes/k8s.io/pull/492 and addressed the questions here, so this is now superseded.
/close

PR #428, opened 2019-10-25
Fix typo in Kas network proxy group

Sh4d1:


Sh4d1:
@cblecker very stupid typo that needs fixing I guess :roll_eyes:

PR #429, opened 2019-10-25
Enable reconcile for distributors-announce

cblecker:
/assign @tallclair


tallclair:
Thanks. I cross-referenced with https://github.com/kubernetes/security/blob/master/private-distributors-list.md#membership
The only difference is `release-managers-private@kubernetes.io` and `security-release-team@kubernetes.io`. I think release-managers-private is correct, and the other list should be updated.

/lgtm

PR #432, opened 2019-10-26
promote promoter

listx:
This brings in the fix at
https://github.com/kubernetes-sigs/k8s-container-image-promoter/pull/146
for credential handling.

Blocked by the Prow job change here https://github.com/kubernetes/test-infra/pull/14988, because the current promoter referenced in the existing Prow job will fail.

/cc @thockin 

So this presents an interesting PR race condition, and am stating this for the record. I want the Prow job I am blocked on to use this image, but obviously it's not promoted to prod yet (that's what this PR is for). To get around this, I just reused the temporary prod cred key on my machine to manually promote the image for the US region (which is what the Prow jobs use). So now when we merge this PR, the promoter should promote to the remaining EU and ASIA regions.


listx:
Wondering if I can self-approve as I'm in the OWNERS...

/approve
/lgtm

listx:
Thanks @cblecker ! And now we wait for the postsubmit to succeed...

listx:
And it worked! https://prow.k8s.io/view/gcs/kubernetes-jenkins/logs/post-k8sio-cip/1187911228912046080

The logs somehow got truncated (I can't see ASIA region's promotion actually running in there) but it did happen because I see it in https://pantheon.corp.google.com/gcr/images/k8s-artifacts-prod/ASIA/artifact-promoter/cip?gcrImageListsize=30. So I'm going to chalk it up to some sort of logging hiccup

PR #434, opened 2019-10-28
releng: Enable elevated privileges for Release Managers on SIG Release GCP projects

justaugustus:
- Add groups to delegate IAM access to Release Managers

  - `k8s-infra-release-admins@`: Partial Admin
  - `k8s-infra-release-editors@`: Editor
  - `k8s-infra-release-viewers@`: Viewer

- Add ensure-release-projects to grant rights to Release Managers
  Establishes rights to GCS, GCB, and KMS for Release Managers

Required for prototyping build/stage/release process on k8s-infra (https://github.com/kubernetes/release/issues/911).

/assign @dims @thockin @cblecker 
cc: @kubernetes/sig-release-admins @kubernetes/release-engineering 
/sig release
/area release-eng

### DIFF
--- a/groups/groups.yaml
+++ b/groups/groups.yaml
@@ -484,6 +484,79 @@ groups:
       - thockin@google.com
       - bartsmykla@gmail.com  # sock-puppet for testing
 
+  - email-id: k8s-infra-release-admins@kubernetes.io
+    name: k8s-infra-release-admins
+    description: |-
+      ACL for Release Engineering subproject owners (partial admin access to Release GCP projects)
+      https://git.k8s.io/sig-release/release-managers.md
+    settings:
+      AllowExternalMembers: "true"
+      WhoCanJoin: "INVITED_CAN_JOIN"
+      WhoCanViewMembership: "ALL_MEMBERS_CAN_VIEW"
+      WhoCanViewGroup: "ALL_MEMBERS_CAN_VIEW"
+      WhoCanDiscoverGroup: "ALL_IN_DOMAIN_CAN_DISCOVER"
+      WhoCanInvite: "ALL_MANAGERS_CAN_INVITE"
+      WhoCanAdd: "ALL_MANAGERS_CAN_ADD"
+      WhoCanApproveMembers: "ALL_MANAGERS_CAN_APPROVE"
+      WhoCanModifyMembers: "OWNERS_AND_MANAGERS"
+      WhoCanModerateMembers: "OWNERS_AND_MANAGERS"
+      ReconcileMembers: "true"
+    owners:
+      - caselim@gmail.com
+      - Stephen@agst.us
+      - tpepper@gmail.com
+
+  - email-id: k8s-infra-release-editors@kubernetes.io
+    name: k8s-infra-release-editors
+    description: |-
+      ACL for Patch Release Team and Branch Managers (edit access to Release GCP projects)
+      https://git.k8s.io/sig-release/release-managers.md
+    settings:
+      AllowExternalMembers: "true"
+      WhoCanJoin: "INVITED_CAN_JOIN"
+      WhoCanViewMembership: "ALL_MEMBERS_CAN_VIEW"
+      WhoCanViewGroup: "ALL_MEMBERS_CAN_VIEW"
+      WhoCanDiscoverGroup: "ALL_IN_DOMAIN_CAN_DISCOVER"
+      WhoCanInvite: "ALL_MANAGERS_CAN_INVITE"
+      WhoCanAdd: "ALL_MANAGERS_CAN_ADD"
+      WhoCanApproveMembers: "ALL_MANAGERS_CAN_APPROVE"
+      WhoCanModifyMembers: "OWNERS_AND_MANAGERS"
+      WhoCanModerateMembers: "OWNERS_AND_MANAGERS"
+      ReconcileMembers: "true"
+    owners:
+      - caselim@gmail.com
+      - Stephen@agst.us
+      - tpepper@gmail.com
+    members:
+      - aleksandram@google.com
+      - feiskyer@gmail.com
+      - ctadeu@gmail.com
+      - dmaceachern@vmware.com
+      - hhorl@pivotal.io
+      - idealhack@gmail.com
+
+  - email-id: k8s-infra-release-viewers@kubernetes.io
+    name: k8s-infra-release-viewers
+    description: |-
+      ACL for Release Manager Associates (view access to Release GCP projects)
+      https://git.k8s.io/sig-release/release-managers.md
+    settings:
+      AllowExternalMembers: "true"
+      WhoCanJoin: "INVITED_CAN_JOIN"
+      WhoCanViewMembership: "ALL_MEMBERS_CAN_VIEW"
+      WhoCanViewGroup: "ALL_MEMBERS_CAN_VIEW"
+      WhoCanDiscoverGroup: "ALL_IN_DOMAIN_CAN_DISCOVER"
+      WhoCanInvite: "ALL_MANAGERS_CAN_INVITE"
+      WhoCanAdd: "ALL_MANAGERS_CAN_ADD"
+      WhoCanApproveMembers: "ALL_MANAGERS_CAN_APPROVE"
+      WhoCanModifyMembers: "OWNERS_AND_MANAGERS"
+      WhoCanModerateMembers: "OWNERS_AND_MANAGERS"
+      ReconcileMembers: "true"
+    owners:
+      - caselim@gmail.com
+      - Stephen@agst.us
+      - tpepper@gmail.com
+
   - email-id: k8s-infra-sig-release-prototype@kubernetes.io
     name: k8s-infra-sig-release-prototype
     description: |-
@@ -1005,9 +1078,9 @@ groups:
   - email-id: release-managers-private@kubernetes.io
     name: release-managers-private
     description: |-
-      Release managers for security releases. Membership should include release leads and patch release managers.
-
-      Membership by release leads and patch release managers.
+      Private communications for SIG Chairs, Patch Release Team, and Branch Managers.
+      (internal scheduling and security release coordination)
+      https://git.k8s.io/sig-release/release-managers.md
     settings:
       AllowExternalMembers: "true"
       WhoCanJoin: "INVITED_CAN_JOIN"
@@ -1039,7 +1112,9 @@ groups:
   - email-id: release-managers@kubernetes.io
     name: release-managers
     description: |-
-      Release Managers communications
+      Release Managers communications.
+      Includes all Release Managers.
+      https://git.k8s.io/sig-release/release-managers.md
     settings:
       AllowExternalMembers: "true"
       WhoCanJoin: "INVITED_CAN_JOIN"
--- a/infra/gcp/ensure-release-projects.sh
+++ b/infra/gcp/ensure-release-projects.sh
@@ -1,0 +1,137 @@
+#!/usr/bin/env bash
+#
+# Copyright 2019 The Kubernetes Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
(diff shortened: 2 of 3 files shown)


cblecker:
/hold
Explicit hold for review.

infra/gcp/ensure-release-projects.sh
```diff
@@ -0,0 +1,140 @@
...
+
+ADMINS="release-managers-admins@kubernetes.io"
+WRITERS="release-managers-private@kubernetes.io"
+VIEWERS="release-managers@kubernetes.io"
+
+for REPO; do
+    color 3 "Configuring: ${REPO}"
+
+    # The GCP project name.
+    PROJECT="${REPO}"
```
hoegaarden: Can we maybe just use `PROJECT` instead of `REPO`? As far as I can tell, we never use `REPO` further down, so can we just do `for PROJECT ...`?

infra/gcp/lib.sh
```diff
@@ -287,6 +287,23 @@ function empower_group_for_gcb() {
         --role roles/serviceusage.serviceUsageConsumer
 }
 
+# Grant KMS admin privileges to a principal
+# $1: The GCP project
+# $2: The group email
+function empower_group_for_kms() {
+    if [ $# -lt 2 -o -z "$1" -o -z "$2" ]; then
```
hoegaarden: Should we make that stricter by doing something like the below?
```bash
if [ $# -ne 2 -o -z "$1" -o -z "$2" ]; then
```

I saw, that other funcs here allow additional arguments, but that might just be a c&p "issue" from cases where additional optional arguments are allowed (e.g. `empower_group_to_gcr` or `empower_gcr_admins`).

infra/gcp/ensure-release-projects.sh
```diff
@@ -0,0 +1,140 @@
...
+set -o pipefail
+
+SCRIPT_DIR=$(dirname "${BASH_SOURCE[0]}")
+. "${SCRIPT_DIR}/lib.sh"
+
+function usage() {
+    echo "usage: $0 [repo...]" > /dev/stderr
+    echo "example:" > /dev/stderr
+    echo "  $0 # do all staging repos" > /dev/stderr
+    echo "  $0 coredns # just do one" > /dev/stderr
```
thockin: change "coredns" to a real (or realer-looking) example?

infra/gcp/ensure-release-projects.sh
```diff
@@ -0,0 +1,140 @@
...
+    k8s-staging-release-test
+    k8s-release-test-prod
+)
+
+if [ $# = 0 ]; then
+    # default to all staging projects
+    set -- "${PROJECTS[@]}"
+fi
+
+ADMINS="release-managers-admins@kubernetes.io"
```
thockin: Casually looking at this, the distinctions are not clear.  Why are release-managers just viewers?  Do we need to rename groups?  Can this be solved with docs here?  Why do we need 3 distinct groups?

infra/gcp/ensure-release-projects.sh
```diff
@@ -0,0 +1,140 @@
...
+    echo "usage: $0 [repo...]" > /dev/stderr
+    echo "example:" > /dev/stderr
+    echo "  $0 # do all staging repos" > /dev/stderr
+    echo "  $0 coredns # just do one" > /dev/stderr
+    echo > /dev/stderr
+}
+
+# NB: Please keep this sorted.
+PROJECTS=(
+    k8s-staging-release-test
```
thockin: currently these are covered by the other ensure-* scripts.  Should we a) drop them from those, b) add this logic to those, c) refactor to make the overlap clear?

infra/gcp/ensure-release-projects.sh
justaugustus: Here's what each group would do:
- `release-managers@`: Includes all Release Managers (including Release Manager Associates), so this needs to be least-privilege (allow Associates to view, but not create, edit, delete)
- `release-managers-private@`: Only includes Patch Release Team, Branch Managers, Build Admins, SIG Chairs (the set of users we want working on GCP, so we need edit here)
- `release-managers-admins@`: Only includes SIG Release Chairs and grants additional access to administer KMS (limited set of users, security event mitigation, key mgmt (store GitHub tokens, deb/rpm signing key))

justaugustus:
@thockin --
> Mostly I question the overall factoring - these don't feel like "staging" and "prod" projects (in the sense that we have scripts that exist) any more.

Yeah, totally agreed. This felt awkward to do, as it felt like shimming something into a model it doesn't really fit. 
Each of the SIG Release GCP projects are essentially near-prod/prod, as various areas of the project depend on the artifacts:
- `k8s-staging-release-test/ci`: Prow/intrepid users testing nightlies
- `k8s-staging-release-test/stage`: Validated prior to promotion (but wider area of effect than other staging projects)
- `k8s-release-test-prod/release`: All Kubernetes users

I think this projects would be more accurately named:
- `k8s-release-staging`
- `k8s-release-prod`

> Maybe they need to be totally distinct (removed from other scripts) and even renamed?
> currently these are covered by the other ensure-* scripts. Should we a) drop them from those, b) add this logic to those, c) refactor to make the overlap clear?

In an ideal world/future refactor, perhaps there could be one script, `configure-projects.sh`?
It would include three arrays, `STAGING_PROJECTS`, `PROD_PROJECTS`, and `RELEASE_PROJECTS` and maybe ensure behavior/project config/principal access on a set of `case` statements?

The overlap here is primarily because of copy/pasta.

I think the future should also include yaml config for custom roles, instead of adhoc IAM attaches.

justaugustus:
(Just also noting that I went the copy/pasta route instead of refactoring because it wouldn't interfere with the other projects.)

justaugustus:
(@hoegaarden -- I'll come back and address your review once we land on an approach.)

infra/gcp/ensure-release-projects.sh
cblecker: We should make distinctions here between "Groups that provide IAM roles" and "E-mail distribution lists". I would be concerned with someone getting access to the release artifacts simply because they become a SIG chair and need access to an e-mail list.

groups/groups.yaml
```diff
@@ -1026,7 +1051,7 @@ groups:
       WhoCanApproveMembers: "ALL_MANAGERS_CAN_APPROVE"
       WhoCanModifyMembers: "OWNERS_AND_MANAGERS"
       WhoCanModerateMembers: "OWNERS_AND_MANAGERS"
-      ReconcileMembers: "false"
+      ReconcileMembers: "true"
```
cblecker: This will remove 30 members from the list. Is this intended?

justaugustus: Dropped this commit.

justaugustus:
@cblecker --
> I've mentioned this before in other contexts, but tying privileges to "chairpersonship" of a SIG/WG is an anti-pattern we should try to avoid. These privileges should be tied to A) a unique role defined by the SIG (like "Release Admins" or whatever you want to call it) or B) ownership of a subproject, or even C) a list of people who are just "keyholders".

Yep, yep. I think the last time we chatted about it was here: https://github.com/kubernetes/org/pull/1205#discussion_r326947353.
SIG Chairs are also Release Engineering subproject owners, so I occasionally interchange the two in documentation. I've updated the verbiage there.

> We should make distinctions here between "Groups that provide IAM roles" and "E-mail distribution lists". I would be concerned with someone getting access to the release artifacts simply because they become a SIG chair and need access to an e-mail list.

Combining the two (communication and access) is exactly what we wanted to do for the Release Manager lists. We also moved to kubernetes.io for a more "official" address and more importantly, to get the benefit of being able to manage this access via config/PR.

More details can be found here: https://github.com/kubernetes/sig-release/issues/732

groups/groups.yaml
```diff
@@ -1002,6 +1002,31 @@ groups:
...
+      WhoCanInvite: "ALL_MANAGERS_CAN_INVITE"
+      WhoCanAdd: "ALL_MANAGERS_CAN_ADD"
+      WhoCanApproveMembers: "ALL_MANAGERS_CAN_APPROVE"
+      WhoCanModifyMembers: "OWNERS_AND_MANAGERS"
+      WhoCanModerateMembers: "OWNERS_AND_MANAGERS"
+      ReconcileMembers: "false"
+    owners:
+      - caselim@gmail.com
+      - cmiles@pivotal.io
+      - saugustus@vmware.com
```
thockin: why are we adding two accounts for Stephen and Tim?  This means twice as many credentials that could leak.  I'd like to make a policy that we explicitly NOT do this.

infra/gcp/lib.sh
thockin: this is probably a good idea, but I'm OK to let this PR go in and fix it up across the board.

infra/gcp/ensure-release-projects.sh
```diff
@@ -0,0 +1,140 @@
...
+set -o nounset
+set -o pipefail
+
+SCRIPT_DIR=$(dirname "${BASH_SOURCE[0]}")
+. "${SCRIPT_DIR}/lib.sh"
+
+function usage() {
+    echo "usage: $0 [repo...]" > /dev/stderr
+    echo "example:" > /dev/stderr
+    echo "  $0 # do all staging repos" > /dev/stderr
```
thockin: s/staging repos/release projects

infra/gcp/ensure-release-projects.sh
```diff
@@ -0,0 +1,140 @@
...
+    echo "  $0 coredns # just do one" > /dev/stderr
+    echo > /dev/stderr
+}
```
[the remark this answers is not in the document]
thockin: it serves as docs, at least :)

infra/gcp/ensure-release-projects.sh
thockin: I agree with @cblecker - there are the social groups (email) and the permissions groups, and I am not sure they are the same.   Should being a SIG-chair automatically grant you access to GCP?

infra/gcp/ensure-release-projects.sh
```diff
@@ -0,0 +1,140 @@
...
+if [ $# = 0 ]; then
+    # default to all staging projects
+    set -- "${PROJECTS[@]}"
+fi
+
+ADMINS="release-managers-admins@kubernetes.io"
+WRITERS="release-managers-private@kubernetes.io"
+VIEWERS="release-managers@kubernetes.io"
+
+for REPO; do
```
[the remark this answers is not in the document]
thockin: Once I learned to embrace "$@" I started to prefer the simpler syntax, and this PR follows wstablished patterns

infra/gcp/ensure-release-projects.sh
thockin: yes, the loop can be `for PROJECT` - "repo" is a holdover from the script he copied, I assume :)

infra/gcp/ensure-release-projects.sh
```diff
@@ -0,0 +1,140 @@
...
+    color 3 "Configuring: ${REPO}"
+
+    # The GCP project name.
+    PROJECT="${REPO}"
+
+    # The names of the buckets
+    STAGING_BUCKET="gs://${PROJECT}" # used by humans
```
thockin: This is the piece I want to discuss most - Do we want these projects "governed" by 2 ensure scripts?  This overlaps directly with the other scripts, which means it will drift.

There's a particular set of semantics that we ascribe to a "staging project" and a "prod project".  You are adding on to that, clearly, but it's not clear if you are changing more than that.  In other words - should we make a clean break and remove these release projects from the other scripts?  Or should we build on those?

For example, this script could do something like:


```
# First ensure that the project is a normal staging project
"${ROOT}/ensure-staging-project.sh" release-test
# Now apply customizations for release staging
```

If it helps, we can refactor the existing scripts to make this more obvious.  I knew this was likely to come up, so we it should not be hard - we left room.

thockin:
Thanks!

/lgtm
/approve

justaugustus:
Awesome. Thanks y'all!
/hold cancel

cblecker:
Groups have been created.

PR #450, opened 2019-10-31
K8s YouTube vanity URL

chris-short:
Hello! Every week I typically have to figure out one URL or another to get people to the YouTube channel. Creating this redirect would make my life easier. The @kubernetes/youtube-admins team might agree too.

Signed-off-by: Chris Short <chris@chrisshort.net>

### DIFF
--- a/k8s.io/configmap-nginx.yaml
+++ b/k8s.io/configmap-nginx.yaml
@@ -291,7 +291,7 @@ data:
       }
 
       server {
-        server_name youtube.k8s.io youtube.kubernetes.io;
+        server_name yt.k8s.io youtube.k8s.io youtube.kubernetes.io;
         listen 80;
 
         location / {
--- a/k8s.io/configmap-nginx.yaml
+++ b/k8s.io/configmap-nginx.yaml
@@ -289,4 +289,13 @@ data:
           rewrite ^/.*$  https://prow.k8s.io/tide redirect;
         }
       }
+
+      server {
+        server_name youtube.k8s.io youtube.kubernetes.io;
+        listen 80;
+
+        location / {
+          rewrite ^/.*$  https://www.youtube.com/c/kubernetescommunity redirect;
+        }
+      }
     }
--- a/k8s.io/README.md
+++ b/k8s.io/README.md
@@ -30,6 +30,7 @@ Vanity URL(s)
 | Tide status (formerly submit queue) | https://prow.k8s.io/tide | https://prow.kubernetes.io/tide |
 | TestGrid | https://testgrid.k8s.io | https://testgrid.kubernetes.io |
 | YUM downloads | https://yum.k8s.io | https://yum.kubernetes.io |
+| Kubernetes YouTube | https://yt.k8s.io | https://youtube.k8s.io | https://youtube.kubernetes.io |
 
 NOTE: please see k8s.io/k8s.io/configmap-nginx.yaml for `server` definitions
 
--- a/k8s.io/test.py
+++ b/k8s.io/test.py
@@ -361,6 +361,11 @@ class RedirTest(unittest.TestCase):
             self.assert_temp_redirect(base + '/$path', 'https://prow.k8s.io/tide',
                 path=rand_num())
 
+    def test_youtube(self):
+        for base in ('yt.k8s.io', 'youtube.k8s.io', 'youtube.kubernetes.io'):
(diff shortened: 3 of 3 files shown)


k8s.io/configmap-nginx.yaml
```diff
@@ -289,4 +289,13 @@ data:
           rewrite ^/.*$  https://prow.k8s.io/tide redirect;
         }
       }
+
+      server {
+        server_name youtube.k8s.io youtube.kubernetes.io;
```
dims: do you want to do `go.k8s.io/youtube` instead? (lines 229-248)

chris-short: youtube.k8s.io is actually shorter.

chris-short: This is a precedent setter though I guess, no? Was there a preferred format that I missed?

mrbobbytables:
Holding for @castrojo to weigh in (he's out of the office but will be back next week). As far as I know, we previously had a single url to use for community meetings but switched to a generated link for each one. I don't recall the specific reason for it however.

/hold

chris-short:
This is for the overarching channel. Not an individual meeting. If memory serves, @castrojo didn't want to have to manage a vanity URL with the Community meeting itself because the URL does change with every meeting. Since this is the high-level channel, that's not a concern here.

mrbobbytables:
> This is for the overarching channel. Not an individual meeting. If memory serves, @castrojo didn't want to have to manage a vanity URL with the Community meeting itself because the URL does change with every meeting. Since this is the high-level channel, that's not a concern here.

Ahh, sorry -- I thought that's what you were getting at with tracking down the url every week.

mrbobbytables:
Forgot to release hold in previous comment:
/hold cancel

chris-short:
Added test for @thockin and updated README after re-reading the right one. It's raining readmes this morning.

k8s.io/configmap-nginx.yaml
mrbobbytables: Thinking about it -- I do kind of like the idea of `yt.k8s.io`/ `youtube.k8s.io` -- the default would go to the landing page for kubernetes, and it opens the door to adding more youtube specific redirects. I'm thinking something like:
`yt.k8s.io/sig-release` -> sig release playlist.

That winds up being a lot friendlier and easier to link to than this:
https://www.youtube.com/playlist?list=PL69nYSiGNLP3QKkOsDsO6A0Y1rhgP84iZ

chris-short: I'm down for that! That does make for a nice clean experience between various platforms (GitHub, YouTube, etc.). I have a to do list item for this.

munnerz:
I may be mistaken, but I think this will also require updates to the DNS config for both k8s.io and kubernetes.io to point it at the redirector service: https://github.com/kubernetes/k8s.io/blob/master/dns/zone-configs/k8s.io.yaml

https://github.com/kubernetes/k8s.io/blob/6e4beb92c9798dd85b3457a3f4ff2ee670502395/dns/zone-configs/k8s.io.yaml#L141-L144

dims:
/hold 

looks like there's some more stuff to be done here per @munnerz

stp-ip:
There are no new DNS entries needed currently, but it might be necessary once we move to the redirector. As this is still running in the old cluster it would need migration to either the new cluster or the new redirector anyway.

Will work on pushing the redirector forward so that we might start right with the new redirector depending on how pressing this vanity url is.

chris-short:
So what should my next steps be? I thought I found where DNS entries were
needed Before KubeCon
-- 

Chris Short
He/Him/His
Web <https://chrisshort.net/> | Newsletter <https://devopsish.com/> |
Podcast <http://podcast.podctl.com/> | Twitter
<https://twitter.com/ChrisShort> | LinkedIn
<https://linkedin.com/in/thechrisshort>

stp-ip:
Missed the use of sub domains. This needs a few CNAMEs to point to he current redirector (nginx) you already configured. According to your entries that would be: 

https://github.com/kubernetes/k8s.io/blob/master/dns/zone-configs/k8s.io.yaml
```
youtube:
  type: CNAME
  value: redirect.k8s.io.
yt:
  type: CNAME
  value: redirect.k8s.io.
```

https://github.com/kubernetes/k8s.io/blob/master/dns/zone-configs/kubernetes.io.yaml
```
youtube:
  type: CNAME
  value: redirect.k8s.io.
```

PR #453, opened 2019-11-05
Enable clusterViewer for k8s users

thockin:
The `gke-security-groups` groups is a list of groups.  Each of those
groups exists to grant RBAC access to a particular namespace.  Ergo, any
human who is a member (transitively) of `gke-security-groups` needs some
basic abilities to get credentials for the cluster.

With this, an otherwise unprivileged user can launch cloud shell,
`gcloud container clusters get-credentials <cluster>` and start working
in their namespace (once Bart's RBAC stuff lands).


thockin:
We don't yet have a script around org-wide permissions, but I manually added this group as "project browser" at the org level, which allows access to the project selector UI.  We may want more work on this (more than just GKE users) and it still doesn't seem to list orgs.

PR #459, opened 2019-11-10
add script to create conformance buckets, add bucket for capo

sbueringer:
For context see: https://github.com/kubernetes/test-infra/issues/15081

### DIFF
--- a/infra/gcp/ensure-conformance-storage.sh
+++ b/infra/gcp/ensure-conformance-storage.sh
@@ -54,10 +54,10 @@ for REPO; do
     color 3 "Configuring conformance: ${REPO}"
 
     # The GCP project name.
-    PROJECT="k8s-cnfm-${REPO}"
+    PROJECT="k8s-conform-${REPO}"
 
     # The group that can write to this conformance repo.
-    WRITERS="k8s-infra-conformance-${REPO}@kubernetes.io"
+    WRITERS="k8s-infra-conform-${REPO}@kubernetes.io"
 
     # The names of the buckets
     CONFORMANCE_BUCKET="gs://${PROJECT}" # used by humans
--- a/infra/gcp/ensure-conformance-storage.sh
+++ b/infra/gcp/ensure-conformance-storage.sh
@@ -14,12 +14,12 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-# This script is used to create a new "conformance" bucket in GCS.
+# This script is used to create the "k8s-conform" project and corresponding buckets in GCS.
 #
-# Each conformance bucket exists in its own GCP project, and is writable by a
-# dedicated Google group.
+# All conformance buckets are created in the "k8s-conform" GCP project, and are writable by
+# dedicated Google groups.
 #
-# It will have a layout which is readable by testgrid.
+# The buckets will have a layout which is readable by testgrid:
 # (See also: https://github.com/kubernetes/test-infra/tree/master/gubernator#gcs-layout)
 # * each test result is stored in a "directory" with some monotonically increasing run identifier (commonly unix epoch time for third party uploads)
 # * they should contain a junit.*.xml file with the structured test results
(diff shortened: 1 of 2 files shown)


dims:
/assign @listx @thockin 

cc @BenTheElder

infra/gcp/ensure-conformance-storage.sh
```diff
@@ -0,0 +1,100 @@
...
+# NB: Please keep this sorted.
+CONFORMANCE_PROJECTS=(
+    capi-openstack
+)
+if [ $# = 0 ]; then
+    # default to all staging projects
+    set -- "${CONFORMANCE_PROJECTS[@]}"
+fi
+
+for REPO; do
```
sbueringer: fyi, I'm not sure how the REPO variable is actually filled. I just replicated the behavior of: https://github.com/kubernetes/k8s.io/blob/master/infra/gcp/ensure-staging-storage.sh#L67

thockin: it iterates "$@", which is either set from commandline or above in the `set --`

infra/gcp/ensure-conformance-storage.sh
```diff
@@ -0,0 +1,100 @@
...
+    color 3 "Configuring staging: ${REPO}"
+
+    # The GCP project name.
+    PROJECT="k8s-conformance-${REPO}"
```
thockin: minor point: there's a limited character count available in some contexts, and "conformance" eats a lot of that.  I suggest "cnfm" or "c9e" or something

infra/gcp/ensure-conformance-storage.sh
```diff
@@ -0,0 +1,100 @@
...
+
+    # The group that can write to this staging repo.
+    WRITERS="k8s-infra-conformance-${REPO}@kubernetes.io"
+
+    # The names of the buckets
+    STAGING_BUCKET="gs://${PROJECT}" # used by humans
+    ALL_BUCKETS=("${STAGING_BUCKET}")
+
+    # A short expiration - it can always be raised, but it is hard to lower
```
thockin: Do we expect to EVER delete conformance results?  That doesn't seem right to me.

infra/gcp/ensure-conformance-storage.sh
```diff
@@ -0,0 +1,100 @@
...
+    ensure_project "${PROJECT}"
+
+    # Enable writers to use the UI
+    color 6 "Empowering ${WRITERS} as project viewers"
+    empower_group_as_viewer "${PROJECT}" "${WRITERS}"
+
+    # Every project gets some GCS buckets
+
+    # Enable GCS APIs
+    color 6 "Enabling the GCS API"
```
thockin: I assume you'll want to `ensure_gcs_bucket_retention` too ?

infra/gcp/ensure-conformance-storage.sh
```diff
@@ -0,0 +1,100 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# This script is used to create a new "conformance" bucket in GCS.
+#
+# Each conformance bucket exists in its own GCP project, and is writable by a
+# dedicated googlegroup.
```
thockin: Can you say more about what gets written here and what the naming conventions should be?

BenTheElder: conformance test results in a way testgrid can consume:
- each is in a "directory" with some monotonically increasing UUID (commonly unix epoch time for third party uploads)
  - they should contain a `junit.*\.xml` file with the structured test results
  - they should (optionally but commonly) contain a build-log.txt with the raw test output
  - they should contain two small json files `started.json` and `finished.json` with some metadata

https://github.com/kubernetes/test-infra/tree/master/gubernator#gcs-layout

BenTheElder: we have a small python tool in test-infra to consume the build log and junit and upload these.
prow will produce results like this.

BenTheElder:
thanks for the PR!!

infra/gcp/ensure-conformance-storage.sh
thockin: I meant "...in the comments" :)

nit: timestamp is not a UUID.

infra/gcp/ensure-conformance-storage.sh
BenTheElder: this is reflecting the current naming scheme for the google.com owned buckets, it was fairly arbitrary at the time.

infra/gcp/ensure-conformance-storage.sh
BenTheElder: > nit: timestamp is not a UUID.

it's true, on prow these are actual snowflake IDs, however third party test results typically use an uploader that just uses the start time in unix epoch, as running two runs at the same start time doesn't make sense for this use case.

BenTheElder: (and we don't expect people to deploy prow to produce these)

BenTheElder: s/UUID/run identifier/

infra/gcp/ensure-conformance-storage.sh
sbueringer: Ah thx didn't know that

infra/gcp/ensure-conformance-storage.sh
sbueringer: I would use "cnfm" but only for the Project I guess? Shouldn't be an issue on the Google group?

infra/gcp/ensure-conformance-storage.sh
sbueringer: Make sense, I removed the auto deletion part. Should I add the retention policy or just keep it as is without auto deletion?

```bash
color 6 "Ensuring the GCS bucket retention policy is set: ${prj}"
RETENTION="10y"
ensure_gcs_bucket_retention "gs://${prj}" "${RETENTION}"
```

infra/gcp/ensure-conformance-storage.sh
sbueringer: I added 10y retention. Did I get it correctly: 
* no files are deleted automatically
* only files older than 10y can be deleted manually?

infra/gcp/ensure-conformance-storage.sh
thockin: I forget the exact limits, but ISTR the list name had to be less than 30 characters, too - "capi-foobar" came about because "cluster-api-provider-foobar" overlfowed.

I value consistency - if we train our brain to read "cnfm" as "conform", let's do it everywhere.

"cnfm" is clunkly.  "cfrm" maybe better?  "cnfrm" is as long as "infra".  "conform" is as long as "staging" - that seems OK and parallel.

infra/gcp/ensure-conformance-storage.sh
thockin: correct

infra/gcp/ensure-conformance-storage.sh
```diff
@@ -0,0 +1,100 @@
...
+CONFORMANCE_PROJECTS=(
+    capi-openstack
+)
+if [ $# = 0 ]; then
+    # default to all staging projects
+    set -- "${CONFORMANCE_PROJECTS[@]}"
+fi
+
+for REPO; do
+    color 3 "Configuring staging: ${REPO}"
```
thockin: s/staging/conformance

infra/gcp/ensure-conformance-storage.sh
thockin: not pushed yet?

infra/gcp/ensure-conformance-storage.sh
sbueringer: I hope I know got them all :)

infra/gcp/ensure-conformance-storage.sh
sbueringer: Good point. So let's use conform

infra/gcp/ensure-conformance-storage.sh
sbueringer: Yup but know. Should start sending comments after I pushed :)

thockin:
sorry, I hit the wrong button - I didn't mean to send, was a note for
myself :)

thockin:
I'm approving, but @BenTheElder gets final LGTM

/approve

justaugustus:
@thockin --
> This does not use GCR - it coud technically be all in one project, different buckets.

I was coming here to say this! 
Seems that we're sprawling a bit here, and I think we'd be just as easily served with a single conformance GCP project with separate buckets or conformance buckets within existing projects e.g., for CAPI, why not store conformance info within their primary GCP project or the respective provider projects?

sbueringer:
@justaugustus 

At least in case of ClusterAPI OpenStack the only GCP project which already exists (afaik) is named `k8s-staging-capi-openstack` which would probably be not the correct on the add a bucket for conformance test results.

Apart from that I don't really know how Kubernetes GCP projects/buckets are usually organized. So I let you discuss and adjust the script accordingly :)

BenTheElder:
Currently I maintain a single GCP project with multiple buckets [for conformance test results[ and service accounts 1:1, with specific IAM permissions to the buckets matched to service accounts.

I'm not sure which pattern we want going forward.

thockin:
I'd be OK with a single `k8s-conformance` or `k8s-conformance-results` project with N buckets and a mailing list IAM to write to each bucket

sbueringer:
> I'd be OK with a single `k8s-conformance` or `k8s-conformance-results` project with N buckets and a mailing list IAM to write to each bucket

Okay so one project with multi buckets. Just that I get it correctly. Only one google group for the project or one per bucket?

BenTheElder:
let's do one project with a group per bucket, someone will need to own the project as well though.

sbueringer:
@BenTheElder Is there's already an existing group I should reuse or should I create a new one?

BenTheElder:
not sure, @dims wdyt?

sbueringer:
@thockin @BenTheElder @dims 

I updated the script to use a single project for all buckets. For now I used the "k8s-infra-conform@kubernetes.io" group for the project. If we want to use this group, instead of reusing an existing one I also have to define this group in the groups.yaml (which I haven't done yet).

I'm not sure which rights are necessary on the project. I used `empower_group_as_viewer` again like in the previous version of the script.

sbueringer:
> @thockin @BenTheElder @dims
> 
> I updated the script to use a single project for all buckets. For now I used the "[k8s-infra-conform@kubernetes.io](mailto:k8s-infra-conform@kubernetes.io)" group for the project. If we want to use this group, instead of reusing an existing one I also have to define this group in the groups.yaml (which I haven't done yet).
> 
> I'm not sure which rights are necessary on the project. I used `empower_group_as_viewer` again like in the previous version of the script.

Just wanted to let you know that I'm on vacation until end of Dezember. I would continue this PR then. If anyone wants to takeover in the meantime, be my guest :)

dims:
@BenTheElder @sbueringer   let's use a new one

BenTheElder:
SGTM

dims:
/hold

let's wait for @sbueringer to get back :)

infra/gcp/ensure-conformance-storage.sh
```diff
@@ -0,0 +1,98 @@
...
+    echo "usage: $0 [repo...]" > /dev/stderr
+    echo "example:" > /dev/stderr
+    echo "  $0 # do all conformance repos" > /dev/stderr
+    echo "  $0 coredns # just do one" > /dev/stderr
+    echo > /dev/stderr
+}
+
+PROJECT="k8s-conform"
+
+PROJECT_VIEWER="k8s-infra-conform@kubernetes.io"
```
thockin: I don't think we need this, do we?  What is it for?

sbueringer: Seems to be a leftover from copying this script. I removed PROJECT_VIEWER

sbueringer:
@thockin please take another look :)

sbueringer:
> @thockin please take another look :)

*push* :) 
@dims @BenTheElder

thockin:
Thanks!

/lgtm
/approve

thockin:
I am actuating this.  One function got renamed, fix coming.

PR #462, opened 2019-11-11
Fix groups settings for steering list

cblecker:
I had fixed settings so anyone could post to the steering public list, but it still wasn't publicly viewable

/assign @dims


dims:
/approve
/lgtm
/hold 

( for actually running the script, please unhold when you are able to @cblecker )

cblecker:
/hold cancel

Thanks @dims!

cblecker:
Applied

Issue #442, opened 2019-10-30
RBAC role for users - are these resources enough?


bartsmykla:
As we discussed in last few zoom calls I was working on creating the role (RBAC) for the users.

After some conversation with @thockin this is the result of my work:
```
---
kind: Role
apiVersion: rbac.authorization.k8s.io/v1beta1
metadata:
  name: namespace-user
  namespace: {{namespace}}
rules:
- apiGroups: [""]
  resources: ["configmaps", "endpoints", "persistentvolumeclaims", "pods", "resourcequotas", "services"]
  verbs: ["*"]
- apiGroups: ["certmanager.k8s.io"]
  resources: ["certificates"]
  verbs: ["*"]
- apiGroups: ["coordination.k8s.io"]
  resources: ["leases"]
  verbs: ["*"]
- apiGroups: ["batch"]
  resources: ["cronjobs", "jobs"]
  verbs: ["*"]
- apiGroups: ["autoscaling"]
  resources: ["horizontalpodautoscalers"]
  verbs: ["*"]
- apiGroups: ["apps"]
  resources: ["deployments"]
  verbs: ["*"]
- apiGroups: ["extensions"]
  resources: ["deployments", "ingresses", "networkpolicies"]
  verbs: ["*"]
- apiGroups: ["networking.k8s.io"]
  resources: ["networkpolicies"]
  verbs: ["*"]
- apiGroups: [""]
  resources: ["secrets"]
  verbs: ["list"]
- apiGroups: ["storage.k8s.io"]
  resources: ["storageclasses"]
  verbs: ["list"]
- apiGroups: ["scheduling.k8s.io"]
  resources: ["priorityclasses"]
  verbs: ["list"]
- apiGroups: ["rbac.authorization.k8s.io"]
  resources: ["clusterrolebindings", "clusterroles", "rolebindings", "roles"]
  verbs: ["get", "list"]
- apiGroups: [""]
  resources: ["events"]
  verbs: ["get", "list"]
---
kind: RoleBinding
apiVersion: rbac.authorization.k8s.io/v1beta1
metadata:
  name: bart-testing-namespace-user-binding
  namespace: bart-testing
subjects:
- kind: Group
  name: k8s-infra-rbac-{{namespace}}@kubernetes.io
  namespace: {{namespace}}
roleRef:
  kind: Role
  name: namespace-user
  apiGroup: rbac.authorization.k8s.io
```
This issue was created to ask you if these resources will be enough?

Bart

[...]

PR #463, opened 2019-11-11
Created script ensure-namespaces.sh

bartsmykla:
Closes: https://github.com/kubernetes/k8s.io/issues/442

Script will automate:
1. Creating namespace for project
2. Applying namespace-user role for project's namespace
3. Applying namespace-user role binding for project's namespace
4. Ensuring there is only one, and proper role binding for
   just created namespace, and if not failing the script

In lines 79-83 there is just example project, we need
to put list of real projects

Cc. @thockin @cblecker @dims @listx 

Signed-off-by: Bart Smykla <bartek@smykla.com>


infra/gcp/namespaces/ensure-namespaces.sh
```diff
@@ -0,0 +1,100 @@
...
+    echo > /dev/stderr
+}
+
+if [ $# != 0 ]; then
+    usage
+    exit 1
+fi
+
+function create_namespace() {
+  kubectl create "$1"
```
thockin: don't you want `create namespace "$1"`?

infra/gcp/namespaces/ensure-namespaces.sh
```diff
@@ -0,0 +1,100 @@
...
+    echo "usage: $0" > /dev/stderr
+    echo > /dev/stderr
+}
+
+if [ $# != 0 ]; then
+    usage
+    exit 1
+fi
+
+function create_namespace() {
```
thockin: Sorry to be a pedant, but shell is so easy to make mistakes in...

each function should validate args:

```
function create_namespace() {
    if [ $# != 1 ]; then
        echo "create_namespace(name) requires 1 argument" >&2
        return 1
    fi
    ns="$1"

}
```

infra/gcp/namespaces/namespace-user-role.yml
```diff
@@ -0,0 +1,46 @@
+---
+kind: Role
+apiVersion: rbac.authorization.k8s.io/v1beta1
+metadata:
+  name: namespace-user
+  namespace: {{namespace}}
+rules:
+  - apiGroups: [""]
```
sonasingh46: nit: Do we need a `*` here too ?

sonasingh46: Or it is fine ? Not sure

infra/gcp/namespaces/ensure-namespaces.sh
```diff
@@ -0,0 +1,100 @@
...
+  kubectl create "$1"
+}
+
+function apply_role() {
+  role=$(sed -e "s/{{namespace}}/$1/" "$role_filename")
+
+  kubectl apply -f "$role"
```
thockin: I don't think this works?  -f wants a filename and you are giving it the content.  You can either do:

`sed ... | kubectl apply -f  -` 

... or save to a temp file.  If you use stdin as above, you'll want to echo something useful - kubectl won't tell you much

thockin: for pedanticness, please use the `-n "${namespace}"` flag - that way if it doesnt match the YAML, kubectl can yell.

infra/gcp/namespaces/ensure-namespaces.sh
```diff
@@ -0,0 +1,100 @@
...
+# for k8s-infra projects
+
+set -o errexit
+set -o nounset
+set -o pipefail
+
+SCRIPT_DIR=$(dirname "${BASH_SOURCE[0]}")
+. "${SCRIPT_DIR}/lib.sh"
+
+role_filename="namespace-user-role.yml"
```
thockin: nit: lower-case names tend to be for local variables, UPPER_CASE_NAMES for globals.

infra/gcp/namespaces/ensure-namespaces.sh
```diff
@@ -0,0 +1,100 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# This script creates namespaces, applies roles and role bindings and
+# checks if assigned role binding is the only one in the namespace
+# for k8s-infra projects
```
thockin: document this assumes kubectl in $PATH

infra/gcp/namespaces/namespace-user-role.yml
sonasingh46: Ignore it. Got the answer.

infra/gcp/namespaces/ensure-namespaces.sh
```diff
@@ -0,0 +1,100 @@
...
+#
+
+# Example project
+EXAMPLE_PROJECT="example_project"
+
+ALL_PROJECTS=(
+    "${EXAMPLE_PROJECT}"
+)
+
+for prj in "${ALL_PROJECTS[@]}"; do
```
thockin: Let's talk about context.  This can either assume a default context/cluster or can take a param explicitly for the cluster name.  The latter is my pref.

```
$ ensure-namespaces.sh
usage: ensure-namespaces.sh <cluster>

$ ensure-namespaces.sh my-renamed-aaa-cluster-entry
Ensuring namespaces in my-renamed-aaa-cluster-entry
```

Then you'll have to set a global CLUSTER variable and use `-c "${CLUSTER}"` in all kubectl calls.  The main benefit (aside from not setting context) is that each call to kubectl has a -n and a -c flag, making it pretty unambiguous what is happening.

infra/gcp/namespaces/ensure-namespaces.sh
```diff
@@ -0,0 +1,100 @@
...
+  role_bindings_count="$(kubectl get rolebindings -o json -n "$project_name" | jq -r '.items | length')"
+
+  if [ "$role_bindings_count" -gt 1 ]; then
+    echo "Only one role binding per namespace is allowed (current: $role_bindings_count)"
+    exit 1
+  fi
+
+  role_binding_name="$(kubectl get rolebindings -o json -n "$project_name" | jq -r '.items[0].metadata.name')"
+
+  if [ "$role_binding_name" != "${project_name}${role_binding_name_postfix}" ]; then
```
thockin: The object names no longer has the project_name in it, right?  So you don't need postfix, just "namespace-user" literal

infra/gcp/namespaces/namespace-user-role-binding.yml
```diff
@@ -0,0 +1,14 @@
+---
+kind: RoleBinding
+apiVersion: rbac.authorization.k8s.io/v1beta1
+metadata:
+  name: namespace-user
+  namespace: {{namespace}}
+subjects:
+  - kind: Group
+    name: k8s-infra-rbac-{{namespace}}@kubernetes.io
+    namespace: {{namespace}}
```
thockin: Docs say:

```
If the object kind is non-namespace, such as "User" or "Group", and this value is not empty the Authorizer should report an error. 
```

So I think we don't want this?

thockin:
I know @cblecker loves bash.  One thing we could explore here is using kustomize to make this less imperative?

I still like the check for other rolebindings

infra/gcp/namespaces/ensure-namespaces.sh
bartsmykla: Ahhh, sorry for that stupid mistake :-(

infra/gcp/namespaces/namespace-user-role-binding.yml
bartsmykla: Agree :-)

infra/gcp/namespaces/ensure-namespaces.sh
bartsmykla: ack

infra/gcp/namespaces/ensure-namespaces.sh
bartsmykla: ack

infra/gcp/namespaces/ensure-namespaces.sh
bartsmykla: agree :-)

bartsmykla: Done

infra/gcp/namespaces/ensure-namespaces.sh
bartsmykla: Done

infra/gcp/namespaces/ensure-namespaces.sh
bartsmykla: I've added the flag `--kubectl-path | -p` where you can specify path for kubectl

infra/gcp/namespaces/ensure-namespaces.sh
bartsmykla: Done :-)

infra/gcp/namespaces/namespace-user-role-binding.yml
bartsmykla: Fixed

bartsmykla:
Hi @thockin thank you for a code review! I included all of your comments in new commit. Please check again :-)

infra/gcp/namespaces/ensure-namespaces.sh
bartsmykla: Fixed

infra/gcp/namespaces/ensure-namespaces.sh
bartsmykla: Fixed

infra/gcp/namespaces/ensure-namespaces.sh
```diff
@@ -0,0 +1,159 @@
...
+  if [[ -z "${KUBECTL}" ]]; then
+    if ! [ -x "$(command -v kubectl)" ]; then
+      echo >&2 "kubectl not found in \$PATH and --kubectl-path flag is not set. Aborting.";
+      exit 1;
+    else
+      KUBECTL="kubectl"
+    fi
+  fi
+
+  if [ ${#args[@]} -lt 1 ]; then
```
thockin: idomatically, any $@ variable should be quoted: `"${#args[@]}"`

infra/gcp/namespaces/ensure-namespaces.sh
```diff
@@ -0,0 +1,159 @@
...
+}
+
+function apply_role() {
+  if [ $# != 1 ]; then
+    echo "apply_role(name) requires 1 argument" >&2
+    return 1
+  fi
+  project_name="$1"
+
+  sed -e "s/{{namespace}}/$project_name/" "$ROLE_FILENAME" | $KUBECTL apply -c $CLUSTER --kubeconfig $KUBECONFIG_PATH -n "$namespace" -f -
```
thockin: nit:  linewrap at `|`:

```
sed -e "s/{{namespace}}/$project_name/" "$ROLE_FILENAME" \
    | $KUBECTL apply -c $CLUSTER --kubeconfig $KUBECONFIG_PATH -n "$namespace" -f -

Comment (thockin) [inline infra/gcp/namespaces/ensure-namespaces.sh]: Always quote variables when they need to be a single token.

`"${KUBECTL}"` and `${CLUSTER}"`

File: infra/gcp/namespaces/ensure-namespaces.sh [hunk 12]
```diff
@@ -0,0 +1,159 @@
...
+}
+
+function ensure_only_one_proper_role_binding() {
+  if [ $# != 1 ]; then
+    echo "ensure_only_one_proper_role_binding(name) requires 1 argument" >&2
+    return 1
+  fi
+  project_name="$1"
+  namespace="$project_name"
+  role_bindings_count="$($KUBECTL get rolebindings -c $CLUSTER --kubeconfig $KUBECONFIG_PATH -o json -n "$namespace" | jq -r '.items | length')"
```
thockin: can linewrap here :)

infra/gcp/namespaces/ensure-namespaces.sh
```diff
@@ -0,0 +1,159 @@
...
+}
+
+parse_args "$@";
+
+#
+# Project names
+#
+
+# Example project
+EXAMPLE_PROJECT="example_project"
```
thockin: no real need for variables here, you can put literals into the array

infra/gcp/namespaces/ensure-namespaces.sh
bartsmykla: Done

infra/gcp/namespaces/ensure-namespaces.sh
bartsmykla: Done

infra/gcp/namespaces/ensure-namespaces.sh
bartsmykla: Done

infra/gcp/namespaces/ensure-namespaces.sh
bartsmykla: Done

infra/gcp/namespaces/ensure-namespaces.sh
bartsmykla: Changed :-)

bartsmykla:
@thockin updated PR after your last comments :-)

infra/gcp/namespaces/ensure-namespaces.sh
```diff
@@ -0,0 +1,160 @@
...
+}
+
+function create_namespace() {
+  if [ $# != 1 ]; then
+    echo "create_namespace(name) requires 1 argument" >&2
+    return 1
+  fi
+  namespace="$1"
+
+  $KUBECTL create -c "$CLUSTER" --kubeconfig "$KUBECONFIG_PATH" namespace "$namespace"
```
thockin: further pedanticness:  KUBECTL should be quoted, in case it includes a space

```
$ X="a b c"

$ $X
bash: a: command not found

$ "$X"
bash: a b c: command not found
```

infra/gcp/namespaces/ensure-namespaces.sh
```diff
@@ -0,0 +1,160 @@
...
+}
+
+function ensure_only_one_proper_role_binding() {
+  if [ $# != 1 ]; then
+    echo "ensure_only_one_proper_role_binding(name) requires 1 argument" >&2
+    return 1
+  fi
+  project_name="$1"
+  namespace="$project_name"
+  role_bindings_count="$($KUBECTL get rolebindings -c "$CLUSTER" --kubeconfig "$KUBECONFIG_PATH" -o json -n "$namespace" \
```
thockin: don't need to quote `$(...)` expressions, which contain quotes themselves

```
$ X=$(echo a b c)

$ echo $X
a b c
```

bartsmykla: Fixed :-)

infra/gcp/namespaces/ensure-namespaces.sh
bartsmykla: Fixed :-)

thockin:
Squash?

bartsmykla:
@thockin done

Issue #415, opened 2019-10-23
dns image is using python2


BenTheElder:
we should probably fix that, looks like the script is probably already py3 compatible
```
WARNING: You are using pip version 19.1.1, however version 19.3.1 is available.
You should consider upgrading via the 'pip install --upgrade pip' command.
freezing deps, please make sure requirements.txt matches
DEPRECATION: Python 2.7 will reach the end of its life on January 1st, 2020. Please upgrade your Python as Python 2.7 won't be maintained after that date. A future version of pip will drop support for Python 2.7.
```

BenTheElder:
blocked on a release of octodns following https://github.com/github/octodns/issues/201

BenTheElder:
thanks!

PR #469, opened 2019-11-14
Update Dockerfile and requirements.txt for python 3.7

puerco:
Hi Everyone,
I'm updating the octodns-docker image and the requirements.txt file to run under Python 3.7. 

octodns 0.9.9 was released and now supports Python 3.7


dns/octodns-docker/Dockerfile
```diff
@@ -12,9 +12,8 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-# python:2-alpine as of July 11th, 2019
-ARG BASE_IMAGE=python@sha256:9a2db929c9e4c65118cb1da9ddb1d64f3ba8e4a722f07f3f2d34b58b1e8556b1
-FROM ${BASE_IMAGE}
+# python:3.7.5-alpine3.10 as of November 14th, 2019
```
cblecker: What's the reason for selecting this tag, as opposed to `python:3-alpine`?

puerco: I tried to mimic the way the last image was selected, the latest image and then frozen with the digest.
It seems that no changes are needed to check-zone.py but io can only test it so far

cblecker: Yes, but the way the tag the digest was based off of was `python:2-alpine`, as noted in the comment. Could we not use `python:3-alpine` to select the new digest, instead of `python:3.7.5-alpine3.10`?

puerco: OK, I've added a new commit with the new image digest for 3-alpine which currently corresponds to 3.8.0-alpine3.10.

BenTheElder:
@cblecker I guess we don't have CI for this ... manual testing then?

dns/octodns-docker/Dockerfile
```diff
@@ -12,9 +12,8 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-# python:2-alpine as of July 11th, 2019
-ARG BASE_IMAGE=python@sha256:9a2db929c9e4c65118cb1da9ddb1d64f3ba8e4a722f07f3f2d34b58b1e8556b1
-FROM ${BASE_IMAGE}
```
cblecker: Why was this abstraction removed?

puerco: No reason in particular, I think I removed it when I was trying out the image digests.

dns/octodns-docker/requirements.txt
```diff
@@ -1,25 +1,42 @@
...
+pluggy==0.13.0
+pprintpp==0.4.0
+protobuf==3.10.0
+py==1.8.0
+pyasn1==0.4.7
+pyasn1-modules==0.2.7
+pycountry==19.8.18
+pycountry-convert==0.7.2
+pyparsing==2.4.5
+pytest==5.2.2
```
cblecker: Looks like there's some unneeded test dependencies.

Try starting up a new python:3-alpine container, running a `pip install octodns==0.9.9`, and then a `pip freeze` and copy the output into this file

dns/octodns-docker/Dockerfile
cblecker: It's helpful to have this there, because you can then test by passing in a --build_arg for BASE IMAGE

dns/octodns-docker/requirements.txt
puerco: Ready, I added the google-cloud-dns dependecy as well.

dns/octodns-docker/Dockerfile
puerco: OK, I've added the abstraction back.

BenTheElder:
I'm not sure how we manage this utility to make sure the changes are safe. Lots of dependency upgrades...

puerco:
I've been testing the script under python3 on a different zone on GCP and it seems to run fine, although I'm not sure if it does what it is supposed to:

What it currently does is read the dns zones **[from Cloud DNS](https://github.com/kubernetes/k8s.io/blob/master/dns/octodns-docker/check-zone.py#L196)** and compare them to what the resolvers are answering. In other words it is just testing if gcp is doing its job.

From what I gather from @cblecker 's comment, it should compare the zones [from the YAML configuration](https://github.com/kubernetes/k8s.io/blob/master/dns/check-zone.sh#L45) against the actual dns responses from the resolvers. Am I wrong? I have a patch ready if this is the intended behavior.

dims:
i just tried this patch and it seems to be working. @puerco please do file a follow up with the patch you mentioned

/approve
/lgtm

puerco:
Thanks @dims i will send the patch in the morning. BTW, this resolves issue #415

BenTheElder:
thanks!

PR #477, opened 2019-11-22
add backup test

listx:
This moves most of the backup logic into a new backup_lib.sh. This logic
is then extended for testing purposes in backup_test.sh (and also tested
there). The backup_prod.sh is the same as the old backup.sh, and is
symlinked from backup.sh so as to not break old jobs.

Please see the README.md for an explanation of how the backup_test.sh
script works.

/cc @thockin @dims @cblecker


listx:
Also, this PR adds a basic credential check at the beginning to fail fast if creds don't exist.

listx:
I have to follow up this PR by modifying the ci-k8sio-backup job in k8s/test-infra to reflect the changes.

PR #484, opened 2019-12-04
Add image-builder.sigs.k8s.io

moshloop:
The netlify project still needs to be created, not sure if there is an automated process for that?

@davidewatson @detiber


detiber:
lgtm assuming the cname matches the eventual netlify URL that is created.

hh:
I'm heading to Seoul, so might not be able to get to this soon.
This process is still manual.
/lgtm

PR #493, opened 2019-12-07
Fix members list for security-release-team

cblecker:
DLs can't be owners of other DLs. This fixes that for security-release-team

/assign @dims @justaugustus 
/approve


justaugustus:
Thanks for catching that, y'all!
/lgtm
/approve

PR #495, opened 2019-12-10
Fixing bug where zone comparison was made from GCP's records instead of YAML confg's

puerco:
Hi infra, while porting this program to python3 I found two bugs:

First, check-zone.py is supposed to check our YAML configuration files against google's DNS resolvers. But the program was instead checking the zones _already configured in Google Cloud_ against its own nameservers. It now correctly compares the YAML zones/records against the actual DNS resolution.

Secondly, a misplaced variable was causing all records in a zone to be reported as being inconsistent after the first failure. .


BenTheElder:
thank you!

PR #496, opened 2019-12-10
Fix bugs in ensure-namespaces

thockin:
* Fix unbound variable $1 while loop+shift
* Use --cluster (-c is not a thing)
* Fix unbound variables with default values ("${X:-}")
* Change -lt to -ne in arg validation

After these I was able to run it successfully.


dims:
/approve

leaving lgtm for @bartsmykla

thockin:
Would be nice to get this in since I keep running it :)

@cblecker ?

PR #498, opened 2019-12-11
wip: testing tag change

listx:
/hold

This is just a test to confirm that tag changes today are approvable (that is, we are missing a test that should check and prevent such PRs from being merged).


listx:
This test is not a priority at this point.

/close

PR #499, opened 2019-12-11
wip: testing tag swap

listx:
This change tests a scenario where a malicious user has already uploaded
an image, and wants to swap the tags with another image.

/hold


listx:
OK so the check should not have passed. I'll work on a fix to address this.

listx:
https://github.com/kubernetes-sigs/k8s-container-image-promoter/pull/158 should fix this.

listx:
So the `pull-k8sio-cip` check correctly detected a tag move. You can grep for "tag move" in the logs here: https://prow.k8s.io/view/gcs/kubernetes-jenkins/pr-logs/pull/k8s.io/499/pull-k8sio-cip/1215806398466625536.

Unfortunately the error message in the logs says `<nil>`, but that is not blocking this test and can be fixed later. Closing.

PR #511, opened 2019-12-12
Update audit data, fixup script

thockin:
I ran the audit script and realized a) data was very stale and b) it was dumping stuff that wasn't super useful.  I went down the rabbithole.

There's a lot more it could do, but for now it seems strictly better.


thockin:
I tried to make the commits reviewable one-by-one

Issue #504, opened 2019-12-11
Move k8s.io from old Google-only cluster to aaa


thockin:
Other than the existing TLS secrets which will need to be manually copied, the YAML exists.

1) PR to make a group and namespace

2) Load YAML in aaa cluster, get new IP

3) figure out what tests we want to do to convince ourselves it is sane

4) Monitoring (at least up-checks)

5) Flip DNS to new IP

6) Turn down old

ThomasWo:
I don't currently have access to either aaa or google clusters, but I'll do what I can on this one.

@thockin For `1. PR to make a group and namespace`, any idea on what the group name should be/who should be in it?  I have a first pass in #513 (just took a guess at a name and added usernames that I recognized).
/assign

BenTheElder:
noting that 1. / https://github.com/kubernetes/k8s.io/issues/504#issuecomment-565773199 was done in https://github.com/kubernetes/k8s.io/pull/513

2\. seems pretty straightforward, 3/4 may need some thought still. 5.6 will be straightforward...

BenTheElder:
GCP project for old cluster is `kubernetes-site`, I can access that and see the deployments etc.
The new `AAA` cluster is in `kubernetes-public`, which I think I actually can't.

Not sure what we want to name these for the new deployment.

thockin:
Add thineself to k8s-infra-cluster-admins@kubernetes.io in groups.yaml and you will be able to drive all over aaa.

PR #513, opened 2019-12-15
Add k8s.io namespaces and gsuite group

ThomasWo:
Adds:
- k8s.io prod and canary namespace yamls
- k8s-infra-k8s-io-admins group

Part of #504

### DIFF
--- a/k8s.io/namespace-canary.yaml
+++ b/k8s.io/namespace-canary.yaml
@@ -1,4 +1,0 @@
-apiVersion: v1
-kind: Namespace
-metadata:
-    name: k8s-io-canary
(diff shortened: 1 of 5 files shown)


ThomasWo:
@k8s-ci-robot I've signed the CLA

k8s.io/namespace-canary.yaml
```diff
@@ -0,0 +1,4 @@
+apiVersion: v1
```
thockin: As nted - we should not need either of these files.

ThomasWo:
@thockin I've addressed your feedback (moving the namespace changes into `ensure-namespaces.sh`), PTAL

thockin:
Activated

Issue #516, opened 2019-12-16
Refactor infra/gcp/...


thockin:
Right now it is set up as "concept-first".  For example "ensure-staging" says "all of these are staging-like" and "ensure-prod" says "all of these are prod-like".  That makes it hard to get a sense of what any one project has going on.

I propose to refactor it to "project-first". One list of projects and each one says "I am prod like" or "I am staging like".  Then I could simply say `ensure-project k8s-foo-bar` and all of the properties would be asserted.

This is very close to terraform territory, but I don't know TF well enough to make the "utility" functions to not be so duplicated.  @cblecker - is this worth pursuing?

cblecker:
Yes, it is absolutely worth pursuing. The bash ensure stuff is getting out of hand IMO.

thockin:
Shall we discuss the idea sometime and you can either volunteer yourself or
write enough that we can solicit other volunteers?

rikatz:
@thockin just taking a walk into the issues, maybe this is related: https://github.com/kubernetes/k8s.io/pull/523

I also saw the discussion into the mailing list, in my opinion Terraform is the best way to do this :)

I just don't think I have enough knowledge into that to help with the Terraform stuff, but anyway just putting the PR here (again) so we may have a follow up.

spiffxp:
Having dipped my toes into adding to this mess:
- I hated concept-first at first, and thought project-first was a great idea
- I'm now less sure, as some concept span projects (image promotion needs access to all these, prow/boskos will need access to all those)
- I suspect the parts of our shell scripts that are for loops could correspond well to terraform modules
- I just need to say out loud that I am still a little freaked out by terraform.  It's this whole other ecosystem that has churn and will need to be kept current. I wasn't confident enough to try migrating the google provider from 2.x to 3.x for `aaa` without maybe accidentally blowing away the cluster.  OTOH as we write more bash in `lib*.sh` files we're also creating our own ecosystem with possibly inconsistent naming, lack of testing, etc.
- I fell back to using a shell script for creating projects, but when I have time would be willing to see what rewriting that as terraform would be like

spiffxp:
/remove-lifecycle rotten
/lifecycle frozen
/priority important-longterm
I don't have the bandwidth for this, and this issue is maybe too broad to stay open, but I think the point stands that:
- continuing with our bash as-is will deepen our tech debt
- we aren't comfortable enough with our bash to allow automation to run it

Whatever we use, even if it's bash, we need:
[...]

PR #523, opened 2019-12-18
Refactor infra/gcp #516

sgmac:
I think it would make sense to have terraform modules and define deployments per environment,  I work mainly on AWS with Terraform 0.11.X so still trying to catch up with the new syntax, but I would like help with this. WDYT?

### DIFF
--- a/infra/gcp/deployments/staging/resources.tf
+++ b/infra/gcp/deployments/staging/resources.tf
@@ -1,0 +1,29 @@
+module "project-artifact-promoter" {
+  source       = "../tf_modules/project"
+  project_name = "artifcat-promoter"
+  project_id   = var.project_id
+  env_name     = "staging"
+  writer       = "k8s-infra-${env_name}-artifact-promoter@kubernetes.io"
+  enable_api   = "containerregistry.googleapis.com"
+  billing_account =  var.gcp_billing
+}
+
+module "build-image" {
+  source       = "../tf_modules/project"
+  project_name = "build-image"
+  project_id   = var.project_id
+  env_name     = "staging"
+  writer       = "k8s-infra-${env_name}-build-image@kubernetes.io"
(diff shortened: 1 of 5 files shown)


cblecker:
Conceptually: Yes, I'd love to be able to define this as modules. My bandwidth will be limited until the new year to review things, but I'd love to get the conversation started. Perhaps we can get a few Terraform minded folks to help community review these things and set some standards.

Would it be possible to send a note out to the WG e-mail list linking to this PR and lets get some ideas going.

cc: @thockin @dims @bartsmykla @spiffxp

sgmac:
@cblecker I just send the mail to the WG mailing list.

ameukam:
@sgmac Thank you for doing this!
Is it possible to split this PR in smaller PRs ? Ideally a PR per module would be great. This allow people with low bandwidth to review the suggested modules ? 
Happy to give help if needed.

sgmac:
@ameukam  Indeed, it would be much easier to review.  This was more a proof of concept,  but if we agree on moving forward, glad to help with this :)

thockin:
So , my terraform skills are nil, but here's how I was thinking about refactoring the shell.

Today we have scripts per-archetype (prod storage, staging, etc) and a list of projects/buckets etc per script.  This makes it weird to figure out what applies to a given project.

What I think we should do is pivot this to be project-centric.  Define archetypes as "mixins" and start with the project.  "Project X has prod-storage enabled, GKE clusters enabled, is cloud-run enabled, runs the GCR auditor".  "Project Y is a staging project".  "Project Z is a staging project and has the following special cases..."

sgmac:
It makes sense to define a module that manages the different components as a unit, per environment. I'll try to see how to approach this next week, I'm on-call right now, so I have almost no time.

PR #524, opened 2019-12-18
Add node-feature-discovery (nfd) to k8s.gcr.io

marquiz:
We would like to host images for [node-feature-discovery](https://github.com/kubernetes-sigs/node-feature-discovery) in k8s.gcr.io, see kubernetes-sigs/node-feature-discovery#177

Upstream project name (**node-feature-discovery**) exceeds the 18 character limit mentioned in [k8s.gcr.io/README.md](https://github.com/kubernetes/k8s.io/blob/master/k8s.gcr.io/README.md) so it was abbreviated to **nfd** here. I don't know if it would be possible/make sense to use the full name in the name of the registry (i.e. in `ensure-staging-storage.sh` and `k8s-staging-nfd/promoter-manifest.yaml`)(?)

/sig node

### DIFF
--- a/groups/groups.yaml
+++ b/groups/groups.yaml
@@ -684,6 +684,18 @@ groups:
       - srampal@cisco.com
       - tasha.drew@gmail.com
 
+  - email-id: k8s-infra-staging-nfd@kubernetes.io
+    name: k8s-infra-staging-nfd
+    description: |-
+      ACL for staging node-feature-discovery (nfd)
+    settings:
+      ReconcileMembers: "true"
+    members:
+      - alexander.kanevskiy@intel.com
+      - jordanjacobelli04@gmail.com
+      - markus.lehtonen@intel.com
+      - zkosic@redhat.com
+
   - email-id: k8s-infra-staging-npd@kubernetes.io
     name: k8s-infra-staging-npd
     description: |-
--- a/infra/gcp/ensure-staging-storage.sh
+++ b/infra/gcp/ensure-staging-storage.sh
@@ -61,6 +61,7 @@ STAGING_PROJECTS=(
     kubeadm
     metrics-server
     multitenancy
+    nfd
     npd
     provider-azure
     publishing-bot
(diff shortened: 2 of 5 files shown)


ArangoGutierrez:
++

thockin:
@bartsmykla Will cross Ts and dot Is, but this seems OK.  Problem with OWNERS is real - maybe just do Derek or someone else who can approve until people are org members?

thockin:
I'm not in love with "nfd" - is there anything else we can find?

marquiz:
Rebased.

> I'm not in love with "nfd" - is there anything else we can find?

Me neither but I couldn't come up with anything better as *nfd* or *NFD* is a commonly used acronym for the project.

I repeat my question from the PR submission: Would be possible/make sense to use the full name in the name of the registries (i.e. in `ensure-staging-storage.sh` and `k8s-staging-nfd/promoter-manifest.yaml`)?

thockin:
The name of the subproject (nfd) is used to create GCS buckets, GCP projects, etc.  We are bound by GCP's character limits on those, and (for better or worse) our prefix is established.  To change it now would be forever inconsistent :(

The image name(s) that you use can be longer - I don't think we have a strong opinion on that (e.g. gcr.io/k8s-artifacts-prod/nfd/node-feature-detector and eventually k8s.gcr.io/...).  We would not want to change the "directory" part of it though.  That name-mapping is very convenient

thockin:
BTW, we still have invalid OWNERS :(

bartsmykla:
I don't have strong opinion about the acronym, and everything except the OWNERS file look good to me.
@marquiz could you fix the issue with owners?

marquiz:
> The name of the subproject (nfd) is used to create GCS buckets, GCP projects, etc. We are bound by GCP's character limits on those, and (for better or worse) our prefix is established. To change it now would be forever inconsistent :(

@thockin thanks for the explanation


> @marquiz could you fix the issue with owners?

I fixed that by removing all non-members from the owners file. There's now only one owner, do you think we would require to have more?

thockin:
More would be nice, but maybe isn't a requirement at this point.

thockin:
Actuating now

PR #535, opened 2020-01-08
K8s YouTube vanity URL

chris-short:
First off, was trying to rebase #450 but, I failed horribly so my apologies for the delay in getting this done AND for screwing up the previous PR.

Hello! Every week I typically have to figure out one URL or another to get people to the YouTube channel. Creating this redirect would make my life easier. The @kubernetes/youtube-admins team might agree too.

NOTE: This is for the top-level Kubernetes YouTube channel; not any individual recording/stream/etc.

Signed-off-by: Chris Short <chris@chrisshort.net>


cblecker:
/approve

I'm good with this. Assigning to Tim as I think he's the only one with the keys right now to deploy this.

/assign @thockin

stp-ip:
Looks good.
Just needs Tim to push the DNS changes and reload the nginx config.
/approve

dns/zone-configs/kubernetes.io.yaml
```diff
@@ -183,3 +183,6 @@ pfjqlwt67ty3.testgrid:
 velodrome:
   type: CNAME
   value: velodrome.k8s.io.
+youtube:
```
thockin: Why yt.k8s.io and not yt.kubernetes.io ?

thockin: and why not go.k8s.io/youtube ?

chris-short: The answer to both is fewer characters, easier to remember. I can add them if you'd like but, I don't know if they'd be utilized in lieu of the others.

thockin:
wrt k8s/kubernetes - it is simpler overall to have most names work in both,
so we don't question why it is different.

I don't feel strongly about go/ vs a full name for this case

chris-short:
Consistency makes sense. Will fix up and push changes shortly.

chris-short:
Alright, @thockin. I _think_ this covers all the cases.

k8s.io/configmap-nginx.yaml
```diff
@@ -244,6 +244,8 @@ data:
           rewrite ^/test-health$     http://velodrome.k8s.io/dashboard/db/bigquery-metrics redirect;
           rewrite ^/test-history$    https://storage.googleapis.com/kubernetes-test-history/static/index.html redirect;
           rewrite ^/triage$          https://storage.googleapis.com/k8s-gubernator/triage/index.html redirect;
+          rewrite ^/youtube$          https://www.youtube.com/c/kubernetescommunity redirect;
```
thockin: Sorry, I wasn't clear - we don't need a DNS name *and* a go link.  One or the other :)

chris-short:
Oh Tim...

Good?

k8s.io/README.md
```diff
@@ -52,6 +53,8 @@ Redirections
 - https://go.k8s.io/test-health
 - https://go.k8s.io/test-history
 - https://go.k8s.io/triage
+- https://go.k8s.io/youtube
```
thockin: Forgot these?

thockin:
so....close....

:)

k8s.io/README.md
chris-short: dag nabbit

chris-short:
Maybe?

Maybe

chris-short:
@thockin bump

k8s.io/configmap-nginx.yaml
```diff
@@ -287,6 +287,12 @@ data:
 
         location / {
           rewrite ^/.*$  https://prow.k8s.io/tide redirect;
+
+      server {
+        server_name yt.k8s.io youtube.k8s.io youtube.kubernetes.io yt.kubernetes.io;
```
thockin: We need to add these to the certificate, too.  Followup is OK.

stp-ip: pointer: https://github.com/kubernetes/k8s.io/blob/master/k8s.io/certificate-prod.yaml

thockin:
I'll send the followup as I am deploying this change

thockin:
https://github.com/jetstack/cert-manager/issues/2005  @munnerz

PR #536, opened 2020-01-08
Add BigQuery auditing

puerco:
As we discussed during the meeting, this PR changes the following in the GCP audit script:

1. Adds dumps of BigQuery datasets, policies and dataset acls.
2. Reenables the creation of compute engine information files for diff'ing (usage, isntances, and disks) since auditors seem to have enough permissions now.
3. Fixes a bug where projects where being referenced to gcloud by name instead of projectid
4. Adds a new capability of generating audit files only for one project

@thockin whenever you have time, could you confirm if this is the way you intended to continue building the audit files? If so, I'll add handlers for more services.

### DIFF
--- a/audit/audit-gcp.sh
+++ b/audit/audit-gcp.sh
@@ -98,6 +98,24 @@ gcloud \
     | sed 's/.googleapis.com//' \
     | while read -r SVC; do
         case "${SVC}" in
+            bigquery)
+                mkdir -p "projects/${PROJECT}/services/${SVC}"
+                bq \
+                    --format=prettyjson --project_id=$PROJECT ls
+                    > "projects/${PROJECT}/services/${SVC}/bigquery.datasets.json"                
+                # Only run if there are any datasets
+                if [ -s "projects/${PROJECT}/services/${SVC}/bigquery.datasets.json" ]
+                then
+                    bq \
+                        --project_id="{$PROJECT}" --format=json ls \
+                        | jq -r '.[] | .datasetReference["datasetId"]' \
+                        | while read -r DATASET; do                        
+                            bq \
+                                --project_id="${PROJECT}" --format=json show "${PROJECT}:${DATASET}" \
+                            | jq .access > "projects/${PROJECT}/services/${SVC}/bigquery.datasets.${DATASET}.access.json"
+                        done
+                fi
+                ;;
             compute)
                 mkdir -p "projects/${PROJECT}/services/${SVC}"
                 gcloud \


thockin:
Conflict, sorry.

puerco:
@thockin  I'll review your changes in PR #511 and delete what this is duplicating. If there is anything still worth saving I'll rebase and send a new commit emulating your format. Otherwise I'll just close it.

audit/audit-gcp.sh
```diff
@@ -2,20 +2,32 @@
 # set -x -e
 CNCF_GCP_ORG=758905017065
 
+SOLEPROJECT=""
+if [ $# -ge 1 ]
+then
+   SOLEPROJECT="^$1\s"
```
thockin: what's going on here?

audit/audit-gcp.sh
```diff
@@ -2,20 +2,32 @@
 # set -x -e
 CNCF_GCP_ORG=758905017065
 
+SOLEPROJECT=""
```
thockin: comment on this block

audit/audit-gcp.sh
```diff
@@ -2,20 +2,32 @@
...
 
+# Do we need other formats? bq does not support other formats very well
 format=json
-echo "# Auditing CNCF CGP Org: ${CNCF_GCP_ORG} #"
-gcloud iam roles list --organization=$CNCF_GCP_ORG --format=$format \
-       > cncf-org.roles.$format
-gcloud organizations get-iam-policy $CNCF_GCP_ORG --format=$format \
-       > cncf-org.policy.$format
-echo "## Iterating over Projects ##"
+if [ ! -z ""]
```
thockin: ?

audit/audit-gcp.sh
```diff
@@ -2,20 +2,32 @@
...
+    gcloud organizations get-iam-policy $CNCF_GCP_ORG --format=$format \
+        > cncf-org.policy.$format
+    echo "## Iterating over Projects ##"
+else
+    echo "## Generating audit files only for $1 ##"
+fi
 gcloud projects list \
        --filter "parent.id=$CNCF_GCP_ORG" \
-       --format "value(name, projectNumber)" \
+       --format "value(project_id, projectNumber)" | grep "$SOLEPROJECT" \
```
thockin: why change this?  Not saying it is wrong (I think we always set them the same in fact)

audit/audit-gcp.sh
```diff
@@ -35,17 +47,18 @@ gcloud projects list \
...
+        echo "##### Processing: $service #####"
         case $service in
             compute)
-                echo TODO: $service Needs compute.projects.get
-                #### gcloud compute project-info describe
-                #### gcloud compute instances list --format=$format > $PROJECT/services/compute.instances.$format
-                #### gcloud compute disks list --format=$format > $PROJECT/services/compute.disks.$format
+                #echo TODO: $service Needs compute.projects.get
+                # Only record usages so that quota increases/decreases do not alter the audit
+                gcloud compute project-info describe --format="json(quotas)" | jq ".quotas[] |= del(.limit)" > $PROJECT/services/compute.resources.json
```
thockin: `compute.quotas.json`

audit/audit-gcp.sh
```diff
@@ -35,17 +47,18 @@ gcloud projects list \
...
+                gcloud compute instances list --format=$format > $PROJECT/services/compute.instances.$format
```
thockin: I don't think we should do this - it will change, for example as clusters are upgraded

audit/audit-gcp.sh
```diff
@@ -35,17 +47,18 @@ gcloud projects list \
...
+                gcloud compute disks list --format=$format > $PROJECT/services/compute.disks.$format
```
thockin: same

bartsmykla:
Hi @puerco! Great job! :-) Could you maybe address some of @thockin comments?

PS. Can you also do a rebase?

thockin:
ping on this?  I mad some surrounding changes, so you'll have to rebase - sorry

puerco:
@bartsmykla @thockin  I'm looking into it now, it got run over pretty badly. Let me see if I can still salvage parts of it

puerco:
OK, most of my changes were superseded by Tim's so I only kept the Big Query audit part of the script.

audit/audit-gcp.sh
puerco: I removed the service quota audit because the quotas are already contained inside `compute/project-info.json`

audit/audit-gcp.sh
puerco: Since auditing a project is fairly expensive and time consuming, I tried to add an option to audit just one project at a time. Removed.

audit/audit-gcp.sh
puerco: Makes sense, any autoscaling would also trigger a difference in the audit files. Removed.

audit/audit-gcp.sh
puerco: Removed as well.

audit/audit-gcp.sh
```diff
@@ -98,6 +98,29 @@ gcloud \
...
+                mkdir -p "projects/${PROJECT}/services/${SVC}"
+                rm -f projects/${PROJECT}/services/${SVC}/bigquery.datasets.* || :
+                bq \
+                    --format=json --project_id=$PROJECT ls
+                    > "projects/${PROJECT}/services/${SVC}/bigquery.datasets.json"
+                
+                # Only run if there are any datasets
+                if [ -s "projects/${PROJECT}/services/${SVC}/bigquery.datasets.json" ]
+                then
+                    for DATASET in `bq --project_id=$PROJECT --format=json ls  | jq -r '.[] | .datasetReference["datasetId"]'`
```
bartsmykla: I'm not an expert in Bash, but few years ago I read using the backticks is being discouraged, and I think this stackoverflow has all the arguments and links about why which would be muhc better than me trying to write some reasons: https://unix.stackexchange.com/a/126928
but maybe there is a reason why you used them in the first place, so it's just a comment and maybe invitation to discussion about it. Wdyt?

bartsmykla:
@puerco thank you for your work! It's much appreciated. Would you be so kind and change the topic of the PR to just `Add BigQuery auditing` maybe?

audit/audit-gcp.sh
puerco: Nice observation Bart. I had not heard of this before but I will of course take that into account. I'll refactor that code to get rid of them.

bartsmykla: Thank you! :-)

puerco:
Done @bartsmykla removed the backticks and fixed the title 
:bash_fire:

bartsmykla:
/lgtm
/ok-to-test

@puerco thank you! It looks good to me but for final approval let's wait for @thockin, who is expert in bash :-)

audit/audit-gcp.sh
```diff
@@ -98,6 +98,31 @@ gcloud \
     | sed 's/.googleapis.com//' \
     | while read -r SVC; do
         case "${SVC}" in
+            bigquery)
+                mkdir -p "projects/${PROJECT}/services/${SVC}"
+                rm -f projects/${PROJECT}/services/${SVC}/bigquery.datasets.* || :
```
thockin: Do you need this?  There's an "rm -rf" at the top of this script

audit/audit-gcp.sh
```diff
@@ -98,6 +98,31 @@ gcloud \
...
+                bq \
+                    --format=json --project_id=$PROJECT ls
+                    > "projects/${PROJECT}/services/${SVC}/bigquery.datasets.json"
```
thockin: `| jq` just for pretty formatting?  I notice BQ JSON is not indented..

audit/audit-gcp.sh
```diff
@@ -98,6 +98,31 @@ gcloud \
...
+                # Only run if there are any datasets
+                if [ -s "projects/${PROJECT}/services/${SVC}/bigquery.datasets.json" ]
+                then
+                    bq \
+                        --project_id="{$PROJECT}" --format=json ls \
+                        | jq -r '.[] | .datasetReference["datasetId"]' \
+                        | while read -r DATASET; do                        
+                            bq \
+                                --project_id="${PROJECT}" --format=json show "${PROJECT}:${DATASET}" \
+                            | jq .access > "projects/${PROJECT}/services/${SVC}/bigquery.datasets.access.json"
```
thockin: We should make one file per dataset

audit/audit-gcp.sh
```diff
@@ -98,6 +98,31 @@ gcloud \
...
+                        done
+                fi
+                gcloud \
```
thockin: This is cute.  We don't do it elsewhere and it is duplicative of the overall IAM dumps above - do you think it's notably better?

audit/audit-gcp.sh
puerco: Done, removed

audit/audit-gcp.sh
puerco: well there is --prettyjson

audit/audit-gcp.sh
puerco: This was a leftover from the previous version , before I had to rebase. I though that every service would store its iam data. Removed.

audit/audit-gcp.sh
puerco: Done

puerco:
Sorry for taking so long to respond to this, I hope all issues are resolved now.

PR #542, opened 2020-01-22
DNS TTL -> 3600

thockin:
Other than prow, any other special cases?


BenTheElder:
are we waiting for input on possible further snowflakes, or on actuation?

stp-ip:
Fine with the change, but would love to hear the actual insight from the DNS people Tim looped in before doing preventive measures that might not address the actual problem.

thockin:
I will actuate

thockin:
Pushed.  I had to add `--force` in push.sh because of the number of changes, which is by intention.

BenTheElder:
ACK

PR #545, opened 2020-01-24
Fix k8s.io changes for youtube.

thockin:
Add youtube names to SSL cert.

Fix broken nginx config.

Fix teh canary test to use the right python version.

Make the canary test better so failures are more obvious when nginx is broken.

### DIFF
--- a/k8s.io/Makefile
+++ b/k8s.io/Makefile
@@ -2,7 +2,7 @@ all: test
 
 .PHONY: test
 test:
-	python test.py -q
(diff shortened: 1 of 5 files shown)


dims:
/approve
/lgtm
/hold

please remove hold when appropriate.

thockin:
I've already activated this (though the cert is having trouble), so I'll remove hold now.

thockin:
Cert issue seems resolved

PR #548, opened 2020-01-24
Fix DNS lame delegation of canary.canary

thockin:
To fix this, we need to not delegate canary.canary, which means loading
SLIGHTLY different yaml in canary than prod.

This change makes the push script "pre-cook" config files by joining
them together.  These are saved in a temporary dir and octodns reads
from there.  This works because records are top-level items in the YAML
schema.

I could not find a way to make octodns do this itself.

Dry-run shows 2 deletions, as expected.

### DIFF
--- a/dns/push.sh
+++ b/dns/push.sh
@@ -15,6 +15,20 @@
 # limitations under the License.
 # This runs as you.  It assumes you have built an image named ${USER}/octodns.
 
+PROD_ZONES=(
+    k8s.io.
+    kubernetes.io.
+    x-k8s.io.
+    k8s-e2e.com.
+)
+
+CANARY_ZONES=("${PROD_ZONES[@]/#/canary.}")
+
+ALL_ZONES=(
+    "${CANARY_ZONES[@]}"
+    "${PROD_ZONES[@]}"
+)
+
 # Pushes config to zones.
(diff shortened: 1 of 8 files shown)


dns/push.sh
```diff
@@ -15,14 +15,31 @@
 # limitations under the License.
 # This runs as you.  It assumes you have built an image named ${USER}/octodns.
 
+CANARY_ZONES=(
```
BenTheElder: should we auto derive these from PROD_ZONES? they are all `"canary.${prod_zone}"` right?

BenTheElder:
btw: if you put users into reviewers when filing the PR prow won't assign more :upside_down_face:

dns/push.sh
thockin: Done

dns/push.sh
```diff
@@ -38,17 +52,33 @@ if [ ! -f octodns-config.yaml -o ! -d zone-configs ]; then
     exit 1
 fi
 
+# Where to hold cooked configs.
+TMPCFG=$(mktemp -d /tmp/octodns.XXXXXX)
+
+# Pre-cook our configs into $TMPCFG
```
BenTheElder: might be worth mentioning what "cook" is -- concatenate?

thockin:
Better words :)

thockin:
pushing now

PR #550, opened 2020-01-27
Promote k/k test e2e images

claudiubelu:
Adds the images.yaml file containing the E2E test images to be promoted, their digests, and versions.

Adds the OWNERS file, based on the kubernetes/test/images/OWNERS file.

Moves the manifest.yaml file to its appropriate location.

### DIFF
--- a/k8s.gcr.io/images/k8s-staging-e2e-test-images/images.yaml
+++ b/k8s.gcr.io/images/k8s-staging-e2e-test-images/images.yaml
@@ -1,0 +1,15 @@
+- name: agnhost
+  dmap:
+    "sha256:3a09e7cc09aab76ac52f0447c16d6a6ec3e0537aa405301b165f5823aa6047e0": ["2.9"]
+- name: jessie-dnsutils
+  dmap:
+    "sha256:31183fd28aa9b4af81329e1886060492cfa67dd94e9d1132e00ed032ce685187": ["1.2"]
+- name: nonewprivs
+  dmap:
+    "sha256:553fcda2222ccea7cd534c8da34b709feb768d6389eb7333444b13f296bd168f": ["1.1"]
+- name: resource-consumer
+  dmap:
+    "sha256:74e800780d7656d42fb0f28e153619044a826e5f532d1871cac4b7c4d66d946f": ["1.6"]
+- name: sample-apiserver
+  dmap:
+    "sha256:eb820c8133a6bbe63a97268713fb85bfc9129e380e8c06f99a34ba0051665c92": ["1.17.2"]
--- a/k8s.gcr.io/images/k8s-staging-e2e-test-images/OWNERS
+++ b/k8s.gcr.io/images/k8s-staging-e2e-test-images/OWNERS
@@ -1,0 +1,5 @@
+# See the OWNERS docs at https://go.k8s.io/owners
+
+approvers:
+- listx
+- mkumatag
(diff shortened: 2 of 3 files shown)


k8s.gcr.io/images/k8s-staging-e2e-test-images/images.yaml
```diff
@@ -0,0 +1,21 @@
+- name: agnhost
```
mkumatag: Why only these many images? where are the rest of the images?
Did we run e2e with the staging images? can we just run the tests modifying the repo to staging repo to see if they really works :)

claudiubelu: Plenty of questions to unpack here. :)

1. Currently, the Image Promoter was set to work with the Conformance Images. I've left some comments here: https://github.com/kubernetes/kubernetes/blob/30058c9d356c5c10c6273b58a41057d9ab252815/test/images/image-util.sh#L164 . Basically, for the 1st implementation, we decided to only go with the Conformance Images, since those are most commonly used in tests and those are the images that are commonly updated, and some non-conformance images take an extremely long time to build.
2. The rest of the images we could probably add in the future, maybe, if there's a need for it, but it's a lower priority item (see above).
3. We have not yet. I could send a PR to check that, sounds like a good idea. :)

mkumatag: 👍

claudiubelu: It seems that it was a good idea to actually check those images. It seems that some issue slipped through, especially due to the lack of logging (a ``-x`` option would help with that). Basically, the job builds ``all-conformance`` images, but incidentally, it just publishes the ``agnhost image`` under different names. Increasing the verbosity, I could see why:

    ++ docker manifest create --amend claudiubelu/jessie-dnsutils:1.2 claudiubelu/agnhost-amd64:2.9 claudiubelu/agnhost-arm:2.9 claudiubelu/agnhost-arm64:2.9 claudiubelu/agnhost-ppc64le:2.9 claudiubelu/agnhost-s390x:2.9 claudiubelu/jessie-dnsutils-amd64:1.2 claudiubelu/jessie-dnsutils-arm:1.2 claudiubelu/jessie-dnsutils-arm64:1.2 claudiubelu/jessie-dnsutils-ppc64le:1.2 claudiubelu/jessie-dnsutils-s390x:1.2
    Created manifest list docker.io/claudiubelu/jessie-dnsutils:1.2

Seems that the ``manifest`` list that's being used in the ``image-util.sh`` is stateful, and the images just pile on in that list. That list needs to be reset. Will send a PR soon, but the fix is simple.

claudiubelu: Sent PR: https://github.com/kubernetes/kubernetes/pull/87647

claudiubelu: This PR tests the staging images: https://github.com/kubernetes/kubernetes/pull/87607

It seems there was an issue with the kitten and nautilus images. Because the agnhost version was bumped a few times before the Image Centralization part 4 merged, the kitten and nautilus base versions were never updated, and thus, didn't have the test-webserver functionality they needed. This PR fixes that: https://github.com/kubernetes/kubernetes/pull/87722

I think after we merge that and update the digests here, we can merge this PR.

listx:
@claudiubelu Can you change the promoter manifest to match the format of the existing subprojects? There must be 2 files, one called `images.yaml` and another called `promoter-manifest.yaml`. For example, the promoter subproject has:

- https://github.com/kubernetes/k8s.io/blob/master/k8s.gcr.io/images/k8s-staging-artifact-promoter/images.yaml
- https://github.com/kubernetes/k8s.io/blob/master/k8s.gcr.io/manifests/k8s-staging-artifact-promoter/promoter-manifest.yaml

These files must be under their correct `images` and `manifests` subdirectories under `k8s.gcr.io`.

claudiubelu:
^ Done

Issue #554, opened 2020-01-28
Wrong spf registry into kubernetes.io domain


rikatz:
As stated by @tallclair at mailing list:

```
We've been getting a lot of reports through the Kubernetes bug bounty that kubernetes.io and k8s.io are missing valid SPF records. It looks like there may be a typo? "~all" should be "-all". Does anyone know if this is correct, and if not, how to fix it?

Here's the generic report:

SPF record lookup and validation for: kubernetes.io
SPF records are published in DNS as TXT records.

The TXT records found for your domain are:
v=spf1 include:_spf.google.com ~all
google-site-verification=oPORCoq9XU6CmaR7G_bV00CLmEz-wLGOL7SXpeEuTt8

Checking to see if there is a valid SPF record.

Found v=spf1 record for kubernetes.io:
v=spf1 include:_spf.google.com ~all

evaluating...
SPF record passed validation test with pySPF (Python SPF library)!

Use the back button on your browser to return to the SPF checking tool without clearing the form.

Remediation :

Replace ~all with -all to prevent fake email.
```

So this issue is to replace the ~all to -all into all kubernetes domains.

puerco:
No, they are wrong. GSuite domains are supposed to have the ~ in the SPF. In the real world I think it would be better to have the `-all`. But, strictly speaking, [GSuite docs specify the SPF](https://support.google.com/a/answer/33786?hl=en) with ~:
```
 v=spf1 include:_spf.google.com ~all
```

ixdy:
FYI `~` means soft fail. `-` is hard fail.

rikatz:
IMO in this case hard is better than soft. ~all will let the message pass and rely on the server / user MTA side, -all will tell him to drop. Some might say ~all can be used only in servers transitioning to SPF.

As we're very aware of which are the authorized domains from kubernetes.io domain I think the best approach is to hard fail the SPF

cblecker:
I'm inclined to go along with the current gsuite recommendations as @puerco noted.

rikatz:
Some thoughts:

* Google and GSuite rely on ~all instead of -all trusting the MTA on the other side (receiving a fake kubernetes.io email) will have a threshold good enough to drop the message. 

An example:
- In case a user from Gmail or other good mail provider receives a mail from a non-authorized kubernetes.io IP it have the hability to receive and mark this as SPAM or phishing (softfail), so Google deals nicelly with soft fails

[...]

PR #555, opened 2020-01-28
Changes SPF validation from soft to hard fail

rikatz:
Fixes: #554 

This PR changes the SPF validation in domains from soft to hard fail (~all to -all) enforcing the validation to came only from the desired TXT record from _spf.google.com ("v=spf1 include:_netblocks.google.com include:_netblocks2.google.com include:_netblocks3.google.com ~all") but overriding the soft fail in the end of this record.


cblecker:
@rikatz Thanks for jumping on this! We're going to hold off on this change for now, as described here: https://github.com/kubernetes/k8s.io/issues/554#issuecomment-581559349

PR #562, opened 2020-02-07
Promote k/k e2e test image: agnhost:2.10

claudiubelu:


claudiubelu:
/cc @listx 

The Image Promoter is back up. The promotion process seems to have worked, we can proceed to promote the 2.10 image as well.

PR #565, opened 2020-02-11
releng: Add Daniel Mangum (hasheddan) as a Release Manager Associate

justaugustus:
- Add Daniel Mangum (hasheddan) as a Release Manager Associate (ref: https://github.com/kubernetes/sig-release/issues/978)
- Temporarily grant hasheddan access to cut v1.18.0-alpha.4 (ref: https://github.com/kubernetes/sig-release/issues/985)

/assign @dims @cblecker
cc: @hasheddan @kubernetes/release-engineering

PR #566, opened 2020-02-11
releng: Revert temporary Branch Manager access for hasheddan

justaugustus:
Reverts elevated privileges that will be granted in https://github.com/kubernetes/k8s.io/pull/565.
/hold until https://github.com/kubernetes/sig-release/issues/985 is closed and this is rebased

cc: @hasheddan @kubernetes/release-engineering


justaugustus:
I needed to do this release personally, due to the issues described in https://github.com/kubernetes/release/issues/1080.
Let's try again with this later in the cycle.

/hold cancel

PR #568, opened 2020-02-12
Bump network proxy to v0.0.6

Jefftree:
/cc @caesarxuchao 
/assign @cheftako


Jefftree:
/hold
There seems to be something off with the docker build script after the client/proxy module split.

Sh4d1:
So the push images fails is due to https://github.com/kubernetes-sigs/apiserver-network-proxy/issues/72

Jefftree:
@cheftako We fixed the image push pipeline and I changed this PR to bump to v0.0.6. PTAL

/unhold

PR #572, opened 2020-02-13
Add thockin to OWNERS

cblecker:
This is obvious, and so Tim doesn't have to keep adding the "approved" label like a naughty goose.

cc: @dims @bartsmykla @nikhita @spiffxp


thockin:
I wondered why that was...

/lgtm
/approve

Issue #444, opened 2019-10-30
give others access to the promoter test projects (for backups, auditing work)


listx:
See https://github.com/kubernetes-sigs/k8s-container-image-promoter/wiki/Container-Image-Promoter:-The-State-of-Things--Today for a description of the backups and auditing work that needs to be completed.

We need a way to give people access to GCP resources to be able to test things.

I don't fully understand what needs to be done other than editing https://github.com/kubernetes/k8s.io/tree/master/groups. We should start from there.

listx:
At a minimum the people working on the promoter should have `roles/viewer` access for the projects used for backups/auditing.

listx:
The test projects are defined here: https://github.com/kubernetes/k8s.io/blob/6a8b73dde2fe08c2be9f5f512254fb5158a00633/infra/gcp/ensure-prod-storage.sh#L66-L75

rajibmitra:
are we going to create any service account for this ?  @listx

thockin:
@listx is this done now?

listx:
> @listx is this done now?

Yes, because the already-existing group `k8s-infra-artifact-admins` got access to administrate the auditor in https://github.com/kubernetes/k8s.io/pull/571 (this group already had storage admin access to `k8s-artifacts-prod-bak`).

listx:
Oops, sorry, misread this issue. Currently the `k8s-infra-artifact-admins` group has access to administrate both the production instance of the auditor as well as to view/administer the backups in `k8s-artifacts-prod-bak`. However this issue is for the *test* projects. I can create a PR for this now that I understand how groups work in the `infra/gcp` folder of this repo.

listx:
It appears that the existing `ensure-prod-storage.sh` script already empowers everyone in the `k8s-infra-staging-cip-test@kubernetes.io` group to at least Project Viewer access. So it's just a matter of expanding this group.

To start, @rajibmitra can you add yourself to https://github.com/kubernetes/k8s.io/blob/26cd0c9a4f21f2bfe7a60a8582adafe02243116b/groups/groups.yaml#L509-L513 ?

rajibmitra:
Done, thanks @listx  
Here is the PR #576  , let me know if I need to anything else.

listx:
> Done, thanks @listx
> Here is the PR #576 , let me know if I need to do anything else.

Can you now try to see if you can see e.g. the `k8s-staging-cip-test` GCP project on the web console?

[...]

PR #575, opened 2020-02-14
ensure-env-cip-auditor.sh: fail early if auditor service is not found

listx:


listx:
This PR has been expanded to also fix https://github.com/kubernetes/k8s.io/issues/444 in part. PTAL

PR #587, opened 2020-02-20
Add munnerz to k8s-infra-team-private

munnerz:
Requesting to be added to this list in order to receive Let's Encrypt account & expiry related emails.

Right now the configured address in our ClusterIssuer resources is `k8s-infra-team-private+letsencrypt`.

If there's sensitive conversation going on there, can we consider changing the email address used to something that can be shared more widely/with me? 😅

### DIFF
--- a/groups/groups.yaml
+++ b/groups/groups.yaml
@@ -849,6 +849,7 @@ groups:
       - spiffxp@google.com
       - thockin@google.com
       - bartek@smykla.com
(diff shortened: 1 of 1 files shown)


cblecker:
/lgtm
/approve
/hold

/assign @bartsmykla @dims @thockin 
Looking for a thumbs up from someone else

PR #600, opened 2020-02-25
Another pass at audit cleanup

thockin:
This cleans up the script and the output, and audits a few more things
like service-account IAM.


audit/audit-gcp.sh
```diff
@@ -6,96 +6,151 @@ set -o pipefail
...
-
 echo "# Auditing CNCF CGP Org: ${CNCF_GCP_ORG} #"
 mkdir -p org_kubernetes.io/roles
-for ROLE_PATH in $(gcloud iam roles list --organization="${CNCF_GCP_ORG}" --format="value(name)")
-do
+gcloud \
+    iam roles list \
+    --organization="${CNCF_GCP_ORG}" \
+    --format="value(name)" \
+| while read ROLE_PATH; do
```
BenTheElder: read without `-r` mangles `\` fwiw, you almost always want `-r`

BenTheElder: same for the other `read`s

audit/k8s-artifacts-prod/buckets/k8s-artifacts-cni/bucketpolicyonly.json
```diff
@@ -0,0 +1,4 @@
+Bucket Policy Only setting for gs://k8s-artifacts-cni:
```
justinsb: Is this json?

audit/k8s-cip-test-prod/services/enabled.json
```diff
@@ -114,6 +117,122 @@
     "parent": "projects/693665670941",
     "state": "ENABLED"
   },
+  {
+    "config": {
+      "authentication": {},
+      "documentation": {
+        "summary": "A data platform for customers to create, manage, share and query data."
```
justinsb: For services in particular, but also in general, I wonder if we should filter this through some sort of simplication filters?  services: I think we only care about the list of enabled service names.  In other places, we likely can filter out etags etc?

audit/audit-gcp.sh
```diff
@@ -6,96 +6,151 @@ set -o pipefail
...
-gcloud projects list \
-       --filter "parent.id=${CNCF_GCP_ORG}" \
-       --format "value(name, projectNumber)" \
-    | sort \
-    | while read PROJECT NUM; do
+gcloud \
+    projects list \
+    --filter="parent.id=${CNCF_GCP_ORG}" \
+    --format="value(name, projectNumber)" \
+| sort \
```
justinsb: FYI, sort is locale sensitive, so can behaves differently in cloudbuilder.  We likely want `export LC_ALL=C` if the order matters (not that it looks like you actually need a sort here)?

audit/k8s-artifacts-prod/buckets/k8s-artifacts-cni/bucketpolicyonly.json
thockin: fixed :)

audit/audit-gcp.sh
thockin: I will remove 1 sort, the other is just for human-output as it iterates

thockin:
Github is borked - I pushed an update but this PR isn't updating

thockin:
Recreated as https://github.com/kubernetes/k8s.io/pull/602

PR #602, opened 2020-02-25
Another pass at audit cleanup

thockin:
This cleans up the script and the output, and audits a few more things
like service-account IAM.


audit/audit-gcp.sh
```diff
@@ -6,96 +6,150 @@ set -o pipefail
...
             container)
-                echo "${service}"
+                mkdir -p "${PROJECT}/services/${SVC}"
                 # Don't do a JSON dump here - too much changes without human
                 # action.
-                mkdir -p "${PROJECT}/services/container"
-                gcloud container clusters list > "${PROJECT}/services/container/clusters.txt"
+                gcloud \
+                    container clusters list \
+                    --format="value(name, location, locations, currentNodeCount, status)" \
```
justinsb: Do we want `currentNodeCount` here?

audit/k8s-gcr-audit-test-prod/iam.json
```diff
@@ -8,16 +14,79 @@
...
+        "serviceAccount:375340694213@cloudservices.gserviceaccount.com",
         "serviceAccount:service-375340694213@containerregistry.iam.gserviceaccount.com"
       ],
       "role": "roles/editor"
     },
     {
       "members": [
+        "serviceAccount:k8s-infra-gcr-promoter@k8s-gcr-audit-test-prod.iam.gserviceaccount.com"
+      ],
+      "role": "roles/errorreporting.admin"
```
justinsb: Well this is unexpected ("admin"?) ... but I guess we can review the actual permissions separately!

justinsb:
One tiny nit, and then I looked through the output and it looked fine - I presume we're not trying to review the security implications here!  I think we should prune the enabledservices for our own sanity.   But nothing we can't iterate on.

/lgtm

audit/k8s-gcr-audit-test-prod/iam.json
thockin: This is in the test-prod project, which is covered by a special-case.

audit/audit-gcp.sh
thockin: I added it on purpose, but we may want to change it if it becomes a problem.

thockin:
We should do an IAM audit as a group and then fix the scripts to reflect anomalies.  We know, for example, that whoever creates a project is listed as owner.  We should be able to automatically remove that.

justinsb:
BTW I think the syntax using jq to remove blocks is:

```
gcloud services list --format=json | jq 'del(.[].config,.[].dependencyConfig)'
```

justinsb:
Oh.... we should `rm -rf` the output directory first, so we don't keep old files.

thockin:
rm is a good point - will retool that a bit.

thockin:
This would make a nice iteration, though maybe we want whitelist rather
than blacklist.

Both services and clusters could use this treatment

justinsb:
Don't have a strong feeling about excluding fields vs including fields.  I thought excluding known-volatile fields was perhaps the safer approach in the face of an evolving schema, though it does present a higher chance of false-positives.

audit/audit-gcp.sh
justinsb: Agree with merging - easy enough to remove it should we choose to, as you say.

thockin:
Added a cleanup step.

WRT `gcloud services list` - I am not sure we actually need anything but name.  The rest seems to all be machinery that we never touch...

thockin:
ping for re-lgtm

PR #619, opened 2020-03-02
Fix vanity URLs & redirections

olivierlemasle:
- Changelog URL has changed since https://github.com/kubernetes/kubernetes/pull/87879
- https://kubernetes.io/docs/setup/pick-right-solution/ does not exist anymore
- Gubernator has a new .k8s.io URL


thockin:
I can activate this

PR #620, opened 2020-03-02
Fix tests after #619

thockin:
I need to remember that this test is overly duplicative. :(


BenTheElder:
I'd almost forgotten about this thing :-)

PR #621, opened 2020-03-02
Set a longer grace period for k8s.io nginx

thockin:


BenTheElder:
xref: https://github.com/nginxinc/docker-nginx/issues/377

thockin:
I'm not sure even that is what we want - SIGQUIT will finish open
connections, but we want it to be willing to accept new connections while
things like LBs are programmed asynchronously

BenTheElder:
I think this change on the surface looks reasonable and I certainly trust your judgement with kubernetes networking services :upside_down_face: but I'm not personally confident enough to say what we want here.

I feel like what we _really_ want is for the pod grace period to start AFTER the LB is finished programming but ;-)
/lgtm

PR #642, opened 2020-03-07
re-add some deleted image references

listx:
This effectively reverts the following commits:

    7958a485433a2cf54f1b618b0d041f8708ca8a29
    fcc4b7af0ac9f683dc36690196c51687d928d693
    ca4467dd0e9f16e2950980ac6b5a9248aa9b31e6
    ba005de40c78faf32cd7a65fd8d588b902d51483

This is to make the promoter manifests more accurate, to account for
what has already been promoted into production (otherwise, when auditing
the production GCR contents, one is forced to dig into the commit
history).

Deletions of items in the promoter manifests shouldn't normally be
allowed because they make the manifests ignore the reality of the images
actually still existing in production.


thockin:
We'll have to socialize the idea that you don't "clean up" in these - whatever gets committed is forever.

/lgtm
/approve

PR #641, opened 2020-03-07
releng: Enable KMS API for all release staging projects and GCS write to test prod

justaugustus:
(Follow-up to https://github.com/kubernetes/k8s.io/pull/624)

- Enable KMS API for all staging projects

  Even though staging projects may not contain their own keys, their
  service accounts could still require the KMS API to be enabled to
  allow decryption using the keys on other GCP projects.

  One such case is the staging release projects requiring access to
  decrypt keys on the `k8s-releng-prod` GCP project.

  (This was originally special-cased for just releng projects, but expanded to 
  every staging project on @thockin's [suggestion in Slack](https://kubernetes.slack.com/archives/CCK68P2Q2/p1583771092152500).)

- Grant `k8s-staging-kubernetes` GCB account GCS write to test prod
  
  gcbmgr+anago kick off a Cloud Build job on the staging project and then
  writes staging assets to the production GCS bucket
  (currently `gs://kubernetes-release`).

  To mirror this behavior in testing, the `k8s-staging-kubernetes` Cloud
  Build service account must have access to the primary GCS bucket on test
  prod (`k8s-release-test-prod`).

Signed-off-by: Stephen Augustus <saugustus@vmware.com>

/hold
ref: https://github.com/kubernetes/release/pull/957#issuecomment-596020336, https://github.com/kubernetes/release/pull/1163

PR #650, opened 2020-03-10
Require gs:// prefix

thockin:


justaugustus:
ref: https://github.com/kubernetes/k8s.io/pull/641

PR #655, opened 2020-03-10
empower KSA used in the promoter's prow jobs to use Workload Identity

listx:
This gives the "k8s-prow-builds.svc.id.goog[test-pods/k8s-artifacts-prod]"
service account powers to act as the
"k8s-infra-gcr-promoter@k8s-artifacts-prod.iam.gserviceaccount.com" for
purposes of Workload Identity. This way, the Prow jobs for the promoter (which live in the `k8s-prow-builds` GKE project, and run in the `test-pods` namespace with the `k8s-artifacts-prod` KSA)
won't have to have service account keys injected into them.

/cc @thockin @fejta 

More context: https://github.com/kubernetes-sigs/k8s-container-image-promoter/issues/180#issuecomment-597334257

### DIFF
--- a/infra/gcp/ensure-prod-storage.sh
+++ b/infra/gcp/ensure-prod-storage.sh
@@ -221,8 +221,10 @@ empower_artifact_auditor "${PROD_PROJECT}"
 empower_artifact_auditor_invoker "${PROD_PROJECT}"
 
 # Special case: empower Prow account to have access to GCR for Workload Identity.
-empower_prowacct_for_workload_identity \
-    "k8s-prow-builds.svc.id.googtest-pods/k8s-artifacts-prod" \
+empower_ksa_to_svcacct \
+    "k8s-prow-builds" \
+    "test-pods" \
+    "${PROD_PROJECT}" \
     $(svc_acct_email "${PROD_PROJECT}" "${PROMOTER_SVCACCT}")
 
 color 6 "Done"
--- a/infra/gcp/ensure-prod-storage.sh
+++ b/infra/gcp/ensure-prod-storage.sh
@@ -223,8 +223,7 @@ empower_artifact_auditor_invoker "${PROD_PROJECT}"
 # Special case: empower Kubernetes service account to authenticate as a GCP
 # service account.
 empower_ksa_to_svcacct \
-    "k8s-prow-builds" \
-    "test-pods" \
+    "k8s-prow-builds.svc.id.goog[test-pods/k8s-artifacts-prod]" \
     "${PROD_PROJECT}" \
     $(svc_acct_email "${PROD_PROJECT}" "${PROMOTER_SVCACCT}")
 
--- a/infra/gcp/ensure-prod-storage.sh
+++ b/infra/gcp/ensure-prod-storage.sh
@@ -220,7 +220,8 @@ empower_group_to_admin_artifact_auditor \
 empower_artifact_auditor "${PROD_PROJECT}"
 empower_artifact_auditor_invoker "${PROD_PROJECT}"
 
-# Special case: empower Prow account to have access to GCR for Workload Identity.
+# Special case: empower Kubernetes service account to authenticate as a GCP
+# service account.
 empower_ksa_to_svcacct \
     "k8s-prow-builds" \
     "test-pods" \
--- a/infra/gcp/ensure-prod-storage.sh
+++ b/infra/gcp/ensure-prod-storage.sh
@@ -220,4 +220,9 @@ empower_group_to_admin_artifact_auditor \
 empower_artifact_auditor "${PROD_PROJECT}"
 empower_artifact_auditor_invoker "${PROD_PROJECT}"
 
+# Special case: empower Prow account to have access to GCR for Workload Identity.
+empower_prowacct_for_workload_identity \
+    "k8s-prow-builds.svc.id.googtest-pods/k8s-artifacts-prod" \
+    $(svc_acct_email "${PROD_PROJECT}" "${PROMOTER_SVCACCT}")
+
 color 6 "Done"
--- a/infra/gcp/lib.sh
+++ b/infra/gcp/lib.sh
@@ -439,24 +439,28 @@ function ensure_dns_zone() {
 # the given GCP service account for the given GKE Project's Kubernetes
 # namespace; this is also called Workload Identity.
 #
-# $1: The GKE project wherein the KSA resides; used to construct the KSA.
-# $2: The K8s namespace; used to construct the KSA.
-# $3: The GCP project that owns the GCP service account; used to construct the KSA.
-# $4: The GCP service account to draw powers from
+# $1: The service account scoped to a (1) GKE project, (2) K8s namespace, and
+#     (3) K8s service account. The format is currently
+#
+#       "${gke_project}.svc.id.goog[${k8s_namespace}/${k8s_svcacct}]"
+#
+#     This is used to scope the grant of permissions to the combination of the
+#     above 3 variables
+# $2: The GCP project that owns the GCP service account.
+# $3: The GCP service account to draw powers from.
 function empower_ksa_to_svcacct() {
-    if [ ! $# -eq 4 -o -z "$1" -o -z "$2" -o -z "$3" -o -z "$4" ]; then
-        echo "empower_ksa_to_svcacct(gke_project, k8s_namespace, gcp_project, gcp_scvacct) requires 4 arguments" >&2
+    if [ ! $# -eq 3 -o -z "$1" -o -z "$2" -o -z "$3" ]; then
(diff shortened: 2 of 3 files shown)


infra/gcp/ensure-prod-storage.sh
```diff
@@ -220,4 +220,9 @@ empower_group_to_admin_artifact_auditor \
 empower_artifact_auditor "${PROD_PROJECT}"
 empower_artifact_auditor_invoker "${PROD_PROJECT}"
 
+# Special case: empower Prow account to have access to GCR for Workload Identity.
+empower_prowacct_for_workload_identity \
```
fejta: IMO this should be more like:
* `empower_pods_to_identify_as_gcp_service_account`
* `bind_k8s_service_account_to_act_as_gcp_service_account`

infra/gcp/lib_gcr.sh
```diff
@@ -149,3 +149,22 @@ function empower_svcacct_to_admin_gcr () {
...
+# service account (for Workload Identity).
+# $1: The Prow service account being empowered
+# $2: The GCP service account to draw powers from
+function empower_prowacct_for_workload_identity() {
+    if [ ! $# -eq 2 -o -z "$1" -o -z "$2" ]; then
+        echo "empower_prowacct_for_workload_identity(prow_acct, gcp_acct) requires 2 arguments" >&2
+        return 1
+    fi
+
+    prow_acct="$1"
```
fejta: I'd consider separating out these variables into their parts so its more clear, I'd recommend this:

```shell
gke_project=$1
namespace=$2
name=$3
gcp_acct=$4
gcp_project=$5

gcloud iam service-accounts add-iam-policy-binding \
  --role roles/iam.workloadIdentityUser \
  --member "serviceAccount:${gke_project}.svc.id.goog[${namespace}/${name}]" \
  "$gcp_acct" --project "$gcp_project"
```

infra/gcp/ensure-prod-storage.sh
```diff
@@ -220,4 +220,9 @@ empower_group_to_admin_artifact_auditor \
...
+    "k8s-prow-builds.svc.id.googtest-pods/k8s-artifacts-prod" \
```
fejta: `k8s-prow-builds.svc.id.goog[test-pods/k8s-artifacts-prod]` (you need the `[]`)

infra/gcp/lib_gcr.sh
```diff
@@ -149,3 +149,22 @@ function empower_svcacct_to_admin_gcr () {
 
     empower_svcacct_to_admin_gcs_bucket "${group}" "${bucket}"
 }
+
```
thockin: First note - this does not belong in this file.  This file just for very low-level GCR-centric primitives.  Put this back in lib.sh until that can be broken up more.

infra/gcp/lib_gcr.sh
```diff
@@ -149,3 +149,22 @@ function empower_svcacct_to_admin_gcr () {
...
+function empower_prowacct_for_workload_identity() {
+    if [ ! $# -eq 2 -o -z "$1" -o -z "$2" ]; then
+        echo "empower_prowacct_for_workload_identity(prow_acct, gcp_acct) requires 2 arguments" >&2
+        return 1
+    fi
+
+    prow_acct="$1"
+    gcp_acct="$2"
+
+    gcloud iam service-accounts add-iam-policy-binding \
```
thockin: you need --project, don't you?

infra/gcp/lib_gcr.sh
thockin: I don;t think we need to go quite this far.

I would position this function as `empower_ksa_to_svcacct(project, k8s_svcacct, gcp_svcacct)`

listx:
@fejta Thanks for the wording clarifications --- hopefully I captured the intent more accurately. But please take another look!

thockin:
@fejta This says that "test-pods/k8s-artifacts-prod" is the namespace/sa right?  That implies that other pods run is the same namespace ("test-pods"), is that right?  If so, can you compare this to running in a dedicated namespace that doesn't risk cross-pod leaks of the service account?

infra/gcp/lib.sh
```diff
@@ -434,3 +434,29 @@ function ensure_dns_zone() {
...
+    fi
+
+    gke_project="$1"
+    k8s_namespace="$2"
+    gcp_project="$3"
+    gcp_svcacct="$4"
+
+    gcloud iam service-accounts add-iam-policy-binding \
+        --role roles/iam.workloadIdentityUser \
+        --member "serviceAccount:${gke_project}.svc.id.goog[${k8s_namespace}/${gcp_project}]" \
```
thockin: Rather than construct the IDNS this way, I'd rather we pass it in fully formed.  This ensures that, should that format ever change or extend, we don't need to change this function or duplicate it or anything.

Also I think the assumption that there's one SA per project is dubious - why couldn't we have 2 GSAs in k8s-artifacts-prod that were each used by different prow jobs?  This conflates the name of the KSA in the GKE cluster with the name of the project the GSA exists in.  It's fine to name them the same, but don't assume they are ALWAYS named the same.

IOW, change args 2 and 3 into a single string, with the documented example "${gke_project}.svc.id.goog[${k8s_namespace}/${k8s_svcacct}]"

infra/gcp/lib.sh
```diff
@@ -434,3 +434,29 @@ function ensure_dns_zone() {
...
+        "${gcp_svcacct}" \
+        --project="${project}"
```
fejta: `gcp_project`

infra/gcp/lib.sh
fejta: An [existing example](https://github.com/kubernetes/test-infra/blob/c5d80629ff8a2eb577d05b3dd88986ff2e5a0c48/prow/cluster/build_serviceaccounts.yaml#L17-L18
) where the service account name in the GKE cluster (`k8s-kops-test`) does not match the GCP project name (`kubernetes-jenkins-pull`).

An [existing example](https://github.com/kubernetes/test-infra/blob/c5d80629ff8a2eb577d05b3dd88986ff2e5a0c48/prow/cluster/trusted_serviceaccounts.yaml#L6-L14) where we are using mulitple GCP service accounts (`resultstore`, `pusher`, `deployer`) in the same project (`k8s-prow`).

fejta:
> This says that "test-pods/k8s-artifacts-prod" is the namespace/sa right? That implies that other pods run is the same namespace ("test-pods"), is that right?

Yes and yes (although we could use unit tests to mitigate against configuring jobs to use this SA).

> If so, can you compare this to running in a dedicated namespace that doesn't risk cross-pod leaks of the service account?

Can we use the new #wg-k8s-infra scripting to schedule these pods in a community owned cluster somewhere? It is easy to configure prow to schedule pods in any conformant k8s cluster.

Yes, my request has been from the start (and continues to be) that we run these jobs outside of prow's trusted context, ideally in their own GKE cluster.

The trusted context is for jobs which prow fully trusts, which should not include these jobs: the core prow oncall/maintainer team does not own/review/maintain the code behind these jobs.

These belong in a different trust domain, which means in a different context than the prow control plane, ideally their own cluster.

infra/gcp/lib.sh
listx: > Also I think the assumption that there's one SA per project is dubious - why couldn't we have 2 GSAs in k8s-artifacts-prod that were each used by different prow jobs? This conflates the name of the KSA in the GKE cluster with the name of the project the GSA exists in. It's fine to name them the same, but don't assume they are ALWAYS named the same.

As I was writing the code, I felt that this assumption was on shaky ground. Thanks for the guidance!

listx:
Updated... PTAL, esp. how I expanded the comments.

thockin:
@fejta yes and no.  In theory yes.  In practice the apiserver IP is firewalled off and so we'd have to think about how to actually do it.

fejta:
> In practice the apiserver IP is firewalled off

Interesting. Which API server? How are things deployed/upgraded/etc when `kubectl` cannot talk to it?

thockin:
Cloud shell is how we manage it.

trying to follow "most secure" practices.

listx:
I think we have to actuate this first, then un-revert https://github.com/kubernetes/test-infra/pull/16463 and see if the ci-k8sio-cip job (or post-k8sio-cip job) still works. Once all the jobs have migrated to WI, then the JSON keys can be revoked.

I think we need to set aside a day to test all of the above; maybe next week? Busy today and most likely tomorrow with https://github.com/kubernetes-sigs/k8s-container-image-promoter/issues/191

thockin:
file a followup bug? I know we have other things to finish (auditor -> alerts :)

listx:
> file a followup bug? I know we have other things to finish (auditor -> alerts :)

https://github.com/kubernetes-sigs/k8s-container-image-promoter/issues/180

thockin:
applied

PR #658, opened 2020-03-12
Add secrets-store-csi-driver project

aramase:
- `secrets-store-csi-driver` was recently moved to kubernetes-sigs
- onboarding the project, so we can start publishing the images to `gcr.io`


groups/groups.yaml
```diff
@@ -1137,3 +1137,13 @@ groups:
       - mpatel@redhat.com
       - pehunt@redhat.com
       - saschagrunert@gmail.com
+
+  - email-id: k8s-infra-staging-secrets-store-csi@kubernetes.io
```
aramase: used `secrets-store-csi` instead of `secrets-store-csi-driver` because of the limitation provided -

```
Create a google group for granting push access by adding an email alias for it in groups.yaml. The email alias should be of the form k8s-infra-staging-<project-name>@kubernetes.io. The project name can have a maximum of 18 characters.
```

groups/groups.yaml
```diff
@@ -1137,3 +1137,13 @@ groups:
...
+    name: k8s-infra-staging-secrets-store-csi
+    description: |-
+      ACL for staging secrets-store-csi-driver
+    settings:
+      ReconcileMembers: "true"
+    members:
+      - anish.ramasekar@gmail.com
+      - ritazh@microsoft.com
```
ritazh: can you pls update to `rita.z.zhang@gmail.com` instead?

aramase: updated!

infra/gcp/ensure-staging-storage.sh
```diff
@@ -70,6 +70,7 @@ STAGING_PROJECTS=(
     release-test
     releng
     scl-image-builder
+    secrets-store-csi-driver
```
thockin: This is going to be too long.  Drop "driver" at least.  That ripples to all the other files in here.

k8s.gcr.io/images/k8s-staging-secrets-store-csi-driver/OWNERS
```diff
@@ -0,0 +1,5 @@
+# See the OWNERS docs at https://go.k8s.io/owners
```
thockin: You'll need the equivalent file in manifests, which defines the paths for promoter to operate on.

infra/gcp/ensure-staging-storage.sh
thockin: I might also suggest "csi-secrets-store" so other csi can be grouped together?

listx:
> So we finally get to have the granularity discussion. :)
> 
> On its face this seems fine and consistent with precedent, but...
> 
> Should we force grouping? E.g. should sig- get a quota on stagings? Or something?
> 
> Or should we nest "directories" in GCR? E.g. instead of k8s.gcr.io/csi-foobar and k8s.gcr.io/csi-quuxzorb, should we encourage k8s.gcr.io/csi/{foobar,quuxzorb}?

+1 for nesting, seems like the right paradigm, at least from a technical standpoint

thockin:
The main argument against nesting is that we didn't already do it for capi
):)

detiber:
I've added an agenda item to the next Cluster API meeting to discuss the topic of nesting staging repos.

thockin:
Note that there are 2 questions and they are orthogonal.

1) Should we merge smaller-scoped staging projects (probably no, IMO)
2) Should we "nest" related prod GCR directories?

On #2 specifically, the nesting would be in the FINAL, public, prod
container image name.

infra/gcp/ensure-staging-storage.sh
aramase: @thockin Thank you for the suggestion. I'll make the change.

aramase: done!

k8s.gcr.io/images/k8s-staging-secrets-store-csi-driver/OWNERS
aramase: thank you! updated it

aramase:
@thockin Updated the PR based on the comments, PTAL!

aramase:
@dims @thockin PTAL when you get a chance!

ncdc:
@thockin re capi, I'd be ok with all the sig-sponsored cluster api images going into a single bucket, with distinct paths. e.g. we currently have k8s-staging-cluster-api/cluster-api-controller and k8s-staging-cluster-api-aws/cluster-api-aws-controller (just to name 2). I'm fine moving everything under a cluster-api bucket (core cluster api + all sig-sponsored providers), if that works for everyone else.

Also xref #671, which is requesting a staging bucket for the digitalocean capi provider.

cc @timothysc @vincepri @CecileRobertMichon @yastij @randomvariable and I'm sure there are others I'm leaving off

listx:
If we do end up reorganizing some of the final prod image paths, we should not delete the old (existing) promoter manifests because they will serve as a historical record. (I.e., let's not be forced to do [this](https://github.com/kubernetes/k8s.io/pull/642) again).

vincepri:
I'm +1 to organize all CAPI projects under the CAPI staging project. In terms of timeline, where are we looking to do this change?

detiber:
If we do organize under a single staging bucket for cluster-api, building on what @ncdc mentioned, I think it should be organized something like this:

cluster-api/ <- hosts all images that are built from cluster-api project directly
cluster-api/<provider-type>/<provider-name> <- hosts images related to that provider

This would allow for separation of ownership/access for sig-sponsored sub-projects that might have separate ownership on individual repos.

Should we rally around a doc/issue for hashing a concrete proposal, including transition plan and maintaining the historical record (as well as maintaining backward compat)?

dims:
/hold 

looks like lots of discussions here

spiffxp:
@detiber 
>Should we rally around a doc/issue for hashing a concrete proposal, including transition plan and maintaining the historical record (as well as maintaining backward compat)?

Yes.  I just opened https://github.com/kubernetes/k8s.io/issues/696 let's take the discussion over there.

/assign @thockin @dims 
I question whether we should continue to block this PR. It's unclear to me how quickly we're going to sort this out, and we're going to have to migrate things anyway.  What do you think?

aramase:
@dims @thockin Can we unblock this PR as per @spiffxp comment?

aramase:
Thank you @thockin!

PR #660, opened 2020-03-15
WIP: Script creation of AWS test accounts

justinsb:
These scripts are how we currently create these accounts; much
opportunity to bring it under better management.


infra/aws/e2e/accounts/attach.sh
```diff
@@ -0,0 +1,14 @@
+#!/bin/bash
+
+set -e
+set -x
+
+for s in a b c d; do
+  ID=e2e-kops-2020-03-15-$s
+
+  for r in AmazonEC2FullAccess IAMFullAccess AWSDeepRacerCloudFormationAccessPolicy; do
```
detiber, line 9: Not directly related to this PR, but if we want to also use accounts created by this script for cluster-api-provider-aws, it would be nice to also add `AmazonSSMFullAccess` here, or a custom policy that would at least allow creation of Session Manager-initiated remote sessions.

infra/aws/e2e/accounts/bulk-create.sh
```diff
@@ -0,0 +1,32 @@
+#!/bin/bash
+
+set -e
+set -x
+
+for s in a b c d; do
+  ID=e2e-kops-2020-03-15-$s
+  aws organizations create-account --email k8s-infra-aws-admins+${ID}@kubernetes.io --account-name ${ID}
```
listx, line 8: Did you run this through https://www.shellcheck.net/ ?

infra/aws/e2e/accounts/bulk-create.sh
```diff
@@ -0,0 +1,32 @@
+#!/bin/bash
+
+set -e
+set -x
```
listx, line 4: You should also have `set -o pipefail` because of piped commands. And also `set -o nounset`.

infra/aws/e2e/accounts/to-yaml.sh
```diff
@@ -0,0 +1,37 @@
+#!/bin/bash
+
+set -e
+#set -x
```
listx, line 4: Remove?

infra/aws/e2e/accounts/attach.sh
```diff
@@ -0,0 +1,14 @@
+#!/bin/bash
+
+set -e
+set -x
+
+for s in a b c d; do
```
listx, line 6: I don't understand how `a b c d` here is used. Can you add comments to explain?

Also, copyright info.

infra/aws/e2e/accounts/to-yaml.sh
```diff
@@ -0,0 +1,37 @@
...
+
+
+toyaml() {
```
listx, line 7: Nit: `to_yaml` instead

spiffxp:
Obviated by https://github.com/kubernetes/k8s.io/pull/1016 ?

ameukam:
@justinsb still want to do this ?

PR #628, opened 2020-03-04
Add cluster-api-provider-vsphere to image promoter

randomvariable:
Adding https://github.com/kubernetes-sigs/cluster-api-provider-vsphere to image promotion, as per https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/issues/795

/assign @dims 
cc @yastij

PR #671, opened 2020-03-19
Add cluster-api-provider-digitalocean to image promoter

xmudrii:
Add a staging GCR bucket for [cluster-api-provider-digitalocean](https://github.com/kubernetes-sigs/cluster-api-provider-digitalocean).

### DIFF
--- a/groups/groups.yaml
+++ b/groups/groups.yaml
@@ -571,6 +571,15 @@ groups:
       - ha.chuck@gmail.com
       - hi@amycod.es
       - prignanov@vmware.com
+  
+  - email-id: k8s-infra-staging-cluster-api-do@kubernetes.io
+    name: k8s-infra-staging-cluster-api-do
+    description: |-
+      ACL for staging CAPI for DigitalOcean
+    settings:
+      ReconcileMembers: "true"
+    members:
+      - mudrinic.mare@gmail.com
 
   - email-id: k8s-infra-staging-capi-vsphere@kubernetes.io
     name: k8s-infra-staging-capi-vsphere
--- a/groups/groups.yaml
+++ b/groups/groups.yaml
@@ -579,6 +579,7 @@ groups:
     settings:
       ReconcileMembers: "true"
     members:
+      - detiber@gmail.com
       - mudrinic.mare@gmail.com
 
   - email-id: k8s-infra-staging-capi-vsphere@kubernetes.io
--- a/infra/gcp/ensure-staging-storage.sh
+++ b/infra/gcp/ensure-staging-storage.sh
@@ -46,6 +46,7 @@ STAGING_PROJECTS=(
     cluster-api
     cluster-api-aws
     cluster-api-azure
+    cluster-api-do
     cluster-api-gcp
     capi-openstack
     capi-kubeadm
--- a/k8s.gcr.io/images/k8s-staging-cluster-api-do/images.yaml
+++ b/k8s.gcr.io/images/k8s-staging-cluster-api-do/images.yaml
@@ -1,0 +1,2 @@
+#- name: cluster-api-do-controller
+#  dmap: {}
--- a/k8s.gcr.io/manifests/k8s-staging-cluster-api-do/promoter-manifest.yaml
+++ b/k8s.gcr.io/manifests/k8s-staging-cluster-api-do/promoter-manifest.yaml
@@ -1,0 +1,11 @@
+# google group for gcr.io/k8s-staging-cluster-api-do is k8s-infra-staging-cluster-api-do@kubernetes.io
+registries:
+- name: gcr.io/k8s-staging-cluster-api-do
+  src: true
+- name: us.gcr.io/k8s-artifacts-prod/cluster-api-do
+  service-account: k8s-infra-gcr-promoter@k8s-artifacts-prod.iam.gserviceaccount.com
+- name: eu.gcr.io/k8s-artifacts-prod/cluster-api-do
+  service-account: k8s-infra-gcr-promoter@k8s-artifacts-prod.iam.gserviceaccount.com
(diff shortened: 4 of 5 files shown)


k8s.gcr.io/images/k8s-staging-cluster-api-do/images.yaml
```diff
@@ -0,0 +1,3 @@
+- name: cluster-api-do-controller
+  dmap:
+    "sha256:": ["v0.3.0"]
```
xmudrii: What is the process to get the SHA256? Should I first cut the release, build the image locally, and then past the SHA256? Is there some other process?

groups/groups.yaml
```diff
@@ -571,6 +571,15 @@ groups:
...
       - prignanov@vmware.com
+  
+  - email-id: k8s-infra-staging-cluster-api-do@kubernetes.io
+    name: k8s-infra-staging-cluster-api-do
+    description: |-
+      ACL for staging CAPI for DigitalOcean
+    settings:
+      ReconcileMembers: "true"
+    members:
+      - mudrinic.mare@gmail.com
```
xmudrii: Right now, I'm the only one maintainer of the CAPI-DO project ([SCL mailing list post for more details](https://groups.google.com/forum/#!topic/kubernetes-sig-cluster-lifecycle/iKvpeKAJmNo)). We will add two new maintainers after the lazy consensus passes on 3/25. 

Is it okay that I'm the only member/owner for now? Should I wait for new maintainers to get promoted or add someone from SCL/Cluster-API?

k8s.gcr.io/images/k8s-staging-cluster-api-do/images.yaml
detiber: You shouldn't actually need to list any images prior to having an image pushed to staging first.

detiber: After you have a published staging image, you can get the sha either from the output of the job that is pushing the image, or you can look in the GCR registry through the console, for example the staging repo for cluster-api-provider-aws is viewable here: https://console.cloud.google.com/gcr/images/k8s-staging-cluster-api-aws/GLOBAL/cluster-api-aws-controller?project=k8s-staging-cluster-api-aws&gcrImageListsize=30 (may not be viewable if you are not in the group membership)

groups/groups.yaml
detiber: A single owner/member here should be fine for now, I'm also more than happy to be a backup resource to assist if needed as well.

k8s.gcr.io/images/k8s-staging-cluster-api-do/images.yaml
xmudrii: That makes sense! Thanks for the explanation! I've dropped this file from the PR.

detiber:
@xmudrii I think you'll need the image yaml file, just with an empty dmap dict

xmudrii:
@detiber I've updated the PR to look like #628

/cc @dims @cblecker @thockin 
we will need the google group and bucket created

dims:
@xmudrii please add @detiber as well. 

/hold

xmudrii:
@dims Done, PTAL.

infra/gcp/ensure-staging-storage.sh
```diff
@@ -46,6 +46,7 @@ STAGING_PROJECTS=(
     cluster-api
     cluster-api-aws
     cluster-api-azure
+    cluster-api-do
```
thockin: @detiber Why are some "cluster-api" and others "capi" ?  Is there a principle as to which to use?

k8s.gcr.io/manifests/k8s-staging-cluster-api-do/promoter-manifest.yaml
```diff
@@ -0,0 +1,11 @@
+# google group for gcr.io/k8s-staging-cluster-api-do is k8s-infra-staging-cluster-api-do@kubernetes.io
+registries:
+- name: gcr.io/k8s-staging-cluster-api-do
+  src: true
+- name: us.gcr.io/k8s-artifacts-prod/cluster-api-do
```
thockin: @detiber Have you given thought to "nesting" of cluster-api in GCR-space?

infra/gcp/ensure-staging-storage.sh
xmudrii: > I think it's related to the max length of the names, but I'm not sure if it's related to GCR names or GCS Bucket ones?

Yes, I believe the same. `cluster-api-do` is less than 18 characters, so I went with that name.

k8s.gcr.io/manifests/k8s-staging-cluster-api-do/promoter-manifest.yaml
xmudrii: @thockin It seems like there is some discussion ongoing here https://github.com/kubernetes/k8s.io/pull/658#issuecomment-601421263. However, I believe that is out of the scope of this PR and out of the scope for CAPI-DO maintainers.

xmudrii:
> Before we actuate this - I have questions on name.

@thockin @dims told me that that the bucket is ready. Is there anything else that should be done before we can use it?

infra/gcp/ensure-staging-storage.sh
detiber: @bartsmykla is correct, the main reason for the bifurcation was due to character length. We'll work to unify the naming as we work on the proposal for nesting.

detiber:
@xmudrii once the bucket has been created it should be all ready for use now.

Will reach out over slack to point you towards the automation of how we are building images and pushing to staging for other providers today.

Issue #666, opened 2020-03-18
optimize k8s.gcr.io backup script


listx:
We are exceeding quota for GCR by doing a simple copy of ~30K images in each of the 3 regions (for a total of ~90K image copies).

We have spoken with GCR and there is currently no way of giving special quota privileges to particular GCP projects. While this might be worth pursuing (simply increasing quota), I think we can just work on optimizing the backup scripts. There are a couple reasons for this:

- We currently do a simple `gcrane cp -r <prod> <prod-backup>/<timestamp>` and while this works, it is *slow* (on the order of taking *hours* to complete), even if there are 0 changes from the point of the last backup!
- The prod GCR (`k8s-artifacts-prod`) will be immutable (the promoter does not allow mutations), so essentially it is an ever-growing-GCR. Given this nature, we can just make sure that the backup always just adds a delta of new images since the last snapshot, as the old backed-up copies will never change or be removed.

I'll come up with a more detailed design soon.

/assign @listx

stp-ip:
First suggestion would be to only backup a single region and add a `sync-check` separately. We should be checking that all regions are in sync anyway.


Is copying over into the same prefix/timestamp faster?

We could copy into the same location gcrane cp -r <prod> <prod-backup>/ and set a long term retention to prevent modifications, but that goes against the GCR teams recommendation to assume or work with anything in GCS, but we already do this for prod afaik so could do here as well.

Splitting or stretching the backup window had issues as gcrane doesn't allow rate-limits afaik and with running into quotas we block updates to say prow.

listx:
> First suggestion would be to only backup a single region and add a `sync-check` separately. We should be checking that all regions are in sync anyway.
> 
> Is copying over into the same prefix/timestamp faster?

How do you mean, exactly? (You can take a look [here](https://console.cloud.google.com/gcr/images/k8s-artifacts-prod-bak/US/2020/03/05?gcrImageListsize=30) for some examples of timestamped prefixes.)

> 
> We could copy into the same location gcrane cp -r / and set a long term retention to prevent modifications, but that goes against the GCR teams recommendation to assume or work with anything in GCS, but we already do this for prod afaik so could do here as well.
 
I'd rather not deal with GCS (implementation) details. We have been advised a number of times by the GCR team to treat GCS as opaque.

> Splitting or stretching the backup window had issues as gcrane doesn't allow rate-limits afaik and with running into quotas we block updates to say prow.

[...]

PR #677, opened 2020-03-20
Optimize backups

listx:
This implements the optimizations described in https://github.com/kubernetes/release/issues/270#issuecomment-601481641 and https://github.com/kubernetes/k8s.io/issues/666.

/hold
/wip


listx:
Looks like the secret was deleted (error from the `pull-k8sio-backup` test):

```
The pod could not start because it could not mount the volume "creds": secret "k8s-gcr-backup-test-prod-bak-service-account" not found
```

I'll take this setback as an opportunity to move to Workload Identity.

listx:
I am going with https://github.com/kubernetes/test-infra/pull/16883 first to see how this Workload Identity stuff plays out in a simpler context. The context there is simpler because we already actuated WI KSA empowerment stuff in #655. For us to use WI here, I would first have to add another special case for the promoter GCP SA (specifically, `k8s-infra-gcr-promoter@k8s-gcr-backup-test-prod-bak.iam.gserviceaccount.com` as per https://github.com/kubernetes/test-infra/pull/15398#issue-345513930).

Once https://github.com/kubernetes/test-infra/pull/16883 is merged, the other changes described above will follow. After all that, then the steps to re-enable backups are as follows:

(1) create another workload identity empowerment PR like #655 for `k8s-infra-gcr-promoter@k8s-gcr-backup-test-prod-bak.iam.gserviceaccount.com`
(2) update the `pull-k8sio-backup` job in test-infra repo to use Workload Identity
(3) re-test the the `pull-k8sio-backup` check to make sure we are not breaking any backup behaviors
(4) merge this PR
(5) re-enable the periodic backup job (`ci-k8sio-backup`; i.e., revert https://github.com/kubernetes/test-infra/pull/16835) --- to run 1x per day to ensure we stay beneath quota limits (the initial runs will take hours and fail with a job timeout, most likely, because we would need to copy all 30K images for the very first time to each region). I can also run the backups on a separate machine to speed up the process.
(6) observe how long backups take once the initial set of images have been copied over; that is, how long would a NOP backup job (due to no delta) take? Experimentally we have seen numbers around the ~5min range
(7) if (6) shows ~5min, then we can increase the backup frequency back to hourly.

spiffxp:
/approve
/lgtm
I leave the `/hold cancel` to you @listx

listx:
@spiffxp I need to wait for the workload identity stuff to get submitted first, namely https://github.com/kubernetes/k8s.io/pull/710 and https://github.com/kubernetes/test-infra/pull/17048. At that point, I can use the `pull-k8sio-backup` presubmit check here to verify that WI is correctly working (and that `gcrane` can handle it --- which it should).

This PR is blocked by `pull-k8sio-backup` passing.

listx:
@jonjohnsonjr I've moved `pull-k8sio-backup` to use Workload Identity instead of `GOOGLE_APPLICATION_CREDENTIALS=...` and it appears that even though it authenticates as the same service account it used to auth as ADC, it's not the same as ADC and thus failing.

For example I can see in the `pull-k8sio-backup` test's logs:

```
+ gcloud auth list
                               Credentialed Accounts
ACTIVE  ACCOUNT
*       k8s-infra-gcr-promoter@k8s-gcr-backup-test-prod-bak.iam.gserviceaccount.com
```

but it is failing with:

```
+ /home/prow/go/src/github.com/google/go-containerregistry/cmd/gcrane/gcrane ls -r us.gcr.io/k8s-gcr-backup-test-prod
+ xargs -n1 /home/prow/go/src/github.com/google/go-containerregistry/cmd/gcrane/gcrane delete
2020/04/02 07:13:14 No matching credentials were found, falling back on anonymous
2020/04/02 07:13:15 deleting us.gcr.io/k8s-gcr-backup-test-prod/artifact-promoter/cip@sha256:f61e2ff9d38993b0e8acd93a8d3d9fbafedf67d97845c244bf4d329aadf0e3c4: GET https://us.gcr.io/v2/token?scope=repository%3Ak8s-gcr-backup-test-prod%2Fartifact-promoter%2Fcip%3Apush%2Cpull&service=us.gcr.io: UNAUTHORIZED: You don't have the needed permissions to perform this operation, and you may have invalid credentials. To authenticate your request, follow the steps in: https://cloud.google.com/container-registry/docs/advanced-authentication
```

I guess `gcrane` always requires ADC? Do you know if it is possible to make Workload Identity work with ADC?

listx:
For more context the gcrane version was built from 3d03ed9b1ca2ad5d78d43832e8e46adc31d2b961 (master HEAD).

listx:
/wip

Still working on resolving the auth issues post-WI...!

listx:
Woohoo, it passed!

FTR the auth was not working for gcrane because we didn't do `gcloud auth configure-docker`. This is now necessary because we no longer pass in ADC creds as an env var.

/hold cancel

listx:
Just following up on the work items:

> (5) re-enable the periodic backup job (`ci-k8sio-backup`; i.e., revert [kubernetes/test-infra#16835](https://github.com/kubernetes/test-infra/pull/16835)) --- to run 1x per day to ensure we stay beneath quota limits (the initial runs will take hours and fail with a job timeout, most likely, because we would need to copy all 30K images for the very first time to each region). I can also run the backups on a separate machine to speed up the process.

This was done here: https://github.com/kubernetes/test-infra/pull/17150

> (6) observe how long backups take once the initial set of images have been copied over; that is, how long would a NOP backup job (due to no delta) take? Experimentally we have seen numbers around the ~5min range

The first successful run, which backed up a handful of images, is here: https://prow.k8s.io/view/gcs/kubernetes-jenkins/logs/ci-k8sio-backup/1248214661472456704

This took 19 minutes, which isn't too bad because, keep in mind that is for 3 regions, serially. So it took ~6-7 minutes per region. We could increase the speed by making the 3 copies parallel (using something like GNU parallel), but I don't think it's worth it (we have to most likely install it at the beginning of the job, because I doubt it's in the image we use).

> (7) if (6) shows ~5min, then we can increase the backup frequency back to hourly.

According to [this comment thread](https://docs.google.com/document/d/11eiosJvm2xEVUhPRU3-luANxxTPL5FqQdJXVrLPImyQ/edit?disco=AAAAGUzb7fY), for a NOP copy we would only be using ~3K requests to list all images for 1 region. If we did the job hourly, then there would be a minimum of 3K * 3 regions, or ~9K requests per hour, or 24 (hours) * 9K or ~216K requests to GCR per day. These are both well under the ~50K requests per 10 minutes, or 1000K requests per day quota limits [here](https://cloud.google.com/container-registry/quotas). However, the request count still feels heavy to me to be doing this hourly.

I think it's OK to leave it at 12h intervals for now.

Issue #681, opened 2020-03-22
apiGroup in role applied by ensure-namespaces.sh script is probably wrong


bartsmykla:
Doing some research today one thought came to me.

@munnerz: [This Pull Request](https://github.com/kubernetes/k8s.io/pull/635) changes the apiGroups from `certmanager.k8s.io` to `cert-manager.io`. I think we should also change it at [`infra/gcp/namespaces/namespace-user-role.yml`](https://github.com/kubernetes/k8s.io/blob/master/infra/gcp/namespaces/namespace-user-role.yml#L11) which is applied for every project at [`aaa` cluster](https://github.com/kubernetes/k8s.io/tree/master/infra/gcp/clusters/kubernetes-public/aaa) when [`ensure-namespaces.sh`](https://github.com/kubernetes/k8s.io/blob/master/infra/gcp/namespaces/ensure-namespaces.sh] script is ran.

I'll followup with the PR, but please confirm my thoghts are right

/assign @munnerz

munnerz:
I've responded over in #682 :)

PR #635, opened 2020-03-06
Update apiVersion on all cert-manager resources

munnerz:
Following on from #586, the apiVersion for all cert-manager resources needs updating.

I have applied this change, along with #586, to AAA already (as v1alpha1 does not exist since I have upgraded to v0.13.1, meaning these manifests cannot otherwise be applied).

```
kubectl get clusterissuer,certificate --all-namespaces
NAME                                                READY   AGE
clusterissuer.cert-manager.io/letsencrypt-prod      True    3m
clusterissuer.cert-manager.io/letsencrypt-staging   True    3m
clusterissuer.cert-manager.io/selfsigning-issuer    True    3m
NAMESPACE       NAME                                                  READY   SECRET              AGE
gcsweb          certificate.cert-manager.io/gcsweb-k8s-io             True    gcsweb-k8s-io-tls   2m
k8s-io-canary   certificate.cert-manager.io/k8s-io-selfsigned         True    k8s-io-tls          1m
k8s-io-prod     certificate.cert-manager.io/k8s-io-letsencrypt-prod   True    k8s-io-tls          1m
```

🎉 

/assign @bartsmykla @dims

PR #682, opened 2020-03-22
Fixed apiGroup of certificates in namespaces role

bartsmykla:
Closes: https://github.com/kubernetes/k8s.io/issues/681

It looks like the apiGroups changed https://github.com/kubernetes/k8s.io/pull/635
Updated the one in infra/gcp/namespaces/ensure-namespaces.sh
for certificates

Holding for @munnerz to look
/assign @munnerz 
/hold


munnerz:
It may also be worth including `acme.cert-manager.io` in here too, as 'Order' and 'Challenge' resources now exist in a separate API group to the core 'Issuer' and 'Certificate' resource types :)

PR #479, opened 2019-11-27
Add external-dns to k8s.gcr.io

njuettner:

PR #689, opened 2020-03-24
Add a DNS staging repo

prameshj:
@MrHohn @bowei @listx

### DIFF
--- a/groups/groups.yaml
+++ b/groups/groups.yaml
@@ -698,6 +698,17 @@ groups:
       - ravisantoshgudimetla@gmail.com
       - spiffxp@google.com
 
+  - email-id: k8s-infra-staging-dns@kubernetes.io
+    name: k8s-infra-staging-dns
+    description: |-
+      ACL for staging kube-dns and node-local-dns images.
+    settings:
+      ReconcileMembers: "true"
+    members:
+      - zihongz@google.com
+      - bowei@google.com
+      - pavithrar@google.com
+
   - email-id: k8s-infra-staging-e2e-test-images@kubernetes.io
     name: Kubernetes E2E test image staging registry owners
     description: |-
--- a/infra/gcp/ensure-staging-storage.sh
+++ b/infra/gcp/ensure-staging-storage.sh
@@ -56,6 +56,7 @@ STAGING_PROJECTS=(
     coredns
     csi
     descheduler
+    dns
     e2e-test-images
     external-dns
     kas-network-proxy
--- a/k8s.gcr.io/images/k8s-staging-dns/OWNERS
+++ b/k8s.gcr.io/images/k8s-staging-dns/OWNERS
@@ -1,0 +1,6 @@
+# See the OWNERS docs at https://go.k8s.io/owners
+
+approvers:
+- bowei
+- mrhohn
+- prameshj
--- a/k8s.gcr.io/images/k8s-staging-dns/images.yaml
+++ b/k8s.gcr.io/images/k8s-staging-dns/images.yaml
@@ -1,0 +1,1 @@
+# No images yet
--- a/k8s.gcr.io/manifests/k8s-staging-dns/promoter-manifest.yaml
+++ b/k8s.gcr.io/manifests/k8s-staging-dns/promoter-manifest.yaml
@@ -1,0 +1,10 @@
+# google group for gcr.io/k8s-staging-dns is k8s-infra-staging-dns@kubernetes.io
+registries:
+- name: gcr.io/k8s-staging-dns
+  src: true
+- name: us.gcr.io/k8s-artifacts-prod/dns
+  service-account: k8s-infra-gcr-promoter@k8s-artifacts-prod.iam.gserviceaccount.com
+- name: eu.gcr.io/k8s-artifacts-prod/dns
+  service-account: k8s-infra-gcr-promoter@k8s-artifacts-prod.iam.gserviceaccount.com
+- name: asia.gcr.io/k8s-artifacts-prod/dns
+  service-account: k8s-infra-gcr-promoter@k8s-artifacts-prod.iam.gserviceaccount.com


bartsmykla:
Hi @prameshj currently you are just creating the google group. What do you want to achieve? Maybe I'll be able to help with some instructions? :-)

prameshj:
> Hi @prameshj currently you are just creating the google group. What do you want to achieve? Maybe I'll be able to help with some instructions? :-)

Thanks! I just included the other files in this PR too.. it would be great if you can review the changes.

k8s.gcr.io/images/k8s-staging-dns/promoter-manifest.yaml
```diff
@@ -0,0 +1,11 @@
...
+registries:
+- name: gcr.io/k8s-staging-dns
+  src: true
+- name: us.gcr.io/k8s-artifacts-prod/dns
+  service-account: k8s-infra-gcr-promoter@k8s-artifacts-prod.iam.gserviceaccount.com
+- name: eu.gcr.io/k8s-artifacts-prod/dns
+  service-account: k8s-infra-gcr-promoter@k8s-artifacts-prod.iam.gserviceaccount.com
+- name: asia.gcr.io/k8s-artifacts-prod/dns
+  service-account: k8s-infra-gcr-promoter@k8s-artifacts-prod.iam.gserviceaccount.com
+imagesPath: "../../images/k8s-staging-dns/images.yaml"
```
listx: This has to be split up into 2 files, the `promoter-manifest.yaml` and the `images.yaml`.

I will have to update the docs to reflect this.

bartsmykla:
@prameshj could you point me to things which will be placed in this repository?

k8s.gcr.io/images/k8s-staging-dns/promoter-manifest.yaml
listx: Actually, I see that the instructions https://github.com/kubernetes/k8s.io/tree/master/k8s.gcr.io#creating-staging-repos already state this.

What docs did you follow? I need to know so I can update them as well.

stp-ip:
To answer this to the best of my knowledge until @prameshj comes into the office.
kube-dns (https://github.com/kubernetes/kubernetes/tree/master/cluster/addons/dns/kube-dns) and the node local containers (https://github.com/kubernetes/kubernetes/tree/master/cluster/addons/dns/nodelocaldns) are the subjects here.

We do have a separate coredns staging group (sorta replacement of kube-dns) and I agree with having a node-local one as well.
Suggestion would be to split this up into node-local and kube-dns staging repo.

Maybe I'm missing why it should be a single staging repo.

k8s.gcr.io/images/k8s-staging-dns/promoter-manifest.yaml
prameshj: I used the docs you linked. I did see that images.yaml should contain the images info and promoteer-manifest.yaml should contain the credentials. I did create images.yaml and left it empty since i do not need to promote an image right away. I was planning to send a separate PR when needed.

I got the content of the promoter-manifest.yaml file by looking at other staging repos examples, as the docs stated. Sorry if I missed something here. Which part of these contents should i have moved into images.yaml?

prameshj:
> To answer this to the best of my knowledge until @prameshj can answer.
> kube-dns (https://github.com/kubernetes/kubernetes/tree/master/cluster/addons/dns/kube-dns) and the node local containers (https://github.com/kubernetes/kubernetes/tree/master/cluster/addons/dns/nodelocaldns) are the subjects here.
> 
> We do have a separate coredns staging group (sorta replacement of kube-dns) and I agree with having a node-local one as well.
> Suggestion would be to split this up into node-local and kube-dns staging repo.
> 
> Maybe I'm missing why it should be a single staging repo.

Yes, kube-dns and node-local-dns images will be in this repo. The reason I used a single repo for all of them is because they are all built as part of the same command and the sources are all in k8s.io/dns. I have the repo a generic name "dns" so it would be applicable even if we used it just for node-local-dns in the future.

k8s.gcr.io/images/k8s-staging-dns/promoter-manifest.yaml
listx: For the promoter-manifest.yaml, if you look at e.g. https://github.com/kubernetes/k8s.io/blob/master/k8s.gcr.io/manifests/DO-NOT-MODIFY-legacy-backfill/promoter-manifest.yaml you can see how `imagesPath` does not exist.

Do you have a link to where you found the promoter-manifest.yaml that has the `imagesPath` still in it? I need to update that so that other people don't make the same mistake.

For the `images.yaml` you can use https://github.com/kubernetes/k8s.io/blob/master/k8s.gcr.io/images/k8s-staging-apisnoop/images.yaml which has an empty file. You still need an empty file.

listx: Also, just in case you didn't notice it: you created `k8s.gcr.io/images/k8s-staging-dns/promoter-manifest.yaml` which is in the wrong path. It should be `k8s.gcr.io/manifests/k8s-staging-dns/promoter-manifest.yaml`.

prameshj: > For the promoter-manifest.yaml, if you look at e.g. https://github.com/kubernetes/k8s.io/blob/master/k8s.gcr.io/manifests/DO-NOT-MODIFY-legacy-backfill/promoter-manifest.yaml you can see how `imagesPath` does not exist.
> 
> Do you have a link to where you found the promoter-manifest.yaml that has the `imagesPath` still in it? I need to update that so that other people don't make the same mistake.
> 

I followed the example from k8s-staging-external-dns. It does not have the imagesPath now though. My mistake, i took the example from the first PR that added the repo - https://github.com/kubernetes/k8s.io/pull/479 , didn't realise the file was modified to remove 
> For the `images.yaml` you can use https://github.com/kubernetes/k8s.io/blob/master/k8s.gcr.io/images/k8s-staging-apisnoop/images.yaml which has an empty file. You still need an empty file.

I have created an empty file. Will add some text indicating this is an empty file, so the diff shows up. Thanks!

listx: Awesome, thanks for the clarifications.

listx:
@thockin Do you have hany reservations about naming conventions here?

listx:
@prameshj I leave it up to you whether you think this should be split up into 2 stagings instead of 1.

/lgtm

prameshj:
> "dns" seems somewhat vague. I would suggest breaking it down, but I'm not sure it's really worthwhile. The fact that kube-dns and the cache share a repo is maybe unfortunate. If we had 2 stagings, you could post-submit the same repo twice into two GCBs and two pushes to different staging and thus 2 subdirs of prod.
> 
> The git repo is "dns" so I guess this is fine, unless we have a more general name we like. "dns-tools"? "dns-utils"? Neither is better, IMO.

I agree "dns" is vague. I used that name since the repo name is "dns" also. I was hoping to avoid pushing to 2 different staging repos since the images are being built by a single workflow today.

prameshj:
@thockin Is it ok to keep a single name 'dns' for the repo? Since it is the same workflow to push the images, I am trying to keep things simple. please approve or let me know if i can make more changes here. Thanks!

thockin:
Thanks!

/lgtm
/approve

PR #691, opened 2020-03-24
Promote CoreDNS images 1.6.7 and 1.6.9

stp-ip:


stp-ip:
Sidenote:

General convention seems to be "vX.Y.Z" that's what we are using in the promoter.
The image tags pushed to the staging repo are of the format "X.Y.Z" similar to the coredns tags on docker hub.

We don't have specific version conventions on tags.
Adding in "X.Y.Z" later on could always be done in my view.

PR #694, opened 2020-03-25
Fix Joel's email, add Marky and Josh to moderators

castrojo:
Signed-off-by: Jorge O. Castro <jorgec@vmware.com>


spiffxp:
Changes are live

Issue #340, opened 2019-08-21
k8s.io redirector should document itself


verb:
I can never remember the short links. There should be a short link for listing the short links, something like https://go.k8s.io/help or a bare https://go.k8s.io/.

It would probably be easiest to just link to https://github.com/kubernetes/k8s.io/tree/master/k8s.io

I'm willing to do this work, just opening this issue for coordination and feedback.

bartsmykla:
Hi @verb. I like the idea, do you still want to do it? :-)

verb:
Hi @bartsmykla, I've had trouble finding time to do this, so a few days ago I resolved to add this to the list of good first issues... which I've also had trouble finding time to do. 😀 

Are you interested in taking this over?

bartsmykla:
I'm doing some research for issues which we can quickly implement and close. I'm gonna create PR soon :-)

bartsmykla:
Done

PR #702, opened 2020-03-30
Added go.k8s.io/redirects redirection

bartsmykla:
Closes https://github.com/kubernetes/k8s.io/issues/340

go.k8s.io/redirects will point to the list of recirections inside README

### DIFF
--- a/k8s.io/README.md
+++ b/k8s.io/README.md
@@ -48,6 +48,7 @@ Redirections
 - https://go.k8s.io/owners/${GITHUB_USER}
 - https://go.k8s.io/partner-request
 - https://go.k8s.io/pr-dashboard
+- https://go.k8s.io/redirects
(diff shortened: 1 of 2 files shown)


stp-ip:
Looks like a good first iteration for #340.

We can wait for the full generate list from redirect entries once we have the redirector in place.

One note. I think /help might be too broad. How about /redirects?

nikhita:
> I think /help might be too broad. 

I agree, might end up being confusing since we have `/help-wanted` too. :thinking:

bartsmykla:
@nikhita what do you think of @stp-ip suggestion about `/redirects`?

nikhita:
`/redirects` sounds good to me :)

bartsmykla:
@stp-ip @nikhita I updated the PR. Can you look now? :-)

PR #708, opened 2020-03-31
Empower the backup-test SA to write to backup-prod

thockin:
The "prod" GCR backup syncs from $PROD_PROJECT into $PRODBAK_PROJECT.

The backup system itself has tests, which backup from
$GCR_BACKUP_TEST_PROD_PROJECT into $GCR_BACKUP_TEST_PRODBAK_PROJECT.

This commit allows the ServiceAccount which does backups to ALSO write
to the prod project, which is not what we want for "real prod" but makes
the bacup testing easier.

### DIFF
--- a/infra/gcp/ensure-prod-storage.sh
+++ b/infra/gcp/ensure-prod-storage.sh
@@ -191,6 +191,16 @@ empower_service_account_for_cip_auditor_e2e_tester \
     $(svc_acct_email "${GCR_AUDIT_TEST_PROD_PROJECT}" "${PROMOTER_SVCACCT}") \
     "${GCR_AUDIT_TEST_PROD_PROJECT}"
 
+# Special case: grant the GCR backup-test svcacct access to the "backup-test
+# prod" project (which models the real $PROD_PROJECT) so it can write the
+# source images and then execute tests of the backup system.  This svcacct
+# already has access to the "backup-test prod backup" project (which models the
+# real $PRODBAK_PROJECT).  We don't want this same power for the non-test
+# backup system, so a compromised promoter can't nuke backups.
+empower_svcacct_to_write_gcr \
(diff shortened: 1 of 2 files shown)


infra/gcp/ensure-prod-storage.sh
```diff
@@ -191,6 +191,16 @@ empower_service_account_for_cip_auditor_e2e_tester \
...
     "${GCR_AUDIT_TEST_PROD_PROJECT}"
 
+# Special case: grant the GCR backup-test svcacct access to the "backup-test
+# prod" project (which models the real $PROD_PROJECT) so it can write the
+# source images and then execute tests of the backup system.  This svcacct
+# already has access to the "backup-test prod backup" project (which models the
+# real $PRODBAK_PROJECT).  We don't want this same power for the non-test
+# backup system, so a compromised promoter can't nuke backups,
+empower_svcacct_to_write_gcr \
+    $(svc_acct_email "${GCR_BACKUP_TEST_PRODBAK_PROJECT}" "${PROMOTER_SVCACCT}") \
```
listx, line 201: It would be better to rephrase this to be simpler but I can't think of a better way at the moment (also, s/,/. at the end).

I think this means I can use `k8s-infra-gcr-promoter@k8s-gcr-backup-test-prod-bak.iam.gserviceaccount.com` for the backup tests (authenticate as this for workload identity) --- right?

listx:
Sorry, just read your email response.

/lgtm

infra/gcp/ensure-prod-storage.sh
thockin: Yes,  the prod-bak SA is the "right" one to use and it has special-case powers to the prod GCR

thockin:
Errant comma removed.

thockin:
Too slow on the hold.

PR #718, opened 2020-04-04
allow workload identity for prod backups

listx:
This empowers the `k8s-infra-gcr-promoter-bak` KSA in the `test-pods`
K8s namespace in the `k8s-prow` GCP Project (where the Prow trusted
cluster lives) to authenticate as
`k8s-infra-gcr-promoter@k8s-artifacts-prod-bak.iam.gserviceaccount.com`.

The `k8s-infra-gcr-promoter-bak` KSA does not exist yet and will be
created when we re-introduce the backup job (pulled on 2020-03-18 due to
quota issues). The backup job itself was optimized in
https://github.com/kubernetes/k8s.io/pull/677.


spiffxp:
```
$ time ./infra/gcp/ensure-prod-storage.sh
# ...
Done

real	10m52.116s
user	5m27.995s
sys	1m51.737s
```

PR #719, opened 2020-04-05
Promote agnhost 2.13

fedepaol:
This PR is the second step to promote & use agnhost 2.13, to address https://github.com/kubernetes/kubernetes/issues/89782


claudiubelu:
The agnhost image version has been bumped, the image building job was succesful, and the manifest digest is correct. Looks promotable.

/lgtm

PR #720, opened 2020-04-06
Added new members to community group

rajula96reddy:
Added new members as discussed in the contrib-ex marketing call

cc: @parispittman

### DIFF
--- a/groups/groups.yaml
+++ b/groups/groups.yaml
@@ -38,6 +38,10 @@ groups:
     members:
       - dgiles@linuxfoundation.org
       - jberkus@redhat.com
+      - chris@chrisshort.net 
+      - jrosland@vmware.com
+      - samudralavamshi@gmail.com
+      - matthewbbroberg@gmail.com
 
   - email-id: conduct@kubernetes.io
     name: conduct
--- a/groups/groups.yaml
+++ b/groups/groups.yaml
@@ -35,13 +35,13 @@ groups:
     managers:
       - pal.nabarun95@gmail.com
       - ameukam@gmail.com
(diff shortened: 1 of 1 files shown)


spiffxp:
/ok-to-test
/approve
/cc @parispittman 
can you confirm with an /lgtm

spiffxp:
@justaugustus Yeah I hear you, we've cut out github team managers to stop manual changes, and we already have ownerless/managerless groups for k8s-infra staging, why not do the same here.  

That said, it's not clear to me at this time whether owners or managers are needed, I could see the need if certain permissions that this tool doesn't reconcile are locked to those roles. I think some outreach is needed before throwing the hammer down.

If you want to give it a go with release-managers I'm in favor of it.

justaugustus:
@spiffxp -- Hmmmm, the first thing that comes to mind as they're mail-enabled groups is moderation.
Should all members be able to moderate messages? 

Anywho, I don't want to distract from this PR. Will open an issue about this.

parispittman:
@spiffxp @justaugustus - all of our managers should be able to moderate messages while the members should only view. it's not public email as individual matters, but not coc, are discussed here. will the new set up mean that everyone listed can view only and/or moderate all the things no matter what?

spiffxp:
@parispittman I think this isn't the place to try and figure out a new owners/manager/members policy, if these changes look good as-is /lgtm and /hold cancel

parispittman:
sure just to keep this going and if there is an issue we will PR in a change. if the folks listed read this, test it; i'll reach out individually. 

/lgtm

spiffxp:
ran `make run -- --confirm` in `/groups`

PR #725, opened 2020-04-07
Add a staging repo for cluster-proportional-vertical-autoscaler

MrHohn:
I largely followed https://github.com/kubernetes/k8s.io/tree/master/k8s.gcr.io#creating-staging-repos and https://github.com/kubernetes/k8s.io/pull/689.

This is to unblock a new release on CPVA: https://github.com/kubernetes-incubator/cluster-proportional-vertical-autoscaler/issues/33

/cc @bowei @thockin

### DIFF
--- a/groups/groups.yaml
+++ b/groups/groups.yaml
@@ -687,6 +687,18 @@ groups:
       - jbelamaric@google.com
       - cohaver@infoblox.com
 
+  - email-id: k8s-infra-staging-cpva@kubernetes.io
+    name: k8s-infra-staging-cpva
+    description: |-
+      ACL for staging cluster-proportional-vertical-autoscaler image.
+    settings:
+      ReconcileMembers: "true"
+    members:
+      - bowei@google.com
+      - dnardo@google.com
+      - thockin@google.com
+      - zihongz@google.com
+
   - email-id: k8s-infra-staging-csi@kubernetes.io
     name: k8s-infra-staging-csi
     description: |-
--- a/infra/gcp/ensure-staging-storage.sh
+++ b/infra/gcp/ensure-staging-storage.sh
@@ -54,6 +54,7 @@ STAGING_PROJECTS=(
     capi-docker
     capi-vsphere
     coredns
+    cpva
     csi
     csi-secrets-store
     descheduler
--- a/k8s.gcr.io/images/k8s-staging-cpva/OWNERS
+++ b/k8s.gcr.io/images/k8s-staging-cpva/OWNERS
@@ -1,0 +1,8 @@
+# See the OWNERS file documentation:
+#  https://github.com/kubernetes/kubernetes/blob/master/docs/devel/owners.md
+
+approvers:
+  - thockin
+  - dnardo
+  - MrHohn
+  - bowei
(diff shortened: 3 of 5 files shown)


listx:
Only nit I have here is regarding the naming --- "cpva" sounds like it's the actual image name, not the name of the subproject responsible for building the image. It might be that you could name it more generically, or maybe not. Wdyt?

MrHohn:
> Only nit I have here is regarding the naming --- "cpva" sounds like it's the actual image name, not the name of the subproject responsible for building the image. It might be that you could name it more generically, or maybe not. Wdyt?

Thanks for the feedback. I mistakenly thought it has to be a one-to-one mapping between project and image. We can probably rename it to "cpa" so that it can handle the image for both [cpa](https://github.com/kubernetes-incubator/cluster-proportional-autoscaler) and [cpva](https://github.com/kubernetes-incubator/cluster-proportional-vertical-autoscaler)? I find it hard to come up with a meaningful name without exceeding the 18 chars limit while not conflicting the actual [k8s autoscaler project](https://github.com/kubernetes/autoscaler
) (different than cpa or cpva)...

listx:
> > Only nit I have here is regarding the naming --- "cpva" sounds like it's the actual image name, not the name of the subproject responsible for building the image. It might be that you could name it more generically, or maybe not. Wdyt?
> 
> Thanks for the feedback. I mistakenly thought it has to be a one-to-one mapping between project and image. We can probably rename it to "cpa" so that it can handle the image for both [cpa](https://github.com/kubernetes-incubator/cluster-proportional-autoscaler) and [cpva](https://github.com/kubernetes-incubator/cluster-proportional-vertical-autoscaler)? I find it hard to come up with a meaningful name without exceeding the 18 chars limit

TIL there is a limit --- do you have a link?

I agree "cpa" is better than "cpva"

> while not conflicting the actual [k8s autoscaler project](https://github.com/kubernetes/autoscaler) (different than cpa or cpva)...

thockin:
It should have been "cpvpa" to go with "vpa".  Given that the repo is cpva, I'll let this go.

thockin:
Thanks!

/lgtm
/approve

MrHohn:
Thanks both for reviewing!

> TIL there is a limit --- do you have a link?
>
> I agree "cpa" is better than "cpva"

I saw the limit from https://github.com/kubernetes/k8s.io/tree/master/k8s.gcr.io#creating-staging-repos: "The project name can have a maximum of 18 characters."

I will send a follow up to fix-up the name.

thockin:
I think you are fine on name length?

listx:
Ah IIRC there is a project name max limit. So anything beyond 18 chars will result in a failure to create the GCP project. I recall this from the time @thockin tried to create `-backup` project names but ended up going with `-bak` to fit into the name limit.

MrHohn:
> I think you are fine on name length?

Sorry I was referring to cpva v.s. cpa (was hoping to have one group that handles both). Length wide we are fine.

spiffxp:
FYI: I just reconciled groups and picked up the changes from this PR

PR #729, opened 2020-04-07
Rename project cpva to cpa

MrHohn:
...and put promoter-manifest.yaml to the correct location.

This is a follow up of https://github.com/kubernetes/k8s.io/pull/725
/assign @listx


MrHohn:
Re https://github.com/kubernetes/k8s.io/pull/725#issuecomment-610650037, @spiffxp I guess it may be too late now to rename the group? If so I can revert this PR to only fix the promoter-manifest.yaml path.

spiffxp:
I think a rename of the group will work, but I'm less clear whether all of the other staging resources will be auto-deleted, let me take a look

listx:
Belated /lgtm from me

spiffxp:
groups reconciled, now to reconcile the other gcp resources

spiffxp:
Finished run of `./infra/gcp/ensure-staging-storage.sh cpa`

MrHohn:
Thank you @spiffxp!

spiffxp:
I guess the resources didn't get created under the old name, since I was unable to delete what the old project would have been named
```$ gcloud projects delete k8s-staging-cpva
ERROR: (gcloud.projects.delete) User [spiffxp@google.com] does not have permission to access project [k8s-staging-cpva] (or it may not exist): The caller does not have permission```

PR #730, opened 2020-04-07
Update audit as of 2020-04-07

spiffxp:
Specifically after reconciling as of 4b2a6dc

### DIFF
--- a/audit/projects/kubernetes-public/services/container/clusters.txt
+++ b/audit/projects/kubernetes-public/services/container/clusters.txt
@@ -1,2 +1,2 @@
-aaa	us-central1	us-central1-c;us-central1-a;us-central1-f	8	RUNNING
+aaa	us-central1	us-central1-c;us-central1-a;us-central1-f	6	RUNNING
 development2	us-central1-b	us-central1-b	1	RUNNING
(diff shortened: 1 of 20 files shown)


audit/projects/kubernetes-public/services/container/clusters.txt
```diff
@@ -1,2 +1,2 @@
-aaa	us-central1	us-central1-c;us-central1-a;us-central1-f	8	RUNNING
+aaa	us-central1	us-central1-c;us-central1-a;us-central1-f	6	RUNNING
```
spiffxp, line 1: looks like we shrank the node count for aaa? but haven't adjusted terraform files at all?

spiffxp:
/cc @thockin @bartsmykla 
since I missed the meeting this morning where you probably made some of the changes in https://github.com/kubernetes/k8s.io/pull/730/commits/d8ead609763fe0cb6f1973eea741a02893795478

thockin:
Cluster autoscaling to the rescue!

audit/projects/kubernetes-public/services/container/clusters.txt
bartsmykla: It's just a result of autoscalling I think. We didn't change anything particular during the meeting. Just one service account was created for me playing with some automation testing. I'm gonna give you some quick update about what we were discussing at a meeting, but please give me few hours to finish other stuff today.

bartsmykla:
I'm curious though about what we should do with the older project ([`4b2a6dc`](https://github.com/kubernetes/k8s.io/commit/4b2a6dcee90d92ec56f32629102dbd5f6032d840)). Doesn't we currently have unmanaged by our reconciliation script google group?

spiffxp:
The scripts don't delete what they don't know about.  I tried manually deleting the project, and was going to use audit to see if I'd missed anything.  But it turned out there was nothing to delete (ref: https://github.com/kubernetes/k8s.io/pull/729#issuecomment-610659710), and this audit picked up no detritus.

PR #734, opened 2020-04-08
Added missing annotation to cert manifests

bartsmykla:
During perfdash deployment we found with munnerz
if there is no existing secret for certificate,
the process for obtaining it will fail,
so if we want to issue the new certificate
we have to add annotation:
[`cert-manager.io/issue-temporary-certificate: "true"`](https://cert-manager.io/docs/usage/certificate/#temporary-certificates-whilst-issuing)
If somehow we would delete existing secrets
they wouldn't be recreated without this annotation.

/cc @munnerz 


munnerz:
lgtm 😄 

/hold cancel

PR #741, opened 2020-04-10
Promote etcd 3.4.7 staging images

jingyih:
Ref: https://github.com/kubernetes/kubernetes/pull/89895

### DIFF
--- a/k8s.gcr.io/images/k8s-staging-etcd/images.yaml
+++ b/k8s.gcr.io/images/k8s-staging-etcd/images.yaml
@@ -1,1 +1,7 @@
-# No images yet
+- name: etcd
+  dmap:
+    "sha256:80d539b791923b37fb501166729307c987a237bf9a72cef3dbfca24744b5cb76": ["3.4.7-0", "3.4.7"]
+- name: etcd-empty-dir-cleanup
+  dmap:
+    "sha256:f805272b4422efa92e20789295c343ed26ff36ddc4f70eeb5d7890d6b758b0fc": ["3.4.7.0"]
+


jingyih:
cc @YoyinZyc

bartsmykla:
Just a question to @listx because I'm not sure and I can't find information about architecture spsecific builds. Should we explicitly put them in manifests?

k8s.gcr.io/images/k8s-staging-etcd/images.yaml
```diff
@@ -1 +1,22 @@
-# No images yet
+- name: etcd
+  dmap:
+    "sha256:80d539b791923b37fb501166729307c987a237bf9a72cef3dbfca24744b5cb76": ["3.4.7-0", "3.4.7"]
+- name: etcd-amd64
```
dims, line 4: etcd-amd64, etcd-arm, etcd-arm64, etcd-ppc64le, etcd-s390x are all listed in the manifest for the `etcd` image:
https://console.cloud.google.com/gcr/images/k8s-staging-etcd/GLOBAL/etcd@sha256:80d539b791923b37fb501166729307c987a237bf9a72cef3dbfca24744b5cb76/details?tab=info

So we don't need to individually list them here in this yaml file.

jingyih: Thanks! I have a question on backward compatibility. If we only list `etcd` in this yaml, and user later ask for ` k8s.gcr.io /etcd-amd64:3.4.7`, will it work?

dims: @jingyih yep. that will work

jingyih: Thanks! PR updated.

jingyih:
Hi @dims, I noticed the new images for etcd 3.4.7 are pushed to `k8s-artifacts-prod/etcd/etcd`, whereas previously the images url are something like `google-containers/etcd`. Is this expected?
https://pantheon.corp.google.com/gcr/images/k8s-artifacts-prod/US/etcd?gcrImageListsize=30

dims:
@jingyih there's a transition in progress, please ping @listx on when it may be available for us to switch over the PR that uses the new image.

listx:
> Hi @dims, I noticed the new images for etcd 3.4.7 are pushed to `k8s-artifacts-prod/etcd/etcd`, whereas previously the images url are something like `google-containers/etcd`. Is this expected?
> https://pantheon.corp.google.com/gcr/images/k8s-artifacts-prod/US/etcd?gcrImageListsize=30

Yes, this is working as intended. We originally commenced the flip on [April 1](https://groups.google.com/forum/#!topic/kubernetes-sig-release/ew-k9PEBckQ), but it was rolled back due to production issues. I'll be sending out a wider comms there today; here is an internal [status doc](http://go/vdf). See https://github.com/kubernetes/k8s.io/blob/master/k8s.gcr.io/Vanity-Domain-Flip.md for a general overview.

For now if you must update to k8s.gcr.io/etcd, use the [internal promoter](https://source.corp.google.com/piper///depot/google3/cloud/kubernetes/releng/registry_promoter/manifest/k8s-addon-container_manifest.yaml).

PR #520, opened 2019-12-17
Set Up Publishing-Bot

ameukam:
Part of #503.

Configs and deployments files are from Publishing-Bot [repo](https://github.com/kubernetes/publishing-bot/tree/master/artifacts).

/assign @dims @nikhita

PR #743, opened 2020-04-14
Added ssd storage class to aaa cluster conf

bartsmykla:
Applies to [discussion](https://github.com/kubernetes/k8s.io/pull/520/files#r379147033)

As we are close to deploy publishing bot at `aaa`
we need this storage class and it will be created
when provisioning `aaa` cluster via terraform now

/cc @ameukam @nikhita @sttts @thockin 

### DIFF
--- a/infra/gcp/clusters/kubernetes-public/aaa/10-cluster-configuration.tf
+++ b/infra/gcp/clusters/kubernetes-public/aaa/10-cluster-configuration.tf
@@ -158,3 +158,26 @@ resource "google_container_cluster" "cluster" {
     enabled = true
   }
 }
+
+
+provider "kubernetes" {
+  host                   = "https://${google_container_cluster.cluster.endpoint}"
+  client_certificate     = base64decode(google_container_cluster.cluster[0].client_certificate)
+  client_key             = base64decode(google_container_cluster.cluster[0].client_key)
+  cluster_ca_certificate = base64decode(google_container_cluster.cluster[0].cluster_ca_certificate)
+  config_context         = "gke_kubernetes-public_us-central1_aaa"
+  load_config_file       = "true"
+}
+
+// Create ssd storage class (initialy for Publishing Bot)
+resource "kubernetes_storage_class" "ssd" {
+  metadata {
+    name = "ssd"
+  }
+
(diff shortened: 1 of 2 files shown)


ameukam:
@bartsmykla The [provider kubernetes](https://www.terraform.io/docs/providers/kubernetes/guides/getting-started.html) required for this resource seems to be missing in the repo.

bartsmykla:
@ameukam you are right! I was a little to fast with the PR. Thank you for catching that! :-)

infra/gcp/clusters/kubernetes-public/aaa/10-cluster-configuration.tf
```diff
@@ -158,3 +158,18 @@ resource "google_container_cluster" "cluster" {
     enabled = true
   }
 }
+
+provider "kubernetes" {}
```
ameukam, line 162: GKE cluster `aaa` auth. creds need to be passed to the provider. A good example : 
https://www.terraform.io/docs/providers/kubernetes/index.html#statically-defined-credentials.

bartsmykla: Yes, I forgot to change the title to `[WIP]` cause I'm gonna start discussion about it tomorrow. We currently are running terraform locally. With kubernetes provider we would have to use google console only because ingress traffic to `aaa` is restricted to locale one only.

spiffxp:
I am back to not being sold on the necessity of terraform to deploy a `StorageClass` resource.  There's no other GCP / infra-level twiddling that's happening here, we're just deploying what could effectively be a .yaml file.

ameukam:
a storageclass `ssd` is now created on aaa. Done during the publishing-bot's migration.

spiffxp:
I'm getting a little more comfortable with terraform, and am now less opposed to using terraform to deploy kubernetes resources, but still think we should be using yaml files as the source

I leave it to @bartsmykla whether to keep this open and iterate, or close for now since the immediate need has been satisfied

bartsmykla:
Let's close it for now and get back in the future if there'll be need

/close

PR #746, opened 2020-04-15
Removed duplicated perfdash manifests

bartsmykla:
/cc @thockin @mm4tt 

- To make it easier to maintain when the new version of perfdash
  will be released and not have to update manifests in two places
  I removed deployment and service manifests from this repo
  and changed the indstruction about how to apply manifests
  from perfdash repository.
- As perfdash is still not fully moved to the `aaa` cluster
  (some issues with permissions to perfdash namespace)
  I didn't want to change the type of service from `LoadBalancer`
  to `NodePort` in source repository, so instead anyone
  who will be deploying perfdash to the `aaa` cluster have to
  dynamicaly change the type (I have updated command which do so
  in README). It will be changed when we'll switch subdomain
  perf-dash.k8s.io to point to `aaa`.
- I didn't like the hardcoded namespace names, so I removed it
  and put the namespace in CLI arguments in the README
- I will followup with the PR adding comments to the source
  perfdash repository with informations we are using these
  two files here, so they should be careful with moving them around.


perf-dash.k8s.io/README.md
```diff
@@ -1,8 +1,19 @@
...
-kubectl apply -f certificate.yaml
-kubectl apply -f deployment.yaml
-kubectl apply -f ingress.yaml
-kubectl apply -f service.yaml
+PERFDASH_REPO_URL="https://raw.githubusercontent.com/kubernetes/perf-tests/master/perfdash"
+PERFDSAH_MANIFEST_URL_DEPLOYMENT="${PERFDASH_REPO_URL}/deployment.yaml"
+PERFDASH_MANIFEST_URL_SERVICE="${PERFDASH_REPO_URL}/perfdash-service.yaml"
+PERFDASH_NAMESPACE="perfdash"
+
+kubectl apply -n "$PERFDASH_NAMESPACE" -f "$PERFDSAH_MANIFEST_DEPLOYMENT"
```
dims, line 9: DSAH -> DASH

bartsmykla: Thank you for catching that and sorry for mistake 😳

bartsmykla: Done

mm4tt:
/lgtm

Thanks! I love de-deduplicating the manifests and keeping them in perf-tests repo, it simplifies pushing new versions of perf-dash

PR #750, opened 2020-04-15
Add cluster resources for slack-infra

spiffxp:
Basically a mirror of sigs.k8s.io/slack-infra/cluster/* minus the managed certs and ingress in root

Secrets encrypted with git-crypt

/assign @bartsmykla


spiffxp:
Copy-pasting resources over here seems like a brute-force approach, prone to drift if that repo sees further development.  As a followup it might be worth investigating an approach that reuses what's there, and patches in what we need.

bartsmykla:
I think we can iterate as I did with perf-dash ([#746](https://github.com/kubernetes/k8s.io/pull/746))

spiffxp:
ref: https://github.com/kubernetes/k8s.io/issues/751

Issue #751, opened 2020-04-15
Migrate slack-infra services to aaa cluster


spiffxp:
- origin gcp-project: kubernetes-tools
- origin cluster-name: tools-cluster
- apps:
  - [ ] slackin (slack.k8s.io)
  - [ ] slack-event-log
  - [ ] slack-moderator
  - [ ] slack-welcomer
- repo: https://github.com/kubernetes-sigs/slack-infra

/wg k8s-infra
/area cluster-infra
/assign @bartsmykla @ameukam

nikhita:
fyi @Katharine @BenTheElder @jeefy @mrbobbytables

BenTheElder:
there's also the tempellis image hosting, we should move that to a staging project probably.

bartsmykla:
My step by step plan to move slack infra to the `aaa` cluster:

1. Creating PR with updated instructions about [how to deploy the `slack-infra`](/slack-infra#todo-how-to-deploy) and two missing resource manifests (ingress and certificate)
2. Deploying ingress resource to receive efemeral IP address
3. Updating DNSes with temporary domains (slack-staging.k8s.io and slack-staging.kubernetes.io) pointing to the address from the point above
4. Asking a person with permission (@thockin probably) to mark received, efemeral address as static (with name: `slack-infra-ingress-prod`)
5. Updating the ingress resource created in point 1 with annotation: `kubernetes.io/ingress.global-static-ip-name: slack-infra-ingress-prod`
6. Asking one of the people who have access to encrypted secrets (@dims, @spiffxp, @thockin) to deploy them to the `slack-infra` namespace:
```shell
# Commands needs to be run from the slack-infra directory
# Secrets
    git-crypt unlock

    # slack-event-log
    kubectl apply -n slack-infra -f slack-event-log/slack-event-log-config-secret.yaml

    # slack-moderator
    kubectl apply -n slack-infra -f slack-moderator/slack-moderator-config-secret.yaml

    # slack-welcomer
    kubectl apply -n slack-infra -f slack-welcomer/slack-welcomer-config-secret.yaml

    # slackin
    kubectl apply -n slack-infra -f slackin/recaptcha-secret.yaml
    kubectl apply -n slack-infra -f slackin/slackin-token-secret.yaml

    git-crypt lock
```
7. Figuring out if deploying `slack-event-log`, `slack-moderator` and `slack-welcomer` without first turning off the already existing instances is possible
   - [if not]:
[...]

PR #757, opened 2020-04-16
Added RBAC groups and namespace for slack infra

bartsmykla:
/cc @dims @ameukam 

### DIFF
--- a/groups/groups.yaml
+++ b/groups/groups.yaml
@@ -143,6 +143,10 @@ groups:
       - k8s-infra-rbac-k8s-io-prod@kubernetes.io
       - k8s-infra-rbac-perfdash@kubernetes.io
       - k8s-infra-rbac-publishing-bot@kubernetes.io
+      - k8s-infra-rbac-slack-event-log@kubernetes.io
+      - k8s-infra-rbac-slack-moderator@kubernetes.io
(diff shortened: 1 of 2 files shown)


stp-ip:
looks good from a code perspective and what we discussed in the call aka keeping separate namespaces.

groups/groups.yaml
```diff
@@ -405,6 +409,57 @@ groups:
...
+
+  # Collective group for RBAC groups for slack infrastructure
+  # created to not duplicate members
+  - email-id: k8s-infra-slack-infra-admins@kubernetes.io
+    name: k8s-infra-slack-infra-admins
+    description: |-
+      ACL for Slack Infra Admins
+    settings:
+      ReconcileMembers: "true"
+    members:
```
ameukam, line 456: @spiffxp is not missing in this list ?

bartsmykla: Right, sorry :-D

bartsmykla: Fixed

spiffxp:
ref: https://github.com/kubernetes/k8s.io/issues/751

spiffxp:
FWIW I would have been fine if all this ran in a "slack-infra" namespace instead of giving each deployment/service its own namespace. It all currently runs in the default namespace of its cluster.

PR #758, opened 2020-04-16
Fixed missing var from fn in ensure_namespaces

bartsmykla:
/cc @thockin @mm4tt 

I think I found the place which was causing problems with accessing the namespaces. :-)


mm4tt:
Nice!

/lgtm

Let me know when the change will be applied so I can check whether it fixes the problem

bartsmykla:
@mm4tt it's already applied

bartsmykla:
I think I was to quik with my conclusion as it is still not working...

PR #768, opened 2020-04-20
groups: add SIG-Storage members to k8s-infra-staging-csi

pohly:
Right now, none of the people involved with SIG-Storage have access to
logs when staging
fails (https://github.com/kubernetes/test-infra/pull/16942#issuecomment-613151134). As
suggested in that
discussion (https://github.com/kubernetes/test-infra/pull/16942#issuecomment-616586275),
I am adding the current SIG chairs and tech leads plus myself as
someone who has been working on image publishing.

/cc @saad-ali @msau42 @xing-yang @jsafrane

### DIFF
--- a/groups/groups.yaml
+++ b/groups/groups.yaml
@@ -800,6 +800,11 @@ groups:
       ReconcileMembers: "true"
     members:
       - davanum@gmail.com
+      - saadali@google.com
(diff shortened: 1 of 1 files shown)


spiffxp:
Ran `make run -- --confirm` in `groups`

PR #770, opened 2020-04-21
Allow RBAC group members to view members

thockin:
This is required for RBAC to work.  It used to be true by default, but
it was hanged late last year.  Since then, the people using the cluster
have mostly have cluster-admin access.  I reproduced it by adding my own
gmail to the group and then Mike Danese pointed out this setting was
missing.

### DIFF
--- a/groups/groups.yaml
+++ b/groups/groups.yaml
@@ -136,6 +136,7 @@ groups:
       Security Groups for GKE clusters
     settings:
       ReconcileMembers: "true"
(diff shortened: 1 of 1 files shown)


bartsmykla:
@thockin thank you for fixing this! :-)

Did you reconcilled groups maybe?

mm4tt:
I've just checked and everything works! Thanks @thockin for fixing it!

thockin:
This is all applied already, yes.

PR #778, opened 2020-04-21
Updated deploy instructions for perf-dash

bartsmykla:
When https://github.com/kubernetes/perf-tests/pull/1184 will be merged
there won't be need to manually changing type of service
from LoadBalancer to NodePort so the docs should reflect that

Hold till https://github.com/kubernetes/perf-tests/pull/1184 will be merged
/hold


bartsmykla:
As kubernetes/perf-tests#1184 was merged, I'm unholding this PR

/unhold

Issue #726, opened 2020-04-07
audit: strip out resource usage from project-info.json


spiffxp:
We want to gather limits, but gathering usage results in noise

ref: https://github.com/kubernetes/k8s.io/pull/714#discussion_r404368371

bartsmykla:
I'm assigning myself to this task but I'm open if anyone would like to shadow me and do some pair-programming session solving this. :-)

/assign

PR #780, opened 2020-04-22
Strip out resource usage from audit files

bartsmykla:
Closes: #726 

Stripping out resource usage from `project-info.json` files with audit reflecting the changes.

/cc @thockin @spiffxp @puerco


bartsmykla:
I tried to split some of the files into different commits

bartsmykla:
Actually I'm gonna split this into two Pull Requests

PR #784, opened 2020-04-22
slack-infra added manifests + updated readme

bartsmykla:
/cc @spiffxp @ameukam 

ref. #751 

- Updated slack-infra readme with deploy instructions
- Added missing certificate.yaml and ingress.yaml manifests


slack-infra/README.md
```diff
@@ -12,6 +12,47 @@ Secrets are stored as `{app}/{secret-name}-secret.yaml` and encrypted with git-c
...
+
+    # slack-moderator
+    kubectl apply -n slack-infra -f slack-moderator/slack-moderator-config-secret.yaml
+
+    # slack-welcomer
+    kubectl apply -n slack-infra -f slack-welcomer/slack-welcomer-config-secret.yaml
+
+    # slackin
+    kubectl apply -n slack-infra -f slackin/recaptcha-secret.yaml
+    kubectl apply -n slack-infra -f slackin/slackin-token-secret.yaml
```
spiffxp, line 53:
```
for f in ./*/*-secret.yaml; do kubectl apply -n slack-infra -f $f; done
```

slack-infra/README.md
```diff
@@ -12,6 +12,47 @@ Secrets are stored as `{app}/{secret-name}-secret.yaml` and encrypted with git-c
...
+    kubectl apply -n slack-infra -f slack-moderator/service.yaml
+
+    # slack-welcomer
+    kubectl apply -n slack-infra -f slack-welcomer/deployment.yaml
+    kubectl apply -n slack-infra -f slack-welcomer/message.yaml
+    kubectl apply -n slack-infra -f slack-welcomer/service.yaml
+
+    # slackin
+    kubectl apply -n slack-infra -f slackin/deployment.yaml
+    kubectl apply -n slack-infra -f slackin/service.yaml
```
spiffxp, line 37: ```for f in ./*/*.yaml; do echo $f | grep -q secret || kubectl apply -n slack-infra -f $f; done```

bartsmykla: I wanted to be more explicit and treat it as some kind of documentation too easily see what are gonna to deploy exactly, but I don't have strong opinion about it. I'm gonna update the PR

ameukam:
I would like to suggest that we iterate later with Makefiles. I feel like they can provide simplicity with the deployments.

bartsmykla:
@spiffxp I liked the suggestion of having two directories for the resources so I made the changes and updatej the instruction.

@spiffxp do you think maybe now it's unnecessary to have that explicit names of secrets as they are in `secrets` directory? I mean maybe we could rename them as for example:
```shell
secrets/slack-moderator/slack-moderator-config-secret.yaml -> secrets/slack-moderator/config-secret.yaml
```
or even:
```shell
secrets/slack-moderator/slack-moderator-config-secret.yaml -> secrets/slack-moderator/config.yaml
```
We would have to change the readme from:
```
Secrets are stored as `secrets/{app}/{secret-name}-secret.yaml` and encrypted with git-crypt
```
to:
```
Secrets are stored as `secrets/{app}/{secret-name}.yaml` and encrypted with git-crypt
```
and update the [`.gitattributes`](/kubernetes/k8s.io/blob/master/.gitattributes#L2), from:
```
slack-infra/**/*secret*.yaml filter=git-crypt diff-git-crypt
```
to:
```
slack-infra/secrets/**/*secret*.yaml filter=git-crypt diff-git-crypt
```

ameukam:
@bartsmykla Can you add the ConfigMap `bad-domains` in this PR : https://gist.github.com/spiffxp/e94803079d9556efe28a317f87757f1c#file-bad-domains-configmap-yaml ?

bartsmykla:
Yes, I will do it in a minute

bartsmykla:
@ameukam @spiffxp updated with `bad-domains` configmap

ameukam:
I'll suggest to disable HTTP for the ingress : https://cloud.google.com/kubernetes-engine/docs/concepts/ingress-xlb#disabling_http.

bartsmykla:
@ameukam I added this topic to our todays agenda. I know you can't attend but I'll write you down the results of our discussion

munnerz:
@ameukam I don't think we do want to disable HTTP, else any user accessing the site *without* HTTPS will not be redirected to the HTTPS version. Instead, we should ensure that we redirect users connecting on port 80 *to* 443/HTTPS (which I don't think GCLBs natively support, so will need to be done within our application or some other intermediary proxy).

bartsmykla:
@munnerz I saw somewhere it's in the alpha (http->https redirections) so maybe we can ask to test it. Let's discuss it today

Issue #783, opened 2020-04-22
DNS Update automation


bartsmykla:
As I started working on automating the process of running octodns whenever [DNS configuration](https://github.com/kubernetes/k8s.io/tree/master/dns/zone-configs) will change here is my step by step plan:

1. [Done: #805, #822, [k/test-infra#17449](https://github.com/kubernetes/test-infra/pull/17449), #827, #828] ~Building consensus around if we want our [octodns container image](/dns/octodns-docker) to be build and managed using [our own image promotion process](/k8s.gcr.io)~
   - ~[If yes]~:
      ~1. [Creating staging repository](/k8s.gcr.io#creating-staging-repos)~
      ~2. [Enabling automatic build](/k8s.gcr.io#enabling-automatic-builds)~
   - ~[if not]: Figure out which image you should use to proceed further~

2. [Done: #949, Closed: #862] ~Modifying current [`dns/push.sh`](https://github.com/kubernetes/k8s.io/blob/master/dns/push.sh) script to be able to run update without docker involved and modifying `make` - ref. [@spiffxp comment](https://github.com/kubernetes/k8s.io/issues/783#issuecomment-618007941)~

3. [Done: #865] ~Ensuring the service account which will be used to run `octodns` exists and have enough permissions to do it's job~

   ~Initially proposed (adding `empower_dns_updater` logic to [lib.sh](/kubernetes/k8s.io/blob/master/infra/gcp/lib.sh), and call the function in [ensure-main-project.sh](/kubernetes/k8s.io/blob/master/infra/gcp/ensure-main-project.sh)):~
   ```bash
   # Ensure the dns_updater service account exists and has the ability to update DNSes.
   # $1: DNS Updater Service Account name
   # $2: The GCP project
   function empower_dns_updater() {
       if [ ! $# -eq 2 -o -z "$1" -o -z "$2" ]; then
           echo "empower_dns_updater(service_account_name, project) requires 2 arguments" >&2
           return 1
       fi
       local service_account_name="$1"
       local project="$2"
       local service_account=$(svc_acct_email "$project" "$service_account_name")

       ensure_service_account "$project" "$service_account_name" "k8s-infra dns updater"

       # Allow dns_updater to adminsiter DNSes.
       gcloud projects add-iam-policy-binding "$project" \
           --member "serviceAccount:$service_account" \
           --role roles/dns.admin
   }
   ```

   ```bash
   PROJECT="kubernetes-public"
[...]

PR #786, opened 2020-04-22
Adding dry-run option when running dns update

### LINKED ISSUE #782: Add dry-run option when running dns update
As it was useful for me whenever I was [performing the DNS update](https://github.com/kubernetes/k8s.io/tree/master/dns#performing-an-update) to run it in `dry-run` mode first and I was always hacking the [`push.sh`](https://github.com/kubernetes/k8s.io/blob/master/dns/push.sh) script to achieve that I think it's the good thing to have a `make dry-run` option.

EricLemieux:
Fixes #782  

Saw the issue had help wanted, and seemed easy enough to implement as someone with little context on the inner workings of this project.

* Added a dry-run make target
* Added an argument to allow for the exclusion of actual deployments

Let me know if a different approach is preferred, or if I need to add anything else.

### DIFF
--- a/dns/push.sh
+++ b/dns/push.sh
@@ -29,6 +29,27 @@ ALL_ZONES=(
     "${PROD_ZONES[@]}"
 )
 
+DRY_RUN=false
+
+function parse_args() {
+  # positional args
+  args=()
+
+  # named args
+  while [ "$#" -gt 0 ]; do
+      case "$1" in
+          -d | --dry-run )          DRY_RUN=true;           ;;
+          * )                       args+=("$1")            # if no match, add it to the positional args
+      esac
+      shift # move to next kv pair
+  done
+
+  # restore positional args
+  set -- "${args[@]}"
+}
+
+parse_args "$@";
+
 # Pushes config to zones.
 #   args: args to pass to octodns (e.g. --doit, --force, a list of zones)
 push () {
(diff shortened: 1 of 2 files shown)


EricLemieux:
CLA signed

EricLemieux:
I signed it

bartsmykla:
Hi @EricLemieux! It's great to see your contribution! Please give me some time to look at it though as I feel a little bit sick today so my brain works on slower pace

dns/push.sh
```diff
@@ -29,6 +29,19 @@ ALL_ZONES=(
...
+while getopts "d" o; do
+  case "${o}" in
+    d)
+      DRY_RUN=true
+      ;;
+    *)
+      ;;
+  esac
+done
+
```
bartsmykla, line 44: I'm not familiar with this kind of arguments parsing but can we do `d | dry-run )` instead of `d`?

That's my way of parsing arguments as a reference (with some more flags/arguments which we'll have to add in the future (ref. #783):

```bash
function parse_args() {
  # positional args
  args=()

  # named args
  while [ "$#" -gt 0 ]; do
      case "$1" in
          -d | --dry-run )          DRY_RUN=true;           ;;
          -f | --from-container )   FROM_CONTAINER=true;    ;;
          -c | --config-file )      CONFIG_FILE="$2";       shift;;
          * )                       args+=("$1")            # if no match, add it to the positional args
      esac
      shift # move to next kv pair
  done

  # restore positional args
  set -- "${args[@]}"

  # set defaults
  if [[ -z "${CONFIG_FILE:-}" ]]; then
      CONFIG_FILE="octodns-config.yaml";
  fi
}

parse_args "$@";
```

bartsmykla:
@EricLemieux just one question from me and everything looks good to me :-)

Adding @thockin as our bash expert to give his opinion
/cc @thockin

EricLemieux:
Hey @bartsmykla,
Thanks for your feedback, I implemented the change that you suggested, the only difference being that I pushed the config file variable default to outside the function rather than checking if it's empty after.
It just made more sense to me to define it with the other variables.

dns/push.sh
bartsmykla: It looks good but I would liek to ask you for one more thing. As we are currently not using arguments `--from-container` and `--config-file` could you remove it for now? It would look like:

```bash
DRY_RUN=false

function parse_args() {
  # positional args
  args=()

  # named args
  while [ "$#" -gt 0 ]; do
      case "$1" in
          -d | --dry-run )          DRY_RUN=true;           ;;
          * )                       args+=("$1")            # if no match, add it to the positional args
      esac
      shift # move to next kv pair
  done

  # restore positional args
  set -- "${args[@]}"
}

parse_args "$@";
```

Following up #783 (step no. 2 in particulary) we'll add this functionality and I would love to work with you on that topic too if you want. :-)

EricLemieux: I have removed the two unused flags.
I'd be happy to help with adding the that functionality

bartsmykla:
@EricLemieux thank you for changes, when we are waiting for @thockin could you squash the commits? (If you need more informations about how to do it here you can find a good instruction: https://github.com/wprig/wprig/wiki/How-to-squash-commits)

bartsmykla:
@EricLemieux thank you again for your work. Today is our community call where during which I will discuss this PR and ask if we are good to go to get it marged. :-)

ref. http://bit.ly/wg-k8s-infra-notes

bartsmykla:
Ok, we didn't manage to discuss this topic yesterday but I think I can accept the PR and if there will be some concerns we'll iterate.

/lgtm
/approve

Issue #775, opened 2020-04-21
Static IP Management


bartsmykla:
When you want to run something in community cluster and need subdomain, we are recomending to mark IP which will be assigned to `Ingress` resource you'll deploy as static.
Currently it's not clear to me who except @thockin can do it and it needs to be done manually.
I think doing it manually is unnecessary and I would like to suggest using our current terraform configuration to handle that.

There are currently [4 static addresses](https://console.cloud.google.com/networking/addresses/list?authuser=0&organizationId=758905017065&project=kubernetes-public):

```shell
gcsweb-k8s-io                 -> 35.190.8.208
k8s-io-ingress-canary         -> 34.102.239.89
k8s-io-ingress-prod           -> 34.107.204.206
perf-dash-k8s-io-ingress-prod -> 34.102.200.94
```

and moving them to be managed by terraform would look something like:

1. Adding blocks for each of the addresses into [the cluster config file](https://github.com/kubernetes/k8s.io/blob/master/infra/gcp/clusters/kubernetes-public/aaa/10-cluster-configuration.tf):
```terraform
resource "google_compute_global_address" "gcsweb-k8s-io" {
  name = "gcsweb-k8s-io"
}

resource "google_compute_global_address" "k8s-io-ingress-canary" {
  name = "k8s-io-ingress-canary"
}

resource "google_compute_global_address" "k8s-io-ingress-prod" {
  name = "k8s-io-ingress-prod"
}

resource "google_compute_global_address" "perf-dash-k8s-io-ingress-prod" {
  name = "perf-dash-k8s-io-ingress-prod"
}
```
2. Running import commands to include our already existing IP addresses in terraform's state file
```shell
terraform import google_compute_global_address.gcsweb-k8s-io kubernetes-public/gcsweb-k8s-io
terraform import google_compute_global_address.k8s-io-ingress-canary kubernetes-public/k8s-io-ingress-canary
terraform import google_compute_global_address.k8s-io-ingress-prod kubernetes-public/k8s-io-ingress-prod
terraform import google_compute_global_address.perf-dash-k8s-io-ingress-prod kubernetes-public/perf-dash-k8s-io-ingress-prod
```
3. Running the `terraform plan` command to see how the configuration compares to the imported resources
4. Changing [the documentation about how to run something in community clusters](https://github.com/kubernetes/k8s.io/blob/master/running-in-community-clusters.md#public-ips)

In first step we are not specyfiing the exact addresses but there are two approaches we can take:

[...]

PR #798, opened 2020-04-23
Added ensure-static-ips.sh script

bartsmykla:
/cc @dims 

ref. https://github.com/kubernetes/k8s.io/issues/775

### DIFF
--- a/infra/gcp/ensure-static-ips.sh
+++ b/infra/gcp/ensure-static-ips.sh
@@ -1,0 +1,50 @@
+#!/usr/bin/env bash
+#
+# Copyright 2020 The Kubernetes Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# This script creates static ip addresses for services deployed in
+# the `aaa` cluster
+
+set -o errexit
+set -o nounset
+set -o pipefail
+
+SCRIPT_DIR=$(dirname "${BASH_SOURCE[0]}")
+. "${SCRIPT_DIR}/lib.sh"
+
+PROJECT_NAME="kubernetes-public"
+
+function usage() {
+    echo "usage: $0" > /dev/stderr
+    echo > /dev/stderr
+}
+
+if [ $# != 0 ]; then
+    usage
+    exit 1
+fi
+
+ADDRESSES=(
+    "gcsweb-k8s-io"
+    "k8s-io-ingress-canary"
+    "k8s-io-ingress-prod"
+    "perf-dash-k8s-io-ingress-prod"
+    "slack-infra-ingress-prod"
+)
+
+for address in "${ADDRESSES[@]}"; do
+    color 6 "Ensure address: $address"
+    ensure_global_ip "$address" "$PROJECT_NAME"
+done
--- a/infra/gcp/lib.sh
+++ b/infra/gcp/lib.sh
@@ -469,3 +469,23 @@ function empower_ksa_to_svcacct() {
         "${gcp_svcacct}" \
         --project="${gcp_project}"
 }
+
+# Ensure that global IP address exists, creating one if need.
+# $1 The IP name (e.g. k8s-io-ingress-prod)
+# $2 The GCP project
+function ensure_global_ip() {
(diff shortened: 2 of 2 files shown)


infra/gcp/ensure-static-ips.sh
```diff
@@ -0,0 +1,68 @@
...
+    echo "usage: $0" > /dev/stderr
+    echo > /dev/stderr
+}
+
+if [ $# != 0 ]; then
+    usage
+    exit 1
+fi
+
+function create_address() {
```
spiffxp, line 39: nit: naming
```suggestion
function ensure_address() {
```

infra/gcp/ensure-static-ips.sh
```diff
@@ -0,0 +1,68 @@
...
+    "gcsweb-k8s-io"
+    "k8s-io-ingress-canary"
+    "k8s-io-ingress-prod"
+    "perf-dash-k8s-io-ingress-prod"
+    "slack-infra-ingress-prod"
+)
+
+for address in "${ADDRESSES[@]}"; do
+    color 6 "Create address: $address"
+    create_address "$address"
```
spiffxp, line 67:
```suggestion
    ensure_address "$address"
```

infra/gcp/ensure-static-ips.sh
```diff
@@ -0,0 +1,68 @@
...
+        return 1
+    fi
+    local address_name="$1"
+
+    # Create address if it doesn't exist
+    if ! gcloud compute addresses describe "$address_name" --global \
+        --project "$PROJECT_NAME" > /dev/null 2>&1; then
+        gcloud compute addresses create "$address_name" --global \
+            --project "$PROJECT_NAME"
+        echo "Address: $address_name created successfully"
```
thockin, line 51: Maybe we should print the value of it?

infra/gcp/ensure-static-ips.sh
```diff
@@ -0,0 +1,68 @@
...
+
+function create_address() {
+    if [ $# != 1 ]; then
+        echo "create_address(name) requires 1 argument" >&2
+        return 1
+    fi
+    local address_name="$1"
+
+    # Create address if it doesn't exist
+    if ! gcloud compute addresses describe "$address_name" --global \
```
thockin, line 47: in the past we had some Service with external IPs (as opposed to Ingress), which requires a slightly different command (no --global, not sure if there's more).

I don't think we have any of those left, but maybe a comment here acknowledging it

infra/gcp/ensure-static-ips.sh
```diff
@@ -0,0 +1,68 @@
...
+fi
+
+function create_address() {
+    if [ $# != 1 ]; then
+        echo "create_address(name) requires 1 argument" >&2
+        return 1
+    fi
+    local address_name="$1"
+
+    # Create address if it doesn't exist
```
thockin, line 46: I could see this moving to lib.sh as `ensure_global_ip(name)`

infra/gcp/ensure-static-ips.sh
```diff
@@ -0,0 +1,68 @@
...
+    if ! gcloud compute addresses describe "$address_name" --global \
+        --project "$PROJECT_NAME" > /dev/null 2>&1; then
+        gcloud compute addresses create "$address_name" --global \
+            --project "$PROJECT_NAME"
+        echo "Address: $address_name created successfully"
+    else
+        echo "Address: $address_name already exists"
```
thockin, line 53: we don't usually print on success

infra/gcp/ensure-static-ips.sh
bartsmykla: We actually can't get it easily, as it looks like it's async operation and in response we are receiving the url address only:
```
Created [https://www.googleapis.com/compute/v1/projects/kubernetes-public/global/addresses/foo].
```

infra/gcp/ensure-static-ips.sh
bartsmykla: I'm not sure what should I put in the comment now, when the logic was moved to separate function in `lib.sh` file

infra/gcp/ensure-static-ips.sh
bartsmykla: Moved the function to the `lib.sh` file, and addressed the name change

infra/gcp/ensure-static-ips.sh
bartsmykla: I moved the function to Moved the function to the `lib.sh` file as @thockin suggested, and addressed the name change

infra/gcp/ensure-static-ips.sh
bartsmykla: Addressed

infra/gcp/ensure-static-ips.sh
bartsmykla: Done

spiffxp:
whoops, I ended up redoing some of this in https://github.com/kubernetes/k8s.io/pull/806

I'll merge this and rebase

PR #808, opened 2020-04-27
Update README

spiffxp:
I didn't include artifactserver because it's the one thing I believe
isn't actively maintained or supported by anyone right now.

There are probably better ways to organize this information (and this
repo) but I just continued the pattern that was there for now

Really I just wanted to add a link to the billing report, but did the
rest since this README is pretty stale

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,10 +2,27 @@
 
 Kubernetes files for various sites and infrastructure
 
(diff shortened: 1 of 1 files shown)


bartsmykla:
Thanks @spiffxp! :-)

Issue #809, opened 2020-04-28
Strip out commonInstanceMetadata.fingerprint from audit's compute project-info.json file


bartsmykla:
This data makes unnecessary noise, so we should strip them out
ref. https://github.com/kubernetes/k8s.io/pull/807#pullrequestreview-401373154

PR #807, opened 2020-04-27
Update audit as of 2020-04-27

spiffxp:
This was intended to pick up the results of running scripts/terraform introduced in https://github.com/kubernetes/k8s.io/pull/806

It also picked up:
- removing some stray `projectView` permissions on `k8s-conform` buckets
- that time where I accidentally the `kubernetes-public` project-wide `ssh-keys` metadata
  - I was trying to manually delete an old key from `spiffxp-node-e2e-project` but forgot the `--project` flag and was defaulted to `kubernetes-public`
  - I would restore the keys but they're listed as expiring on `2020-01-28` 

/cc @dims 
to ack the ssh-keys removal
/cc @thockin @bartsmykla @cblecker 
whomever is interested in reviewing the audit

PR #810, opened 2020-04-28
Stripped out fingerprint properties from audit

bartsmykla:
Closes: #809

Added logic to remove .commonInstanceMetadata.fingerptint
property from audit/projects/*/services/compute/project-info.json

As #807 is not yer merged I haven't included audit files after this change as it would cause conflicts. We can followup after this PR will be merged or if @spiffxp he can redo audit (I don't think it's necessary though as followup PR should make less effort).

/cc @spiffxp @thockin 

### DIFF
--- a/audit/audit-gcp.sh
+++ b/audit/audit-gcp.sh
@@ -127,7 +127,7 @@ gcloud \
                     compute project-info describe \
                     --project="${PROJECT}" \
                     --format=json \
-                    | jq 'del(.quotas[].usage)' \
+                    | jq 'del(.quotas[].usage, .commonInstanceMetadata.fingerprint)' \
                     > "projects/${PROJECT}/services/${SVC}/project-info.json"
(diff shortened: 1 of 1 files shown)


puerco:
I've ran the audit and con confirm the compute dumps dont have the fingerprint anymore.

```json
{
  "commonInstanceMetadata": {
    "kind": "compute#metadata"
  },
  "creationTimestamp": "2019-04-30T17:34:01.258-07:00",
  "defaultNetworkTier": "PREMIUM",
  "defaultServiceAccount": "388270116193-compute@developer.gserviceaccount.com",
  "id": "2088537435958237078",
  "kind": "compute#project",
```

thockin:
Do we want to make an audit commit that just strips those out?

puerco:
I think it could wait until the next big change in the audit script or periodic run

Issue #654, opened 2020-03-10
community infra for kubernetes/KIND CI


BenTheElder:
We require:
- a GCP project to host:
  - GCB (for trusted building, currently we have a prototype running from prow in `bentheelder-kind-dev`)
  - a GCS bucket to host CI binary builds of `kind` for use in Kubernetes testing (currently we use `gs://bentheelder-kind-ci-builds`). These are ~9MB / platform

Prow needs access to GCB, and GCB needs write access to the bucket. The bucket should be public read.

xref: https://github.com/kubernetes-sigs/kind/issues/1392
cc @justaugustus @spiffxp @dims

bartsmykla:
@BenTheElder I think I can help with that. I'll try to write soon how we should proceed.

BenTheElder:
gentle ping, I'd really like to get this out of google.com infra and stop having github.com/kubernetes/kubernetes CI download from a `bentheelder...` bucket w/o detrimentally impacting CI...

we essentially just need a staging project 😅 
cc @dims @spiffxp @justaugustus

listx:
Our existing instructions for subprojects needing a [GCR](https://github.com/kubernetes/k8s.io/tree/master/k8s.gcr.io#creating-staging-repos), have [provisions](https://github.com/kubernetes/k8s.io/blob/master/infra/gcp/ensure-staging-storage.sh#L104) for creating GCS buckets for humans and for GCB. Also, Katharine's automated [cloudbuild instructions](https://github.com/kubernetes/test-infra/blob/master/config/jobs/image-pushing/README.md) have detailed instructions for setting up Prow integration for building in GCB (in the subproject). So it looks like just creating the subproject using the GCR directions will give you what you need. I think we should minimize special-casing our infra rules where possible.

bartsmykla:
@BenTheElder I'm gonna get back to this today, I promise!

bartsmykla:
@BenTheElder so [every staging project](https://github.com/kubernetes/k8s.io/blob/master/infra/gcp/ensure-staging-storage.sh#L41-L81) as @listx wrote have access to two GCS buckets - [one for humans](https://github.com/kubernetes/k8s.io/blob/master/infra/gcp/ensure-staging-storage.sh#L104) and [one for GCB](https://github.com/kubernetes/k8s.io/blob/master/infra/gcp/ensure-staging-storage.sh#L105). They have retention policy set to 60 days though.
[...]

PR #811, opened 2020-04-28
Add staging project for etcdadm

justinsb:
Per https://github.com/kubernetes/k8s.io/tree/master/k8s.gcr.io#creating-staging-repos

fellow owners: @neolit123 @dlipovetsky

### DIFF
--- a/groups/groups.yaml
+++ b/groups/groups.yaml
@@ -839,6 +839,17 @@ groups:
       - jingyih@google.com
       - yczhou@google.com
 
+  - email-id: k8s-infra-staging-etcadm@kubernetes.io
+    name: k8s-infra-staging-etcdadm
+    description: |-
+      ACL for staging etcdadm artifacts.
+    settings:
+      ReconcileMembers: "true"
+    members:
+      - justinsb@google.com
+      - dlipovetsky@d2iq.com
+      - neolit123@gmail.com
+
   - email-id: k8s-infra-staging-external-dns@kubernetes.io
     name: k8s-infra-staging-external-dns
     description: |-
--- a/groups/groups.yaml
+++ b/groups/groups.yaml
@@ -839,7 +839,7 @@ groups:
       - jingyih@google.com
       - yczhou@google.com
 
-  - email-id: k8s-infra-staging-etcadm@kubernetes.io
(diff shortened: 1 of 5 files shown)


dlipovetsky:
Thank you @justinsb! This brings etcdadm one step closer to binary releases (https://github.com/kubernetes-sigs/etcdadm/issues/100)

stp-ip:
/hold

This feels like user facing and binary artifacts could have a need to be stored longer or?
Thanks for giving some context on the usage and therefore needed retention.

Wanted to mention that GCS buckets created within the staging repos do not have extended retention. It comes with similar retention promises as staging GCR images aka months not years.

There is a plan for binary artifact promotion, but that's not being worked on/available yet.
For longer retention a production GCS bucket could potentially "manually"  be created similar to https://github.com/kubernetes/k8s.io/issues/654.

stp-ip:
We briefly discussed this within the k8s-infra call.

Justin will be working on an artifact promoter, if a staging bucket is enough considering it doesn't have guaranteed long retention than the PR is fine, if we want this to have longer retention it should get a manually created prod bucket until we have binary promotion working.

spiffxp:
/approve cancel
Somehow I thought this was an ask for GCR images and a forward looking statement for GCS for binary artifacts.  Let me know what needs to be done here.

justinsb:
I'm not sure what's needed here; we'd like a staging project for etcdadm like other k-sigs projects.

We understand that there isn't yet a fully baked promotion process for binary artifacts (though I'm working on that); when there is though we would like to take advantage of that.  etcdadm will benefit from container distribution in the meantime, along with having the staging locations to get builds going.

groups/groups.yaml
```diff
@@ -839,6 +839,17 @@ groups:
       - jingyih@google.com
       - yczhou@google.com
 
+  - email-id: k8s-infra-staging-etcadm@kubernetes.io
```
spiffxp, line 842:
```suggestion
  - email-id: k8s-infra-staging-etcdadm@kubernetes.io
```

spiffxp:
(I ran the scripts yesterday, forgot to comment here)

Issue #816, opened 2020-04-29
DNS REQUEST: slack-staging.k8s.io


bartsmykla:
### Type of DNS update:
Delete

### Domain being modified:
e.g. `k8s.io`

### Existing DNS Record:

```yaml
slack-staging:
  type: A
  value: 34.107.195.71
```

### Reason for update:

When we'll switch [slack.k8s.io](https://slack.k8s.io) and [slack.kubernetes.io](https://slack.kubernetes.io) to point to the `aaa` and we'll test everything works as expected, we''ll want to get rid of this unnecessary subdomain.

ref. #814, #815
/cc @munnerz

bartsmykla:
As we already switched dnses for slack.k8s.io and slack.kubernetes.io I think we are ready to proceed with removing the slack-staging.k8s.io subdomain

PR #818, opened 2020-04-29
Remove staging slack dns and update certificate

bartsmykla:
Closes: #816

Removing unnecessary anymore dns record for slack-staging.k8s.io
as we switched slack.k8s.io and slack.kubernetes.io to point to
the `aaa` cluster.

Also updated slack-infra's certificates to reflect these changes


bartsmykla:
Ran `make push` at the [`k/k8s.io/dns`](https://github.com/kubernetes/k8s.io/tree/master/dns) so the dns are reconcilled and also applied updated certificate manifest to the `aaa` cluster.

Issue #813, opened 2020-04-28
Proposal: switch to using Secret Manager instead of git-crypt


spiffxp:
We are increasingly relying on `git-crypt` to store shared secrets in this (or other) repos.  My concern is that `git-crypt` is one-repo-fits-all.

WDYT about using `gcloud secrets`? We can specify who's allowed to access a secret by user, group, or serviceAccount

```sh
# create a secret
gcloud secrets create --project=my_project --replication-policy=automatic \
  "my-secret" --data-file="my-secret.json"
 
# allow my-group@kubernetes.io to read the secret
gcloud secrets add-iam-policy-binding my-secret \
  member="group:my-group@kuberentes.io" \
 --role="roles/secretManager.secretAccessor"

# read the secret
gcloud secrets versions access latest --secret="my-secret"
```

I feel like it wouldn't be too hard to rip out our `git-crypt` usage and replace it with this

/wg k8s-infra
/area access

bartsmykla:
lgtm

spiffxp:
/assign @cblecker 
I'd like to demonstrate what this would look like by getting rid of our git-crypt integration for groups

I need some kind of go-ahead to try this out by enabling the secretmanager api on a project; I was going to use k8s-gsuite

cblecker:
I'm on board for running a POC of this :)

dims:
+1 from me as well!

spiffxp:
Opened https://github.com/kubernetes/k8s.io/pull/821

dims:
#821 works great. looks like i did have to run `gcloud auth application-default login` additionally though. we need to touch up readme etc.

spiffxp:
@spiffxp to move slack-infra secrets out of git-crypt to google secret manager

spiffxp:
/close
#913 closed this out

PR #821, opened 2020-04-30
Use gcloud secrets instead of git-crypt for groups

spiffxp:
I have already run the reconciler so that everyone else
should have access to the secret necessary to run the
reconciler

ref: https://github.com/kubernetes/k8s.io/issues/813


stp-ip:
Awesome. Code looks good.

bartsmykla:
@spiffxp lgtm, the only thing we can also do is to remove the first line from the [`.gitattributes`](https://github.com/kubernetes/k8s.io/blob/master/.gitattributes#L1) file, but I don't think it should block this PR.

/lgtm
/hold for @dims

infra/gcp/ensure-gsuite.sh
```diff
@@ -55,13 +59,33 @@ enable_api "${PROJECT}" admin.googleapis.com
...
+    color 6 "A human with admin privileges needs to run the following:"
+    color 4 "  gcloud iam service-accounts keys create tmp.json --project=${PROJECT} \\"
+    color 4 "    --iam-account=$(svc_acct_email "${PROJECT}" "${GSUITE_SVCACCT}") && \\"
+    color 4 "  gcloud secrets create --project=${PROJECT} --replication-policy=automatic \\"
+    color 4 "    --data-file=tmp.json ${GSUITE_SVCACCT}_key && \\"
+    color 4 "  rm tmp.json"
+else
+    color 6 "Empowering ${GSUITE_GROUP_ADMINS} to access the ${GSUITE_SVCACCT}_key secret"
+    gcloud --project="${PROJECT}" \
+        secrets add-iam-policy-binding "${GSUITE_SVCACCT}_key" \
```
cblecker, line 84: Instead of add, should this be a "set"?

spiffxp:
/hold cancel
merging this so I can reconcile other groups

dims:
LGTM

PR #826, opened 2020-05-01
Added substitution to octodns cloudbuild.yaml

bartsmykla:
/cc @BenTheElder 

It is done only to trigger initial push job


BenTheElder:
IIRC this should be fine :shipit: 
/lgtm

PR #742, opened 2020-04-12
Update Release Engineering lists

justaugustus:
- Update Release Manager groups
  - Remove Jorge Alarcon Ochoa (@alejandrox1) (previous RT Lead (1.18)) (ref: https://github.com/kubernetes/sig-release/issues/933)
  - Update Release Manager Associates:
    - Add (ref: https://github.com/kubernetes/sig-release/issues/1047):
      - Marky Jackson (@markyjackson-taulia)
      - Verónica López (@Verolop)
    - Remove (ref: https://github.com/kubernetes/sig-release/pull/1046):
      - Ace Eldeib (@alexeldeib)
      - Kendrick Coleman (@kacole2)
      - Paul Bouwer (@paulbouwer)
      - Taylor Dolezal (@onlydole)
- Update access to Release Manager GCP projects
  - Add editors to `k8s-infra-staging-kubernetes`
  - Deprecate IAM groups for (ref: https://github.com/kubernetes/release/issues/1161):
    - `k8s-infra-sig-release-prototype`
    - `k8s-infra-staging-release-test`
- Grant new RT Lead Shadows (@jeremyrickard) view access to Releng projects (ref: https://github.com/kubernetes/sig-release/issues/1031)

/assign @tpepper @dims @cblecker 
cc: @onlydole @kubernetes/release-managers @kubernetes/release-engineering

PR #857, opened 2020-05-09
Allow k8s-staging-e2e-test-images GCB to access secrets

spiffxp:
[This job](https://github.com/kubernetes/test-infra/blob/master/config/jobs/image-pushing/k8s-staging-e2e-test-images.yaml) which uses [this cloudbuild.yaml](https://github.com/kubernetes/kubernetes/blob/master/test/images/cloudbuild.yaml) needs to be able to call out to a remote docker instance that can build windows images. We had originally been doing this by using kubernetes secrets, but now that the image build has been moved to GCB, we need to pass in secrets to GCB somehow.

This takes the approach of using Secret Manger, and enabling the secretAccessor role for the GCB service account as described here: https://cloud.google.com/cloud-build/docs/securing-builds/use-encrypted-secrets-credentials#store_credentials

~Note: I have only run `ensure-staging-storage.sh e2e-test-images`, and since our billing quota is maxed out, I had to comment out the `gclouc beta billing projects link` command in `lib.sh` to get it to work.~

~I'd like to hear whether folks think:~
~- we should tighten this down further and only allow secret access on a per-secret basis~
~- we should leave as-is; enable secretmanager for all projects, but special-case per-project secret access~
~- we should make this the default for all staging projects, and ensure each project's GCB can access secrets in that project~

EDIT: narrowed the use case down to https://github.com/kubernetes/k8s.io/pull/857#issuecomment-626838955

It's expected that N staging projects will need their GCB service account to access 3 secrets to talk to a remote windows docker instance.  The secrets are hosted in the k8s-infra-prow-build-trusted project, and IAM policies are added to each of those specific secrets for each of the GCB service accounts (currently N=1)

/cc @claudiubelu @dims

### DIFF
--- a/infra/gcp/clusters/modules/gke-project/main.tf
+++ b/infra/gcp/clusters/modules/gke-project/main.tf
@@ -68,6 +68,11 @@ resource "google_project_service" "stackdriver" {
   service = "stackdriver.googleapis.com"
   disable_dependent_services = true
 }
+resource "google_project_service" "secretmanager" {
+  project = google_project.project.project_id
+  service = "secretmanager.googleapis.com"
+  disable_dependent_services = true
+}
 
 
 // "Empower cluster admins" is what ensure-main-project.sh says
--- a/infra/gcp/ensure-staging-storage.sh
+++ b/infra/gcp/ensure-staging-storage.sh
@@ -231,12 +231,12 @@ done
 
 # Special case: Empower GCB in k8s-staging-e2e-test-images to access secrets
 #               that were manually added to k8s-infra-prow-trusted
-color 6 "Configuring special cases for GCB access to windows-image-promoter-cert secrets"
+color 6 "Configuring special cases for GCB access to windows-img-promoter-cert secrets"
 for repo in "${WINDOWS_REMOTE_DOCKER_PROJECTS[@]}"; do
     (
         PROJECT="k8s-staging-${repo}"
         SECRET_PROJECT="k8s-infra-prow-build-trusted"
-        SECRET_GROUP="windows-image-promoter-cert"
+        SECRET_GROUP="windows-img-promoter-cert"
         for secret in $(gcloud secrets list \
                         --format="value(name)" \
                         --project="${SECRET_PROJECT}" \
--- a/infra/gcp/ensure-staging-storage.sh
+++ b/infra/gcp/ensure-staging-storage.sh
@@ -225,4 +225,22 @@ for repo in "${RELEASE_STAGING_PROJECTS[@]}"; do
     ) 2>&1 | indent
 done
 
-
+# Special case: Empower GCB in k8s-staging-e2e-test-images to access secrets
+#               that were manually added to k8s-infra-prow-trusted
+color 6 "Configuring special case for k8s-staging-e2e-test-images"
+(
+    PROJECT="k8s-staging-e2e-test-images"
+    SECRET_PROJECT="k8s-infra-prow-build-trusted"
+    SECRET_GROUP="windows-image-promoter-cert"
+    for secret in $(gcloud secrets list \
+                    --format="value(name)" \
+                    --project="${SECRET_PROJECT}" \
+                    --filter="labels.secret-group=${SECRET_GROUP}"); do
+      color 6 "Empowering ${PROJECT}'s GCB service account to access secret ${secret} in ${SECRET_PROJECT}"
+      gcloud secrets add-iam-policy-binding \
+        "${secret}" \
+        --project="${SECRET_PROJECT}" \
+        --member="serviceAccount:$(gcb_service_account_email "k8s-staging-e2e-test-images")" \
+        --role="roles/secretmanager.secretAccessor"
+    done
+) 2>&1 | indent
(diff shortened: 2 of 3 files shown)


claudiubelu:
Thanks for doing this. This looks pretty much like what we need in order to get GCB to use the sig-windows Windows build nodes for the Windows images.

However, if I'm not mistaken, this only empowers GCB to have access to the ``k8s-staging-e2e-test-images`` secrets, and the actual secret containing the Windows build nodes certificates was not created yet, right? If you need them, you can find them in prow under the name ``windows-img-promoter-cert``, or I can send them to you if needed.

> 
> I'd like to hear whether folks think:
> 
>     * we should tighten this down further and only allow secret access on a per-secret basis
> 
>     * we should leave as-is; enable secretmanager for all projects, but special-case per-project secret access
> 
>     * we should make this the default for all staging projects, and ensure each project's GCB can access secrets in that project
> 

FWIW, we are going to need to do this in multiple jobs / projects. For example, one other projects I've looked into is CSI secret store driver (https://github.com/kubernetes/test-infra/pull/17335), and other sig-windows folks are looking into other CSI drivers as well, and I'm going to have to do the same for the pause image. So, it would be helpful for multiple projects to be able to access the same secret in order to get the necessary certs.

spiffxp:
I purposely haven't put in the secret yet until I get some consensus on the approach we'd like to take here

If the intent is to share secrets across GCB jobs, then I think I'd rather store the secret in the project hosting the prow build cluster, and special-case access from specific project GCB service account to  specific secrets, eg:

```bash
for prj in "${staging_projects_that_need_windows_secret[@]}"; do
  gcloud secrets add-iam-policy-binding \
    --project="k8s-infra-prow-builds-trusted" windows-img-promoter-cert  \
    --member="serviceAccount:$(gcb_service_account_email "${prj}")" \
    --role="roles/secretmanager.secretAccessor" 
done
```

claudiubelu:
So, if another project will also require access to a specific secret, it will have to send a PR for it, which will have to be approved, right? Yeah, that sounds reasonable to me.

spiffxp:
I have access to the `windows-img-promoter-cert` secret, what I'm less clear on is whether I should jam in a yaml file with base64 encoded keys, or create three secrets for each file contained within

claudiubelu:
Hm, they were sent as files initially, so my original thought was a .tar.gz. I would say that 3 secrets would work best, it's easier to just fetch those secrets and store the contents into the necessary files.

spiffxp:
Stored as three secrets labeled after the old secret name
```console
$ gcloud secrets list --project=k8s-infra-prow-build-trusted --filter=labels.secret-group=windows-img-promoter-cert --format="value(name)"
windows-remote-docker_ca-pem
windows-remote-docker_cert-pem
windows-remote-docker_key-pem
```

spiffxp:
@dims and @thockin any objections to this approach? I've stored the secrets as described in the latest comments, but won't run the script in this PR until it's merged

dims:
No objections from me. feel free to remove hold when ready.

/lgtm
/hold

spiffxp:
Ran `terraform apply` for `infra/gcp/clusters/k8s-infra-prow-build/prow-build` (I needed to update resource definitions to prevent terraform from nuking a nodepool, opened https://github.com/kubernetes/k8s.io/pull/946)
Ran `terraform apply` for `infra/gcp/clusters/k8s-infra-prow-build-trusted/prow-build-trusted`

spiffxp:
Running `./infra/gcp/ensure-staging-storage.sh` ...

spiffxp:
The run bailed out here
```
Configuring staging: release-test
  Ensuring project exists: k8s-staging-release-test
  billingAccountName: billingAccounts/018801-93540E-22A20E
  billingEnabled: true
  name: projects/k8s-staging-release-test/billingInfo
  projectId: k8s-staging-release-test
  Empowering k8s-infra-staging-release-test@kubernetes.io as project viewers
  ERROR: Policy modification failed. For a binding with condition, run "gcloud alpha iam policies lint-condition" to identify issues in condition.
  ERROR: (gcloud.projects.add-iam-policy-binding) INVALID_ARGUMENT: Group k8s-infra-staging-release-test@kubernetes.io does not exist.
```

Which is true, `k8s-infra-staging-release-test@kubernetes.io` doesn't exist in groups.yaml

It looks like https://github.com/kubernetes/k8s.io/pull/742 removed it.  @justaugustus what should we be doing here? I followed a trail of links to https://github.com/kubernetes/release/issues/1161 which indicates it may be safe to delete `release-test` from the list of staging projects and delete the GCP project?

spiffxp:
I'm commenting out mentions of `release-test` out and re-running `./infra/gcp/ensure-staging-storage.sh`

justaugustus:
> It looks like #742 removed it. @justaugustus what should we be doing here? I followed a trail of links to [kubernetes/release#1161](https://github.com/kubernetes/release/issues/1161) which indicates it may be safe to delete `release-test` from the list of staging projects and delete the GCP project?

@spiffxp -- Yep, both `k8s-staging-release-test` and `k8s-infra-sig-release-prototype` are safe to delete.

Active releng projects are listed here: https://github.com/kubernetes/sig-release/blob/master/release-engineering/gcp.md#projects

I think at the time that we had initially raised deleting these projects, @thockin mentioned that there wasn't really a deletion flow for the scripts, so we punted and filed the issue to come back to later.

PR #871, opened 2020-05-15
Moved creation of iam roles to separate function

bartsmykla:
As there was multiple times discussion about possibility of
creating custom iam roles with more granular permissions,
and there is ongoing work to automate some of our scripts
which will probably need custom roles it's good time
to move out the logic to the separate function

Also created one more helper function `join_by` to joing strings
together with the provided separator

/cc @thockin @spiffxp @munnerz 

### DIFF
--- a/infra/gcp/lib.sh
+++ b/infra/gcp/lib.sh
@@ -517,3 +517,44 @@ function ensure_regional_address() {
         --region="${region}"
     fi
 }
+
+# Ensure that custom IAM role exists, creating one if needed
+# Arguments:
+#   $1:  The GCP project
+#   $2:  The role name (e.g. "ServiceAccountLister")
+#   $3:  The role title (e.g. "Service Account Lister")
+#   $4:  The role description (e.g. "Can list ServiceAccounts.")
+#   $5+: The role permissions (e.g. "iam.serviceAccounts.list")
+# Example usage:
+#   ensure_custom_iam_role \
+#       kubernetes-public \
+#       ServiceAccountLister \
+#       "Service Account Lister" \
+#       "Can list ServiceAccounts." \
+#       iam.serviceAccounts.list
+function ensure_custom_iam_role() {
+    if [ $# -lt 5 ] || [ -z "${1}" ] || [ -z "${2}" ] || [ -z "${3}" ] \
+        || [ -z "${4}" ] || [ -z "${5}" ]
+    then
+        echo -n "ensure_custom_iam_role(gcp_project, name, title," >&2
+        echo    " description, permission...) requires at least 5 arguments" >&2
+        return 1
+    fi
+
+    local gcp_project="${1}"; shift
+    local name="${1}"; shift
+    local title="${1}"; shift
+    local description="${1}"; shift
+    local permissions; permissions=$(join_by , "$@")
+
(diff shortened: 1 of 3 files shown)


infra/gcp/lib.sh
```diff
@@ -517,3 +517,44 @@ function ensure_regional_address() {
...
+        echo -n "ensure_custom_iam_role(gcp_project, name, title," >&2
+        echo    " description, permission...) requires at least 5 arguments" >&2
+        return 1
+    fi
+
+    local gcp_project="${1}"; shift
+    local name="${1}" shift
+    local title="${1}" shift
+    local description="${1}" shift
+    local permissions="$(join_by , "$@")"
```
thockin, line 548: small nit: I don't think you need the outer quotes, though I think in this case bash manages to get the inner quotes right

infra/gcp/lib.sh
```diff
@@ -517,3 +517,44 @@ function ensure_regional_address() {
...
+    if [ $# -lt 5 ] || [ -z "${1}" ] || [ -z "${2}" ] || [ -z "${3}" ] \
+        || [ -z "${4}" ] || [ -z "${5}" ]
+    then
+        echo -n "ensure_custom_iam_role(gcp_project, name, title," >&2
+        echo    " description, permission...) requires at least 5 arguments" >&2
+        return 1
+    fi
+
+    local gcp_project="${1}"; shift
+    local name="${1}" shift
```
thockin, line 545: missing semi-colons on a bunch of these

infra/gcp/lib_util.sh
```diff
@@ -67,3 +67,19 @@ fi
...
+# Example usage:
+#   join_by , foo bar baz
+function join_by() {
+  if [ $# -lt 2 ] || [ -z "${1}" ] || [ -z "${2}" ]; then
+      echo "join_by(separator, string...) requires at least 2 arguments" >&2
+      return 1
+  fi
+
+  local IFS="${1}"; shift
+  echo "$*"
```
thockin, line 84: To ensure someone doesn't "fix" this, please comment on why this is $* rather than $@ ?  $@ doesn't seem to respect IFS.

infra/gcp/lib_util.sh
```diff
@@ -67,3 +67,19 @@ fi
 function indent() {
     ${SED} -u 's/^/  /'
 }
+
+# Join things with separator
+# Arguments:
+#   $1:  Separator
```
thockin, line 73: comment that this must be a single character

infra/gcp/lib.sh
bartsmykla: Changed

infra/gcp/lib.sh
bartsmykla: Ah, thank you for catching that. Somehow my `shellcheck` stopped working with VSCode (fixed now) and I missed it

infra/gcp/lib_util.sh
bartsmykla: Done

infra/gcp/lib_util.sh
bartsmykla: Done

bartsmykla:
@thockin ping, could you check if everything is fine now? :-)

Issue #604, opened 2020-02-26
Need a script to generate service account keys for k8s-conform project


dims:
Thanks to https://blog.csanchez.org/2019/06/11/google-container-registry-service-account-permissions/ for coming up with the following snippet

```
#!/bin/bash

set -xeu

FOR=cri-o
PROJECT_ID=k8s-conform

ACCOUNT=service-${FOR}
BUCKET=${PROJECT_ID}-${FOR}

# enable Google Cloud Storage JSON API for project
gcloud services enable storage-api.googleapis.com --project=$PROJECT_ID

# delete the existing service account and policy binding
gcloud -q iam service-accounts delete ${ACCOUNT}@${PROJECT_ID}.iam.gserviceaccount.com || true
gsutil iam ch -d serviceAccount:${ACCOUNT}@${PROJECT_ID}.iam.gserviceaccount.com gs://${BUCKET} || true

# create a new one
gcloud -q iam service-accounts create ${ACCOUNT} --project ${PROJECT_ID} --display-name ${ACCOUNT}
gcloud -q iam service-accounts keys create "${ACCOUNT}.json" --iam-account=${ACCOUNT}@${PROJECT_ID}.iam.gserviceaccount.com
gsutil iam ch serviceAccount:${ACCOUNT}@${PROJECT_ID}.iam.gserviceaccount.com:objectCreator gs://${BUCKET}
gsutil iam get gs://${BUCKET}

echo "key json is ${ACCOUNT}.json"


```

dims:
related to https://github.com/kubernetes/k8s.io/pull/501#issuecomment-591332061

dims:
Another tip from https://stackoverflow.com/questions/51678745/how-to-restrict-a-google-service-account-to-a-single-bucket-in-google-cloud-stor

dims:
also related to https://github.com/kubernetes/k8s.io/pull/459

dims:
I've generated a key for each of
- cri-o ( handed to @saschagrunert )
- huaweicloud ( handed to @RainbowMango )
- capi-openstack ( handed to @sbueringer )

dims:
related to https://github.com/kubernetes/test-infra/pull/16515

thockin:
Can you say more about why they need keys rather than just being in the correct googlegroup?

thockin:
@saschagrunert @RainbowMango @sbueringer

[...]

PR #873, opened 2020-05-15
Added key generation for k8s-conform service accounts

bartsmykla:
My idea is to use GCP Secret Manager to store the keys, and I didn't want to implement in bash any complicated logic to check if there is already existing secret and/or key for the service account so I just make simple check if the service account already exist or not. It would also mean for current service accounts we would have to run script like:

```bash
#!/usr/bin/env bash

readonly PROJECT="k8s-conform"
readonly REPO="${1}"
readonly KEY_FILE="${2}"
readonly BUCKET_WRITERS="k8s-infra-conform-${REPO}@kubernetes.io"
readonly BUCKET="gs://${PROJECT}-${REPO}" 
readonly SECRET_ID="service-${REPO}-key"

gcloud --project "${PROJECT}" services enable secretmanager.googleapis.com

gcloud secrets create "${SECRET_ID}" \
    --project "${PROJECT}" \
    --replication-policy "automatic"

gcloud secrets versions add "${SECRET_ID}" \
    --project "${PROJECT}" \
    --data-file "${KEY_FILE}"

gcloud secrets add-iam-policy-binding "${SECRET_ID}" \
    --member "group:${BUCKET_WRITERS}" \
    --role "roles/secretmanager.secretAccessor"
```

For example:
```shell
~ ./script.sh capi-openstack ./path/to/the/key.json
~ ./script.sh cri-o ./path/to/the/key.json
~ ./script.sh huaweicloud ./path/to/the/key.json
```

/cc @dims @thockin @spiffxp 

Closes: #604

### DIFF
--- a/infra/gcp/ensure-conformance-storage.sh
+++ b/infra/gcp/ensure-conformance-storage.sh
@@ -62,6 +62,10 @@ ensure_project "${PROJECT}"
 color 6 "Enabling the GCS API"
 enable_api "${PROJECT}" storage-component.googleapis.com
 
+# Enable Secret Manager API
+color 6 "Enabling the Secret Manager API"
+enable_api "${PROJECT}" secretmanager.googleapis.com
+
 color 6 "Ensuring all conformance buckets"
 for REPO; do
     color 3 "Configuring conformance bucket for ${REPO}"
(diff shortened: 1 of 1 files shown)


dims:
@bartsmykla looks good to me. Will let @thockin sign off. From a practical point of view, do you need the 3 key json files we minted earlier for the 3 service accounts?

infra/gcp/ensure-conformance-storage.sh
```diff
@@ -90,5 +94,51 @@ for REPO; do
     # Enable writers on the bucket
     color 6 "Empowering ${BUCKET_WRITERS} to GCS"
     empower_group_to_write_gcs_bucket "${BUCKET_WRITERS}" "${BUCKET}"
+
+    readonly SERVICE_ACCOUNT_NAME="service-${REPO}"
```
spiffxp, line 98: nit: I would either
- drop the `service-` prefix since it's too generic
- rename to `${REPO}-writer` to hint at what this account is used for

(I don't care strongly enough to block if you ignore this though)

bartsmykla: So I used this pattern because I deduced from #604 these pattern was used for `capi-openstack`, `cri-o`, `huaweicloud` already and changing it would require some work, but I'm open to changing it (I also prefer the way you suggested it). @dims wdyt?

spiffxp: if we're stuck with the names so be it; per @dims' question do you need the existing keys?

bartsmykla: I think it would be good to have it consistent if for example we would want to change something in the future

dims: note from slack `And also if you could look once again here: https://github.com/kubernetes/k8s.io/pull/873 as we discussed this yesterday and we can proceed without storing these secrets (which were currently given to three projects)`

PR #892, opened 2020-05-22
Node perf dash small changes

bartsmykla:
- Removed node-perf-dash.k8s.io dns record from the certificate manifest as it was blocking certificate from being issued by cert-manager
- Renamed certificate names from `node-perf-dash-*.yaml` to `*.yaml` as I think it's unnecessary to duplicate the names inside the manifests names when the directory has the name in it already
- Changed service type to NodePort as it is unnecessary and would generate costs when GKE Ingress Controller + Cert Manger are handling our casese
- Changed bootstraping instructions to be more generic using for loop do apply manifests instead of listing all of them individually

/cc @ameukam @dims 

ref. @lorqor


node-perf-dash.k8s.io/deployment.yaml
```diff
@@ -2,7 +2,6 @@ apiVersion: extensions/v1beta1
 kind: Deployment
 metadata:
   name: node-perf-dash
-  namespace: node-perf-dash
```
ameukam, line 5: Is it not better to keep the namespace ? I prefer to avoid using the default namespace for everything deployed on aaa.

bartsmykla: Removed commit with this changes

PR #895, opened 2020-05-26
node-feature-discovery: promote v0.5.0

marquiz:
Drop previously merged entry which had an invalid digest.


kad:
let's try second time...
/lgtm

Issue #851, opened 2020-05-07
Identify quota changes needed for scalability jobs, create pool of scalability projects


spiffxp:
The default quotas for an e2e project (eg: k8s-infra-e2e-gce-project) are insufficient to run ci-kubernetes-e2e-gci-gce-scalability

Currently this job runs in the google.com k8s-prow-builds cluster, using a project from that boskos' `scalability-project` pool

- Identify what quotas are set that make these projects differ from a default e2e project
- Identify what jobs this will allow us to migrate
- Identify pool size
- Provision pool in CNCF org
- Add pool to k8s-infra-prow-build's boskos

spiffxp:
I created k8s-infra-e2e-scale-project and tried manually running ci-kubernetes-e2e-gci-gce-scalability in k8s-infra-prow-build (with some local modifications not to dump results to gs://kubernetes-jenkins)

It got stuck on CPU quota

I was able to retrieve project info for k8s-periodic-scale-1, so I compared with k8s-infra-e2e-scale-project globally and in us-east1.  CPUs and in-use addresses were bumped up to 125 in us-east1.

I filed a request to raise quota for k8s-infra-e2e-scale-project with the following info:
```
OSS Kubernetes scalability testing, 100 node clusters

https://testgrid.k8s.io/sig-release-master-blocking#gce-cos-master-scalability-100

Goal is to migrate OSS Kubernetes tests to use GCP projects under CNCF org instead of google.com org

Trying to match quotas of google.com project "k8s-periodic-scale-1". Anticipate 10 more projects like this.
```

The quota got bumped during the job run, and the job was able to complete successfully.

spiffxp:
Jobs that use scalability-project

100 node release-blocking tests (definitely migrate these)
- ci-kubernetes-e2e-gci-gce-scalability - https://testgrid.k8s.io/sig-release-master-blocking#gce-cos-master-scalability-100
- ci-kubernetes-e2e-gci-gce-scalability-stable3 - https://testgrid.k8s.io/sig-release-1.15-blocking#gce-cos-1.15-scalability-100
- etc for each release branch

100 node kubemark tests (these look like CI variants of merge-blocking jobs so make room for them)
- ci-kubernetes-kubemark-100-gce - https://testgrid.k8s.io/sig-scalability-kubemark#kubemark-100
- ci-kubernetes-kubemark-500-gce-stable3 - https://testgrid.k8s.io/sig-scalability-kubemark#kubemark-1.15-500
[...]

PR #898, opened 2020-05-26
add e2e scale projects

spiffxp:
one manual project, and a pool of 5 projects

these are intended to match the projects that currently live in
prow.k8s.io's "scalability-project" boskos pool, e.g.
- us-east1 CPU quota raised to 125
- us-east1 in-use addresses quota raised to 125

ref: https://github.com/kubernetes/k8s.io/issues/851


spiffxp:
forgot to uncomment the other e2e projects

mm4tt:
FYI, we don't need the 125 in-use addresses quota. We migrated our tests to private IPs in https://github.com/kubernetes/kubernetes/issues/76374

Issue #896, opened 2020-05-26
Google Cloud Bucket change for cri-tools


saschagrunert:
The cri-tools update PR to v1.18.0 (https://github.com/kubernetes/kubernetes/pull/91239) got reverted because I changed the download URL to point to GitHub, which caused an API rate limit hit. The main reason to change that URL was that release managers (and myself, even as maintainer of cri-tools) do not have the right set of permissions to upload the cri-tools artifacts to that bucket.

The bucket where `ctictl` was originally downloaded is `kubernetes-release/crictl`: https://console.cloud.google.com/storage/browser/kubernetes-release/crictl

The idea is to use the new [gh2gc](https://github.com/kubernetes/release/pull/1311) tool in kubernetes-release, so that release managers can upload the binary to a bucket which can be accessed by every release manager. This way we could use `kubepkg` to also build the artifacts (basically tarballs) and upload them in one command.

/cc @feiskyer @justaugustus

justaugustus:
cc: @dims (who was on today's RelEng call)

justaugustus:
Opened a PR here to add the new bucket: https://github.com/kubernetes/k8s.io/pull/908
cc: @thockin (since he's been working on the other push buckets)

justaugustus:
> The idea is to use the new [gh2gcs](https://github.com/kubernetes/release/pull/1311) tool in kubernetes-release, so that release managers can upload the binary to a bucket which can be accessed by every release manager. This way we could use `kubepkg` to also build the artifacts (basically tarballs) and upload them in one command.

@saschagrunert -- To clarify, cri-tools maintainers will still need to use whatever mechanisms are in place today to package tarballs for their releases. There will be a config in k/release hooked up to a Prow job to upload their releases to the new bucket.

`kubepkg` will then eventually be used to package the cri-tools during the k/k release process. This may change in the future, but today, creating tarballs is not an intended `kubepkg` feature.

saschagrunert:
Alright thanks for the clarification.

feiskyer:
Thanks @saschagrunert @justaugustus for moving this forward.

justaugustus:
For sure! https://github.com/kubernetes/k8s.io/pull/908 has merged and the bucket has been created, so we're done from the k/k8s.io perspective (thanks @dims!).

I've opened https://github.com/kubernetes-sigs/cri-tools/issues/616 to continue tracking this from the k-sigs/cri-tools project instead.

/close

PR #908, opened 2020-05-28
Add k8s-artifacts-cri-tools prod push bucket and IAM group

justaugustus:
ref: https://github.com/kubernetes/kubernetes/pull/91239, https://github.com/kubernetes/k8s.io/issues/896

/assign @thockin @dims @cblecker 
cc: @saschagrunert @feiskyer @dchen1107 @derekwaynecarr @kubernetes/release-engineering

### DIFF
--- a/groups/groups.yaml
+++ b/groups/groups.yaml
@@ -305,6 +305,7 @@ groups:
       - davanum@gmail.com
       - spiffxp@google.com
       - thockin@google.com
+
   #
   # Push groups: k8s-infra-push-*
   #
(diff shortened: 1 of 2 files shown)


groups/groups.yaml
```diff
@@ -318,23 +319,32 @@ groups:
     settings:
       ReconcileMembers: "true"
     members:
+      - k8s-infra-release-admins@kubernetes.io
```
justaugustus, line 322: This change is not directly related to CRI tools. Just doing it as a cleanup since I was already in the file.

justaugustus:
The current IAM group does not include everyone from https://github.com/kubernetes-sigs/cri-tools/blob/master/OWNERS, simply because I don't have their email addresses on hand.

Happy to add them here once I do or we can address that as a follow-up.

PR #912, opened 2020-05-28
Add cert-manager image bucket and project

vincepri:
Signed-off-by: Vince Prignano <vincepri@vmware.com>


groups/groups.yaml
```diff
@@ -745,6 +745,20 @@ groups:
...
+      ACL for staging certmanager
+    settings:
+      ReconcileMembers: "true"
+    members:
+      - davanum@gmail.com
+      - detiber@gmail.com
+      - goldsteina@vmware.com
+      - prignanov@vmware.com
+      - james@munnelly.eu
+      - Cecile.RobertMichon@microsoft.com
```
CecileRobertMichon, line 760:
```suggestion
      - cerobert@microsoft.com
```

thockin:
Hold up, we discussed this yesterday but did not come to consensus.

The issue here is whether CNCF should really be expected to foot the bill for serving cert-manager images publicly.  We do this for coredns and etcd because a) they are integral to k8s installs and b) they are part of CNCF.

Given that cert-manager is neither (though increasingly true for A), it's not clear that we can host this, without opening the door to all sorts of other projects.  One could argue that if cert-manager wants to be widely used, it should be reliable, and the owners of it should be funding that reliability.

According to @munnerz this is a non-trivial number of downloads.

IF we are going to go here, we need to understand why we would say yes to cert-manager and no to some other project.

Can we put the brakes on here, and get to that consensus?  Maybe roll this back and clean up any activations?

thockin:
The problem here is that there's no technical distinction between "being a mirror" and "serving the image for free to anyone who wants it".

PR #913, opened 2020-05-28
remove git-crypt support from repo

spiffxp:
Removing the keys means encrypted data in older commits cannot be
decrypted. I did this to limit exposure. I don't think it's necessary
to go back and rewrite history to remove the encrypted files in the
first place, but am open to suggestion.

/hold
depends on: #911

This should close out #813 but I'm not using the keyword to give us time to confirm


spiffxp:
/hold cancel
#911 merged

/uncc @mikedanese @nikhita 
/cc @cblecker @dims @thockin 
fellow git-crypt consumers

spiffxp:
ping for review

PR #914, opened 2020-05-28
cni: Use releng IAM groups instead of individual addresses

justaugustus:
Split out from https://github.com/kubernetes/k8s.io/pull/908.

/assign @dims

### DIFF
--- a/groups/groups.yaml
+++ b/groups/groups.yaml
@@ -332,23 +332,16 @@ groups:
     settings:
       ReconcileMembers: "true"
     members:
+      - k8s-infra-release-admins@kubernetes.io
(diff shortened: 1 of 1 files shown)


groups/groups.yaml
```diff
@@ -332,23 +332,16 @@ groups:
     settings:
       ReconcileMembers: "true"
     members:
+      - k8s-infra-release-admins@kubernetes.io
+      - k8s-infra-release-editors@kubernetes.io
```
justaugustus, line 336: Matches Patch Release Team + Branch Managers listed here: https://github.com/kubernetes/sig-release/blob/master/release-managers.md

Issue #915, opened 2020-05-28
Decide and document rules for what can be hosted as staging projects


thockin:
Context: #912

thockin:
Copying some:

The issue here is whether CNCF should be expected to foot the bill for serving other projects' container images publicly. We do this for some (coredns and etcd) because a) they are integral to k8s installs and b) they are part of CNCF.

It's not clear to me that those criteria are correct or sufficient.

We need to understand why we would say yes to one project and no to another.

Is CNCF ownership a requirement?

Is the fact that some k8s project took a dependency on it sufficient?

Should we assert that all binaries in our GCR are built by mechanisms we control and can audit?

spiffxp:
> Is CNCF ownership a requirement?

I would argue Kubernetes ownership.  This WG's scope is about Kubernetes project infra.  If we're talking about CNCF project infra, this WG should be re-homed under CNCF.

> Is the fact that some k8s project took a dependency on it sufficient?

Not in my opinion.  I would argue Kubernetes releases are a special exception..

> Should we assert that all binaries in our GCR are built by mechanisms we control and can audit?

That would be my preference.

My strawman proposal:
- the source used to create the artifact is part of a Kubernetes subproject (owned by a SIG, or in exceptional cases by a Committee)
- the artifact is built using mechanisms wg-k8s-infra owns
- mumble something about not mirroring, with exceptions given on a case-by-case basis (e.g. Kubernetes' coredns, etcd artifacts, which are shipped as part of a Kubernetes release)

thockin:
CoreDNS is the obvious outlier among existing staging repos.

We could fork it and cut builds from our fork.  A lot of work for low return...

stp-ip:
On the CoreDNS side of things we have a (long term) tracking issue: #533 
TL;DR: It's more about the possibility to mirror code, and control the build (incl. process) aka have more flexibility what goes into the artifact, but not a full fork only a build wrapper repo of sorts. As we already do mirror the actual artifact and control the endpoint.

On the main issue:
From a user's point of view it seems not mirroring might lead to a bad experience with K8s. So this would mean anything that could be said is contained or needed for "plain" Kubernetes should be mirrored or at least be prevented from breaking.

[...]

PR #917, opened 2020-05-28
Revert "Add cert-manager image bucket and project"

BenTheElder:
Reverts kubernetes/k8s.io#912

suggested by @justaugustus 

We don't have consensus on this (hosting / providing resources for projects that are not even CNCF, let alone in-org) yet. See also https://github.com/kubernetes/k8s.io/issues/915

/cc @thockin @dims @spiffxp


spiffxp:
/approve
/lgtm
rolling the gcp infra changes back will require some surgery

PR #918, opened 2020-05-28
Reconcile security@kubernetes.io members

tallclair:
We've removed opsgenie, so we can reconcile this list now.


justaugustus:
Thanks @tallclair!
/lgtm

Issue #844, opened 2020-05-06
Develop better ACLs for prow build clusters and e2e projects


spiffxp:
At present, I basically copied what was done for the `aaa` cluster and the `kubernetes-public` project, and applied that to:
- k8s-infra-prow-build/prow-build
- k8s-infra-prow-build-trusted/prow-build-trusted

There are some problems with this setup:
- I'm listed as the only owner (org owners incidentally have owner privileges, I think)
- Nobody else has the project editor role
- The existing set of people who have access to k8s-prow-builds/prow don't have access to these clusters (IIRC they basically have the project owner role for their clusters)

I'd like to get us to the point where we have something better, possibly:
- the ability for a larger group of people to be able to view things (logging, monitoring) for both prow build clusters
- the ability for a smaller trusted group of people to be able to write to the prow-build cluster
- the ability for a (even smaller? different?) trusted group of people to read secrets and write to the prow-build-trusted cluster
- the ability for people to see what's happening in a given e2e cluster (maybe v0 of this is predefining a role that can be manually/selectively granted)

ref: https://github.com/kubernetes/k8s.io/issues/752, followup to: https://github.com/kubernetes/k8s.io/pull/830

spiffxp:
Spoke with test-infra-oncall team, if I can't add the group they use, I have permission to create a new @kubernetes.io group for them here

spiffxp:
Will work on this over next two weeks

spiffxp:
#919 setup the following:
- k8s-infra-prow-oncall@kubernetes.io - oncall/break-glass access to prow project and e2e projects
- k8s-infra-sig-scalability-oncall@kubernetes.io - oncall/break-glass access to k8s-infra-e2e-boskos-scale-* projects
- k8s-infra-prow-viewers@kubernetes.io - readonly access to prow project and e2e projects

They're only using primitive roles (owner and viewer)

spiffxp:
k8s-infra-prow-viewers@kubernetes.io is significantly more useful now thanks to https://github.com/kubernetes/k8s.io/pull/1061

spiffxp:
Revisiting the description

>I'd like to get us to the point where we have something better, possibly:
> - the ability for a larger group of people to be able to view things (logging, monitoring) for both prow build clusters

Members of k8s-infra-prow-viewers@kubernetes.io can do this

> - the ability for a smaller trusted group of people to be able to write to the prow-build cluster
[...]

PR #919, opened 2020-05-28
Add oncall groups and bind to projective primitive roles

spiffxp:
This mirrors what is done today for the google.com-equivalent projects
- test-infra-oncall has owner access to the prow projects, and all e2e projects
- scalability-oncall has owner access to all scalability e2e projects

I also added a k8s-infra-prow-viewer group to see if that would give sufficient
access for troubleshooting

I anticipate dialing these down to custom roles to be done as followup

ACLs ref: https://github.com/kubernetes/k8s.io/issues/844
scalability ref: https://github.com/kubernetes/test-infra/pull/17725

### DIFF
--- a/groups/groups.yaml
+++ b/groups/groups.yaml
@@ -549,6 +549,51 @@ groups:
       - simony@google.com
       - sumitranr@google.com
 
+  - email-id: k8s-infra-prow-oncall@kubernetes.io
+    name: k8s-infra-prow-oncall
+    description: |-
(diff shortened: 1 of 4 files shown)


spiffxp:
ran `make -C groups run -- --confirm`

working on the rest...

spiffxp:
ran `terraform apply` for both build clusters

find a typo (ref: https://github.com/kubernetes/k8s.io/pull/922) so edited locally to workaround

ran `./infra/gcp/ensure-e2e-projects.sh`

Issue #843, opened 2020-05-06
Setup ghproxy in k8s-infra-prow-build-trusted


spiffxp:
(ghProxy)[https://github.com/kubernetes/test-infra/tree/master/ghproxy] is a reverse proxy HTTP cache optimized for use with the GitHub API.  Jobs that manage our github orgs and repos (label_sync, peribolos) greatly benefit from this, and at this point probably won't run without it.

Today, the google.com prow.k8s.io cluster acting as "trusted build cluster" (k8s-prow/prow)has it deployed to its own nodepool (label dedicate:ghproxy)
- node: n1-standard-8, 100 GB pd-standard, COS
- volume: 100Gi, pd-ssd, gce-ssd-retain storageclass (defined in test-infra/config/prow/cluster)

ref: https://github.com/kubernetes/k8s.io/issues/752, followup to: https://github.com/kubernetes/k8s.io/pull/830

ameukam:
@spiffxp I started to look at ghproxy and I'm curious about something: is [pushgateway](https://github.com/kubernetes/test-infra/blob/master/config/prow/cluster/ghproxy.yaml#L60) a hard requirement for ghproxy ?

bartsmykla:
@ameukam it looks like it's an endpoint for prometheus metrics to be exposed, but looking at: [`k/test-infra/ghproxy/ghproxy.go#L122`](https://github.com/kubernetes/test-infra/blob/d7e9546a3edf043ef8ab6a74a79ce2416e810b85/ghproxy/ghproxy.go#L122) I would understand it can be ommited and I would suggest to to it at this point.

Looking at [`k/test-infra/prow/metrics/metrics.go`](https://github.com/kubernetes/test-infra/blob/d7e9546a3edf043ef8ab6a74a79ce2416e810b85/prow/metrics/metrics.go#L40-L45) to not expose the metrics there still needs to be passed `--push-gateway` flag with no `--serve-metrics` flag or it set to `false` as when `--push-gateway` is not set it would ignore `--serve-metrics` set to `false`.

BTW. @fejta should it work that way or is it a bug?

spiffxp:
/assign
need to provide token to trusted cluster for ghproxy

spiffxp:
Blocked on https://github.com/kubernetes/test-infra/issues/11402#issuecomment-760350837

ameukam:
ghproxy is up and running on `k8s-infra-prow-builder-cluster-trusted`. 
[...]

PR #921, opened 2020-05-29
Add nodepool for ghproxy

ameukam:
Add a GKE nodepool to `prow-build-trusted` GKE cluster.

Ref: #843 

Similar work: https://github.com/kubernetes/test-infra/blob/master/config/clusters/k8s-prow/cluster-configuration.tf#L94-L136

### DIFF
--- a/infra/gcp/clusters/projects/k8s-infra-prow-build-trusted/prow-build-trusted/main.tf
+++ b/infra/gcp/clusters/projects/k8s-infra-prow-build-trusted/prow-build-trusted/main.tf
@@ -104,3 +104,20 @@ module "prow_build_nodepool" {
   disk_type       = "pd-standard"
   service_account = module.prow_build_cluster.cluster_node_sa.email
 }
+
+module "ghproxy_nodepool" {
+  source       = "../../modules/gke-nodepool"
+  project_name = local.project_id
+  cluster_name = module.prow_build_cluster.cluster.name
+  location     = module.prow_build_cluster.cluster.location
+  name         = "ghproxy"
+  labels       = { dedicated = "ghproxy" }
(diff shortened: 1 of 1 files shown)


infra/gcp/clusters/projects/k8s-infra-prow-build-trusted/prow-build-trusted/main.tf
```diff
@@ -104,3 +104,20 @@ module "prow_build_nodepool" {
...
+  location     = module.prow_build_cluster.cluster.location
+  name         = "ghproxy"
+  labels       = { dedicated = "ghproxy" }
+  # NOTE: taints are only applied during creation and ignored after that, see module docs
+  taints          = [{ key = "dedicated", value = "ghproxy", effect = "NO_SCHEDULE" }]
+  min_count       = 1
+  max_count       = 1
+  machine_type    = "n1-standard-8"
+  disk_size_gb    = 100
+  disk_type       = "pd-ssd"
```
spiffxp, line 121: FWIW it looks like k8s-prow has this as pd-standard but I'm fine with this

spiffxp:
Found a typo, corrected locally and opened #923

ran:
```
terraform init
terraform plan
terraform apply
```

thockin:
Can we document somewhere (like in these files) WHY we need dedicated nodepools?  This is going to allocate hundreds of dollars/month of CPU and memory.

Do we need all 8 cores?
Do we need all 30GB of RAM?
Do we just need dedicated machines?  why?
Do we really need nodes in all zones, or are those "wasted"
Would a custom machine shape be more appropriate?

Also, we should start switching to E2 machines when possible.  More effective than N1 all around.

PR #924, opened 2020-05-29
Setup for groups reconciliation to happen via prow

spiffxp:
This just allows the service account to be used on k8s-infra-prow-build-trusted

Will modify groups in followup PR to allow running without a secret in config file


spiffxp:

PR #925, opened 2020-05-29
groups: add more convention tests, fixup groups.yaml

spiffxp:


spiffxp:

PR #926, opened 2020-05-29
Proof of concept for post-k8sio-groups

spiffxp:
Polish the README to mention post-k8sio-groups and add some members to trigger it

### DIFF
--- a/groups/Makefile
+++ b/groups/Makefile
@@ -22,5 +22,9 @@ endif
 default: run
 
 .PHONY: run
-run:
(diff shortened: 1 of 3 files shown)


groups/Makefile
```diff
@@ -22,5 +22,9 @@ endif
 default: run
 
 .PHONY: run
-run:
-	GO111MODULE=on go run reconcile.go $(runargs)
```
spiffxp, line 26: I think it's safe to assume this is on by default now

spiffxp:

Issue #460, opened 2019-11-10
groups/groups.yaml can become difficult to maintain


neolit123:
i foresee that groups/groups.yaml will face the fate of k/test-infra/testgrid/config.yaml :) - i.e. one big yaml with a few owners that is difficult to edit and causes conflicts in PRs.

ideally groups should be sharded on the owner side.
can group settings be delegated on the side of group owners and perhaps only the group names can be enumerated in groups/groups.yaml ?

/kind feature
/priority important-longterm

justaugustus:
Strong +1 for this!

cblecker:
Maybe longterm, but right now the token to manage groups is extremely highly privileged.. I think only dims and I have access to it (the process is a manual run). So either way, we have to funnel approvals through a couple folks anyways. Sharding the config will only get us partially the way there.

nikhita:
> So either way, we have to funnel approvals through a couple folks anyways. 

Can we go around this by setting up the sync as a postsubmit like we do for peribolos?

cblecker:
Peribolos has a bunch of safeguards in place to prevent locking itself out from GitHub. We don't have those kinds of safeguards in place on the groups sync script.. right now, it requires a dry run, manual verification of the changes, and then a real live run.

If we had better testing and some of those safeguards in place, I'd be way more comfortable/confident with us doing this.

thockin:
maybe we can break the monolith file into smaller files which are merged at run-time.  At least it would be easier to read?  Maybe even directory structure to separate RBAC groups from role groups from administrative groups

spiffxp:
https://github.com/kubernetes/k8s.io/pull/925 is a start at enforcing some conventions, though I'm not sure what "don't lock us out" tests we should have

spiffxp:
I added https://github.com/kubernetes/k8s.io/pull/927 to address @cblecker's particular concern above

Longer term I agree sharding the config would be cool.

I personally think we need more conventions documented and enforced before we take that next step.

https://github.com/kubernetes/k8s.io/pull/928 should reduce some of the maintenance burden

spiffxp:
Oh, one other note to address https://github.com/kubernetes/k8s.io/issues/460#issuecomment-552658697

There are four humans who directly have the org admin role, in the event that the @kubernetes.io groups disappear or stop working

nikhita:
Created https://github.com/kubernetes/k8s.io/pull/996

/assign

spiffxp:
[...]

PR #927, opened 2020-05-29
groups: hardcode some groups for paranoia

spiffxp:
ref: https://github.com/kubernetes/k8s.io/issues/460#issuecomment-552658697


spiffxp:

PR #930, opened 2020-05-31
scale prow-build pool 3x

spiffxp:
this is in response to @BenTheElder suggesting the build cluster is
running out of resources (ref: https://github.com/kubernetes/kubernetes/issues/91548#issuecomment-636502815)

had to expose initial_count or terraform was going to delete/recreate
nodepools

autoscaling should be growing nodes for us if we're out of resources,
which leads me to believe prowjobs aren't declaring the resources they
actually need

### DIFF
--- a/infra/gcp/clusters/modules/gke-nodepool/main.tf
+++ b/infra/gcp/clusters/modules/gke-nodepool/main.tf
@@ -32,7 +32,7 @@ resource "google_container_node_pool" "node_pool" {
   }
 
   // Autoscale the cluster as needed. Note if using a regional cluster these values will be multiplied by 3
-  initial_node_count = var.min_count
+  initial_node_count = var.initial_count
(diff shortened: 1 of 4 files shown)


spiffxp:
I ran `terraform apply` on this at ~3:20pm 2020-05-31, waiting to see if autoscaling bumps nodes to min_count

spiffxp:
https://cloud.google.com/kubernetes-engine/docs/concepts/cluster-autoscaler
> If the current node pool size is lower than the specified minimum or greater than the specified maximum when you enable autoscaling, the autoscaler waits to take effect until a new node is needed in the node pool or until a node can be safely deleted from the node pool.

So I might be waiting a while.  Resizing through the UI from 8 (2-3 per zone) to 18 (6 per zone)

spiffxp:
resize complete

PR #835, opened 2020-05-03
releng: Backfill most recent debian-base, debian-iptables, and pause

justaugustus:
- Backfill most recent `debian-base`, `debian-iptables`, and `pause`

  `pull-kubernetes-cross` expects `debian-base` and `debian-iptables` to be in
  the same registry. As we haven't promoted a new `debian-iptables` image
  yet, we need to backfill the most recent version into community infra.
  
  We do the same for the `debian-base` and pause images to support anyone
  who wants to move to the new image registry without changing image
  versions as well.

  Versions backfilled:
  - `debian-base` (all arches): v2.0.0
  - `debian-iptables` (all arches): v12.0.1
  - `pause` (all arches): 3.2

- Promote all arches for `debian-base-**:v2.1.0`
  
  `pull-kubernetes-cross` expects base images in the form:
  `image-<arch>:version`, which means we need to push all image names,
  despite the fat manifest.

  We can refactor the release libs at a later date, but for now, this
  should unblock build/promotion of `debian-iptables`.

Signed-off-by: Stephen Augustus <saugustus@vmware.com>

/hold **until I actually backfill the images to staging**
/assign @dims @cblecker @BenTheElder @listx 
cc: @kubernetes/release-engineering 
/sig release
/area release-eng dependency security

PR #934, opened 2020-06-02
kubernetes-csi: promote existing releases

pohly:
These images were previously published on quay.io.

This is the first time that we are using the image promoter and we still had to define approvers.

### DIFF
--- a/k8s.gcr.io/images/k8s-staging-csi/images.yaml
+++ b/k8s.gcr.io/images/k8s-staging-csi/images.yaml
@@ -1,0 +1,31 @@
+- name: csi-attacher
+  dmap:
+    "sha256:2ffa647e8107cfd39e5f464e738dce014c9f5e51b108da36c3ab621048d0bbab": [ "v2.2.0" ]
+- name: csi-node-driver-registrar
+  dmap:
+    "sha256:9622c6a6dac7499a055a382930f4de82905a3c5735c0753f7094115c9c871309": [ "v1.3.0" ]
+    "sha256:273175c272162d480d06849e09e6e3cdb0245239e3a82df6630df3bc059c6571": [ "v1.2.0" ]
+- name: csi-provisioner
+  dmap:
+    "sha256:78e3393f5fd5ff6c1e5dada2478cfa456fb7164929e573cf9a87bf6532730679": [ "v1.6.0" ]
+- name: csi-resizer
+  dmap:
+    "sha256:6c6a0332693a7c456378f6abd2bb40611826c1e1a733cadbdae2daab3125b71c": [ "v0.5.0" ]
+    "sha256:43195976fb9f94d943f5dd9d58b8afa543be22d09a1165e8a489b7dfe22c657a": [ "v0.4.0" ]
+- name: csi-snapshotter
+  dmap:
+    "sha256:35ead85dd09aa8cc612fdb598d4e0e2f048bef816f1b74df5eeab67cd21b10aa": [ "v2.1.0" ]
+    "sha256:1530db9fd380ecee86b48a7b3540568c46adb5a64b7fe718c51d9260c4834fd8": [ "v2.0.1" ]
+- name: hostpathplugin
+  dmap:
+- name: livenessprobe
+  dmap:
+    "sha256:f8cec70adc74897ddde5da4f1da0209a497370eaf657566e2b36bc5f0f3ccbd7": [ "v1.1.0" ]
+- name: mock-driver
+  dmap:
+    "sha256:0b4273abac4c241fa3d70aaf52b0d79a133d2737081f4a5c5dea4949f6c45dc3": [ "v3.1.0" ]
+- name: nfs-provisioner
+  dmap:
+    "sha256:c1bedac8758029948afe060bf8f6ee63ea489b5e08d29745f44fab68ee0d46ca": [ "v2.2.2" ]
+- name: snapshot-controller
+  dmap:
--- a/k8s.gcr.io/images/k8s-staging-csi/images.yaml
+++ b/k8s.gcr.io/images/k8s-staging-csi/images.yaml
@@ -24,8 +24,5 @@
 - name: mock-driver
   dmap:
     "sha256:0b4273abac4c241fa3d70aaf52b0d79a133d2737081f4a5c5dea4949f6c45dc3": [ "v3.1.0" ]
-- name: nfs-provisioner
-  dmap:
-    "sha256:c1bedac8758029948afe060bf8f6ee63ea489b5e08d29745f44fab68ee0d46ca": [ "v2.2.2" ]
 - name: snapshot-controller
   dmap:
--- a/k8s.gcr.io/images/k8s-staging-csi/generate.sh
+++ b/k8s.gcr.io/images/k8s-staging-csi/generate.sh
@@ -1,0 +1,21 @@
+#! /bin/sh
+
+# List of repos under https://console.cloud.google.com/gcr/images/k8s-staging-csi/GLOBAL
+repos="
+csi-attacher
+csi-node-driver-registrar
+csi-provisioner
+csi-resizer
+csi-snapshotter
+hostpathplugin
+livenessprobe
+mock-driver
+snapshot-controller
+"
+
+for repo in $repos; do
+    echo "- name: $repo"
+    echo "  dmap:"
+    gcloud container images list-tags gcr.io/k8s-staging-csi/$repo --format='get(digest, tags)' --filter='tags~v.* AND NOT tags~v2020.* AND NOT tags~.*rc.*' |
+        sed -e 's/\([^ ]*\)\t\(.*\)/    "\1": [ "\2" ]/'
+done


pohly:
@msau42: the `images.yaml` file was created with this script:
```
#! /bin/sh

# List of repos under https://console.cloud.google.com/gcr/images/k8s-staging-csi/GLOBAL
repos="
csi-attacher
csi-node-driver-registrar
csi-provisioner
csi-resizer
csi-snapshotter
hostpathplugin
livenessprobe
mock-driver
nfs-provisioner
snapshot-controller
"

for repo in $repos; do
    echo "- name: $repo"
    echo "  dmap:"
    gcloud container images list-tags gcr.io/k8s-staging-csi/$repo --format='get(digest, tags)' --filter='tags~v.* AND NOT tags~v2020.*' |
        sed -e 's/\([^ ]*\)\t\(.*\)/    "\1": [ "\2" ]/'
done
```

I am not sure how you want to handle this in the future; if this script would be useful, then I can add it as k8s.gcr.io/images/k8s-staging-csi/generate.sh to this repo.

Beware that I have *not* verified whether the hashes match the ones from quay.io. I am assuming that whoever pushed to staging hasn't made a mistake or maliciously pushed different images.

pohly:
/hold

I think it makes sense to promote these images, but let's hear from @msau42 whether she agrees.

pohly:
> Beware that I have not verified whether the hashes match the ones from quay.io. I am assuming that whoever pushed to staging hasn't made a mistake or maliciously pushed different images.

I think we can trust @dims: https://github.com/kubernetes/kubernetes/pull/91257

:grin:

dims:
LOL, i am 100% legit :)

msau42:
I don't think we should promote these images that were copied over from quay because they were not built using the same infra. Also, we shouldn't promote the "-rc*" images which are not meant for public consumption.

pohly:
> I don't think we should promote these images that were copied over from quay because they were not built using the same infra.

Does that matter? We expect consumers of those images to trust that they contain what we claim that they contain, regardless of how they were built or where they are currently hosted.

The advantage of promoting the images is that consumers get a choice between pulling from the potentially flaky quay.io and the (hopefully) more reliable gcr.io; that's my motivation for this PR. We still see test failures from intermittent image pull errors for the sidecars from quay.io. We could do the same as the Kubernetes CI and pull from the staging area, but that feels wrong.

msau42:
Hosting an image in k8s infra means that we've used k8s infra to build them. We can't guarantee that with the images we temporarily copied from quay.

If we don't want to wait until we release the next set of sidecars to use this, then can we manually trigger building and pushing these tags in cloudbuild (and update k8s CI to not use any "rc" tags)?

justaugustus:
> Hosting an image in k8s infra means that we've used k8s infra to build them. We can't guarantee that with the images we temporarily copied from quay.

Not exactly. AFAIK, it means that an image was pushed to staging and promoted to prod, but some trusted set of people. There are likely a few subprojects that have a staging project and don't yet use GCB for building/pushing.

> If we don't want to wait until we release the next set of sidecars to use this, then can we manually trigger building and pushing these tags in cloudbuild (and update k8s CI to not use any "rc" tags)?

Agreed on the fact that you probably don't need the RC tags, but I don't think the images should be rebuilt.

For context, I've backfilled a few images for RelEng.
Here's one of those PRs: https://github.com/kubernetes/k8s.io/pull/835

I'd suggest backfilling one or two of the most recent images; we did this to ensure that people who might be changing their registry endpoint could move to K8s Infra w/o necessarily needing to move to a newer image at the same time.

One thing to be careful of, as @listx [mentions](https://github.com/kubernetes/k8s.io/pull/835#issuecomment-623781909):
> @justaugustus Your backfill script should use gcloud container images add-tag in the future to avoid pulling/pushing with docker. Using docker risks changing the digests if your workstation's gzip lib is different than the one used by GCR. I just did a check now and indeed the digest changed, for example from gcr.io/google-containers/debian-base@sha256:ebda8587ec0f49eb88ee3a608ef018484908cbc5aa32556a0d78356088c185d4 to gcr.io/k8s-staging-build-image/debian-base@sha256:d7be39e143d4e6677a28c81c0a84868b40800fc979dea1848bb19d526668a00c for the v2.0.0 tag.

There's a good chance your digests will change, which might be unavoidable, but be sure to check up on any code you have that's doing digest verification.

pohly:
> There's a good chance your digests will change, which might be unavoidable, but be sure to check up on any code you have that's doing digest verification.

We have no such code in Kubernetes or the sidecars. We have advocated using the version tags and promised that we will never rebuilt those images, so I suspect most users will use that to pull a deterministic image.

But it wouldn't matter anyway, because it seems that the sha256 digests have not changed. I only checked one image (csi-node-driver-registrar:v1.3.0) and it had the same digest on quay.io (https://quay.io/repository/k8scsi/csi-node-driver-registrar?tab=tags) and gcr.io (https://console.cloud.google.com/gcr/images/k8s-staging-csi/GLOBAL/csi-node-driver-registrar@sha256:9622c6a6dac7499a055a382930f4de82905a3c5735c0753f7094115c9c871309/details?tab=info).

pohly:
If there's consensus that promoting the release images is worthwhile, then I can update the PR to exclude the -rc images.

msau42:
Sounds like there's precedent to backfilling images when migrating over to gcr, so promoting this older tags (minus rc) sgtm. Thanks @justaugustus for the clarification!

pohly:
I've used an updated script to re-generate the images.yaml file. The script now compares the checksums or (if those differ) the content of the images to verify that they are the same as the ones previously hosted on quay.io. I know that we can trust @dims, but it doesn't hurt to be extra careful, right? :stuck_out_tongue: 

The result is that newer images were moved entirely unchanged, while older ones were apparently repacked and thus have a different sha256. The content is exactly the same, though. nfs-provisioner wasn't on quay.io; I've included it anyway in the images.yaml:
```
OKAY: csi-attacher:v2.2.0: sha56 matches
OKAY: csi-node-driver-registrar:v1.3.0: sha56 matches
WARNING: csi-node-driver-registrar:v1.2.0: sha56 does not match.
         Content is the same.
OKAY: csi-provisioner:v1.6.0: sha56 matches
OKAY: csi-resizer:v0.5.0: sha56 matches
WARNING: csi-resizer:v0.4.0: sha56 does not match.
         Content is the same.
OKAY: csi-snapshotter:v2.1.0: sha56 matches
WARNING: csi-snapshotter:v2.0.1: sha56 does not match.
         Content is the same.
WARNING: livenessprobe:v1.1.0: sha56 does not match.
         Content is the same.
OKAY: mock-driver:v3.1.0: sha56 matches
Error response from daemon: unauthorized: access to the requested resource is not authorized
WARNING: quay.io/k8scsi/nfs-provisioner:v2.2.2 not found.
```

Script:
```sh
#! /bin/sh

# List of repos under https://console.cloud.google.com/gcr/images/k8s-staging-csi/GLOBAL
repos="
csi-attacher
csi-node-driver-registrar
csi-provisioner
csi-resizer
csi-snapshotter
hostpathplugin
livenessprobe
mock-driver
nfs-provisioner
snapshot-controller
"

tmp=$(mktemp -d)
trap "rm -rf $tmp" EXIT

for repo in $repos; do
    echo "- name: $repo"
    echo "  dmap:"
    gcloud container images list-tags gcr.io/k8s-staging-csi/$repo --format='get(digest, tags)' --filter='tags~v.* AND NOT tags~v2020.* AND NOT tags~.*-rc.*' | while read sha tag; do
        quay=$(docker pull quay.io/k8scsi/$repo:$tag | grep ^Digest: | sed -e 's/^Digest: //')
        if [ "$sha" != "$quay" ]; then
            if [ "$quay" ]; then
                echo >&2 "WARNING: $repo:$tag: sha56 does not match."
                container=$(docker create gcr.io/k8s-staging-csi/$repo:$tag)
                mkdir $tmp/gcr.io
                docker export $container | tar -C $tmp/gcr.io -xf -
                docker rm $container >/dev/null
                container=$(docker create quay.io/k8scsi/$repo:$tag)
                mkdir $tmp/quay.io
                docker export $container | tar -C $tmp/quay.io -xf -
                docker rm $container >/dev/null
                if diff -r $tmp/quay.io $tmp/gcr.io >&2; then
                    echo >&2 "         Content is the same."
                else
                    echo >&2 "         Content differs."
                fi
                rm -rf $tmp/quay.io $tmp/gcr.io
            else
                echo >&2 "WARNING: quay.io/k8scsi/$repo:$tag not found."
            fi
        else
            echo >&2 "OKAY: $repo:$tag: sha56 matches"
        fi
        echo "    \"$sha\": [ \"$tag\" ]"
    done
done
```

pohly:
Also note that we need to do something about Kubernetes CI: the staging images will be deleted after 60 days, i.e. around July 19th. Let's promote the images and then switch the CI to use the promoted images.

dims:
@pohly 100% +1 to "trust but verify" :)

k8s.gcr.io/images/k8s-staging-csi/images.yaml
```diff
@@ -0,0 +1,31 @@
...
+    "sha256:1530db9fd380ecee86b48a7b3540568c46adb5a64b7fe718c51d9260c4834fd8": [ "v2.0.1" ]
+- name: hostpathplugin
+  dmap:
+- name: livenessprobe
+  dmap:
+    "sha256:f8cec70adc74897ddde5da4f1da0209a497370eaf657566e2b36bc5f0f3ccbd7": [ "v1.1.0" ]
+- name: mock-driver
+  dmap:
+    "sha256:0b4273abac4c241fa3d70aaf52b0d79a133d2737081f4a5c5dea4949f6c45dc3": [ "v3.1.0" ]
+- name: nfs-provisioner
```
msau42, line 27: I'm not sure we should promote nfs-provisioner since it's not related to csi. We should find a better permanent home for it. cc @kmova

msau42:
I think the script would be useful to check in, to make promoting future images easier.

pohly:
Removed nfs-provisioner, added the script in the version without the comparison against quay.io (that was a one-time thing).

PR #939, opened 2020-06-04
Deploy ghproxy

ameukam:
Part of https://github.com/kubernetes/k8s.io/issues/843.

### DIFF
--- a/infra/gcp/clusters/projects/k8s-infra-prow-build-trusted/prow-build-trusted/main.tf
+++ b/infra/gcp/clusters/projects/k8s-infra-prow-build-trusted/prow-build-trusted/main.tf
@@ -113,20 +113,3 @@ module "prow_build_nodepool" {
   service_account = module.prow_build_cluster.cluster_node_sa.email
 }
 
-module "ghproxy_nodepool" {
-  source          = "../../../modules/gke-nodepool"
-  project_name    = local.project_id
-  cluster_name    = module.prow_build_cluster.cluster.name
-  location        = module.prow_build_cluster.cluster.location
-  name            = "ghproxy"
-  labels          = { dedicated = "ghproxy" }
-  # NOTE: taints are only applied during creation and ignored after that, see module docs
-  taints          = [{ key = "dedicated", value = "ghproxy", effect = "NO_SCHEDULE" }]
-  initial_count   = 1
-  min_count       = 1
-  max_count       = 1
-  machine_type    = "n1-standard-8"
-  disk_size_gb    = 100
-  disk_type       = "pd-ssd"
-  service_account = module.prow_build_cluster.cluster_node_sa.email
-}
--- a/infra/gcp/clusters/projects/k8s-infra-prow-build-trusted/prow-build-trusted/resources/ghproxy-service.yaml
+++ b/infra/gcp/clusters/projects/k8s-infra-prow-build-trusted/prow-build-trusted/resources/ghproxy-service.yaml
@@ -1,0 +1,32 @@
+---
+apiVersion: v1
+kind: Service
+metadata:
+  namespace: default
+  name: ghproxy
+  labels:
+    app: ghproxy
+spec:
+  ports:
+    - name: main
+      port: 80
+      protocol: TCP
+      targetPort: 8888
+  selector:
+    app: ghproxy
+  type: ClusterIP
+---
+apiVersion: v1
+kind: Service
+metadata:
+  name: ghproxy-metrics
+  namespace: default
+  labels:
+    app: ghproxy
+spec:
+  selector:
+    app: ghproxy
+  ports:
+    - name: metrics
+      port: 9090
+  type: LoadBalancer
--- a/infra/gcp/clusters/projects/k8s-infra-prow-build-trusted/prow-build-trusted/resources/ghproxy-deployment.yaml
+++ b/infra/gcp/clusters/projects/k8s-infra-prow-build-trusted/prow-build-trusted/resources/ghproxy-deployment.yaml
@@ -1,0 +1,36 @@
+---
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  namespace: default
+  name: ghproxy
+  labels:
+    app: ghproxy
+spec:
+  selector:
+    matchLabels:
+      app: ghproxy
+  replicas: 1 # TODO(fejta): this should be HA
+  template:
+    metadata:
+      labels:
+        app: ghproxy
+    spec:
+      containers:
+        - name: ghproxy
+          image: gcr.io/k8s-prow/ghproxy:v20200611-4c51f3fcb1
+          args:
+            - --cache-dir=/cache
+            - --cache-sizeGB=99
+            # TODO: uncomment once the monitoring.prow.k8s.io and pushgateway are migrated
+            #- --push-gateway=pushgateway
+            - --serve-metrics=true
+          ports:
+            - containerPort: 8888
+          volumeMounts:
+            - name: cache
+              mountPath: /cache
+      volumes:
+        - name: cache
+          persistentVolumeClaim:
+            claimName: ghproxy
--- a/infra/gcp/clusters/projects/k8s-infra-prow-build-trusted/prow-build-trusted/resources/ghproxy-storage.yaml
+++ b/infra/gcp/clusters/projects/k8s-infra-prow-build-trusted/prow-build-trusted/resources/ghproxy-storage.yaml
@@ -1,0 +1,14 @@
+kind: PersistentVolumeClaim
+apiVersion: v1
+metadata:
+  namespace: default
+  labels:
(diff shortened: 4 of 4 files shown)


infra/gcp/ensure-prow-build-trusted.sh
```diff
@@ -0,0 +1,47 @@
...
+    exit 1
+fi
+
+ADDRESSES=(
+    "ghproxy-ingress-prod"
+)
+
+for address in "${ADDRESSES[@]}"; do
+    color 6 "Ensure address: $address"
+    ensure_global_address "$PROJECT_NAME" "$address" "IP for ghproxy Ingress"
```
spiffxp, line 45: How would you feel about trying to do this via terraform instead of via bash?

Alternatively, I guess I'd rather see `ensure-e2e-projects.sh` renamed to something like `ensure-prow-infra.sh` and have this moved into it

infra/gcp/clusters/projects/k8s-infra-prow-build-trusted/prow-build-trusted/main.tf
```diff
@@ -125,7 +125,7 @@ module "ghproxy_nodepool" {
   initial_count   = 1
   min_count       = 1
   max_count       = 1
-  machine_type    = "n1-standard-8"
+  machine_type    = "e2-standard-4"
```
spiffxp, line 128: What's the reasoning behind this change? I suspect this will constrain IOPS which may not be good for a caching service.

infra/gcp/ensure-prow-build-trusted.sh
spiffxp: ok wait actually I don't think we want this at all, we shouldn't be exposing ghproxy

infra/gcp/clusters/projects/k8s-infra-prow-build-trusted/prow-build-trusted/resources/ghproxy-ingress.yaml
```diff
@@ -0,0 +1,11 @@
...
+kind: Ingress
+metadata:
+  name: ghproxy
+  namespace: default
+  annotations:
+    kubernetes.io/ingress.global-static-ip-name: "ghproxy-ingress-prod"
+spec:
+  backend:
+    serviceName: ghproxy
+    servicePort: main
```
spiffxp, line 11: We do not want this exposed as a service outside of the cluster at all; we should only be caching for services and jobs deployed in this cluster, not the internet.

Please take a look at how ghproxy is currently deployed for prow: https://github.com/kubernetes/test-infra/blob/master/config/prow/cluster/ghproxy.yaml

infra/gcp/clusters/projects/k8s-infra-prow-build-trusted/prow-build-trusted/resources/ghproxy-service.yaml
```diff
@@ -0,0 +1,31 @@
...
+  ports:
+    - name: main
+      port: 80
+      protocol: TCP
+      targetPort: 8888
+    - name: metrics
+      port: 9090
+  selector:
+    app: ghproxy
+  type: NodePort
```
spiffxp, line 19: This should be `ClusterIP`, we don't want this exposed outside of the cluster

We may want metrics available outside of this cluster, I would create a separate service for that endpoint.  See how this is done for boskos or greenhouse for examples

infra/gcp/clusters/projects/k8s-infra-prow-build-trusted/prow-build-trusted/main.tf
spiffxp: OK, I see https://github.com/kubernetes/k8s.io/pull/921#issuecomment-636195722

Thinking about this a bit more, I suppose we can try not blindly replicating the setup used by prow.k8s.io for this setup.  

The caching here is going to be for the benefit of jobs, so I'm not sure I expect super high performance or super high availability.  How about we consider dropping this separate pool entirely, and just deploying as a resource on the existing node pool?

infra/gcp/ensure-prow-build-trusted.sh
ameukam: I was looking for a way to use google/triage-party running in `aaa` again ghproxy. If ghproxy is not exposed how workloads like triage-party can use it ?

spiffxp: Sorry it wasn't clear to me triage party is why you were setting this up.  The recommendation I got from prow folks was to keep ghproxy constrained to within a cluster.  So if your intent is to run ghproxy for triage party, I would run an instance of ghproxy on the `aaa` cluster.

We still need an instance of `ghproxy` on this cluster for the benefit of peribolos jobs etc.

infra/gcp/clusters/projects/k8s-infra-prow-build-trusted/prow-build-trusted/main.tf
thockin: I'm in favor of less dedicated pools, when possible.  It should be last-resort in most cases.

infra/gcp/clusters/projects/k8s-infra-prow-build-trusted/prow-build-trusted/resources/ghproxy-statefulset.yaml
```diff
@@ -0,0 +1,62 @@
...
+  volumeClaimTemplates:
+    - metadata:
+        name: ghproxy
+      spec:
+        accessModes:
+          - ReadWriteOnce
+        resources:
+          requests:
+            storage: 100Gi
+        storageClassName: gce-ssd-retain
```
thockin, line 62: Why do we want to retain these PVs after they have been released?   Are we under some litigation-hold or do we do some post-facto analysis?

infra/gcp/ensure-prow-build-trusted.sh
thockin: Or we can expose it as an "internal" Service on an ILB, which lets a service be used by other members of the VPC?

I don't quite know what is needed here, so guessing a bit.

infra/gcp/clusters/projects/k8s-infra-prow-build-trusted/prow-build-trusted/resources/ghproxy-statefulset.yaml
ameukam: I'm not sure about that, but I guess it prevents the cache from being cleared when ghproxy is upgraded by k8s-ci-robot.

infra/gcp/clusters/projects/k8s-infra-prow-build-trusted/prow-build-trusted/main.tf
ameukam: PR updated with nodepool for ghproxy removed.

infra/gcp/clusters/projects/k8s-infra-prow-build-trusted/prow-build-trusted/resources/ghproxy-statefulset.yaml
spiffxp: I think it's so we can avoid having to warm the cache if we have to kill/reschedule the ghproxy instance, @cjwagner does that sound right to you?

thockin: That's not what retain means.  As long as you don't delete a PVC, the PV is yours to use and reuse.  Once you delete the PVC, the retention policy kicks in.  If you say "retain", you will "leak" the underlying disk until someone comes along and manually deletes it.

cjwagner: Tim is right, retain shouldn't be used here. That was originally included because I misunderstood that concept.
Other Prow instances just use the default storage class (non-SSD) and it hasn't been a problem. If we want an SSD we'll still need to define a storage class, but it should not include `reclaimPolicy: Retain`.

ameukam: @cjwagner @thockin thank you for your explanations!
I updated the PR. the PVC will use the default `StorageClass`.

spiffxp:
Will try to deploy this week

spiffxp (approved):
/approve
/lgtm
Merging this for the nodepool changes, I'm not sure that I have time to verify that ghproxy is working correctly, may revert that part

spiffxp:
Ran `./infra/gcp/ensure-staging-storage.sh addon-manager`

spiffxp:
I ran `terraform apply`, have not checked the deployment itself

PR #940, opened 2020-06-04
kubernetes-csi: promote first multi-arch mock-driver release

pohly:
csi-test v3.1.1 produced the first multi-arch image. We probably want
to promote only the combined image (tag: v3.1.1) and not the
individual ones for each architecture (tags: amd64-linux-v3.1.1,
ppc64le-linux-v3.1.1, s390x-linux-v3.1.1).

This showed that the `tags~` regex matching does a sub-string match
and not a full match, so the script gets updated a bit to simplify
some terms and tighten the one that matches versions.

/cc @msau42

### DIFF
--- a/k8s.gcr.io/images/k8s-staging-csi/OWNERS
+++ b/k8s.gcr.io/images/k8s-staging-csi/OWNERS
@@ -1,0 +1,9 @@
+# See the OWNERS docs at https://go.k8s.io/owners
+
+# Current set of people in sig-storage-leads from
+# https://github.com/kubernetes/community/blob/master/OWNERS_ALIASES.
+approvers:
+- jsafrane
+- msau42
+- saad-ali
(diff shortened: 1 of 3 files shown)


pohly:
The OWNERS file should have been in the previous PR #934 but I accidentally dropped it while reworking that PR. Now we need approval from one of the repo owners one more (and hopefully then the last) time.

k8s.gcr.io/images/OWNERS
```diff
@@ -0,0 +1,9 @@
+# See the OWNERS docs at https://go.k8s.io/owners
```
msau42, line 1: should this be one level down in "k8s-staging-csi"?

pohly: Oops, yes. Fixed.

Issue #823, opened 2020-04-30
Refactor dns/push.sh script to let call octodns directly - without docker involved


bartsmykla:
As @spiffxp suggested at [#783#issuecomment-618007941](https://github.com/kubernetes/k8s.io/issues/783#issuecomment-618007941) we should refactor the [`dns/push.sh`](https://github.com/kubernetes/k8s.io/blob/master/dns/push.sh)  script to be able to call it from the container or locally which would involve invoking container.

My suggestion is to add few flags which would let us use make like:
```shell
PUSH_FLAGS="--from-container --dry-run --config-file=./dns/octodns-config.yaml" make push
```

but I'm open to suggestions

/cc @spiffxp

PR #949, opened 2020-06-10
Removing docker from dns shell scripts

bartsmykla:
- Moved out check-zone.py from octodns-docker 
   - Logically when we'll have the option to run the octodns locally and
      not only via docker container it's better place to store it
   - Modified dns/octodns-docker/Dockerfile and removed copying of the
      check-zone.py file, which is now one level above
- Removed docker from push.sh and check-zone.sh
   - Now the script can be called locally without involvement of docker,
     which will be useful in automation of dns reconciliation
   - As precooking of zone config files is now done at the same layer as
     pushing and without involvement of docker had to include logic to
     dynamically generate octodns-config.yaml file with proper
     providers.config.directory parameter pointing to directory with cooked
     zone config files
   - The same reason stands behind the additional "-o" flag I had to add to
     dns/check-zone.sh script, which now had to have octod
- Added option to specify logs path to push.sh
   - Now if there is $ARTIFACTS set it will be used as a place to store log
      files, which is important for automation purposes as prow is providing
      that variable and then it will store files from this path and we'll be
      able to see them at prow.k8s.io
   - Also aded logic to check if LOGS_PATH is an existing directory and if
      our script has permissions to write to it
- Updated Makefile with two new targets - dry-run-local and push-local
- Moved out config precooking from `push.sh` to `lib.sh`
   - As now check-zone.sh has to have precooked zone configs and also
      octodns config it was working only if called from push.sh but not from
      make so I moved the functions responsible for precooking to lib.sh file
      and updated push.sh and also Makefile respectively

/cc @thockin @spiffxp 
/closes: #823

ref. #783


dns/Makefile
```diff
@@ -63,3 +68,8 @@ push-local:
 .PHONY: dry-run-local
 dry-run-local: ARGS := --dry-run
 dry-run-local: push-local
+
+precook-configs:
+	source ./lib.sh; \
```
thockin, line 73: I have to admit, I have never in my life `source`d a shell script from within a `make` rule.  It never would have crossed my mind.  You win, Bart.

spiffxp:
```
$ make dry-run
docker run -ti \
		--user 579674 \
		--volume ~/.config/gcloud:/.config/gcloud:ro \
		--volume /Users/spiffxp/w/kubernetes/k8s.io/dns:/octodns \
		--workdir /octodns \
		us.gcr.io/k8s-artifacts-prod/infra-tools/octodns:v20200501-36789b1 \
		./push.sh \
			--dry-run
env: can't execute 'bash': No such file or directory
make: *** [push] Error 127
```

spiffxp:
OK well I guess I answered @thockin's question of why the bash.  A better answer might be whether there is bash-specific stuff happening inside the script that's been added.  I'm happy with this once that's been addressed

bartsmykla:
@spiffxp @thockin the flow changed, as before it was:
```
push.sh -> docker -> octodns-sync + check-zone.py
```
and now there are two flows (local and from container):
```
push.sh -> octodns-sync + check-zone.py
docker -> push.sh -> octodns-sync + check-zone.py
```
which means that now container is running our scripts. It is possible to rewrite the scripts to use only POSIX compatible features, but it would mean we have to get rid of arrays and rely on strings only, which I'm not saying is terrible but we need to decide between just having bash in container or refactoring the scripts and removing the arrays + few smaller tweaks (removing flags from `echo`s, removing `function` keywords, changing `source` to `.` etc.)

BenTheElder:
Using it this way seems like a recipe for non-reproducible behavior? Looks like we don't even do a simple version check? 😬

bartsmykla:
@BenTheElder I’m not sure what you mean by version check in this context, could you write more?

bartsmykla:
You mean version check of octodns?

bartsmykla:
The mail goal is to automate it so in the future people would not have to call it manually and if there will be the case the persol woll have to do it it still can be done via docker as the main make targets are called via it, I think it is acceptable at this stage

thockin:
Thanks!

/lgtm
/approve

BenTheElder:
> The main goal is to automate it so in the future people would not have to call it manually and if there will be the case the person will have to do it, it still can be done via docker as the main make targets are using docker. I think it is acceptable but don’t have anything against removing the local make targets

that's fair, my concern was that people might be using this locally for actual deployments w/o any control over the versions etc..

it also seems fairly reasonable to require containers in our space :+)

PR #955, opened 2020-06-14
backfill from google-containers

listx:
Generated with:

    cip-mm -base_dir $PWD/k8s.gcr.io -staging_repo gcr.io/google-containers

This is another backfill of the contents from gcr.io/google-containers. The diff is rather large because of the stylistic difference with the new cip-mm command.


bartsmykla:
I don't think it's possible to review it so I trust you @listx everything is fine :-)

/approve
/lgtm

thockin:
@listx this triggered a whole load of CIP alerts last night - Can you PTAL at why?

thockin:
(not on the weekend)

listx:
Briefly, there is at least 1 issue which might be related to these log events (I need to dig a little more to establish the causation link). But we are exceeding GCR quota, leading to GCR read errors.

thockin:
Cool.  Are we rate-limiting?  Is this going to trigger on every push?

listx:
This will trigger on any push that pushes up too many new images at once. We are not rate-limiting ourselves, no --- we can certainly add that logic to mitigate this.

Every single image insertion triggers the auditor, which, *for every insertion*, reads all repos defined in the promoter manifests. This means, for example, reading the entirety of all staging repos (in particular, the massive gcr.io/google-containers). We are forced to do this read because promoter manifests often only specify the parent digest of a fat manifest. So when the auditor sees a child image with an unknown SHA not seen in the promoter manifest yamls, it has to read the staging repo responsible for the image to see if its *parent* fat manifest digest is tracked in the promoter manifests.

If we forced users to also track child images in the manifests (this is possible if, for example, we popularize a tool to automate this), then we wouldn't even need to read from GCR at all and the network read failures here would be categorically eliminated. This tool can happen in the future (maybe?) but for now I think we should just work on better rate-limiting and other measures.

thockin:
Should we look into batching of updates (e.g. collect all changes over the
last 10 minutes), too?

listx:
Batching sounds good but it also adds complexity. There is another issue which is the proliferation of identical pubsub messages (we keep verifying the same message multiple times, because pub/sub keeps sending us the same message); need to fix this as well (I think it can be reproduced in an e2e test).

listx:
OK so I understand why we are seeing duplicate messages. It's because we are not always responding with a 2xx or 4xx/5xx (error) for each request within the 10-second deadline (current deadline for the cip-auditor-invoker subscriber). We can easily go beyond the 10-second deadline if a GCR read fails and we retry the GCR read call with ExponentialBackoff.

We can fix this by increasing the PubSub ack deadline beyond the limit we have set for the API call retries. It appears we wait ~15 minutes before calling it quits for each GCR read call https://github.com/kubernetes-sigs/k8s-container-image-promoter/blob/078d8ab561fd60ffe8e460cc4ba91e765a5716de/lib/stream/types.go#L58. So, the subscription deadline should be increased to >15 minutes (how about 20 minutes?). This will give the auditor time to "tap out" from retry failures before PubSub sends another copy of the same message (although in practice, if the auditor does "tap out" in time, it will send a 4xx code back, letting PubSub know to re-send the message again).

Also I found that Cloud Run has a quota of 15 min max for timeouts per request (https://cloud.google.com/run/quotas). So it looks like we cannot go higher than that for purposes of "tapping out" for GCR read failures. If we don't respond within Cloud Run's 15 min quota limit, our CPU resources get severely restricted (this is by design --- Cloud Run doesn't want us to run arbitrarily long computations). https://stackoverflow.com/q/57365044/437583. In other words, we got 15 minutes of compute time for each PubSub message before PubSub automatically resends the message to the auditor (assuming that the message is not over 7 days old, which is the current retention period we have set for the subscription).

**I think we want our PubSub ack deadline to be 20 minutes**. Then if we exceed the Cloud Run quota of 15 min per request, and for example if we take more than the 20 minute ack deadline to return an error code, presumably PubSub will know that for that *particular* message, the deadline has expired and will ignore our HTTP error code which normally should trigger a resend. On the other hand, if we take less than the 15 min quota to process the message, then it's fine because we'll either succeed (message is processed once and for all), or fail (message is asked to be re-sent by Cloud Run to PubSub).

I think that takes care of the message duplication issues. The other problem is how to batch requests. I'll think more about this tomorrow.

listx:
Another thing. I've noticed that even with Cloud Run autoscaling set to 1 (we don't scale! on purpose!), we see different UUIDs in the logs (UUIDs are assigned 1x at startup). This is most likely due to containers hitting memory limits. I've seen this in the graphs where we hit 100% memory utilization. According to the docs https://cloud.google.com/run/docs/configuring/memory-limits#understanding containers that hit the limit are terminated. This would further compound the error because it would nullify all requests that are in the middle of being handled within the 15min Cloud Run request response window. I've increased the memory allocation to 2GiB, up from 244MiB. As I cannot see any logs about a failure to delete Git clones (after cloning for each request), I think the increased memory usages are due to the outstanding PubSub messages that are being handled by the auditor. So increasing the PubSub ack deadline to 20 minutes as stated in my comment above should alleviate memory usage issues as well.

listx:
It appears that if we take more than 15 minutes to process a PubSub message, Cloud Run will terminate the request (whatever that means --- perhaps terminating our container...) and return a 504. This will in turn tell PubSub to retry the message, which is good. Docs here: https://cloud.google.com/run/docs/configuring/request-timeout

thockin:
I bumped the timeout to 10mins, which is the max allowed.

listx:
After thinking about this more, your suggestion of batching is the way to go. Until we get that implemented, there are some preventative measures to reduce the number of alerts (to avoid flooding the alert channels) that are being worked on in https://github.com/kubernetes-sigs/k8s-container-image-promoter/pull/223.

Regarding batching, I have a rough idea of what I want to do. The auditor will loop forever, and every 30 seconds (or some other interval), send all newly received HTTP reqs (i.e., the PubSub messages) to a new (different) list of PubSub messages to reconcile. And then it's just a matter of making sure that each list is resolved with a single GCR read. The important thing is that the GCR read happens at the end, so that we know for a fact that we are resolving old transactions.

PR #958, opened 2020-06-16
promote agnhost 2.19

BenTheElder:
.... I'm not sure how we got to the current state. I don't see 2.17 promoted but the tests use this??

cc @justaugustus xref https://github.com/kubernetes/kubernetes/pull/92119 https://github.com/kubernetes/kubernetes/issues/92059 https://github.com/kubernetes/kubernetes/pull/87105


spiffxp:
The current OWNERS for this image is not what I would have expected, opened https://github.com/kubernetes/k8s.io/pull/959

PR #963, opened 2020-06-16
Added dummy.k8s.io dns record to trigger prow job

bartsmykla:
As there is `k8sio-dns-update` job which should run dns updates in
dry-mode this temporary, dummy record is to check if the job will be
properly triggered and work as expected

/assign @spiffxp 

### DIFF
--- a/dns/zone-configs/k8s.io._0_base.yaml
+++ b/dns/zone-configs/k8s.io._0_base.yaml
@@ -22,6 +22,10 @@ www:
   type: CNAME
   value: k8s.io.
 
+dummy:
+  type: CNAME
+  value: k8s.io.
+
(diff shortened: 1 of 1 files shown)


bartsmykla:
Job succeeded: [prow](https://prow.k8s.io/view/gcs/kubernetes-jenkins/logs/k8sio-dns-update/1272824812418371584) and logs can be found in [artifacts](https://gcsweb.k8s.io/gcs/kubernetes-jenkins/logs/k8sio-dns-update/1272824812418371584/artifacts/)

bartsmykla:
I will remove the dummy record after #962 will be merged and I'll update the prow job to use `make` instead of calling the scripts directly, to also test it

PR #970, opened 2020-06-17
groups: add leads@kubernetes.io

nikhita:
Fixes https://github.com/kubernetes/community/issues/4673

https://groups.google.com/forum/#!forum/kubernetes-sig-leads has been prone
to getting stale and manually updating and pruning is tedious. Instead,
move this to gsuite so that it can be handled by automation.

---

The current google group has lots of stale members who are no longer SIG leads. They have been pruned from this list in groups.yaml (thanks for helping, @mrbobbytables :heart: ).

This list also contains 2 non SIG leads - @caniszczyk and `cncf-speakers` (Nanci) - to help with notifications regarding KubeCon, etc.

/cc @mrbobbytables @parispittman @spiffxp @cblecker @dims

### DIFF
--- a/groups/groups.yaml
+++ b/groups/groups.yaml
@@ -22,6 +22,86 @@ groups:
       - codyclark@google.com
       - vonguard@gmail.com
 
+  - email-id: leads@kubernetes.io
+    name: leads
+    description: |-
+      SIG/WG/UG Leads
+    settings:
+      WhoCanViewGroup: "ANYONE_CAN_VIEW"
+      WhoCanDiscoverGroup: "ANYONE_CAN_DISCOVER"
+      ReconcileMembers: "true"
+    owners:
+      - contributors@kubernetes.io
+    members:
+      - ahg@google.com
+      - alarcj137@gmail.com
+      - bartek@smykla.com
+      - bentheelder@google.com
+      - caniszczyk@linuxfoundation.org
+      - casey@tigera.io
+      - cblecker@gmail.com
+      - chiachenk@google.com
+      - cncf-speakers@linuxfoundation.org
+      - community@kubernetes.io
+      - davanum@gmail.com
+      - davidopp@google.com
+      - dawnchen@google.com
+      - dbosanac@redhat.com
+      - dbsmith@google.com
+      - dcbw@redhat.com
+      - deads@redhat.com
+      - decarr@redhat.com
+      - ehashman@redhat.com
+      - fbongiovanni@google.com
+      - frapposelli@vmware.com
+      - hankang@google.com
+      - hpandey@pivotal.io
+      - hweicdl@gmail.com
+      - jameswangel@gmail.com
+      - jbelamaric@google.com
+      - jeef111x@gmail.com
+      - jeremy.r.rickard@gmail.com
+      - jonathan.berkhahn@gmail.com
+      - jorgec@vmware.com
+      - justin@fathomdb.com
+      - kaitlynbarnard10@gmail.com
+      - kbhawkey@gmail.com
+      - killen.bob@gmail.com
+      - kim.andrewsy@gmail.com
+      - klaus1982.cn@gmail.com
+      - lachlan.evenson@gmail.com
+      - liggitt@google.com
+      - maszulik@redhat.com
+      - matt.farina@gmail.com
+      - michmike@gmail.com
+      - mikedanese@google.com
+      - mwielgus@google.com
+      - neolit123@gmail.com
+      - nikitaraghunath@gmail.com
+      - nospam.wong@gmail.com
+      - onlydole@gmail.com
+      - paris.pittman@gmail.com
+      - prydonius@gmail.com
+      - quinton@hoole.biz
+      - saadali@google.com
+      - saugustus@vmware.com
+      - seans@google.com
+      - spiffxp@gmail.com
+      - stephen.k8s@agst.us
+      - szostok.mateusz@gmail.com
+      - tallclair@google.com
+      - tasha.drew@gmail.com
+      - theenjeru@gmail.com
+      - thockin@google.com
+      - tim@scalefactory.com
+      - timothysc@gmail.com
+      - tpepper@vmware.com
+      - wfender@google.com
+      - wojtekt@google.com
+      - xingyang105@gmail.com
+      - ynli@google.com
+      - zcorleissen@linuxfoundation.org
+
   - email-id: community@kubernetes.io
     name: community
     description: |-


cblecker:
Do we want to allow anyone to post?

nikhita:
> Do we want to allow anyone to post?

@mrbobbytables mentioned that they get a lot of spam in contributors@ because of allowing everyone to post but I'll let him confirm

nikhita:
also, question for @parispittman @mrbobbytables @castrojo - what will happen to https://groups.google.com/forum/#!forum/kubernetes-sig-leads once the gsuite is setup? can we "archive" google groups?

mrbobbytables:
We get a lot of spam on community@kubernetes.io which allows anyone to post. The settings for the current group only allow members to post to the list. Here are the current settings:
Screen Shot 2020-06-17 at 3 34 09 PM

nikhita:
Should we also set:

```yaml
WhoCanViewGroup: "ANYONE_CAN_VIEW"
WhoCanDiscoverGroup: "ANYONE_CAN_DISCOVER"
```

so that non-members can still view content? Right now, anyone can view https://groups.google.com/forum/#!forum/kubernetes-sig-leads

nikhita:
just adding an explicit hold until we get the questions answered
/hold

groups/groups.yaml
```diff
@@ -22,6 +22,83 @@ groups:
...
       - vonguard@gmail.com
 
+  - email-id: chairs-and-techleads@kubernetes.io
+    name: chairs-and-techleads
+    description: |-
+      SIG/WG/UG Chairs and TLs
+    settings:
+      ReconcileMembers: "true"
+    owners:
+      - contributors@kubernetes.io
```
spiffxp, line 32: Why does this group need owners? I suspect the answer is "moderation" but it would be helpful to have it documented somehow.

Why is contributors@kubernetes.io not defined in this file?

nikhita: > Why does this group need owners? I suspect the answer is "moderation" but it would be helpful to have it documented somehow.

I'd be ok with removing it if we don't allow everyone to post. I'm not sure if there are other moderation powers that might be useful here.

> Why is contributors@kubernetes.io not defined in this file?

No idea. contributors@ are current owners of the google group.

@parispittman @mrbobbytables @idvoretskyi @jdumars @castrojo (folks from https://github.com/kubernetes/community/issues/4673#issuecomment-623556960) - can we get contributors@ into this file?

nikhita: This is probably blowing scope... what's the difference between community@ and contributors@? I think this was mentioned somewhere/in some meeting but I can't remember :see_no_evil:

mrbobbytables: community is a group account that acts as the point of contact for things like events etc and is shared with a larger group of people. It cannot own docs or other things as it's not an actual user. contributors@ is an actual user account, not a group. It can own docs/drive and a bunch of other things 👍

nikhita: @mrbobbytables Thanks for the context! Three questions:

1. Given that no account needs to own this gsuite email anymore, would it be ok to remove contributors@ from the list?

2. Would it be useful to have community@ here as as member to help with posting information about events, etc?

3. Is contributors@ and the list of people behind that alias documented/can we document it?

nikhita:
> For brevity/"easy-to-type"-ey-ness, could this just be leads@kubernetes.io?

`chairs-and-techleads` seems more detailed and clear to me but don't have strong opinions.

> For continuity, once this group is created, can we add it as a member of the old @googlegroups.com address?

I'm confused...would the old googlegroups still continue to be functional? I assumed it would continue to exist to preserve old messages but we wouldn't be able to post to it anymore.

nikhita:
Updated to make it viewable by anyone (https://github.com/kubernetes/k8s.io/pull/970#issuecomment-645604600). PTAL.

groups/groups.yaml
mrbobbytables: > 1. Given that no account needs to own this gsuite email anymore, would it be ok to remove contributors@ from the list?


I would still leave it in there for moderation purposes. It's the "backup" method we use in case we have to step in.


> 2. Would it be useful to have community@ here as as member to help with posting information about events, etc?


I think that's a good idea 👍 


> 3. Is contributors@ and the list of people behind that alias documented/can we document it?

I don't think it's well documented at this time. >_> As a start, the folks who have access to it could be listed in our [moderation doc](https://github.com/kubernetes/community/blob/master/communication/moderators.md).

nikhita: >> 2. Would it be useful to have community@ here as as member to help with posting information about events, etc?

> I think that's a good idea +1

Added

mrbobbytables:
> For continuity, once this group is created, can we add it as a member of the old @googlegroups.com address?

> I'm confused...would the old googlegroups still continue to be functional? I assumed it would continue to exist to preserve old messages but we wouldn't be able to post to it anymore.

That was my understanding too. The idea will be once this is in place, communicate the change and archive that group. We will need to follow up with docs changes.

spiffxp:
/approve
I have no objections to this merging as is.  I didn't add the `/hold` and it looks like some of nikhita's questions remain unanswered so I leave that and `/lgtm` to someone else.

justaugustus:
Circling back (from Slack):

```
nikhita Today at 2:21 PM
@justaugustus iiuc the only outstanding thing on this PR is the decision on what happens to the old group once the new alias is created. I wanted to confirm if you are ok with the approach @mrbobbytables suggested here - https://github.com/kubernetes/k8s.io/pull/970#issuecomment-645974779
mrbobbytablesmrbobbytables
Comment on #970 groups: add chairs-and-techleads@kubernetes.io
For continuity, once this group is created, can we add it as a member of the old @googlegroups.com address?
I'm confused...would the old googlegroups still continue to be functional? I assumed it would continue to exist to preserve old messages but we wouldn't be able to post to it anymore.
That was my understanding too. The idea will be once this is in place, communicate the change and archive that group. We will need to follow up with docs changes.
<https://github.com/kubernetes/k8s.io|kubernetes/k8s.io>kubernetes/k8s.io | Jun 18th | Added by GitHub

justaugustus:kubecat:  4 minutes ago
@nikhita -- I think I still have a preference towards leads@

justaugustus:kubecat:  3 minutes ago
Just thinking of typo scenarios and not wanting to miss potentially important info

nikhita  3 minutes ago
nods

mrbobbytables:keys:  3 minutes ago
I can see that

mrbobbytables:keys:  3 minutes ago
it was carryover from the current list

justaugustus:kubecat:  2 minutes ago
Like I sweat doing:
/sig contributor-experience

nikhita  2 minutes ago
lol

nikhita  2 minutes ago
cc @castrojo @paris @cblecker @spiffxp @dims in case you have thoughts

nikhita  1 minute ago
stephen -- you good with archiving the old group?

mrbobbytables:keys:  1 minute ago
following normal practice of send notice, etc etc

justaugustus:kubecat:  1 minute ago
Yeah, that's fine. My worry was old messages that might've still been in play.

justaugustus:kubecat:  < 1 minute ago
If someone responds to an old thread, it'll blackhole

justaugustus:kubecat:  < 1 minute ago
then again, it's a pretty low-traffic list, so maybe I'm being paranoid.
```

nikhita:
Updated to leads@kubernetes.io. PTAL :smile:

justaugustus:
Thanks so much for working on this, @nikhita!
/lgtm
/approve
/hold cancel

PR #980, opened 2020-06-24
Revise email address for Tim Bannister

sftim:
Use a subaddress for Kubernetes topics.


sftim:
/close

Just remembered, I don't think Google Groups does subaddresses how I imagined it would.

PR #981, opened 2020-06-24
releng: Promote arch-specific digests for debian-hyperkube-base:v1.1.0

justaugustus:
(Continuation of promotion for https://github.com/kubernetes/kubernetes/pull/92354.)

There are certain scripts that expect the arch-specific image names
e.g., `debian-hyperkube-base-amd64`, so we need to promote the individual
image manifests as well.

/assign @dims @cblecker @BenTheElder @listx @liggitt 

/priority critical-urgent
ref: https://github.com/kubernetes/kubernetes/issues/92275, https://github.com/kubernetes/kubernetes/issues/92272, https://github.com/kubernetes/kubernetes/issues/92250


justaugustus:
Thanks Dan!
/hold cancel

PR #987, opened 2020-06-27
prow-builds: increase loop device nodes

BenTheElder:
see: https://github.com/kubernetes/test-infra/pull/18090
/cc @spiffxp @aojea


spiffxp:
Deployed

BenTheElder:
thank you!

PR #967, opened 2020-06-16
Deploy triage-party

ameukam:
Deploy triage-party for the release team.
This current deployment will only show a dashboard tracking all issues and PRs with the labels: 
`v1.19`
`priority/critical-urgent`.

The rules added for this deployment came from https://github.com/google/triage-party/blob/v1.1.0/config/examples/kubernetes.yaml.
`

Signed-off-by: Arnaud Meukam <ameukam@gmail.com>

PR #988, opened 2020-06-29
Allow namespace-use-role access to secrets

spiffxp:
Putting this up to ask the question: why are cluster admins the only
people allowed to create secrets?  I personally would rather delegate
control of a namespace's resource, including its secrets, to the rbac
group.

Bumped into this while helping @ameukam setup triage-party here: https://github.com/kubernetes/k8s.io/pull/967/files#r445789688

/cc @cblecker @thockin @dims

### DIFF
--- a/infra/gcp/namespaces/namespace-user-role.yml
+++ b/infra/gcp/namespaces/namespace-user-role.yml
@@ -40,7 +40,7 @@ rules:
     verbs: ["get", "list"]
   - apiGroups: [""]
     resources: ["secrets"]
-    verbs: ["list"]
+    verbs: ["*"]
   - apiGroups: ["storage.k8s.io"]
     resources: ["storageclasses"]
     verbs: ["list"]


spiffxp:
/hold
for comment from requested reviewers

thockin:
We left Secret out because we thought a) the vast vast majority of
use-cases was for TLS certs, which cert-manager handles and b) it was risky
for people to exfiltrate TLS certs for something.k8s.io.

Maybe that's overly-paranoid?  It was a case of "we can add it if we need
it.  I don't object to adding it back in, but it is unfortunate that we
can't easily get finer-grained than that.

cblecker:
If we already provide `list`, folks can already exfiltrate secrets. If we want to allow management, but not the ability to view, we could allow create/patch/delete.

dims:
@spiffxp do we still need this?

spiffxp:
/close
I'll close this for now.  If I have time I'll play around with my alternate account to confirm @cblecker's suggestion.  I get the exfiltration of certs concern.  Still seems like the status quo leaves us awfully bottlenecked on a few cluster admins.

PR #990, opened 2020-06-30
Changed ingresses and networkpolicies apiGroup

bartsmykla:
It was changed in `infra/gcp/namespaces/namespace-user-role.yml`. It was
`extensions` before so I changed it to `networking.k8s.io`

/cc @ameukam 


infra/gcp/namespaces/namespace-user-role.yml
```diff
@@ -18,13 +18,13 @@ rules:
     resources: ["cronjobs", "jobs"]
     verbs: ["*"]
   - apiGroups: ["extensions"]
```
ameukam, line 20: We could delete this API group. @thockin @spiffxp WDYT ?

ameukam:
/lgtm
/hold
for comments.

bartsmykla:
@spiffxp what is your opinion on @ameukam's question?

infra/gcp/namespaces/namespace-user-role.yml
spiffxp: Yeah, I think we should, once we've migrated everything away from it

infra/gcp/namespaces/namespace-user-role.yml
```diff
@@ -18,13 +18,13 @@ rules:
...
-    resources: ["deployments", "ingresses", "networkpolicies"]
```
spiffxp, line 21: I think this should be done in a followup PR

spiffxp (changes requested):
I don't think we should remove access to the deprecated extensions group (and resources) until _after_ we've migrated things away from it.  So for this PR I would probably just add access to networking.k8s.io/ingresses

Is anyone interested in doing the actual migration?

spiffxp:
We'll want this done before the cluster is migrated to v1.16

ameukam:
/assign
I'll work on that next week.

spiffxp:
/close
In favor of #1629

Issue #928, opened 2020-05-29
Setup automatic groups reconciliation


spiffxp:
I've dropped the ball a couple of times on merging a PR and then forgetting to `make -C groups run -- --confirm`

Now that we have a trusted prow build cluster, I'd like to start trusting prow to do this for us

I started working on this and realized we don't have an issue to tie this all together, so here it is

spiffxp:
https://github.com/kubernetes/k8s.io/pull/924 - allow prow-build-trusted to use the gsuite-groups-manager service account via workload identity
https://github.com/kubernetes/k8s.io/pull/925 - add tests to enforce some groups.yaml conventions (these get run at presubmit and postsubmit right before actuating)
https://github.com/kubernetes/k8s.io/pull/927 - hardcode some groups for paranoia

https://github.com/kubernetes/test-infra/pull/17760 - getting post-k8sio-groups setup
https://github.com/kubernetes/test-infra/pull/17762 - then
https://github.com/kubernetes/test-infra/pull/17764 - unbreaking things

https://github.com/kubernetes/k8s.io/pull/926 - proof of concept, this was actuated by https://prow.k8s.io/view/gcs/kubernetes-jenkins/logs/post-k8sio-groups/1266505019021070339

spiffxp:
https://github.com/kubernetes/test-infra/pull/17765 - setup a periodic and send alerts to k8s-infra-alerts@kubernetes.io

bartsmykla:
@spiffxp is there anything still to be done in relation to this issue?

spiffxp:
/close
I'm happy enough with this as is.  Please open other issues for followup.

I think the next logical step would be to consider breaking into hierarchical files that can be used with OWNERS to make this more self service (#460).  But I'm also not sure whether we've adequately documented all policies we care to.

Are there more conventions/rules/policies we should be testing for? Current tests are here: https://github.com/kubernetes/k8s.io/blob/master/groups/groups_test.go

PR #996, opened 2020-07-04
groups: support hierarchical groups.yaml files

nikhita:
For https://github.com/kubernetes/k8s.io/issues/460

Requiring all groups be added to a single `groups.yaml` can become
tedious to maintain. This commit allows splitting groups into
hierarchical `groups.yaml` files, akin to OWNERS files.

This commit contains the following changes:

- We now hardcode files containing groups config to have the file name
`groups.yaml`.

- We no longer allow the group config to be specified at an URL.

- The `groups-file` field in config.yaml is renamed to `groups-path`.
`groups-path` is the directory containing groups.yaml files. It must be
an absolute path. If not specified, it defaults to the directory
containing `config.yaml`.

- `readGroupsConfig` uses `groups-path` as the root dir and recursively
walks through all dirs and files. When it sees a `groups.yaml` file,
it reads the config and adds its groups to `config.Groups`.

- `config.Groups` finally contains the merged groups config from all
`groups.yaml` files.

---

This is WIP because:

- As https://github.com/kubernetes/k8s.io/issues/928#issuecomment-640993935 mentions, we probably need documented policies about how we'd split groups.yaml before enabling this. I just wanted to get this PR out to get the ball rolling.

- I moved `community@kubernetes.io` to the `sig-contributor-experience` dir and seeded OWNERS with the root OWNERS in k/community as an example, we'd probably need to tweak OWNERS/groups a bit more.

/hold
to ensure we have policies in place before enabling this

/assign @cblecker @dims @spiffxp

### DIFF
--- a/groups/groups_test.go
+++ b/groups/groups_test.go
@@ -20,28 +20,73 @@ import (
 	"flag"
 	"fmt"
 	"os"
+	"path/filepath"
 	"reflect"
 	"sort"
 	"strings"
 	"testing"
 	"unicode/utf8"
+
+	"k8s.io/apimachinery/pkg/util/sets"
 )
 
 var cfg GroupsConfig
 
-var groupsConfigPath = flag.String("groups-config", "./groups.yaml", "Path to groups config")
+var groupsPath = flag.String("groups-path", "", "Directory containing groups.yaml files")
 
 func TestMain(m *testing.M) {
 	flag.Parse()
-	if *groupsConfigPath == "" {
-		fmt.Println("--groups-config must be set")
+	var err error
+
+	if *groupsPath != "" && !filepath.IsAbs(*groupsPath) {
+		fmt.Printf("groups-path \"%s\" must be an absolute path\n", *groupsPath)
 		os.Exit(1)
 	}
-	if err := readGroupsConfig(".", *groupsConfigPath, &cfg); err != nil {
-		fmt.Printf("Could not load groups-config: %v", err)
+
+	if *groupsPath == "" {
+		*groupsPath, err = os.Getwd()
+		if err != nil {
+			fmt.Printf("Cannot get current working directory: %v\n", err)
+			os.Exit(1)
+		}
+	}
+
+	if err := readGroupsConfig(*groupsPath, &cfg); err != nil {
+		fmt.Printf("Could not load groups-file: %v\n", err)
 		os.Exit(1)
 	}
 	os.Exit(m.Run())
+}
+
+// TestMergedGroupsConfig tests that readGroupsConfig reads all
+// groups.yaml files and the merged config does not contain any duplicates.
+//
+// It tests that the config is merged by checking that the final
+// GroupsConfig contains at least one group that isn't in the
+// root groups.yaml file.
+func TestMergedGroupsConfig(t *testing.T) {
+	var containsMergedConfig bool
+	found := sets.String{}
+	dups := sets.String{}
+
+	for _, g := range cfg.Groups {
+		name := g.Name
+		if name == "community" {
+			containsMergedConfig = true
+		}
+
+		if found.Has(name) {
+			dups.Insert(name)
+		}
+		found.Insert(name)
+	}
+
+	if !containsMergedConfig {
+		t.Errorf("Final GroupsConfig does not have merged configs from all groups.yaml files")
+	}
+	if n := len(dups); n > 0 {
+		t.Errorf("%d duplicate groups: %s", n, strings.Join(dups.List(), ", "))
+	}
 }
 
 // TestStagingEmailLength tests that the number of characters in the
--- a/groups/reconcile.go
+++ b/groups/reconcile.go
@@ -17,23 +17,20 @@ limitations under the License.
 package main
 
 import (
-	"crypto/tls"
 	"encoding/json"
 	"flag"
 	"fmt"
 	"io/ioutil"
 	"log"
 	"net/http"
-	"net/url"
 	"os"
 	"path/filepath"
 	"reflect"
-	"time"
 
 	secretmanager "cloud.google.com/go/secretmanager/apiv1"
 	"golang.org/x/net/context"
 	"golang.org/x/oauth2/google"
-	"google.golang.org/api/admin/directory/v1"
+	admin "google.golang.org/api/admin/directory/v1"
 	"google.golang.org/api/googleapi"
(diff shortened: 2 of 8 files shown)


nikhita:
cc @mrbobbytables @castrojo 
might be useful if we want to include more gsuite accounts (xref https://github.com/kubernetes/steering/issues/166)

groups/reconcile.go
```diff
@@ -202,30 +202,66 @@ func readConfig(configFilePath string, confirmChanges bool) error {
...
+	}
+	return nil
+}
+
+// readMultipleGroupsConfig uses configFilePath's dir as the root dir and
+// recursively walks through all dirs and files. It reads the GroupsConfig
+// from all files that have the name groupsFileName and adds the groups in
+// each GroupsConfig to config.Groups.
+func readMultipleGroupsConfig(configFilePath string, groupsFileName string, config *GroupsConfig) error {
+	rootDir := filepath.Dir(configFilePath)
```
spiffxp, line 230: nit: I think it would be cleaner if this function took `rootDir` directly, and the work of deriving it was moved elsewhere

groups/reconcile.go
```diff
@@ -202,30 +202,66 @@ func readConfig(configFilePath string, confirmChanges bool) error {
...
+
+			for _, g := range groupsConfigAtPath.Groups {
+				config.Groups = append(config.Groups, g)
+			}
 		}
+		return nil
+	})
+}
+
+func readGroupsConfigFromURL(groupsURL *url.URL, config *GroupsConfig) error {
```
spiffxp, line 259: Perhaps we should drop this functionality entirely.  It's not exercised and it's not tested.

groups/reconcile.go
```diff
@@ -202,30 +202,66 @@ func readConfig(configFilePath string, confirmChanges bool) error {
 	return err
 }
 
-func readGroupsConfig(configFilePath string, groupsConfigFilePath string, config *GroupsConfig) error {
+func readGroupsConfig(configFilePath string, groupsFileName string, config *GroupsConfig) error {
```
spiffxp, line 205: Capitalization nit
```suggestion
func readGroupsConfig(configFilePath string, groupsFilename string, config *GroupsConfig) error {
```

As with `readMultipleGroupsConfig` I think this should take an explicit dir instead of passing along `configFilePath` and deriving from it later

groups/groups_test.go
```diff
@@ -25,25 +25,59 @@ import (
...
 	"testing"
 	"unicode/utf8"
+
+	"k8s.io/apimachinery/pkg/util/sets"
 )
 
 var cfg GroupsConfig
 
-var groupsConfigPath = flag.String("groups-config", "./groups.yaml", "Path to groups config")
+var groupsConfigFileName = flag.String("groups-file", "groups.yaml", "Filename containing groups config")
```
spiffxp, line 34: My personal preference would be to alter things to take an explicit `groups-path` or `groups-dir`, instead of auto-deriving it from the location of `config.yaml`

I also wonder if it would be useful to allow files named something other than `groups.yaml`, in the same way that prow jobs can be defined in arbitrarily named files.  Not blocking on this, just thinking out loud. I could foresee wg-k8s-infra using this to have a rbac file, a staging groups file, etc.

nikhita: > explicit groups-path or groups-dir, instead of auto-deriving it from the location of config.yaml

> I also wonder if it would be useful to allow files named something other than groups.yaml

This is interesting! Would something like "read all yaml files under the directory specified by `groups-dir`" make sense? (I still need to look at how prow does it today)

thockin:
Why?  Having a strong one-file-per-directory convention is nice because it
aligns with OWNERS

thockin:
We may even want to link directory-name to group-name to lock it own to
allow OWNERS delegation more

nikhita:
Addressed comments. PTAL. The updates are documented in the commit message and the issue body.

TLDR:

- We now hardcode files containing groups config to have the file name
`groups.yaml`.

- We no longer allow the group config to be specified at an URL.

- The `groups-file` field in config.yaml is renamed to `groups-path`.
`groups-path` is the directory containing groups.yaml files. It must be
an absolute path. If not specified, it defaults to the directory
containing `config.yaml`.

nikhita:
> We may even want to link directory-name to group-name to lock it own to
allow OWNERS delegation more

I haven't added a validation for this because we might want to create directories based on sig names. I've added discussing the policy for splitting groups.yaml to [this week's wg-k8s-infra meeting](https://docs.google.com/document/d/16VBfsFMynA7tObzuZGPpw-sKDKfFc_T5W_E4IeEIaOQ/edit#).

groups/reconcile.go
nikhita: > Perhaps we should drop this functionality entirely. It's not exercised and it's not tested.

Removed.

groups/sig-contributor-experience/OWNERS
```diff
@@ -0,0 +1,10 @@
+# See the OWNERS docs at https://go.k8s.io/owners
+
+approvers:
```
spiffxp, line 3:
```suggestion
reviewers:
```
I would happily merge this PR as-is if this wasn't present.  I do want to open up the pool but we're not there yet.

If the goal is to fast-track membership changes, I would suggest we test against a known list of groups that is under /groups/OWNERS approval only

If the goal is to allow sigs/wgs/whatever to create whatever @kubernetes.io groups they want, I would at a minimum require adherence to common naming conventions and settings

nikhita: Should I just removed the contribex dir and comment out the test for it? I just meant it as an example that this PR works :woman_shrugging: 

I agree that we shouldn't merge this as-is.

spiffxp: Leaving the dir and everything is fine (and preferred!), if the approvers are changed to reviewers for now

nikhita: Done!

bartsmykla:
lgtm

nikhita:
Discussed in today's wg k8s infra meeting, let's merge and iterate :sparkles: 

We discussed a [few approaches towards policy](https://docs.google.com/document/d/16VBfsFMynA7tObzuZGPpw-sKDKfFc_T5W_E4IeEIaOQ/edit), and decided to go with whatever is easiest to setup validation for. I'll take a dab at that over the weekend :)

/hold cancel

PR #1008, opened 2020-07-08
Allow read-only access to monitoring and logging of 'aaa' cluster

ameukam:
Allow operators of the different workloads(except cert-manager) running on `aaa` to have
read-only access to Monitoring and Logging in the Google Cloud
Console and the APIs.

Detailed list of permissions :
- Monitoring : https://cloud.google.com/monitoring/access-control#roles
- Logging : https://cloud.google.com/logging/docs/access-control#permissions_and_roles

### DIFF
--- a/infra/gcp/clusters/projects/kubernetes-public/aaa/12-iam.tf
+++ b/infra/gcp/clusters/projects/kubernetes-public/aaa/12-iam.tf
@@ -1,0 +1,26 @@
+/*
+Ensure google groups defined in 'members' have read-only access to
+all monitoring data and configurations.
+Detailed list of permissions: https://cloud.google.com/monitoring/access-control#roles
+*/
+resource "google_project_iam_binding" "readonlymonitoringbinding" {
+  project = data.google_project.project.id
+  role    = "roles/monitoring.viewer"
+
+  members = [
+    "group:gke-security-groups@kubernetes.io",
+  ]
+}
+
+/*
+Ensure google groups defined in 'members' have read-only access to logs.
+Detailed list of permissions: https://cloud.google.com/logging/docs/access-control#permissions_and_roles
+*/
+resource "google_project_iam_binding" "readonlyloggingbinding" {
+  project = data.google_project.project.id
+  role    = "roles/logging.privateLogViewer"
+
+  members = [
+    "group:gke-security-groups@kubernetes.io",
+  ]
+}


infra/gcp/clusters/projects/kubernetes-public/aaa/12-iam.tf
```diff
@@ -0,0 +1,40 @@
...
+
+  members = [
+    "group:k8s-infra-rbac-gcsweb@kubernetes.io",
+    "group:k8s-infra-rbac-k8s-io-canary@kubernetes.io",
+    "group:k8s-infra-rbac-k8s-io-prod@kubernetes.io",
+    "group:k8s-infra-rbac-perfdash@kubernetes.io",
+    "group:k8s-infra-rbac-publishing-bot@kubernetes.io",
+    "group:k8s-infra-rbac-triageparty-release@kubernetes.io",
+    "group:k8s-infra-rbac-slack-infra@kubernetes.io",
+    "group:k8s-infra-rbac-node-perf-dash@kubernetes.io",
```
spiffxp, line 18: Would it make sense to use `gke-security-groups@kubernetes.io` instead?  Thinking it'd be nice to avoid duplicating this list

infra/gcp/clusters/projects/kubernetes-public/aaa/12-iam.tf
```diff
@@ -0,0 +1,40 @@
...
```
spiffxp, line 38: Same here

infra/gcp/clusters/projects/kubernetes-public/aaa/12-iam.tf
ameukam: Totally make sense to use `gke-security-groups@kubernetes.io`. I'll update the PR.

infra/gcp/clusters/projects/kubernetes-public/aaa/12-iam.tf
ameukam: PR updated with `gke-security-groups@kubernetes.io`.

spiffxp:
I don't have time to setup an unprivileged account to test myself, so I'm fine deploying and reverting/updating if someone finds this doesn't work

spiffxp:
ran `terraform apply`

ameukam:
Can confirm I have access to metrics and logs for`aaa`. Thanks Folks!
Maybe it's a bad idea but can we send an email to those groups about those new privileges ?

Issue #984, opened 2020-06-25
Need IAM permissions for Session Manager added to Boskos-managed AWS accounts


randomvariable:
Porting over from https://github.com/kubernetes/test-infra/issues/17190 by @detiber 

What would you like to be added:

Recently we removed creation of a Bastion host by default from cluster-api-provider-aws and would prefer to avoid creating one for the purposes of testing and would rather use ssh proxied over AWS Session Manager to connect to remote instances for retrieving logs during testing instead.

Will need one of the below done against each AWS subaccount users that is registered as an aws-account resource in Boskos:

Attach the AmazonSSMFullAccess policy
Create a policy with the permissions outlined in docs.aws.amazon.com/systems-manager/latest/userguide/session-manager-getting-started-restrict-access.html and attach it, or add it as an inline policy

randomvariable:
Is there anything we can do to get this moving? We kind of missed an upgrade bug for Kubernetes 1.19 for a few reasons, but would have been compounded by not having the logs that SSM would have provided.

justinsb:
I added the permission to #1016 (and it's in the latest accounts that I just created, but we should wait till #1016 merges before applying to boskos)

spiffxp:
/priority backlog
I have lost track of where we're at with AWS account management

ameukam:
All of the related tests to AWS have been migrated to a dedicated AWS organization.
A new issue will open if needed again.

/close

PR #1016, opened 2020-07-12
Create AWS test accounts using terraform

justinsb:
This brings them under declarative management.  We use a date to
enable rotation of old accounts, as these are limited.  Pre-created 14
accounts, next step is to get these into boskos.

A lot of workarounds for terraform shortcomings, hence the slightly
odd structure.

### DIFF
--- a/infra/aws/e2e/accounts/tf/README.md
+++ b/infra/aws/e2e/accounts/tf/README.md
@@ -1,0 +1,35 @@
+# AWS account scripting
+
+* We create AWS accounts using terraform.
+
+* The state file is stored (encrypted) in GCS.
+
+* The encryption key is available only to those that should be running
+  this, as it contains the AWS account keys.
+
+* We additionally generate the boskos objects, and upload that also to
+  GCS, encrypted using the same key (as it broadly has the same data)
+
+
+## Applying to terraform
+
+Ensure that the key is written to `gcs-encryption-key`
+
+Ensure you are using the correct AWS profile, and then run tf-apply, for example:
+
+```
+export AWS_PROFILE=cncf
+./tf-apply
+```
+
+## Uploading to boskos
+
+Ensure that the encryption key is written to `gcs-encryption-key`
+
+Ensure you are using the correct k8s context, then run `apply-to-k8s`.  For example:
+
+
+```
+gcloud container clusters get-credentials ...
+./apply-to-k8s
+```
--- a/infra/aws/e2e/accounts/tf/accounts.tf
+++ b/infra/aws/e2e/accounts/tf/accounts.tf
@@ -1,0 +1,182 @@
+terraform {
+  required_version = "~> 0.13"
+  backend "gcs" {
+    bucket = "k8s-infra-clusters-terraform"
+    prefix = "aws/accounts"
+    # NOTE: Must specify encryption key
+  }
+}
+
+
+# Terraform doesn't work well with count in modules (even in 0.13)
+
+# Accounts for cluster-api-provider-aws project
+module "capa-user-00" {
+  source      = "./modules/account-and-user"
+  id          = "capa-2020-11-19-00"
+  email       = "k8s-infra-aws-admins+capa2020111900@kubernetes.io"
+  boskos-name = "capa-user-00"
+}
+
+module "capa-user-01" {
+  source      = "./modules/account-and-user"
+  id          = "capa-2020-11-19-01"
+  email       = "k8s-infra-aws-admins+capa2020111901@kubernetes.io"
+  boskos-name = "capa-user-01"
+}
+
+module "capa-user-02" {
+  source      = "./modules/account-and-user"
+  id          = "capa-2020-11-19-02"
+  email       = "k8s-infra-aws-admins+capa2020111902@kubernetes.io"
+  boskos-name = "capa-user-02"
+}
+
+module "capa-user-03" {
+  source      = "./modules/account-and-user"
+  id          = "capa-2020-11-19-03"
+  email       = "k8s-infra-aws-admins+capa2020111903@kubernetes.io"
+  boskos-name = "capa-user-03"
+}
+
+module "capa-user-04" {
+  source      = "./modules/account-and-user"
+  id          = "capa-2020-11-19-04"
+  email       = "k8s-infra-aws-admins+capa2020111904@kubernetes.io"
+  boskos-name = "capa-user-04"
+}
+
+module "capa-user-05" {
+  source      = "./modules/account-and-user"
+  id          = "capa-2020-11-19-05"
+  email       = "k8s-infra-aws-admins+capa2020111905@kubernetes.io"
+  boskos-name = "capa-user-05"
+}
+
+module "capa-user-06" {
+  source      = "./modules/account-and-user"
+  id          = "capa-2020-11-19-06"
+  email       = "k8s-infra-aws-admins+capa2020111906@kubernetes.io"
+  boskos-name = "capa-user-06"
+}
+
+module "capa-user-07" {
+  source      = "./modules/account-and-user"
+  id          = "capa-2020-11-19-07"
+  email       = "k8s-infra-aws-admins+capa2020111907@kubernetes.io"
+  boskos-name = "capa-user-07"
+}
+
+module "capa-user-08" {
+  source      = "./modules/account-and-user"
+  id          = "capa-2020-11-19-08"
+  email       = "k8s-infra-aws-admins+capa2020111908@kubernetes.io"
+  boskos-name = "capa-user-08"
+}
+
+module "capa-user-09" {
+  source      = "./modules/account-and-user"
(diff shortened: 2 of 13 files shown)


justinsb:
cc @BenTheElder @detiber 

I also have a bunch of yaml files ready to upload into the boskos cluster!

BenTheElder:
seems about right, I'm not 100% positive about this though.
go.k8s.io/oncall has the current test-infra oncall for uploading to boskos

detiber:
/lgtm
I'm not overly familiar with terraform, but nothing is jumping out at me.

justinsb:
Storing it in a GCS bucket seems reasonable.  The tfstate file does include AWS account credentials (for now), so we may want to look into encryption.

This must be run by people with administrative access to the AWS CNCF account.  I suggest we aim to follow in the footsteps of the GCP automation - is that still run manually?

The other problem is that we must upload the resulting credentials to boskos (when we create new accounts).  I don't believe we've done that yet here.

In the long term it would be nice to get the prow jobs using workload identity to authenticate to AWS, so there would be no need to store credentials.  But that feels like it's going to require some test-infra changes.

thockin:
On Wed, Aug 12, 2020 at 12:38 PM Justin Santa Barbara <
notifications@github.com> wrote:

> Storing it in a GCS bucket seems reasonable. The tfstate file does include
> AWS account credentials (for now), so we may want to look into encryption.
>
> This must be run by people with administrative access to the AWS CNCF
> account. I suggest we aim to follow in the footsteps of the GCP automation
> - is that still run manually?
>

Yes but the set of people who can run it is governed by GoogleGroups.


> The other problem is that we must upload the resulting credentials to
> boskos (when we create new accounts). I don't believe we've done that yet
> here.
>
> In the long term it would be nice to get the prow jobs using workload
> identity to authenticate to AWS, so there would be no need to store
> credentials. But that feels like it's going to require some test-infra
> changes.
>

Agree


>

infra/aws/e2e/accounts/tf/modules/user/main.tf
```diff
@@ -0,0 +1,39 @@
...
+
+resource "aws_iam_user_policy_attachment" "main-ec2" {
+  user       = aws_iam_user.main.name
+  policy_arn = "arn:aws:iam::aws:policy/AmazonEC2FullAccess"
+}
+
+resource "aws_iam_user_policy_attachment" "main-iam" {
+  user       = aws_iam_user.main.name
+  policy_arn = "arn:aws:iam::aws:policy/IAMFullAccess"
+}
```
randomvariable, line 34: Also add `AmazonSSMFullAccess` as per https://github.com/kubernetes/k8s.io/issues/984

justinsb: Done!

justinsb:
Updated:

* Now stores tfstate encrypted in GCS.
* Also stores the yaml to apply to boskos in GCS, encrypted with the same key, for easier application to the cluster.
* README and scripts to simplify the process

infra/aws/e2e/accounts/tf/accounts.tf
```diff
@@ -0,0 +1,182 @@
...
+    # NOTE: Must specify encryption key
+  }
+}
+
+
+# Terraform doesn't work well with count in modules (even in 0.13)
+
+# Accounts for cluster-api-provider-aws project
+module "capa-user-00" {
+  source      = "./modules/account-and-user"
```
onlydole, line 15: Style Nit: For these module instantiations, you might be able to [take advantage of the `for_each` command](https://www.terraform.io/docs/configuration/meta-arguments/for_each.html) and move these to be variables.

onlydole (changes requested):
One small nit and I'm free to sync on that front if beneficial!

infra/aws/e2e/accounts/tf/accounts.tf
```diff
@@ -0,0 +1,182 @@
+terraform {
+  required_version = "~> 0.13"
```
spiffxp, line 2: All of our other terraform modules require a minimum of 0.12 at present; can this be downgraded to 0.12?

infra/aws/e2e/accounts/tf/README.md
```diff
@@ -0,0 +1,35 @@
...
+* The encryption key is available only to those that should be running
+  this, as it contains the AWS account keys.
+
+* We additionally generate the boskos objects, and upload that also to
+  GCS, encrypted using the same key (as it broadly has the same data)
+
+
+## Applying to terraform
+
+Ensure that the key is written to `gcs-encryption-key`
```
spiffxp, line 16: Who has access to this?

Ideally this key would be stored in google cloud secret manager, and access would be restricted to a @kubernetes.io group we manage.  e.g.

```sh
KEY=$(gcloud secrets --project=kubernetes-public versions access latest --secret=gcs-encryption-key)
```

infra/aws/e2e/accounts/tf/README.md
```diff
@@ -0,0 +1,35 @@
...
+
+Ensure you are using the correct AWS profile, and then run tf-apply, for example:
+
+```
+export AWS_PROFILE=cncf
```
spiffxp, line 21: Who has access to this?

infra/aws/e2e/accounts/tf/modules/account-and-user/main.tf
```diff
@@ -0,0 +1,40 @@
+# Intermediate module; workaround for terraform being bad at multiple providers
```
spiffxp, line 1: I'm not sure I understand what problem this is solving.  Why not merge the account and user modules into this one module?

infra/aws/e2e/accounts/tf/README.md
```diff
@@ -0,0 +1,35 @@
...
+
+## Uploading to boskos
+
+Ensure that the encryption key is written to `gcs-encryption-key`
+
+Ensure you are using the correct k8s context, then run `apply-to-k8s`.  For example:
+
+
+```
+gcloud container clusters get-credentials ...
```
spiffxp, line 33: Which cluster/boskos are these destined for?

Would like input from @kubernetes/test-infra-admins (as proxy for test-infra-oncall)

infra/aws/e2e/accounts/tf/tf-apply
```diff
@@ -0,0 +1,15 @@
...
+
+terraform init -backend-config="encryption_key=${KEY}"
+
+terraform apply
+
+# To view the encrypted state file
+#gsutil -o "GSUtil:decryption_key1=${KEY}" cat gs://k8s-infra-clusters-terraform/aws/accounts/default.tfstate
+
+find boskos/ -name "*.yaml" | sort | xargs cat > boskos/joined
+gsutil -o "GSUtil:encryption_key=${KEY}" cp boskos/joined gs://k8s-infra-clusters-terraform/aws/k8s/accounts-boskos.yaml
```
spiffxp, line 15: IMO it'd be better to store this directly in secret manager as well, then apply via e.g.

```
gcloud secrets --project=kubernetes-public versions access latest --secret=boskos-aws-accounts |\
    kubectl apply -f -
```

infra/aws/e2e/accounts/tf/accounts.tf
```diff
@@ -0,0 +1,182 @@
...
+  email       = "k8s-infra-aws-admins+capa2020111913@kubernetes.io"
+  boskos-name = "capa-user-13"
+}
+
+
+# Accounts for imagebuilder project
+module "image-builder-aws-1" {
+  source      = "./modules/account-and-user"
+  id          = "image-builder-aws-2020-11-19-00"                    # TODO: Switch to -01
+  email       = "k8s-infra-aws-admins+image2020111900@kubernetes.io" # TODO: Switch to 01
```
spiffxp, line 117: Do these need to be addressed?

spiffxp (changes requested):
I want to unblock, but:
- I don't have appropriate access to run all of this
- My terraform environment isn't setup for this (I'll look into whether upgrading to 0.13 or 0.14 for all of our other modules is a heavy lift)
- I would rather see more use of `gcloud secrets` vs. storing encrypted in GCS

I see gs://k8s-infra-clusters-terraform/aws/k8s/accounts-boskos.yaml already exists, so if we want to unblock, I'm assuming it's a matter of getting test-infra-oncall access to that file and manually uploading?

infra/aws/e2e/accounts/tf/README.md
cjwagner: This would need to be destined for the boskos instance in the build cluster where the consuming jobs will be run in order for the jobs to access the boskos instance. Ideally this would be a build cluster besides the default or trusted clusters so that test-infra-oncall doesn't need to act as a middleman for uploading to boskos or debugging related problems.

spiffxp:
/close
Closing this in favor of https://github.com/kubernetes/k8s.io/pull/1839

PR #1028, opened 2020-07-16
releng: Match images.yaml format to cip-mm output

justaugustus:
With VDF (vanity domain flip) around the corner next week, Release
Managers will start to use the 'cip-mm' tool (container image promoter -
merge manifest) to promote images in this file.

The tool outputs the manifest in a slightly different format than is
currently displayed. This commit is just a pre-step to ensure a smaller
diff when we're doing actual release image promotion.

A side-effect here is that the pause:3.3 arch-specific images are
tagged.
(They already exist in the prod registry via the manifest list though.)

Note: the first test run will fail due to a digest/tag mismatch that was a result of the backfill in https://github.com/kubernetes/k8s.io/pull/835.
/hold until I fix that up


justaugustus:
/hold cancel
@kubernetes/release-engineering for approval

PR #1007, opened 2020-07-08
create and authorize service account for promoter vulnerability check

yodahekinsew:
For the purpose of implementing a new vulnerability presubmit check for the image promoter (https://github.com/kubernetes-sigs/k8s-container-image-promoter/issues/235), a new service account needs to be created which is authorized to view the vulnerability scanning results for all of the staging projects. This service account will be used to authenticate a new Prow job - which runs the vulnerability check - via Workload Identity.

PR #1035, opened 2020-07-17
Fix vulnerability scanning binding

spiffxp:
Two things:
- The service account didn't actually exist
- The PROD_PROJECT env var was unbound in ensure-staging-storage.sh

/cc @listx @yodahekinsew


spiffxp:
I should have caught this in review of https://github.com/kubernetes/k8s.io/pull/1007

listx:
This looks fine to me. @yodahekinsew can you test if the bindings are working after this is merged/activated?

/lgtm
/hold

PR #1036, opened 2020-07-17
backfill from gcr.io/google-containers

listx:
This is another backfill before the flip on July 20, 2020. [1] [2]

[1]: https://groups.google.com/d/msg/kubernetes-sig-release/lFmJZKwIxyI/tY6cv5uIAwAJ
[2]: https://github.com/kubernetes/k8s.io/blob/master/k8s.gcr.io/Vanity-Domain-Flip.md

/hold

Please do not release the hold! I need to coordinate with our cip-auditor instance so that this large PR doesn't overwhelm it.


dims:
please don't remove hold without talking to @listx

listx:
Thanks!

/hold cancel

PR #1056, opened 2020-07-23
Add jying to kas-network-proxy members

Jefftree:
Requesting access to view cloudbuild logs for kas-network-proxy.


Sh4d1:
Logs problem is "resolved" but I got no problem against adding jying to the members 😄

/lgtm

PR #1058, opened 2020-07-23
Add test-pods limitranges used by k8s-prow-builds cluster

spiffxp:
I'm seeing some release-blocking jobs fail for what looks like resource contention reasons in k8s-infra-prow-builds (random timeouts depending on which node a pod runs on)

I notice https://github.com/kubernetes/test-infra/pull/18416 recently added default CPU requests for k8s-prow-builds, and there is also a memory limitrange that's been living there for a while

I was tempted to consolidate the two but I'd rather keep as much parity as possible with k8s-prow-builds

(I doubt this will actually solve the contention, I'm tempted to start forcing requests/limits for jobs)

### DIFF
--- a/infra/gcp/clusters/projects/k8s-infra-prow-build/prow-build/resources/cpu-limit-range_limitrange.yaml
+++ b/infra/gcp/clusters/projects/k8s-infra-prow-build/prow-build/resources/cpu-limit-range_limitrange.yaml
@@ -1,0 +1,10 @@
+apiVersion: v1
+kind: LimitRange
+metadata:
+  name: cpu-limit-range
+  namespace: test-pods
+spec:
+  limits:
+  - defaultRequest:
+      cpu: 250m
+    type: Container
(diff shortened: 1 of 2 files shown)


BenTheElder:
I _think_ there's already one you manually deployed here, one other thing that changed recently was https://github.com/kubernetes/test-infra/commit/b74eb1493c9604c0e3022bee54b44ef191e15e54

spiffxp:
Right, the pod_utils defaults are already being applied to pods scheduled here, that's not a build cluster default

/hold cancel
I did not apply anything manually here, so these are going in shortly
```
$ k --context=gke_k8s-infra-prow-build_us-central1_prow-build get limitranges -A
NAMESPACE   NAME     CREATED AT
default     limits   2020-04-30T21:34:14Z
```

spiffxp:
```
$ k --context=gke_k8s-infra-prow-build_us-central1_prow-build get limitranges -A
NAMESPACE   NAME              CREATED AT
default     limits            2020-04-30T21:34:14Z
test-pods   cpu-limit-range   2020-07-23T23:55:02Z
test-pods   mem-limit-range   2020-07-23T23:55:07Z
```

PR #1060, opened 2020-07-24
fix incorrect vulnerability scanning KSA

yodahekinsew:
We are currently empowering an incorrect KSA ("k8s-infra-gcr-promoter-scanning") to the vulnerability scanning GSA. We should be empowering the KSA "k8s-infra-gcr-vuln-scanning".


spiffxp:
Ran `./infra/gcp/ensure-prod-storage.sh`

PR #1073, opened 2020-07-29
Chaging elections list to reflect current officers.

jberkus:
New list reflects current officers: me, Jaice DuMars, Ihor.

Amended and force pushed to get correct list.

### DIFF
--- a/groups/groups.yaml
+++ b/groups/groups.yaml
@@ -183,10 +183,9 @@ groups:
       WhoCanPostMessage: "ANYONE_CAN_POST"
       ReconcileMembers: "true"
     owners:
-      - briangrant@google.com
       - ihor@cncf.io
-      - jorgec@vmware.com
-      - killen.bob@gmail.com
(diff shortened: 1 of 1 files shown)


groups/groups.yaml
```diff
@@ -183,10 +183,10 @@ groups:
       WhoCanPostMessage: "ANYONE_CAN_POST"
       ReconcileMembers: "true"
     owners:
-      - briangrant@google.com
       - ihor@cncf.io
-      - jorgec@vmware.com
+      - josh@berkus.org
       - killen.bob@gmail.com
```
cblecker, line 188: Bob should be removed

jberkus: so just me/ihor/jaice?  OK.  Wasn't clear.

jberkus: Fixed and squash-pushed

dims:
@jberkus @jdumars feel free to remove hold when ready

Issue #1078, opened 2020-07-31
Provision stock e2e projects in anticipation of migrating kubernetes merge-blocking jobs


spiffxp:
This is in anticipation of https://github.com/kubernetes/test-infra/issues/18550

Based on visual inspection of https://monitoring.prow.k8s.io/d/wSrfvNxWz/boskos-resource-usage?orgId=1&from=now-90d&to=now

Source:
- Current k8s-prow-build boskos gce-project pool capacity is 200.  Usage peaks just under 100.

Target:
- Current k8s-infra-prow-build boskos k8s-infra-gce-project pool capacity is 40.  Usage peaks just under 30.

Goal:
- Add at least another 100 projects to k8s-infra-prow-build's boskos instance
- Add the new projects to a gce-project pool
- While doing this, merge the k8s-infa-gce-project pool into the gce-project pool, ref: https://github.com/kubernetes/test-infra/issues/18552
  - Switch jobs away from k8s-infra-gce-project
  - Wait for k8s-infra-gce-project pool usage to be completely clean
  - Merge k8s-infra-gce-project pool into gce-project pool
- At the end of this, ensure boskos dashboard is setup to display the gce-project pool

/area prow
/sig testing
/sig release
/wg k8s-infra

spiffxp:
I've added 40 thus far, so out of 100 that mean 60 remain. Going to add another 40, leaving 20.

It may be that as we're re-evaluating the full set of merge-blocking jobs for kubernetes (ref: https://github.com/kubernetes/test-infra/issues/18729) we may lower demand for the gce-project resource

spiffxp:
/close
OK, I think the gce-project pool is large enough for now.

Over the past 30 days we've migrated over all merge-blocking jobs for kubernetes/kubernetes on top of having migrated most of the release-blocking jobs earlier.  We're going to have plenty of headroom for a while.
![Screen Shot 2020-09-01 at 10 50 07 AM](https://user-images.githubusercontent.com/49258/91888299-ec9d1100-ec40-11ea-9ebb-c150485f716e.png)

PR #1084, opened 2020-07-31
Fix boskos gce-project deployment

spiffxp:
The configmap should be named resources, not boskos-configmap.yaml. Little too fast with kubectl today

The gce-project pool needs to be explicitly referenced by boskos components or it would be appropriately managed

fix/followup to https://github.com/kubernetes/k8s.io/pull/1083

ref: https://github.com/kubernetes/k8s.io/issues/1078


spiffxp:
/hold
missing a few other things

Issue #1095, opened 2020-07-31
Identify quota changes needed for gpu jobs, create pool of gpu projects


spiffxp:
This is like https://github.com/kubernetes/k8s.io/issues/851 but for gpu

The default quotas for an e2e project (eg: k8s-infra-e2e-gce-project) MAY be insufficient to run ci-kubernetes-e2e-gce-device-plugin-gpu

Currently this job runs in the google.com k8s-prow-builds cluster, using a project from that boskos' [gpu-project pool](https://github.com/kubernetes/test-infra/blob/master/config/prow/cluster/boskos-resources.yaml#L430-L449)

- [x] Identify whether ci-kubernetes-e2e-gce-device-plugin-gpu can run against a stock e2e project: it couldn't, needed raised quota
- [x] Identify what quotas are set that make gpu-project pool projects differ from a default e2e project: raised committed k80 gpu's in us-west1 from 0 to 2
- [x] Identify what jobs this will allow us to migrate: jobs mentioned in https://github.com/kubernetes/k8s.io/issues/1095#issuecomment-668866249 plus pull-kubernetes-e2e-gce-device-plugin-gpu
- [x] Identify pool size: went with 10 (reasoning in https://github.com/kubernetes/k8s.io/pull/1118)
- [x] Provision pool in kubernetes.io org (done via: https://github.com/kubernetes/k8s.io/pull/1118)
- [x] Add pool to k8s-infra-prow-build's boskos (done via: https://github.com/kubernetes/k8s.io/pull/1125)

spiffxp:
```
$ for prj in k8s-gke-gpu-boskos-02 k8s-infra-e2e-gpu-project; do
  gcloud compute project-info describe --project=$prj > $prj.compute-project-info.yaml
done
$ diff -yW80 {k8s-infra-e2e-ingress-project,k8s-gke-gpu-boskos-02}.compute-project-info.yaml | grep -A3 "|.*limit"
- limit: 100.0			      |	- limit: 200.0
  metric: SECURITY_POLICY_RULES		  metric: SECURITY_POLICY_RULES
  usage: 0.0				  usage: 0.0
- limit: 45.0				- limit: 45.0
--
- limit: 300.0			      |	- limit: 1000.0
  metric: NETWORK_ENDPOINT_GROUPS	  metric: NETWORK_ENDPOINT_GROUPS
  usage: 0.0				  usage: 0.0
- limit: 6.0				- limit: 6.0
--
[...]

Issue #1093, opened 2020-07-31
Identify quota changes needed for ingress jobs, create pool of ingress projects


spiffxp:
This is like https://github.com/kubernetes/k8s.io/issues/851 but for ingress

The default quotas for an e2e project (eg: k8s-infra-e2e-gce-project) MAY be insufficient to run ci-kubernetes-e2e-gci-gce-ingress

Currently this job runs in the google.com k8s-prow-builds cluster, using a project from that boskos' [ingress-project pool](https://github.com/kubernetes/test-infra/blob/master/config/prow/cluster/boskos-resources.yaml#L450-L472)

- [x] ~Identify whether ci-kubernetes-e2e-gci-gce-ingress can run against a stock e2e project~ (it can! so this is much lower priority, since we don't need ingress projects for release/merge-blocking job migration)
- [ ] Identify what quotas are set that make ingress-projects differ from a default e2e project
- [ ] Identify what jobs this will allow us to migrate
- [ ] Identify pool size
- [ ] Provision pool in kubernetes.io org
- [ ] Add pool to k8s-infra-prow-build's boskos

spiffxp:
```
$ for prj in k8s-ingress-boskos-01 k8s-infra-e2e-ingress-project; do
  gcloud compute project-info describe --project=$prj > $prj.compute-project-info.yaml
done
$ diff -yW80 k8s-infra-e2e-ingress-project.compute-project-info.yaml k8s-ingress-boskos-01.compute-project-info.yaml | grep -A3 "|.*limit"
- limit: 21.0			      |	- limit: 50.0
  metric: STATIC_ADDRESSES		  metric: STATIC_ADDRESSES
  usage: 0.0				  usage: 0.0
- limit: 250.0				- limit: 250.0
--
- limit: 45.0			      |	- limit: 100.0
  metric: FORWARDING_RULES		  metric: FORWARDING_RULES
  usage: 0.0				  usage: 0.0
- limit: 150.0				- limit: 150.0
--
- limit: 150.0			      |	- limit: 55.0
  metric: HEALTH_CHECKS			  metric: HEALTH_CHECKS
  usage: 0.0				  usage: 0.0
- limit: 69.0			      |	- limit: 100.0
  metric: IN_USE_ADDRESSES		  metric: IN_USE_ADDRESSES
  usage: 0.0				  usage: 0.0
- limit: 150.0				- limit: 150.0
--
- limit: 30.0			      |	- limit: 50.0
[...]

PR #1108, opened 2020-08-04
Add manual gpu-project and ingress-project

spiffxp:
This follows the previous pattern of "create a non-boskos managed
project that contributors could use as humans", and sets the stage for
me to work on figuring out what gpu-project and ingress-project will
need from a quota perspective

part of https://github.com/kubernetes/k8s.io/issues/1093
part of https://github.com/kubernetes/k8s.io/issues/1095


spiffxp:
Ran `./infra/gcp/ensure-e2e-projects.sh`

Issue #846, opened 2020-05-06
Migrate jobs away from gs://kubernetes-release-dev


spiffxp:
There are a number of jobs that assume they have write access to `gs://kubernetes-release-dev`, and then a bunch of other jobs that assume they should be downloading builds from `gs://kubernetes-release-dev`

Unfortunately, this bucket will not allow non-google.com service accounts to write to it.  We need to use a new bucket, and develop a plan for using it that doesn't involve all jobs being cut over in lock step.  

I am pretty sure this overlaps with ongoing work by the release-engineering subproject

ref: https://github.com/kubernetes/k8s.io/issues/752, https://github.com/kubernetes/k8s.io/issues/841 - followup to https://github.com/kubernetes/k8s.io/pull/830

spiffxp:
https://cs.k8s.io/?q=latest-green.txt&i=nope&files=&repos=

It _looks_ like `gs://kubernetes-release-dev/ci/latest-green.txt` isn't actually consumed by automation

justaugustus:
@spiffxp -- Carrying the convo from https://kubernetes.slack.com/archives/CJH2GBF7Y/p1590088139329000...

Yep, we've identified this as a need, but slowed down because the Prow bits weren't in place yet (ref: https://kubernetes.slack.com/archives/CJH2GBF7Y/p1590088139329000).

So, broad strokes/first guess on a plan:
- We need a new GCP project (shared ownership between SIG Release (Release Engineering), SIG Testing, WG K8s Infra). I don't know who "owns" `kubernetes-release-dev` today.
- New GCS bucket
- **insert infra script updates/actuation here**
- Add a new [scenario](https://github.com/kubernetes/test-infra/tree/master/scenarios) that defaults to publish to the new bucket
- Create canary `ci-kubernetes-build` jobs for all branches to start seeding the new bucket
- Start cutting over some e2e jobs to pull from the canary buckets (maybe some of the `release-*` e2es?)
- Migrate the other e2es
- Flip the `ci-kubernetes-build` jobs over to point to the new scenario (and turn down the canary build jobs)
- Flip release tooling to use the new bucket

Again, that's an incredible rough draft, but let me know what you think.

justaugustus:
(I should assign myself here as well)
/assign

justaugustus:
FYI: @kubernetes/release-engineering @kubernetes/ci-signal

spiffxp:
Are there any strong opinions about:
[...]

PR #1110, opened 2020-08-04
Add k8s-release project and buckets for ci

spiffxp:
We need a home for community-owned buckets that are equivalent to:
- gs://kubernetes-release-dev  (hosts periodic builds of kubernetes)
- gs://kuberentes-release-pull (hosts pr builds of kubernetes, eg: from e2e test jobs)

I'm suggesting we create a `k8s-release` project, and set it up
identically to the `k8s-release-test-prod` project currently in use by
release-engineering.

Then setup two equivalent buckets:
- gs://k8s-release-dev
- gs://k8s-release-pull

And give prow write access to these buckets. I suspect prow will need
admin access, but let's find out.

ref: https://github.com/kubernetes/k8s.io/issues/846#issuecomment-668795671

### DIFF
--- a/infra/gcp/ensure-release-projects.sh
+++ b/infra/gcp/ensure-release-projects.sh
@@ -34,6 +34,7 @@ function usage() {
 
 # NB: Please keep this sorted.
 PROJECTS=(
+    k8s-release
     k8s-release-test-prod
 )
 
@@ -134,3 +135,41 @@ for PROJECT; do
 
     color 6 "Done"
 done
+
+## Special case: setup buckets that are used by CI
+
+# community-owned equivalents to gs://kubernetes-release-{dev,pull}
+RELEASE_BUCKETS=(
+  "gs://k8s-release-dev"
+  "gs://k8s-release-dev-asia"
+  "gs://k8s-release-dev-eu"
+  "gs://k8s-release-pull"
+)
+PROW_BUILD_SVCACCT=$(svc_acct_email "k8s-infra-prow-build" "prow-build")
+
+for BUCKET in "${ALL_BUCKETS[@]}"; do
+    color 3 "Configuring bucket: ${BUCKET}"
+
+    # Create the bucket
+    color 6 "Ensuring the bucket exists and is world readable"
+    ensure_public_gcs_bucket "k8s-release" "${BUCKET}"
+
+    # Enable admins on the bucket
+    color 6 "Empowering GCS admins"
+    empower_gcs_admins "k8s-release" "${BUCKET}"
+
+    # Enable prow to write to the bucket
+    # TODO(spiffxp): I almost guarantee prow will need admin privileges but 
+    #                let's start restricted and find out
+    empower_svcacct_to_write_gcs_bucket "${PROW_BUILD_SVCACCT}" "${BUCKET}"
+
+    # Enable writers on the bucket
+    for group in ${ADMINS} ${WRITERS}; do
+        color 6 "Empowering ${group} to GCS"
+        empower_group_to_write_gcs_bucket "${group}" "${BUCKET}"
+    done
+done
+
+color 6 "Ensure auto-deletion policies are set for CI buckets"
+ensure_gcs_bucket_auto_deletion "gs://k8s-release-dev" 90
+ensure_gcs_bucket_auto_deletion "gs://k8s-release-pull" 14


spiffxp:
/cc @justaugustus @tpepper 
for input from @kubernetes/release-engineering

/cc @thockin 
FYI

justaugustus:
Matches what Aaron and I chatted about.
/lgtm
/approve

/hold for any other reviewers; lift when ready

thockin:
Why `kubernetes-release-dev` (and I assume you mean the 3 regionals) and `kubernetes-release-pull` and not `kubernetes-release` (and regionals)?

Looking at stats, `kubernetes-release` is the next-highest cost bucket.  

I think there's a way to literally have the bucket moved between projects - would that be a better path?

I know @justinsb started thinking about billing breakdown per-bucket.  Unfortunately, GCP's normal billing breakdown stops at a project+SKU, but we can enable finer-grained logs.  We should decide if it makes sense to lump ANYTHING together, or if we're better off with a project per bucket (or set of common-purpose buckets).  Puke.

In other words, what if we just moved them all into k8s-artifacts-prod (or into a new k8s-release project)?  That would mean that all external tools remain functional.

BTW - I have no idea where kubernetes-release-pull lives.  Sigh.

spiffxp:
>Why kubernetes-release-dev (and I assume you mean the 3 regionals) and kubernetes-release-pull and not kubernetes-release (and regionals)?

I'm not aiming at (monetary) cost right now, but opportunity cost.  We're blocked on moving some of the CI jobs out of google.com because `gs://kuberetes-release-dev` (technically its host project `google-containers`) won't allow non-google.com accounts to be added to IAM (ref: https://github.com/kubernetes/k8s.io/issues/846). 

Moving CI related builds into community owned bucket(s) means we can move all release-blocking/merge-blocking jobs into community owned infra, and will empower more people to help out.

>I think there's a way to literally have the bucket moved between projects - would that be a better path?

Discussed offline.  There does not appear to be such a way, unfortunately.  So migrating piecemeal or stop-the-world-and-sync are our options.

>Unfortunately, GCP's normal billing breakdown stops at a project+SKU, but we can enable finer-grained logs.

The "stops at a project+SKU" is why I'm proposing something (`k8s-release`) instead of `k8s-artifacts-prod`.  The dividing line in my head is `k8s-artifacts-prod` is for final/static release artifacts for every single kubernetes subproject; `k8s-release` is for all the intermediary stuff (per-pr, postsubmits, nightlies, etc) related to kubernetes

>BTW - I have no idea where kubernetes-release-pull lives.

It's part of the `kubernetes-jenkins` project.  That's going to be a fun one to migrate away from.  But we don't have to do that today.

thockin:
>>Why kubernetes-release-dev (and I assume you mean the 3 regionals) and kubernetes-release-pull and not kubernetes-release (and regionals)?
> 
> I'm not aiming at (monetary) cost right now, but opportunity cost.  We're blocked on moving some of the CI jobs out of google.com because `gs://kuberetes-release-dev` (technically its host project `google-containers`) won't allow non-google.com accounts to be added to IAM (ref: https://github.com/kubernetes/k8s.io/issues/846). 

Do we need just k-release-dev or also the -asia and -eu sister-buckets?

> Moving CI related builds into community owned bucket(s) means we can move all release-blocking/merge-blocking jobs into community owned infra, and will empower more people to help out.

ACK

> 
>> I think there's a way to literally have the bucket moved between projects - would that be a better path?
> 
> Discussed offline.  There does not appear to be such a way, unfortunately.  So migrating piecemeal or stop-the-world-and-sync are our options.

Correct.  For kubernetes-release, we might go a long way by changing dl.k8s.io
and seeing how much is left.

>> Unfortunately, GCP's normal billing breakdown stops at a project+SKU, but we can enable finer-grained logs.
> 
> The "stops at a project+SKU" is why I'm proposing something (`k8s-release`) instead of `k8s-artifacts-prod`.  The dividing line in my head is `k8s-artifacts-prod` is for final/static release artifacts for every single kubernetes subproject; `k8s-release` is for all the intermediary stuff (per-pr, postsubmits, nightlies, etc) related to kubernetes

Where would the current 'kubernetes-release' bucket go? Are we OK conflating
 the bill for dev, pull, and release together?  Or should we make a "one bucket
 per project" rule for these?

Regardless, we should put an LB VIP in front of whatever bucket(s) so we don't
have to deal with this again. See ensure-prod-storage-gclb.sh for example.  We
could maybe set up ONE VIP to cover everything (except the existing VIP, which
exposes at the root, sigh).

I am fine with the overall plan to make new buckets and copy stuff over, then
flip pointers, for these.  We will need to consider kubernetes-release soon,
too.

>>BTW - I have no idea where kubernetes-release-pull lives.
> 
> It's part of the `kubernetes-jenkins` project.  That's going to be a fun one to migrate away from.  But we don't have to do that today.

infra/gcp/ensure-release-projects.sh
```diff
@@ -58,6 +59,8 @@ for PROJECT; do
     color 6 "Ensuring project exists: ${PROJECT}"
     ensure_project "${PROJECT}"
 
+    continue
```
MushuEE, line 62: Does this not mean everything below it is never run?

spiffxp: 🤦 good catch

spiffxp:
>Do we need just k-release-dev or also the -asia and -eu sister-buckets?

So I added equivalents for these.  But I have to admit I'm not sure how the existing kubernetes-release buckets are mirrored to their asia/eu siblings.  I would be curious to see how much traffic these buckets are seeing compared to the non-geo (us, I guess) bucket.

I misspoke earlier, gs://kubernetes-release-pull is under kubernetes-jenkins-pull.  I confirmed it has no geo-specific buckets.

spiffxp:
To unblock the current set of release-blocking jobs I don't think mirroring to geo-specific buckets is required.  I think looking at existing traffic will help us decide how much of a priority setting up mirroring is.

spiffxp:
Running `./infra/gcp/ensure-release-projects.sh`

spiffxp:
re: https://github.com/kubernetes/k8s.io/pull/1110#issuecomment-670060188

>So I added equivalents for these. But I have to admit I'm not sure how the existing kubernetes-release buckets are mirrored to their asia/eu siblings. I would be curious to see how much traffic these buckets are seeing compared to the non-geo (us, I guess) bucket.

I did some digging and discovered a few things re: syncing to regional buckets:
- kubernetes-release-dev hasn't been synced to regional buckets for over a year
- kubernetes-release-eu/-asia buckets serve two orders of magnitude less traffic than kubernetes-release (I can only see back by six weeks

Given that, I think regional syncing should not be a blocker to moving forward.  If we decide it's necessary after the fact, we should look at using Storage Transfer Service (https://cloud.google.com/storage-transfer/docs/create-manage-transfer-console) to sync

PR #1118, opened 2020-08-07
Provision 10 gpu GCP projects

spiffxp:
This is to account for:
- the [~5 projects of peak usage k8s-prow-builds' gpu-project pool currently sees](https://monitoring.prow.k8s.io/d/wSrfvNxWz/boskos-resource-usage?orgId=1&from=now-30d&to=now&fullscreen&panelId=6)
- the [`max_concurrency: 15` pull-kubernetes-e2e-gce-device-plugin-gpu presubmit job](https://github.com/kubernetes/test-infra/blob/f33224993b51b111050b865a83686cc810677c49/config/jobs/kubernetes/sig-cloud-provider/gcp/gpu/gcp-gpu-presubmits.yaml#L12-L20)  that is all currently pinned to a single gcp project 

The actual raising of quota for these projects is going to need to be done by hand, unfortunately. So before I do this big ugly manual thing, I wanted to make sure I'm not missing something simpler / more obvious.

We could make a single gcp project to hold all of the presubmits, with a much higher gpu quota.  This is equivalent to what is done now.  But I feel like sticking to pools of projects means we're less likely to develop pets instead of cattle.

ref: https://github.com/kubernetes/k8s.io/issues/1095

/hold
for comment

### DIFF
--- a/infra/gcp/ensure-e2e-projects.sh
+++ b/infra/gcp/ensure-e2e-projects.sh
@@ -92,10 +92,18 @@ for i in $(seq 1 5); do
   E2E_SCALE_PROJECTS+=($(printf "k8s-infra-e2e-boskos-scale-%02i" $i))
 done
 
+# e2e projects for gpu jobs
+# - us-west1 Committed NVIDIA K80 GPUs raised to 2
+E2E_GPU_PROJECTS=()
+for i in $(seq 1 10); do
+  E2E_GPU_PROJECTS+=($(printf "k8s-infra-e2e-boskos-gpu-%02i" $i))
+done
+
 E2E_PROJECTS=(
   "${E2E_MANUAL_PROJECTS[@]}"
   "${E2E_BOSKOS_PROJECTS[@]}"
   "${E2E_SCALE_PROJECTS[@]}"
+  "${E2E_GPU_PROJECTS[@]}"
 )
 
 if [ $# = 0 ]; then


spiffxp:
/cc @BenTheElder @liggitt 
ref: https://github.com/kubernetes/kubernetes/issues/92937#issuecomment-662178519 I like the idea of making pull-kubernetes-e2e-gce-device-plugin-gpu `always_run: false`, then we'd need way fewer of these special gpu projects lying around

liggitt:
+1 for postsubmit and `always_run: false` presubmit for manual testing

BenTheElder:
yes, please.

BenTheElder:
further, I still aim to stop requiring things in presubmit that we can't
reasonably expect contributors to run and debug themselves anyhow.

BenTheElder:
it's also unreasonable for kubeadm to wait for the slow job in CI at least,
it's trivial to support fast and the latency is much better.
kind is kubeadm and does not need to wait.

dims:
/lgtm

based on the feedback above, if you still want to go through with it :) please remove hold

spiffxp:
https://github.com/kubernetes/test-infra/pull/18728 sets `always_run: false` and `max_concurrency: 5`, I'm going to drop down to 10 projects

spiffxp:
OK I cut the number of projects in half

spiffxp:
Ran `./infra/gcp/ensure-e2e-projects.sh`

PR #1123, opened 2020-08-07
bump new image of the k8s container image promoter

yodahekinsew:
Bump a new image of the k8s container image promoter which contains an update for the image vulnerability check


yodahekinsew:
@listx @prathak07

listx:
/hold

Remove hold when ready.

PR #1125, opened 2020-08-07
Add gpu-project pool to prow-build boskos

spiffxp:
This is part of https://github.com/kubernetes/k8s.io/issues/1095

Adds the projects provisioned via https://github.com/kubernetes/k8s.io/pull/1118

PR #1127, opened 2020-08-08
Add gpu-project to boskos components

spiffxp:
The gpu-project pool has been added, but since it defaults to dirty, and
nothing is configured to clean or reap it, boskos alerts are firing


spiffxp:
Was followup to https://github.com/kubernetes/k8s.io/pull/1125

PR #1148, opened 2020-08-13
Add staging project for cluster-addons

justinsb:
As we want to publish the addon manager image.


spiffxp:
Ran './infra/gcp/ensure-staging.sh cluster-addons`

PR #1150, opened 2020-08-14
Add dns records for the contributor site

mrbobbytables:
I believe these are setup correctly.
Netlify will handle the redirect of k8s.dev to kubernetes.dev.

/cc @cblecker

### DIFF
--- a/dns/README.md
+++ b/dns/README.md
@@ -7,6 +7,8 @@ Zones we manage:
   - k8s.io
   - x-k8s.io
   - k8s-e2e.com
+  - kubernetes.dev
+  - k8s.dev
 
 ## How to become an admin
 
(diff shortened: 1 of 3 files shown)


cblecker:
Are we ready to go live with the site?

mrbobbytables:
Just about, there is 1 PR open in k/community and the blog post that needs to be merged - but I'd like to get this going so I can make sure all the netlify configs are set / redirects configured on that side.

Think a quiet launch and then when those other 2 items are sorted out we then announce / socialize it.

mrbobbytables:
To follow-up - here is the soft launch plan:
- DNS records are updated, the links are sent to the ContribEx ML for verification / vetting and **not** broadcast widely
- If after 1 week everything appears fine, any small issues resolved we move to launch with a coordinated effort with the marketing team.

PR #1151, opened 2020-08-14
disable promotions from gcr.io/google-containers

listx:
Now that the domain flip [1] has landed, there is no need to keep these
configs "live" because they are frozen. This makes the cip-auditor [2]
skip these images when considering a pubsub message event, which should
reduce GCR read traffic significantly.

[1]: https://github.com/kubernetes/k8s.io/blob/master/k8s.gcr.io/Vanity-Domain-Flip.md
[2]: https://github.com/kubernetes/k8s.io/blob/master/k8s.gcr.io/Vanity-Domain-Flip.md#auditing-cip-auditor


justaugustus:
> Can you explain this a bit more? Will the promoter no longer know the provnenance of those legacy images and flag them as unknown?

+1. The PR seems to primarily move directories.
Do we have today/could we have a `legacy|disabled|archived-registries` field somewhere? This is something we could test against.

listx:
> Will the promoter no longer know the provnenance of those legacy images

Yes. Those legacy images are completely unknown to the promoter with this change.

> and flag them as unknown?

No, the legacy images will not be flagged. For example, imagine we have 1 image defined declaratively for promotion in the manifests, but also that we have 500 other images in production that are not in the manifests at all. During a promotion, the promoter will only make sure that the 1 image is promoted into prod. The other 500 extraneous images are just noise.

For the auditor, however, if any of the 500 images in prod are mutated in any way, it *will still be flagged* (which is what we want), because the promoter manifests will not have info about these 500.

listx:
> > Can you explain this a bit more? Will the promoter no longer know the provnenance of those legacy images and flag them as unknown?
> 
> +1. The PR seems to primarily move directories.
> Do we have today/could we have a `legacy|disabled|archived-registries` field somewhere? This is something we could test against.

We'd have to add that. This takes less effort for virtually the same result, so that's why I reached for this change first.

listx:
Ping?

thockin (commented):
Thanks!

/lgtm
/approve

spiffxp:
/hold cancel
Discussed with @listx offline, this should be good to go

PR #1159, opened 2020-08-18
Fix DNS records for contrib site

mrbobbytables:
Root level CNAME (CNAME flattening) is not supported.

The A record value was provided by netlify's documentation as an alternative.

/assign @cblecker


BenTheElder:
https://prow.k8s.io/?type=postsubmit&job=*dns* :green_heart: 
https://prow.k8s.io/view/gs/kubernetes-jenkins/logs/post-k8sio-dns-update/1295861440007114752

BenTheElder:
MUCH faster than past runs though ...?

PR #1161, opened 2020-08-19
Use boskos' new image repo

spiffxp:
Followup to https://github.com/kubernetes/k8s.io/pull/1160


BenTheElder:
/lgtm
/approve
deploy from staging seems legit

spiffxp:
>deploy from staging seems legit

Yeah I agree, I think we can ask for a promoted image tomorrow if things look better

PR #1162, opened 2020-08-19
Enable serviceusage api in cluster projects

spiffxp:
This is primarily for boskos' janitor's benefit, so I could see the
case being made that this should be done only for the one project
that is running boskos

This is ostensibly to allow a service account in project A to call
`gcloud services list --project=project-B`

Followup to https://github.com/kubernetes/k8s.io/pull/1160
ref: https://github.com/kubernetes/test-infra/issues/18897


spiffxp:
Meanwhile I manually ran `gcloud services enable serviceusage.googleapis.com --project=k8s-infra-prow-build` to unblock boskos

spiffxp:
Ran `terraform apply` for both prow build clusters

Issue #1132, opened 2020-08-11
Raise k8s-infra-prow-build quotas in anticipation of handling merge-blocking jobs


spiffxp:
The node pool is currently setup as 3 * (6 to 30) n1-highmem-8's.  We don't have enough quota to hit max nodepool size

At least in terms of resources, we need at least:
- `3 * 30 * 8` = 720 CPU's
- `3 * 30 * 250` = 22500 Gi SSD capacity
- `3 * 30` = 90 in-use IP addresses

If we want to match the size of the k8s-prow-builds cluster, which has 160 nodes, we should ask for more

/wg k8s-infra
/area prow

spiffxp:
No quota for memory.

Submitted requests for:
- 1024 CPU's in us-central1
- 100 in-use IP addresses in us-central1

spiffxp:
Well, the 1024 CPU request went through just fine.

The 100 in-use IP's...
>Unfortunately, we are unable to grant you additional quota at this time. If this is a new project please wait 48h until you resubmit the request or until your Billing account has additional history.

So, I'll hold this open and see what comes back in two days.  Quota is 69 in-use IP addresses until then.

spiffxp:
Part of https://github.com/kubernetes/test-infra/issues/18550

spiffxp:
>setup more small build clusters

I tried raising CPU and SSD quota in us-west1 to be able to create an equivalently sized build cluster in the same k8s-infra-prow-build project over there.  Both requests automatically rejected.

spiffxp:
There is suspicion that moving to n1-highmem-16's has actually increased flakiness.  Specifically for these jobs, across release branches:
- https://github.com/kubernetes/kubernetes/issues/94197 - verify
- https://github.com/kubernetes/kubernetes/issues/94184 - integration

spiffxp:
Opened https://github.com/kubernetes/k8s.io/issues/1173 to track the rollback

spiffxp:
>setup more small build clusters

I was able to raise CPU quota in us-east1 to 1024, but was rejected for SSD and IP quota requests.

Next step would be to try raising quotas for a different GCP project, in case k8s-infra-prow-build has gotten flagged for some reason.

spiffxp:
>At least in terms of resources, we need at least:
> - 3 * 30 * 8 = 720 CPU's
> - 3 * 30 * 250 = 22500 Gi SSD capacity
> - 3 * 30 = 90 in-use IP addresses

[...]

PR #1172, opened 2020-08-24
Add an n1_highmem_8 nodepool to k8s-infra-prow-build to prep for migration

spiffxp:
There is a suspicion that switching to n1-highmem-16 nodes has caused more jobs-per-node to be scheduled, thus leading to more contention over resources that aren't accounted for by the scheduler (such as IOPS)

This PR is the first step in undoing that change, by moving us back to n1-highmem-8's.

I am concerned that moving back to more, smaller nodes will cause us to hit our quota of IP's, and we'll have to shift jobs off of this cluster until we can find some other way to mitigate (ref: https://github.com/kubernetes/k8s.io/issues/1132#issuecomment-678389503)

### DIFF
--- a/infra/gcp/clusters/projects/k8s-infra-prow-build/prow-build/main.tf
+++ b/infra/gcp/clusters/projects/k8s-infra-prow-build/prow-build/main.tf
@@ -126,6 +126,25 @@ module "prow_build_nodepool_n1_highmem_16" {
   service_account = module.prow_build_cluster.cluster_node_sa.email
 }
 
+module "prow_build_nodepool_n1_highmem_8" {
+  source = "../../../modules/gke-nodepool"
+  project_name    = local.project_id
+  cluster_name    = module.prow_build_cluster.cluster.name
+  location        = module.prow_build_cluster.cluster.location
+  name            = "pool3"
+  initial_count   = 1
+  min_count       = 1
+  max_count       = 30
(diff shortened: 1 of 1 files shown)


hasheddan:
xref: https://github.com/kubernetes/kubernetes/issues/94184 https://github.com/kubernetes/kubernetes/issues/94197

spiffxp:
Will track migration under https://github.com/kubernetes/k8s.io/issues/1173

infra/gcp/clusters/projects/k8s-infra-prow-build/prow-build/main.tf
```diff
@@ -126,6 +126,25 @@ module "prow_build_nodepool_n1_highmem_16" {
...
+  location        = module.prow_build_cluster.cluster.location
+  name            = "pool3"
+  initial_count   = 1
+  min_count       = 1
+  max_count       = 30
+  # kind-ipv6 jobs need an ipv6 stack; COS doesn't provide one, so we need to
+  # use an UBUNTU image instead. Keep parity with the existing google.com 
+  # k8s-prow-builds/prow cluster by using the CONTAINERD variant
+  image_type      = "UBUNTU_CONTAINERD"
+  machine_type    = "n1-highmem-16"
```
spiffxp, line 142: Welp.  Looks like this is gonna need a redo

PR #1087, opened 2020-07-31
Promote kops 1.19.0-alpha.2 binary artifacts

justinsb:

PR #1176, opened 2020-08-25
Promote kops 1.19.0-alpha.3 binary artifacts

hakman:
/cc @rifelpet @justinsb


hakman:
Depends on https://github.com/kubernetes/k8s.io/pull/1087
/hold

justinsb:
Actually I think this is safe to merge, it won't do anything yet, but there's no reason not to merge them as we verify them IMO.

Verified hashes against the staging build.

I was also able to verify some of the hashes against `make bazel-version-dist && find .bazelbuild/dist/ -type f | sort | xargs  sha256sum`, the images don't mount but I think that's because we're including a timestamp and need to pin that to the commit SHA or similar.

/approve
/lgtm

PR #1179, opened 2020-08-26
Update README.md

wangxy518:
adminstrator->administrator


spiffxp:
/cc @bartsmykla 
we should update the README now that https://github.com/kubernetes/test-infra/pull/18919 has merged

bartsmykla:
Will do soon

PR #1186, opened 2020-08-28
Add n1-highmem-8 nodepool with larger SSDs to k8s-infra-prow-build

spiffxp:
This raises the cost of the nodepool by ~14% but doubles the IOPS available per node.


spiffxp:
More rationale in https://github.com/kubernetes/k8s.io/issues/1187, which I will also use to track migration to this nodepool

PR #1199, opened 2020-09-01
Fix npd image tag and promote v0.8.4

Random-Liu:
The image tag is `v0.8.3` instead of `0.8.3`.


k8s.gcr.io/images/k8s-staging-npd/images.yaml
```diff
@@ -1,3 +1,4 @@
 - name: node-problem-detector
   dmap:
-    "sha256:9343960c22c6033903413d032bdd6b903e34ff69862913aca31f3008a5800c51": ["0.8.3"]
```
listx, line 3: Hmm, the presubmit check should have caught this. Tag deletions are meaningless because we never ever delete tags in production. So the 0.8.3 tag is still up there.

PR #1198, opened 2020-09-01
use Prow trusted cluster for workload identity

listx:
This is a typo fix. The "-builds" suffix is for the regular non-trusted Prow
cluster. However, the Prow job that was created [1] is configured for the
trusted cluster -- hence this change.

[1]: https://github.com/kubernetes/test-infra/pull/19003

PR #1202, opened 2020-09-01
Refactor KSA CIP binding to name intended prow build clusters

spiffxp:
Instead of having to know what the magic names "k8s-prow" vs
"k8s-prow-builds" imply, and having to copy-paste for "and also setup
workload identity for the equivalent clusters in k8s-infra", let's use
arrays

Followup to https://github.com/kubernetes/k8s.io/pull/1198#issuecomment-685004110


spiffxp:
/cc @listx 
I'm hoping this makes thing clearer

listx:
This is a step forward in readability! +1

/lgtm

spiffxp:
I got yanked into internal stuff before I could cap this off.  I ran the script this morning shortly after this merged... but it turns out there was a typo, so what I ended up actually running is now up as https://github.com/kubernetes/k8s.io/pull/1204

PR #1216, opened 2020-09-04
Add staging for git-sync

thockin:


stp-ip:
+1 from me, but as I am one of the reviewers listed not doing to lgtm etc.

justaugustus (changes requested):
Hey @thockin -- I think you'll need to actuate the project in https://github.com/kubernetes/k8s.io/blob/af78aa95a4705166a4b9a920160125fd89c3cf48/infra/gcp/ensure-staging-storage.sh#L43, right?

thockin:
fixed!

justaugustus:
Thanks Tim!
/lgtm

stp-ip:
> /hold what's this for again?

git-sync is a project done at k/git-sync and is therefore in the vicinity of Kubernetes. It's main use is as a git pulling sidecar.
Container builds and publishing is currently done manually by Tim and with this PR would be moved to be governed by 2 people instead of one and also open to non googlers. Also open to be automated as well later etc. pp.

dims:
Ack @stp-ip we can remove hold when someone runs the script for creating the staging repo.

thockin:
Group added.

thockin:
Staging created

PR #1218, opened 2020-09-08
Temporarily remove k8s-infra-e2e-boskos-scale-12 from scalability pool

mm4tt:
Ref. https://github.com/kubernetes/kubernetes/issues/94450


mm4tt:
/assign @bartsmykla

@bartsmykla, could you take a look? It's causing our tests to fail randomly

mm4tt:
Thanks!

PR #1222, opened 2020-09-09
Updated dns/README after implementing automation

bartsmykla:
As there is currently no need to manually update dnses and it's done
automatically (prow manages that), I updated the documentation to
reflect the changes

ref. https://github.com/kubernetes/k8s.io/issues/783#issuecomment-682190399

/cc @spiffxp 


spiffxp (approved):
/approve
/lgtm

PR #1224, opened 2020-09-09
Add rajula96reddy to slack-infra google groups

ameukam:
Request access for @rajula96reddy for slack-infra staging container
registry and slack-infra k8s namespace.


spiffxp (approved):
/approve
/lgtm
I'm going off of the fact that @ameukam is already in these groups.

It would probably be a good idea to get some overlap between these group members and  https://github.com/kubernetes-sigs/slack-infra/blob/master/OWNERS

PR #1226, opened 2020-09-09
adding marosset to leads group

marosset:
Adding marosset (SIG-Windows co-chair) to leads group


marosset:
@ddebroy would you like me to include you in this PR as well?
If so let me know which email address to use for you.

Issue #1231, opened 2020-09-10
Raise k8s-infra-prow-build cluster nodepool max size


spiffxp:
Well, it took about two weeks from the last "let's wait and see" comment (ref: https://github.com/kubernetes/k8s.io/issues/1132#issuecomment-682246486)

This is a regional cluster spread across 3 zones, with a nodepool that will autoscale up to 30 nodes per zone.  So, 90 nodes total.  We've hit that limit a few times now.

[Metrics explorer: VM instance uptime (count)](https://console.cloud.google.com/monitoring/metrics-explorer?project=k8s-infra-prow-build&timeDomain=1h&pageState=%7B%22xyChart%22:%7B%22dataSets%22:%5B%7B%22timeSeriesFilter%22:%7B%22filter%22:%22metric.type%3D%5C%22compute.googleapis.com%2Finstance%2Fuptime%5C%22%20resource.type%3D%5C%22gce_instance%5C%22%20metadata.system_labels.%5C%22state%5C%22%3D%5C%22ACTIVE%5C%22%22,%22minAlignmentPeriod%22:%22120s%22,%22aggregations%22:%5B%7B%22perSeriesAligner%22:%22ALIGN_DELTA%22,%22crossSeriesReducer%22:%22REDUCE_COUNT%22,%22groupByFields%22:%5B%22metadata.system_labels.%5C%22region%5C%22%22%5D%7D,%7B%22crossSeriesReducer%22:%22REDUCE_NONE%22%7D%5D%7D,%22targetAxis%22:%22Y1%22,%22plotType%22:%22STACKED_AREA%22,%22legendTemplate%22:%22$%7Bmetadata.system_labels%5C%5C.region%7D%22%7D%5D,%22options%22:%7B%22mode%22:%22COLOR%22%7D,%22constantLines%22:%5B%5D,%22timeshiftDuration%22:%220s%22,%22y1Axis%22:%7B%22label%22:%22y1Axis%22,%22scale%22:%22LINEAR%22%7D%7D,%22isAutoRefresh%22:true,%22timeSelection%22:%7B%22timeRange%22:%221m%22%7D%7D)

Over the last month
![Screen Shot 2020-09-09 at 6 31 11 PM](https://user-images.githubusercontent.com/49258/92671278-abed6980-f2ca-11ea-8cec-608b4997260f.png)

Let's zoom in on some of those peaks
[...]

PR #1232, opened 2020-09-10
Raise prow-build max nodepool size to 50

spiffxp:
Instead of 3 x 30 = 90 nodes total, let's go up to 150 total

The build cluster we are migrating prow jobs from had a fixed size of
160 nodes.  We're starting to hit the limit of 90 nodes with increased
PR traffic.

Part of https://github.com/kubernetes/k8s.io/issues/1231


spiffxp:
`terraform plan` says this will be an update-in-place operation

spiffxp:
/cc @BenTheElder 
FYI @kubernetes/ci-signal

spiffxp:
Ran terraform apply

Issue #1120, opened 2020-08-07
Upgrade k8s-infra prow build clusters from v1.14 to v1.15


spiffxp:
v1.14 deprecation was announced here: https://cloud.google.com/kubernetes-engine/docs/release-notes#coming-soon-20200722

The clusters should have been automatically upgraded: https://cloud.google.com/kubernetes-engine/docs/release-notes#scheduled_automatic_upgrades

This issue is to confirm whether they have been, and if not, initiate such an upgrade

/area prow
/wg k8s-infra
/sig testing

spiffxp:
We're sitting on 1.14.10-gke.42 for the control plane, and 1.14.10-gke.37 for the main node pool

spiffxp:
We're still sitting on 1.14.  I am waiting until we've drained the bulk of outstanding v1.20 PR's (ref: https://groups.google.com/g/kubernetes-dev/c/YXGBa6pxLzo/discussion) before explicitly triggering this.

There's still a chance it'll happen when we're not watching though.

spiffxp:
Upgrading k8s-infra-prow-build's control plane from 1.14.10-gke.42 to 1.15.12-gke.17

spiffxp:
Control plane upgraded, next the greenhouse nodepool.  This may disrupt bazel-based jobs, though they should fall back to not using the cache when it's unavailable.

spiffxp:
Greenhouse nodepool upgraded.  Waiting until https://github.com/kubernetes/test-infra/pull/19182#issuecomment-691226512 is resolved before proceeding with the main nodepool

spiffxp:
/assign
OK, prow was reverted and sinker is behaving as expected again.

spiffxp:
Upgrading k8s-infra-prow-build's default node pool from 1.14.10-gke.42 to 1.15.12-gke.17

spiffxp:
29/41 nodes done

spiffxp:
And everything is at 1.15.12-gke.17 after ~4 hours

... so I think that node-by-node-upgrade and cluster-autoscaling aren't the best of friends.  I suspect this would have gone more quickly if we had spun up an entirely new nodepool that was on 1.15.12 to begin with, and cordoned the the old nodepool.

I would like to move up to v1.16 next but I think we'll leave it here for the weekend

spiffxp:
[...]

PR #1239, opened 2020-09-11
Add pdb for pods created by prow

spiffxp:
This should allow autoscaling/draining to avoid deleting nodes that
are still running pods scheduled by prow

Intended to make https://github.com/kubernetes/k8s.io/issues/1120 less disruptive


hasheddan (approved):
/lgtm

Thanks @spiffxp!

PR #1243, opened 2020-09-12
Don't allow duplicate members in groups

spiffxp:
I noticed two groups had members flipping back and forth between member
and owner.  Add tests to catch this, and adjust membership accordingly.

### DIFF
--- a/groups/groups.yaml
+++ b/groups/groups.yaml
@@ -76,7 +76,6 @@ groups:
       - jameswangel@gmail.com
       - jbeale@inguardians.com
       - jbelamaric@google.com
-      - jdumars@gmail.com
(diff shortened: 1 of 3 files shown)


spiffxp:
/cc @jdumars @tpepper 
FYI, as the affected members

tpepper:
/lgtm

This probably explains some oddities on me, although many Google web pages' .js don't handle multiple identities well either (eg: 'u/0' in an url vs 'u/1') and I had intentionally attempted to have work-me and non-work-me on high prio things like patch release in order to be able to pitch in even when I was away from work managed computer and identity.  I've given up on that as it was fragile and there should just be work-me now /shrug

PR #1300, opened 2020-10-02
Correct capitialisation of email address

JamesLaverack:
Jacky is unable to access the mailing list, despite his email being in the groups.yaml file. We believe this is caused by a capitalisation difference between his email as presented for authentication (`@outlook.com`) and as written in the file (`@Outlook.com`).


JamesLaverack:
I think the CI Robot has mistakenly inferred that my use of that `@` symbol in a commit message is a GitHub mention. GitHub itself doesn't seem to parse it as one. Is there a way to tell the bot that it's mistaken, or should I amend the commit to remove the `@` symbol?

PR #1324, opened 2020-10-12
Add eddiezane to k8s-infra-prow-viewers

eddiezane:
1.20 CI signal shadow digging in here.

/cc @RobertKielty


hasheddan (approved):
/lgtm

PR #1358, opened 2020-10-23
Add ip-masq-agent v2.6.0 images

varunmar:


k8s.gcr.io/images/k8s-staging-networking/images.yaml
```diff
@@ -1 +1,12 @@
-# No images yet
+- name: ip-masq-agent-amd64
```
MrHohn, line 1: What about ip-masq-agent (the multi-arch image)? I think http://github.com/kubernetes-sigs/kind relies on that.

varunmar: Hmm. Yeah, I see the ip-masq-agent image in the old repository.

Although - how is that multi-arch image built? I don't see the build for that target in the makefile for ip-masq-agent.

MrHohn: Oh, I thought it would be pushed automatically via the `make all-push` command: https://github.com/kubernetes-sigs/ip-masq-agent/blob/f59e0b275e272a823c0c31c57451035eaf90b26c/Makefile#L88-L91

MrHohn: I see now, the release.md is not matching up for the push command :(

wespanther: Looks like it happens here - https://github.com/kubernetes-sigs/ip-masq-agent/blob/master/Makefile#L65

And there's an all-build target that uses ALL_ARCH

varunmar: Ah, it's a manifest list, not an image type. I found it now. Adding it.
Thanks!

MrHohn (approved):
Thanks!
/lgtm
/approve

PR #1369, opened 2020-10-29
promote csi sidecar images

msau42:
This promotes the set of patch releases from 2 versions ago. Note that these images are mirrored from quay.io since we don't have cloudbuild setup in those release branches. Therefore, there is also no multiarch support.


pohly (approved):
/lgtm

PR #1372, opened 2020-10-29
add aojea to k8s.io group

aojea:
let's start adding ipv6 support and see how can I help more later.


thockin (commented):
Thanks!

/lgtm
/approve

PR #1374, opened 2020-10-29
Hello IPv6 I'm kubernetes

aojea:
The ingresses work:

- prod
```
 curl [2600:1901:0:26f3::]
<!DOCTYPE html>
<html lang=en>
  <meta charset=utf-8>
  <meta name=viewport content="initial-scale=1, minimum-scale=1, width=device-width">
  <title>Error 404 (Not Found)!!1</title>
  <style>
```

- canary 

```
curl [2600:1901:0:552b::]
<html>
<head><title>301 Moved Permanently</title></head>
<body bgcolor="white">
<center><h1>301 Moved Permanently</h1></center>
<hr><center>nginx/1.10.3</center>
</body>
</html>
```

the DNS, since all the domains using CNAME, we only need to add an AAAA to the redirect. domain


infra/gcp/ensure-static-ips.sh
```diff
@@ -40,6 +40,8 @@ ADDRESSES=(
     "gcsweb-k8s-io"
     "k8s-io-ingress-canary"
     "k8s-io-ingress-prod"
+    "k8s-io-ingress-canary-v6"
```
thockin, line 43: This calls ensure_global_address from infra/gcp/lib.sh, but that function doesn't understand v6.  Maybe we should just use the "-v6" suffix to tell that function to create IPv6 addresses?

k8s.io/canary.sh
```diff
@@ -34,3 +34,6 @@ while true; do
 done
 
 make test TARGET_IP=34.102.239.89
+
+# it needs IPv6 connectivity
+# make test TARGET_IP=2600:1901:0:26f3::
```
thockin, line 39: trailing newline?

dns/zone-configs/k8s.io._0_base.yaml
```diff
@@ -29,9 +31,10 @@ dummy:
 # Our vanity redirector.  This is not just 'k8s.io', on the off chance this
 # becomes different from the main record. (@thockin)
 redirect:
-  type: A
-  value: 34.107.204.206
-
+  - type: A
```
thockin, line 34: I guess the DNS tests will just work, but I guess we'll find out?

thockin (commented):
To be clear for readers - the ingress has been instantiated but the DNS has not.

dns/zone-configs/k8s.io._0_base.yaml
aojea: this is the part that scaries me the most, since everything depends on redirect and is moving to an array :crossed_fingers: 
I saw similar records before, but :shrug:

thockin:
I can manually add the names and you can run the test

thockin:
OK, I manually added AAAA records for canary.k8s.io and
redirect.canary.k8s.io - see if the tests pass :)

infra/gcp/ensure-static-ips.sh
aojea: check if that is ok with you, my bash foo is not too great :)

aojea:
here you go 
image

PR #1379, opened 2020-10-30
Fix YAML quoting

thockin:
cc @aojea


aojea:
aahg,

 actually I had doubts because I saw other without them 
https://github.com/kubernetes/k8s.io/blob/3263e91505ba5b455462bc2fee5aedba9c5cf5f9/dns/zone-configs/k8s.io._0_base.yaml#L230-L233

anyway, sorry

thockin:
Apparently 2600:1901:0:26f3::1234 is a string but 2600:1901:0:26f3:: is a
key (YAML "mapping value") because of the trailing `:`

PR #1383, opened 2020-11-02
Add Initial Promoter Manifest For scheduler-plugins

seanmalloy:


seanmalloy:
@dims and @bartsmykla would one of you be able to take a look?

PR #1393, opened 2020-11-03
Grant GCR write privileges to prow-build Service Account

ameukam:
Grant prow-build Service Account privileges to write to GCR registries
for the staging projects.

### DIFF
--- a/groups/sig-release/groups.yaml
+++ b/groups/sig-release/groups.yaml
@@ -115,6 +115,7 @@ groups:
       - k8s-infra-prow-oncall@kubernetes.io
       - k8s-infra-release-editors@kubernetes.io
       - ameukam@gmail.com
+      - prow-build@k8s-infra-prow-build.iam.gserviceaccount.com
 
   - email-id: k8s-infra-staging-cip-test@kubernetes.io
     name: k8s-infra-staging-cip-test


ameukam:
Potential Fix for build failure of `ci-kubernetes-build-canary` : https://github.com/kubernetes/kubernetes/issues/95936.

/assign @spiffxp

infra/gcp/ensure-staging-storage.sh
```diff
@@ -173,6 +173,11 @@ for REPO; do
         color 6 "Empowering ${WRITERS} to GCR"
         empower_group_to_write_gcr "${WRITERS}" "${PROJECT}"
 
+        # Enable prow-build Service Account to write to GCR
+        empower_svcacct_to_write_gcr \
+            $(svc_acct_email "k8s-infra-prow-build" "prow-build") \
+            "${PROJECT}"
+
```
justaugustus, line 180: Do we want K8s Infra prow have access to write to _any_ staging GCR?

We might want a special case, explicitly for `gcr.io/k8s-staging-ci-images`, which can be added under this block:

https://github.com/kubernetes/k8s.io/blob/07a3a07626d8733550866abe799cf58241045a18/infra/gcp/ensure-staging-storage.sh#L230-L254

justaugustus (changes requested):
@ameukam -- An idea about special-casing this.

infra/gcp/ensure-staging-storage.sh
ameukam: Fixed.

spiffxp (commented):
/hold
So thank you for submitting this, but I'm trying to figure out if this is really how we want to proceed vs. other alternatives

The k8s-infra-prow-build cluster runs untrusted jobs (PRs), and this would allow _any_ of those to write to gcr.io/k8s-staging-ci-images.

At the same time, we have all these @kubernetes.io groups that try to restrict who can write to these repos.  I'm not sure we want to completely bypass that security mechanism.

spiffxp:
Looks like I reviewed a stale version of this PR.

Two thoughts:
- I think it would be clearer if we added the service account to k8s-staging-ci-images@kubernetes.io
- I think a more sustainable solution would be to setup service accounts for staging repos, and enforce which jobs are allowed to use which service accounts (I won't block this PR for that, if we're willing to allow any job to write to k8s-staging-ci-images temporarily)

justaugustus:
> if we're willing to allow any job to write to k8s-staging-ci-images temporarily

I'm fine with this in the interim.
AFAIK, there's nothing preventing any CI job from writing to gcr.io/kubernetes-ci-images today.

spiffxp (approved):
/approve
/lgtm
/hold cancel

spiffxp:
For reasons I no longer have the visibility to troubleshoot, it appears as though service account membership in a group isn't being picked up.

https://gsuiteupdates.googleblog.com/2020/08/service-accounts-in-google-groups-beta.html leads me to believe it _should_ be, but our gsuite may not be in beta

PR #1400, opened 2020-11-06
Add kubectl and kustomize netlify sites for cli-experimental

pwittrock:


dns/zone-configs/k8s.io._0_base.yaml
```diff
@@ -299,3 +299,13 @@ krew.sigs:
 secrets-store-csi-driver.sigs:
 - type: CNAME
   value: kubernetes-sigs-secrets-store-csi-driver.netlify.app.
+# https://github.com/kubernetes-sigs/cli-experimental docs (@pwittrock @seans3 @eddiezane @soltysh)
+# kustomize and kubectl share the same docs site
+kustomize.sigs:
+  type: CNAME
```
munnerz, line 305: I think this is meant to have a - before it as it is a list of structures (like the entry above)

munnerz: Ah, I see your discussion on slack now and that other entries do not 🙈

pwittrock: :)

munnerz:
@pwittrock seems like the netlify site needs to be properly configured to fetch a certificate that's valid for these new DNS names 😊 https://kubectl.sigs.k8s.io

Also, does one of these two domain names redirect to the other? Else we'll end out with duplicate google search entries/pagerank profiles

pwittrock:
@munnerz looks like it is fixed now

PR #1405, opened 2020-11-10
Promote CoreDNS v1.7.1 and v1.8.0

stp-ip:


rajansandeep (approved):
/lgtm
thank you @stp-ip

rajansandeep:
Is there a time frame as to when the images are promoted?
Seems like the images aren't available in k8s.gcr.io yet

stp-ip:
Should be quite close after merge. Will check what's up.

stp-ip:
Jobs ran without error and checked the repo as well: https://console.cloud.google.com/gcr/images/k8s-artifacts-prod/EU/coredns/coredns?gcrImageListsize=30

The last version v1.7.0 was still using the root due to it using the legacy path as well. 

Should be available under k8s.gcr.io/coredns/coredns:v1.8.0

On mobile so can't check myself.

chrisohaver:
@stp-ip thanks!  I checked and it works.

PR #1429, opened 2020-11-16
Migrate deployments to apps/v1

ameukam:
The `aaa` cluster is in version v1.15 (`v1.15.12-gke.20`) which is the last
version serving deployment resource under extensions/v1beta1.
See: https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG/CHANGELOG-1.16.md#deprecations-and-removals. 

~Adding a new deployment resource under apps/v1.~
EDIT: Migrate all the _Deployments_ to apps/v1.

### DIFF
--- a/gcsweb.k8s.io/deployment.yaml
+++ b/gcsweb.k8s.io/deployment.yaml
@@ -1,4 +1,4 @@
-apiVersion: extensions/v1beta1
+apiVersion: apps/v1
 kind: Deployment
 metadata:
   name: gcsweb
@@ -6,7 +6,9 @@ metadata:
     app: gcsweb
 spec:
   replicas: 2
-  # selector defaults from template labels
+  selector:
+    matchLabels:
+      app: gcsweb
   strategy:
     rollingUpdate:
       maxSurge: 1
--- a/k8s.io/deployment.yaml
+++ b/k8s.io/deployment.yaml
@@ -1,4 +1,4 @@
-apiVersion: extensions/v1beta1
+apiVersion: apps/v1
 kind: Deployment
 metadata:
   name: k8s-io
@@ -7,7 +7,10 @@ metadata:
     version: v1
 spec:
   replicas: 2
-  # selector defaults to template's labels
+  selector:
+    matchLabels:
+      app: k8s-io
+      version: v1
   strategy:
     type: RollingUpdate
     rollingUpdate:
--- a/node-perf-dash.k8s.io/deployment.yaml
+++ b/node-perf-dash.k8s.io/deployment.yaml
@@ -1,4 +1,4 @@
-apiVersion: extensions/v1beta1
+apiVersion: apps/v1
 kind: Deployment
 metadata:
   name: node-perf-dash
@@ -8,6 +8,9 @@ metadata:
     version: "0.3"
 spec:
   replicas: 1
+  selector:
+    matchLabels:
+      app: node-perf-dash
   template:
     metadata:
       labels:


ameukam:
The plan would be :
- deploy this deployment in the canary namespace `k8s-io-canary`
- ensure the traffic is served to the pods created by this new deployment
- run the canary test. 

If the test is successful, 
- deploy in prod
- delete the deployments under extensions/v1beta1.

/assign @thockin @spiffxp

thockin:
We don't need a new resource - it will be readable as v1 automatically.  We should only have to update the `apiVersion` in our YAML files.

```
thockin@cloudshell:~ (kubernetes-public)$ k --cluster=prod-aaa -n k8s-io-prod get deployments.v1.apps k8s-io -o yaml | head -2
apiVersion: apps/v1
kind: Deployment
```

ameukam:
@thockin TIL!
I update the PR with all deployments under extensions/v1beta1.

gcsweb.k8s.io/deployment.yaml
```diff
@@ -7,6 +7,9 @@ metadata:
 spec:
   replicas: 2
   # selector defaults from template labels
```
thockin, line 9: remove this comment?

k8s.io/deployment.yaml
```diff
@@ -8,6 +8,10 @@ metadata:
 spec:
   replicas: 2
   # selector defaults to template's labels
```
thockin, line 10: ditto

k8s.io/deployment.yaml
```diff
@@ -31,7 +35,7 @@ spec:
           # Map all keys to files.
       containers:
       - name: nginx
-        image: nginx:1.10-alpine  # = :stable-alpine as of 8/23/2016
+        image: nginx:1.19-alpine  # = :stable-alpine as of 11/17/2020
```
thockin, line 38: is this already running or is it a change?

k8s.io/deployment.yaml
```diff
@@ -57,8 +61,12 @@ spec:
           failureThreshold: 2
         lifecycle:
           preStop:
-            exec: # nginx likes to terminate ASAP
-              command: [ "/bin/sleep", "30" ]
+            exec:
+              command:
```
thockin, line 65: is this already running or is it a change?

k8s.io/deployment.yaml
ameukam: A change not tested in the canary namespace.

k8s.io/deployment.yaml
ameukam: A change not tested in the canary namespace.

gcsweb.k8s.io/deployment.yaml
ameukam: Fixed

k8s.io/deployment.yaml
ameukam: Fixed

k8s.io/deployment.yaml
ameukam: Change reverted. I'll introduce the change in a followup PR.

spiffxp:
I would suggest urgency on this has increased.  The control plane for aaa was upgraded to 1.16 overnight
```
$ gcloud container operations list --project=kubernetes-public
NAME                              TYPE            LOCATION     TARGET  STATUS_MESSAGE  STATUS  START_TIME                      END_TIME
operation-1607598183239-f6b55616  UPGRADE_MASTER  us-central1  aaa                     DONE    2020-12-10T11:03:03.239556383Z  2020-12-10T11:28:16.789338532
```
```
$ gcloud container clusters list --project=kubernetes-public
NAME  LOCATION     MASTER_VERSION   MASTER_IP     MACHINE_TYPE   NODE_VERSION      NUM_NODES  STATUS
aaa   us-central1  1.16.13-gke.404  34.68.58.204  n1-standard-2  1.15.12-gke.20 *  6          RUNNING
```

node-perf-dash.k8s.io/deployment.yaml
```diff
@@ -8,6 +8,10 @@ metadata:
     version: "0.3"
 spec:
   replicas: 1
+  selector:
+    matchLabels:
+      app: node-perf-dash
+      version: "0.3"
```
thockin, line 14: I don't think you want the version label - it won't match the labels in template below.

ameukam: Fixed.

thockin (commented):
Thanks!

/lgtm
/approve

spiffxp:
IIRC merging these won't auto-deploy them, can you follow up here to confirm @ameukam ?

ameukam:
@spiffxp Effectively, there is no auto-deployment mechanism set for all workloads running on `aaa`. Historically, we have always left this responsibility to the owners of the workloads. 
I suggest that one of cluster-admins manually apply those changes.

PR #1442, opened 2020-11-21
Promote v0.3.1 filestore image

saikat-royc:


saikat-royc:
Following the steps listed in https://github.com/kubernetes/k8s.io/blob/master/k8s.gcr.io/README.md

Issue #1466, opened 2020-12-07
Migrate away from google.com gcp project kubernetes-tools


spiffxp:
Part of umbrella issue to migrate away from google.com gcp projects: https://github.com/kubernetes/k8s.io/issues/1469

We used to host slack-infra on a cluster running in kubernetes-tools.  The cluster has since been turned down: https://github.com/kubernetes/k8s.io/issues/790

However there are still references to images in this project: https://cs.k8s.io/?q=kubernetes-tools&i=nope&files=&repos=

I'm pretty sure we have a staging repo, and can replace the remaining references

spiffxp:
Here's what we have in the staging repository
```
$ gcloud container images list --repository gcr.io/k8s-staging-slack-infra
NAME
gcr.io/k8s-staging-slack-infra/slack-event-log
gcr.io/k8s-staging-slack-infra/slack-moderator
gcr.io/k8s-staging-slack-infra/slack-post-message
gcr.io/k8s-staging-slack-infra/slack-report-message
gcr.io/k8s-staging-slack-infra/slack-welcomer
gcr.io/k8s-staging-slack-infra/tempelis
```

spiffxp:
We are missing slackin, which is referenced here: https://github.com/kubernetes/k8s.io/blob/54e8506f8404206eb15d619ac55324332a47985c/slack-infra/resources/slackin/deployment.yaml#L19

ameukam:
@spiffxp Since the slackin image used is built from a fork (possibly https://github.com/Katharine/slackin), we could manually push the image to `gcr.io/k8s-staging-slack-infra/`.

spiffxp:
Did the manual push as a workaround for now.
```
$ docker pull gcr.io/kubernetes-tools/slackin-kubernetes@sha256:d1a9b02239e690d5cbd78c76475a112aedf279dbd8ef401de1b8394f817a23b3
$ docker tag gcr.io/kubernetes-tools/slackin-kubernetes@sha256:d1a9b02239e690d5cbd78c76475a112aedf279dbd8ef401de1b8394f817a23b3 gcr.io/k8s-staging-slack-infra/slackin-kubernetes
$ docker push gcr.io/k8s-staging-slack-infra/slackin-kubernetes
# ...
latest: digest: sha256:d1a9b02239e690d5cbd78c76475a112aedf279dbd8ef401de1b8394f817a23b3 size: 1583
[...]

PR #1478, opened 2020-12-10
Switch slackin image

ameukam:
Ref: https://github.com/kubernetes/k8s.io/issues/1466

Switch to slackin image pushed in gcr.io/k8s-staging-slack-infra.


spiffxp (approved):
/approve
/lgtm

spiffxp:
Approving to move us away from the google-hosted image, but IMO it would be wise to promote the image to k8s.gcr.io.  This isn't something that changes often.

PR #1482, opened 2020-12-12
Allow bannai@google.com push images to network staging repository

varunmar:
Add bannai@google.com to k8s-infra-staging-networking@kubernetes.io so that they can push to gcr.io/k8s-staging-networking for pre-production ingress-gce images


MrHohn (approved):
/lgtm
/approve

PR #1494, opened 2020-12-18
Promote Images For external-dns v0.7.5 Release

seanmalloy:
This is the first release that supports multi-arch container images.


seanmalloy:
For reference here is the image that is being promoted from the staging registry.

https://console.cloud.google.com/gcr/images/k8s-staging-external-dns/GLOBAL/external-dns@sha256:d57bc04c95513dcfff3531e65fac15346ede7de2542d3479af97571365c0a8fa/details?tab=info

Raffo:
CI failed on the tag (see https://prow.k8s.io/view/gs/kubernetes-jenkins/logs/post-external-dns-push-images/1337774065301065728 ), I asked on slack, but it looks like there is no way to retrigger it. This means that I will have to delete the tag and re-create it to re-run CI. Unfortunately given the lack of access to the google CI/cluster and the slowness of responses on slack, the whole process is taking a while, but I can confirm that I am on it. And slowness is to be expected given the upcoming holidays, so no blame 😉 

Please close this PR if possible to avoid confusion.

Issue #1535, opened 2021-01-11
unknown: Service 'containerregistry.googleapis.com' is not enabled for consumer 'project:k8s-infra-e2e-boskos-scale-Xcd ..'.


mm4tt:
This make all presubmit kubemark jobs (which are now optional, that's why we haven't caught it earlier) to fail

spiffxp:
I opened https://github.com/kubernetes/k8s.io/pull/1536 but it will require a human to run

Scripting something to run just for scale projects to unblock you for now.

spiffxp:
```bash
for i in $(seq 1 30); do
  p=$(printf "k8s-infra-e2e-boskos-scale-%02i" $i);
  echo $(date) $p...;
  gcloud services enable containerregistry.googleapis.com --project=$p;
done
```
```
Thu Jan 14 21:51:12 EST 2021 k8s-infra-e2e-boskos-scale-01...
Operation "operations/acf.p2-40460865376-1066bd2a-f3b7-49eb-b12d-aa3563511143" finished successfully.
Thu Jan 14 21:51:21 EST 2021 k8s-infra-e2e-boskos-scale-02...
Operation "operations/acf.p2-660290009727-d676eda7-7fcd-474a-bc35-3c93d980a9fd" finished successfully.
Thu Jan 14 21:51:28 EST 2021 k8s-infra-e2e-boskos-scale-03...
Operation "operations/acf.p2-647280784790-ffabb77c-daa7-41ac-98d6-ebfcf4da9937" finished successfully.
# etc...
```

mm4tt:
Thanks @spiffxp! Let me check whether it works now

mm4tt:
It works - kubernetes/kubernetes#97858.

spiffxp:
/close
https://github.com/kubernetes/k8s.io/pull/1536 merged and I ran the script

PR #1536, opened 2021-01-15
Enable GCR for e2e projects

spiffxp:
This is needed specifically by the scale projects to be able to run kubemark, but I would like the "push to GCR repo in the project boskos gave to my job, with no expectation the image will live beyond the job" pattern to be possible for all e2e projects.

ref: https://github.com/kubernetes/k8s.io/issues/1535


spiffxp:
/cc @BenTheElder 
any concerns with this?

spiffxp:
Running ./infra/gcp/ensure-e2e-projects.sh

PR #1594, opened 2021-01-29
releng: Provision k8s-staging-experimental project

justaugustus:
To be used to stage experimental images, like support for new
architectures in an existing Tier 1 image.

Signed-off-by: Stephen Augustus <foo@auggie.dev>

/assign @dims @hasheddan @saschagrunert 

Part of https://github.com/kubernetes/sig-release/issues/1337.
Context (https://github.com/kubernetes/release/pull/1880#issuecomment-769074508):

> > ```
> > 1. We shouldn't add any new platforms to base images until there is clear path to at least `Tier 2` support in k/k. This would mean closing this PR and waiting until a functioning build of k/k can be demonstrated with this architecture.
> > ```
> 
> The counter to my feeling is: "How can a platform elevate their tier without supporting artifacts like this?".
> 
> > ```
> > 2. Adding support here allows for folks to more easily build k/k with that architecture to demonstrate its functionality as part of `Tier 3`. This would be accepting this PR and then having @houfangdong (or others) demonstrate ability to build k/k for this architecture, which would get them started on process to higher tiers.
> > ```
> > 
> > 
> > If we are comfortable with taking on the new platform from a maintenance perspective here, I don't think it necessarily needs to be gated on the criteria for tiers in k/k. In fact, I think our aim should be for base images like this to be as broadly useful as possible to the extent that we have bandwidth to maintain.
> 
> So my concern here would another tier impacting our ability to support our current artifacts.
> Example: @saschagrunert @cpanato saw image build errors recently with `s390x`, which cause our `amd64` pushes to fail.
> 
> I think if we can establish a playground variant of sorts, then I'd be open to this:
> 
> Example:
> 
>     * Tier 1: `debian-base:<tag>`
> 
>     * Anything else: `debian-base-experimental-arch:<tag>`
> 
> 
> Or even:
> 
>     * Tier 1: `k8s.gcr.io/build-image/debian-base:<tag>`
> 
>     * Anything else: `k8s.gcr.io/experimental-arch/debian-base:<tag>`
> 
> 
> They would build using the same Dockerfile/Makefiles, but run independently and (maybe) push to different staging projects.

### DIFF
--- a/groups/sig-release/groups.yaml
+++ b/groups/sig-release/groups.yaml
@@ -152,6 +152,18 @@ groups:
       - thockin@google.com
       - rajib.jolite@gmail.com
 
+  - email-id: k8s-infra-staging-experimental@kubernetes.io
+    name: k8s-infra-staging-experimental
+    description: |-
+      ACL for staging experimental images
+
+      This project is used to stage experimental images, for example, support
+      for new architectures.
+    settings:
+      ReconcileMembers: "true"
+    members:
+      - k8s-infra-release-editors@kubernetes.io
+
   - email-id: k8s-infra-staging-kubernetes@kubernetes.io
     name: k8s-infra-staging-kubernetes
     description: |-
--- a/infra/gcp/ensure-staging-storage.sh
+++ b/infra/gcp/ensure-staging-storage.sh
@@ -73,6 +73,7 @@ STAGING_PROJECTS=(
     etcd
     etcdadm
     examples
+    experimental
     external-dns
     git-sync
     infra-tools
@@ -107,6 +108,7 @@ STAGING_PROJECTS=(
 )
 
 RELEASE_STAGING_PROJECTS=(
+    experimental
     kubernetes
     mirror
     releng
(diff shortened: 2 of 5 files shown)


hasheddan (approved):
/lgtm

saschagrunert (approved):
/lgtm

dims:
@justaugustus error running script!

```
davanum@cloudshell:~/k8s.io$ infra/gcp/ensure-staging-storage.sh experimental
Configuring staging: experimental
  Ensuring project exists: k8s-staging-experimental
  Create in progress for [https://cloudresourcemanager.googleapis.com/v1/projects/k8s-staging-experimental].
  Waiting for [operations/cp.5129442009212321338] to finish...
  ..done.
  billingAccountName: billingAccounts/018801-93540E-22A20E
  billingEnabled: true
  name: projects/k8s-staging-experimental/billingInfo
  projectId: k8s-staging-experimental
  Empowering k8s-infra-staging-experimental@kubernetes.io as project viewers
  ERROR: Policy modification failed. For a binding with condition, run "gcloud alpha iam policies lint-condition" to identify issues in condition.
  ERROR: (gcloud.projects.add-iam-policy-binding) INVALID_ARGUMENT: Group k8s-infra-staging-experimental@kubernetes.io does not exist.
```

dims:
This is ready now!

justaugustus:
Thanks Dims!

PR #1614, opened 2021-02-05
Add wg-reliability-sippy namespace.

deads2k:
Sippy is a community tool will be used by wg-reliability to help produce
transparent metrics to assess overall stability and break down by
project area.

Additional details available in #wg-reliability and on the mailing list
at https://groups.google.com/g/kubernetes-sig-testing/c/BxnXXipU42M/m/bWyT-9gtAwAJ

@wojtek-t @stevekuznetsov

PR #1615, opened 2021-02-05
add manifests for initial deployment of sippy

deads2k:
builds on https://github.com/kubernetes/k8s.io/pull/1614

This adds manifests to deploy and run sippy.  This is similar to what I have in my private toy cluster, but I need to do a little bit of alignment and work out ingresses instead of routes.

/assign @wojtek-t 

/hold

explicit hold until I'm able to test these locally.

### DIFF
--- a/sippy/sippy/ingress.yaml
+++ b/sippy/sippy/ingress.yaml
@@ -1,0 +1,17 @@
+apiVersion: networking.k8s.io/v1beta1
+kind: Ingress
+metadata:
+  name: sippy
+  namespace: sippy
+  annotations:
+    kubernetes.io/ingress.global-static-ip-name: "sippy-ingress-prod"
+  labels:
+    app: sippy
+spec:
+  defaultBackend:
+    service:
+      name: sippy
+      port:
+        number: 8080
+  tls:
+  - secretName : sippy-k8s-io-tls
--- a/sippy/sippy/pvc.yaml
+++ b/sippy/sippy/pvc.yaml
@@ -1,0 +1,12 @@
+apiVersion: v1
+kind: PersistentVolumeClaim
+metadata:
+  name: sippy
+  namespace: sippy
+spec:
+  accessModes:
+    - ReadWriteOnce
+  resources:
+    requests:
+      storage: 100Gi
+  storageClassName: standard
--- a/sippy/OWNERS
+++ b/sippy/OWNERS
@@ -1,0 +1,9 @@
+# See the OWNERS docs at https://go.k8s.io/owners
+
+approvers:
+  - wg-reliability-leads
+reviewers:
+  - wg-reliability-leads
+
+labels:
+  - wg/reliability
--- a/sippy/README.md
+++ b/sippy/README.md
@@ -1,0 +1,17 @@
+# sippy
+
+`sippy` is a tool used to aid in transparency of stability metrics.
+
+[sippy](https://github.com/openshift/sippy) summarizes multiple test-grid dashboards with data slicing of related jobs, job runs,
+and tests.  Visit (placeholder for eventual URL) to see the kube instance. 
+
+
+## How to deploy sippy
+
+- Have [access](https://github.com/kubernetes/k8s.io/blob/master/running-in-community-clusters.md) to the GKE cluster `aaa`.
+
+- From the `sippy` directory run:
+
+```console
+kubectl apply -f sippy/
+```
--- a/sippy/sippy/certificate.yaml
+++ b/sippy/sippy/certificate.yaml
@@ -1,0 +1,15 @@
+apiVersion: cert-manager.io/v1
+kind: Certificate
+metadata:
+  name: sippy-k8s-io
+  namespace: sippy
+  annotations:
+    acme.cert-manager.io/http01-override-ingress-name: "sippy"
+    cert-manager.io/issue-temporary-certificate: "true"
+spec:
+  secretName: sippy-k8s-io-tls
+  issuerRef:
+    kind: ClusterIssuer
+    name: letsencrypt-prod
+  dnsNames:
+    - sippy.k8s.io
--- a/sippy/sippy/deployment.yaml
+++ b/sippy/sippy/deployment.yaml
@@ -1,0 +1,66 @@
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  labels:
+    app: sippy
+  name: sippy
+  namespace: sippy
+spec:
+  replicas: 1
+  selector:
+    matchLabels:
+      app: sippy
+  strategy:
+    type: Recreate
+  template:
+    metadata:
+      labels:
+        app: sippy
+    spec:
+      containers:
+        - image: quay.io/bparees/sippy:unstable
+          imagePullPolicy: Always
+          name: sippy
+          ports:
+            - containerPort: 8080
+              protocol: TCP
+          resources:
+            limits:
+              memory: 1500M
+          terminationMessagePath: /dev/termination-log
+          terminationMessagePolicy: File
+          command:
+            - /tmp/src/sippy
+          args:
+            - --server
+            - --end-day
+            - "7"
+            - --local-data
+            - /data
+            - --dashboard
+            - kube-master=sig-release-master-blocking,sig-release-master-informing=
+            - --variant=kube
+          volumeMounts:
+            - mountPath: /data
+              name: data
+        - image: quay.io/bparees/sippy:unstable
+          imagePullPolicy: Always
+          name: fetchdata
+          resources:
+            limits:
+              memory: 500M
+          terminationMessagePath: /dev/termination-log
+          terminationMessagePolicy: File
+          # This runs in a loop to collect data every hour or so.  We run it in the same pod so that both containers can mount
+          # and we don't need r/w many.
+          command:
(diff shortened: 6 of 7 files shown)


groups/groups.yaml
```diff
@@ -241,6 +241,7 @@ groups:
       - k8s-infra-rbac-triageparty-release@kubernetes.io
       - k8s-infra-rbac-slack-infra@kubernetes.io
       - k8s-infra-rbac-node-perf-dash@kubernetes.io
+      - k8s-infra-rbac-wg-reliability@kubernetes.io
```
wojtek-t, line 244: nit: This is already merged - please rebase

wg-reliability/README.md
```diff
@@ -0,0 +1,17 @@
+# wg-reliability
+
+wg-reliability has tools used to aid in transparency of stability metrics.
+
+[sippy](https://github.com/openshift/sippy) summarizes multiple test-grid dashboards with data slicing of related jobs, job runs,
+and tests.  Visit (placeholder for eventual URL) to see the kube instance.
```
wojtek-t, line 6: We should probably request public IP and DNS now (given we already have support for running it).

I'm wondering if we want "sippy.k8s.io" or something more generic?

wg-reliability/sippy/service.yaml
```diff
@@ -0,0 +1,17 @@
+apiVersion: v1
+kind: Service
```
wojtek-t, line 2: We also want Ingress and Certificate too, as e.g. here:
https://github.com/kubernetes/k8s.io/tree/master/perf-dash.k8s.io

wg-reliability/sippy/deployment.yaml
```diff
@@ -0,0 +1,62 @@
...
+          command:
+            - /tmp/src/sippy
+          args:
+            - --server
+            - --end-day
+            - "7"
+            - --local-data
+            - /data
+            - --dashboard
+            - kube-master=sig-release-master-blocking,sig-release-master-informing=
```
wojtek-t, line 39: Do we want both? Or maybe we want to start just with blocking?

Or actually, I guess it's probably moot anyway, because we probably want to create our own dashboard for this purpose...

wg-reliability/sippy/deployment.yaml
```diff
@@ -0,0 +1,62 @@
...
+        - image: quay.io/bparees/sippy:kube-latest
+          imagePullPolicy: Always
+          name: fetchdata
+          resources:
+            limits:
+              memory: 500M
+          terminationMessagePath: /dev/termination-log
+          terminationMessagePolicy: File
+          command:
+            - /tmp/src/scripts/fetchdata-kube.sh
```
wojtek-t, line 53: What is this? I didn't see that anywhere.
I think you can simply run:
./sippy --fetch-data /usr/local/google/home/wojtekt/sippy_data --dashboard=kube-master=sig-release-master-blocking=

Or I guess, you might wanted this to run periodically already.
Do we want a container? Or do we want a CronJob that will be doing this?

wg-reliability/sippy/deployment.yaml
```diff
@@ -0,0 +1,62 @@
...
+          volumeMounts:
+            - mountPath: /data
+              name: data
+      restartPolicy: Always
+      terminationGracePeriodSeconds: 30
```
wojtek-t, line 58: Isn't 30 default?

wg-reliability/sippy/pvc.yaml
```diff
@@ -0,0 +1,11 @@
...
+kind: PersistentVolumeClaim
+metadata:
+  name: sippy
+  namespace: wg-reliability-sippy
+spec:
+  accessModes:
+    - ReadWriteOnce
+  resources:
+    requests:
+      storage: 900M
```
wojtek-t, line 11: I guess we want to set some StorageClass, right?

wg-reliability/sippy/deployment.yaml
wojtek-t: Ahh - I guess with CronJob, we may have a problem with mounting for different pods...

So we probably want this to be a container within a pod.

spiffxp (changes requested):
I would rather see these manifests in a folder called `sippy` or `sippy.k8s.io`, in a `sippy` namespace to match.

If we're going down the path of assigning group ownership to apps deployed in cluster (we should! accountability is great!) I would rather see a `sig: foo` label, given that wg's don't own code.

I would suggest `sig: architecture` or `sig: release` to start... though long term it should be `sig: testing` when we have bandwidth.

sippy/sippy/pvc.yaml
```diff
@@ -0,0 +1,12 @@
...
+metadata:
+  name: sippy
+  namespace: sippy
+spec:
+  accessModes:
+    - ReadWriteOnce
+  resources:
+    requests:
+      storage: 900M
+  # storageClassName: standard
```
wojtek-t, line 12: @spiffxp - what storage class we have in the community cluster?

wojtek-t (commented):
This looks fine to me (modulo missing storage class).

sippy/sippy/pvc.yaml
ameukam: @wojtek-t we have 2 storage classes in the cluster : 

```shell
kubectl get sc
NAME                 PROVISIONER            AGE
ssd                  kubernetes.io/gce-pd   334d
standard (default)   kubernetes.io/gce-pd   551d
```

wojtek-t: Thanks!

@deads2k - standard should be good enough for us;
 Though I would visibly increase the size, say to 10GB or sth.

groups/groups.yaml
```diff
@@ -595,8 +595,8 @@ groups:
       - spiffxp@gmail.com
       - thockin@google.com
 
-  - email-id: k8s-infra-rbac-wg-reliability@kubernetes.io
-    name: k8s-infra-rbac-wg-reliability
+  - email-id: k8s-infra-rbac-sippy@kubernetes.io
```
spiffxp, line 598: You'll probably need to rebase, I did this in https://github.com/kubernetes/k8s.io/pull/1844

spiffxp (approved):
/approve
/lgtm
/hold
Remove /hold when ready to deploy

Thanks for changing the namespace, I ended up setting it up via https://github.com/kubernetes/k8s.io/pull/1844 so you'll likely need to rebase

sippy/sippy/ingress.yaml
```diff
@@ -0,0 +1,17 @@
+apiVersion: networking.k8s.io/v1
```
spiffxp, line 1: I'm not sure if aaa is at a version that supports this

spiffxp (approved):
/lgtm

sippy/sippy/pvc.yaml
wojtek-t: @deads2k - friendly ping

sippy/sippy/ingress.yaml
wojtek-t: good catch - the cluster is still in 1.16 (will be upgraded to 1.17 soon, but ingress/v1 is only in 1.19), so we should use v1beta1 stil...

deads2k:
/hold

We will merge on monday when everyone is back in the office and ready for potential excitement

deads2k:
/hold cancel

here we go!

sbueringer:
@deads2k for even more excitement. Is there a specific issue where I can follow the implementation?

sbueringer:
Ah sorry that was misleading. I meant more like the deployment of sippy for the kubernetes repos. I saw the DNS name `sippy.k8s.io`, but it's not available yet. So I assumed there might be an issue somewhere (which I can subscribe to) to stay up-to-date about the deployment implementation.

deads2k:
> Ah sorry that was misleading. I meant more like the deployment of sippy for the kubernetes repos. I saw the DNS name `sippy.k8s.io`, but it's not available yet. So I assumed there might be an issue somewhere (which I can subscribe to) to stay up-to-date about the deployment implementation.

As it happens, I was just starting to look now and I haven't worked out where my resources went.

```
deads@cloudshell:~$ kubectl --context=prod-aaa -nsippy get pod
No resources found in sippy namespace.
deads@cloudshell:~$ kubectl --context=prod-aaa -nsippy get deployment
No resources found in sippy namespace.
deads@cloudshell:~$ kubectl --context=prod-aaa -nsippy get all
Error from server (Forbidden): replicationcontrollers is forbidden: User "deads@redhat.com" cannot list resource "replicationcontrollers" in API group "" in the namespace "sippy": r
equires one of ["container.replicationControllers.list"] permission(s).
Error from server (Forbidden): daemonsets.apps is forbidden: User "deads@redhat.com" cannot list resource "daemonsets" in API group "apps" in the namespace "sippy": requires one of 
["container.daemonSets.list"] permission(s).
deads@cloudshell:~$ kubectl --context=prod-aaa -nsippy get ingress
No resources found in sippy namespace.
```

sbueringer:
Ah got it :). I'll just follow the update here then.

ameukam:
@deads2k 👋🏾 , resources deployment is still a manual action in `aaa`.

spiffxp:

Issue #1638, opened 2021-02-11
Remove unused windows-img-promoter-cert secrets from k8s-infra-prow-build trusted


spiffxp:
The secrets were added via https://github.com/kubernetes/k8s.io/pull/857 in May 2020

https://github.com/kubernetes/kubernetes/pull/93889 switched to using buildx and dropped dependence on remote windows docker nodes October 2020

I'd like to remove the secrets from secret manager and from the cluster
```
$ gcloud secrets list --project=k8s-infra-prow-build-trusted --filter="labels.secret-group=windows-img-promoter-cert"
NAME                            CREATED              REPLICATION_POLICY  LOCATIONS
windows-remote-docker_ca-pem    2020-05-27T23:37:49  automatic           -
windows-remote-docker_cert-pem  2020-05-27T23:38:13  automatic           -
windows-remote-docker_key-pem   2020-05-27T23:38:32  automatic           -
$ k --context=gke_k8s-infra-prow-build-trusted_us-central1_prow-build-trusted get secrets -n test-pods | grep windows
windows-img-promoter-cert                  Opaque                                3      280d
```

Then remove related special case bash in infra/gcp

/wg k8s-infra
/sig testing
/sig windows
/assign @claudiubelu 
to confirm we can remove these secrets
/assign @spiffxp 
I can take care of this
/priority important-soon
/milestone v1.21

claudiubelu:
Indeed, that is correct. The certificates were being used for the Remote Docker Windows nodes that we were using to build Windows images, but we've switched to docker buildx in the meanwhile. We can remove it, since I'm sure no one else are using it anymore (might be wrong, but I think the certificates also expired in December).

/lgtm

spiffxp:
Removed from secretmanager and cluster
```bash
p=k8s-infra-prow-build-trusted
for s in $(gcloud secrets list --project=$p --filter="labels.secret-group=windows-img-promoter-cert" --format="value(name)"); do 
  gcloud secrets delete --project=$p $s
done
kubectl delete secret -n test-pods windows-img-promoter-cert
```

spiffxp:
Opened https://github.com/kubernetes/k8s.io/pull/1639 to remove script, will auto-close this once merged

PR #1639, opened 2021-02-11
Remove windos-img-promoter-cert secret

spiffxp:
The secret was manually added, and has been manually removed. This is
removing the supporting infra/gcp scripting.

Fixes https://github.com/kubernetes/k8s.io/issues/1638


spiffxp:
/assign @claudiubelu 
For lgtm

PR #1650, opened 2021-02-16
Audit Update for kubernetes-public

hh:
This is not a full audit update. It's missing secrets and is only for Kubernetes-public.

More to come as I work through them in manageable batches.

This is the first PR authored by @cncf-ci bot, so we'll need to figure out how to get the CLA signed... by the CNCF?

PR #1534, opened 2021-01-15
results of running audit script as of 2021-01-13

spiffxp:
It's been way too long since this was last run and the results checked in.  Last commit to `audit/` is dated 2020-05-06.  So this is ~7mo of changes

I tried to group changes into logical commits, but I won't be able to exhaustively tie each commit back to the issues/PRs that spawned it.

This is WIP because I want to either resolve or open up followup issues for commits that have `QQ` or `TODO` in their messages.

I really, really, really want us to get this running and automatically PR'ed by prow (ref: https://github.com/kubernetes/k8s.io/issues/244).  Current state of affairs is broken.

PR #1651, opened 2021-02-16
Audit Update for k8s-infra-* (non-boskos)

hh:
Followup to #1650


spiffxp:
/hold
I'd like to first merge the initial audit PR I opened since it has much more granular comments https://github.com/kubernetes/k8s.io/pull/1534#issuecomment-777248548

hh:
Now that we have current merged status, I'm not sure it's worth it to break these up.

PR #1652, opened 2021-02-16
Audit Update for k8s-infra-boskos 64/160

hh:
There are a 160 boskos projects... so I'm going to do them in batches... likely in 3 sets.

It's already just over 1000+ files changed. I'll stop here for feedback before processing any more.

IF we want to go smaller batch, I can try and break them up. It's pretty much a manual run with a limit on the projects it iterates over, then cleaning up deletions for a PR.


spiffxp:
/hold
I'd like to first merge the initial audit PR I opened since it has much more granular comments https://github.com/kubernetes/k8s.io/pull/1534#issuecomment-777248548

hh:
Since we've merged all current changes, let's leave this until a normal PR submission from @cncf-ci

PR #1668, opened 2021-02-17
Adding gateway-api.sigs.k8s.io for Gateway API

robscott:
Closely related to https://github.com/kubernetes/org/issues/2506

### DIFF
--- a/dns/zone-configs/k8s.io._0_base.yaml
+++ b/dns/zone-configs/k8s.io._0_base.yaml
@@ -313,3 +313,7 @@ kustomize.sigs:
 kubectl.sigs:
   type: CNAME
   value: cli-experimental.netlify.app.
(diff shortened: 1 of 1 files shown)


dns/zone-configs/k8s.io._0_base.yaml
```diff
@@ -313,3 +313,8 @@ kustomize.sigs:
 kubectl.sigs:
   type: CNAME
   value: cli-experimental.netlify.app.
+# https://github.com/kubernetes-sigs/gateway-api docs (@bowei @hbagdi @danehans @jpeach @robscott)
+kustomize.sigs:
```
hbagdi, line 317:
```suggestion
gateway-api.sigs:
```

robscott: 🤦 woops, thanks for catching that!

thockin (commented):
Thanks!

/lgtm
/approve

PR #1680, opened 2021-02-18
Enable nodelocal dnscache on prow build clusters

chaodaiG:

Issue #1541, opened 2021-01-15
Update k8s-infra prow build clusters from v1.16 to release channels


spiffxp:
Upgrading cluster versions is currently done at a time of our choosing, but we've let our clusters fall out of supported versions.  Using GKE release channels means cluster version will get upgraded as versions fall out of support.  

For more info on channels/schedule, see https://cloud.google.com/kubernetes-engine/docs/concepts/release-channels#channels

We're now on v1.16.  If we move to release channels, we may bump versions again
- stable: 1.16
- regular: 1.17
- rapid: 1.18

Due to lack of skip-level upgrades, if we want to move to rapid, we'll have to move to regular first.

Hopefully terraform supports doing this in-place.

ameukam:
#1402 give the possibility to change the release channel in the terraform gke module `gke-cluster`. Once the change is applied, Terraform will register the cluster to the release channel defined.

spiffxp:
https://github.com/kubernetes/k8s.io/pull/1402 - added release channel support to modules
https://github.com/kubernetes/k8s.io/pull/1631 - enroll clusters in stable channel

Just applied the changes https://github.com/kubernetes/k8s.io/pull/1631#issuecomment-777219336

spiffxp:
So it looks like we'll be upgrading to 1.17.15-gke.800 sometime over the next few days https://cloud.google.com/kubernetes-engine/docs/release-notes#february_9_2021_2021-r5

spiffxp:
We still remain at v1.16
```
$ for s in build build-trusted; do gcloud container clusters list --project=k8s-infra-prow-$s; done
NAME        LOCATION     MASTER_VERSION    MASTER_IP      MACHINE_TYPE    NODE_VERSION      NUM_NODES  STATUS
prow-build  us-central1  1.16.15-gke.6000  34.69.231.159  n1-standard-32  1.16.15-gke.6000  106        RUNNING
NAME                LOCATION     MASTER_VERSION    MASTER_IP    MACHINE_TYPE   NODE_VERSION      NUM_NODES  STATUS
prow-build-trusted  us-central1  1.16.15-gke.6000  34.66.53.94  n1-standard-8  1.16.15-gke.6000  3          RUNNING
```

spiffxp:
Going to force an upgrade to v1.17, we've added node local dns, and that requires node recreation to take effect.  Since upgrading to v1.17 requires the same, let's batch the changes before updating node pools.

[...]

PR #1686, opened 2021-02-19
Enable dns cache for k8s-infra-prow-build, cleanup gke-cluster module

spiffxp:
Followup to https://github.com/kubernetes/k8s.io/pull/1680 which added the dns cache option, and enabled for k8s-infra-prow-build-trusted

I did some cleanup while I was here, see commits for details

### DIFF
--- a/infra/gcp/clusters/modules/gke-cluster/main.tf
+++ b/infra/gcp/clusters/modules/gke-cluster/main.tf
@@ -178,7 +178,7 @@ resource "google_container_cluster" "prod_cluster" {
       disabled = false
     }
     dns_cache_config {
-      enabled = var.enable_node_local_dns_cache
+      enabled = var.dns_cache_enabled
     }
   }
 
@@ -279,6 +279,9 @@ resource "google_container_cluster" "test_cluster" {
     }
     network_policy_config {
       disabled = false
+    }
+    dns_cache_config {
+      enabled = var.dns_cache_enabled
     }
   }
 
--- a/infra/gcp/clusters/modules/gke-cluster/main.tf
+++ b/infra/gcp/clusters/modules/gke-cluster/main.tf
@@ -186,11 +186,6 @@ resource "google_container_cluster" "prod_cluster" {
     channel = var.release_channel
   }
 
-  // Enable PodSecurityPolicy enforcement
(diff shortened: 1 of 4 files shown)


spiffxp:
/cc @chaodaiG 
you opened previous pr
/cc @michelle192837 
test-infra-oncall

chaodaiG (approved):
LGTM

/hold
Do we also need to specify minimum terraform version so that the changes from this PR always work?

spiffxp:
/hold cancel

> Do we also need to specify minimum terraform version so that the changes from this PR always work

We use version constraints in the individual clusters, e.g. https://github.com/kubernetes/k8s.io/blob/e5f748a117d15a7c3716089a43faa82948e1121b/infra/gcp/clusters/projects/k8s-infra-prow-build/prow-build/00-provider.tf#L8-L20

But, you're right, there are definite gaps:
- we don't specify constraints within our terraform modules, since they're intended to be internal-only I have been lazy about this (ref: https://www.terraform.io/docs/language/modules/develop/providers.html#provider-version-constraints-in-modules)
- we're not doing any kind of presubmit testing of our terraform, I glanced at starting with `terraform validate` but won't have time today (will try to open an issue)

spiffxp:
Ran `terraform apply`, timestamps for updates to the control plane
```
$ gcloud container operations list --project=k8s-infra-prow-build --filter="startTime>='2021-02-19'" --format="table(startTime,endTime,operationType,name,status)"
START_TIME                      END_TIME                        TYPE            NAME                              STATUS
2021-02-19T17:29:29.862954192Z  2021-02-19T17:29:30.046259487Z  UPDATE_CLUSTER  operation-1613755769862-3158504f  DONE
2021-02-19T17:29:40.181746963Z  2021-02-19T17:57:21.542807273Z  UPDATE_CLUSTER  operation-1613755780181-48a49af1  DONE
```

Since nodes need to be recreated for this to take effect, I would rather batch this up with upgrading the cluster to v1.17

spiffxp:
https://github.com/kubernetes/k8s.io/issues/1541#issuecomment-782356369 - upgrading node pool now, this will take some time to complete, will verify whether addon has deployed afterward

PR #1728, opened 2021-02-26
Add Rajula to sig-contributor-experience group as manager

rajula96reddy:
I am part of the Upstream marketing team and need these permissions to send out comms related to contributor survey for this year

CC: @mrbobbytables


spiffxp (approved):
/approve
/lgtm

PR #1729, opened 2021-02-26
prow-staging: Fix spec of the namespaces

ameukam:


spiffxp (approved):
And this is why we need a presubmit that validates these as valid k8s manifests

(not a knock on you, I missed it too)

PR #1732, opened 2021-02-27
Add script validation for Terraform

ameukam:
Download terraform and validate Terraform configuration for
a given path assuming terraform files are presen in this path.

### DIFF
--- a/hack/verify-terraform.sh
+++ b/hack/verify-terraform.sh
@@ -1,0 +1,65 @@
+#!/usr/bin/env bash
+# Copyright 2021 The Kubernetes Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+set -o errexit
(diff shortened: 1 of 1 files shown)


ameukam:
The presubmit job will have a command like: 
```console
hack/verify-terraform.sh infra/gcp/clusters/projects/k8s-infra-prow-build/prow-build
```

ameukam:
/hold
/assign @spiffxp 
cc @BenTheElder @thockin

hack/verify-terraform.sh
```diff
@@ -0,0 +1,57 @@
...
+  echo >&2 "Usage: $0 <CLUSTER_NAME>"
+  exit 1
+}
+
+if [ $# != 1 ]; then
+    usage
+    exit 1
+fi
+
+TERRAFORM_ZIP="terraform_${TERRAFORM_VERSION}_linux_amd64.zip"
```
spiffxp, line 34: I'd rather this work on my laptop too.  

I think a version check and erroring out if it doesn't exist/match would be a good start.  Let the developer work out how to get the correct version, and assume when this runs in a container it will have the correct version already installed

Handy for managing locally, maybe useful to allow CI to auto-install / switch:
- [`tfenv`](https://github.com/tfutils/tfenv) + `.terraform-version` files fit a pattern I know , but it's lots of bash
- [`tfswitch`](https://github.com/warrensbox/terraform-switcher) uses `version.tf` files which could be another option

ameukam: I'll go for `tfswitch` since it supports `version.tf`.

nikhita:
For the record - `verify-terraform.sh` was excluded in https://github.com/kubernetes/k8s.io/pull/1770 so `pull-k8sio-verify` does not run this script.

spiffxp (approved):
/approve
/lgtm

Issue #1659, opened 2021-02-16
audit followup: organization resources should be managed by script


thockin:
For example, there's a Role "prow.viewer" that is not set up anywhere.

justaugustus:
@thockin -- I'm working on infra bring up for another project, which included some crude copy/pasting of what's in k/k8s.io.

As part of that, I have a WIP refactor of the `groups/**` logic (working title: `ggreconcile`): https://github.com/justaugustus/ggreconcile/pull/1

Want me to see how far I can take that?

thockin:
Always interested in refactor, though groups is a bit different from the org-roles?

justaugustus:
Yep, yep, but I did have to end up copying in snippets of the infra scripts, so replacing some portion of them is on my list too for that other project. 🙃 

Longer term --> https://twitter.com/stephenaugustus/status/1361686269823225859

spiffxp:
@thockin that role is created here: https://github.com/kubernetes/k8s.io/blob/main/infra/gcp/prow/ensure-e2e-projects.sh#L35-L40

I was going to shuffle things around so k8s-infra-prow-oncall@ had /approve rights to scripts they had permissions to run, I can undo if you'd rather leave everything in a flat namespace until we sort out how to organize

spiffxp:
https://github.com/kubernetes/k8s.io/issues/1656 - help wanted, script creation of the custom role definition from primitive roles

spiffxp:
/assign
https://github.com/kubernetes/k8s.io/pull/1679 is a start

spiffxp:
There are individuals assigned roles at the org level that I'm less inclined to script.

The TODO's I dropped in that PR are what I consider sufficient to close this out

spiffxp:
/retitle audit followup: organization resources should be managed by script

spiffxp:
https://github.com/kubernetes/k8s.io/pull/1726 addresses the majority of this, one more PR for cleanup after that merges

spiffxp:
/reopen
not quite done

spiffxp:
https://github.com/kubernetes/k8s.io/pull/1859 contains changes to ensure-main-project.sh that remove the project-level ServiceAccountLister role in favor of the org-level ServiceAccountLister role.  Change was deployed per https://github.com/kubernetes/k8s.io/pull/1859#issuecomment-814369851

spiffxp:
/close
At this point `infra/gcp/ensure-organization.sh` does the following:
[...]

PR #1726, opened 2021-02-26
Manage organization roles

spiffxp:
This addresses most of https://github.com/kubernetes/k8s.io/issues/1659

To do so I added a script to generate roles out of other roles/permissions (https://github.com/kubernetes/k8s.io/issues/1656).

A role spec looks like the following:
```yaml
# role name, e.g.
name: "foo.bar"
# human readable title for the role, e.g.
title: "Foo Barrer"
# human readable description for the role, e.g. 
description: "Allows doing Bar on Foo resources"
# include permissions from the following...
include:
  # a list of permissions, e.g.
  permissions:
  - foo.bar.doSomething
  # a list of roles, e.g.
  roles:
  - roles/foo.bar
  # only include permissions matching any extended regex in this list, e.g.
  permissionRegxes:
  - ^foo.bar.(get|list)
exclude:
  # exclude any permissions matching any extended regex in this list, e.g.
  permissionRegexes:
  - SuperDangerousOperation$
```

Sadly, I did this with bash and yq.  Redoing this with go would be a great help-wanted.

Updates the following managed org roles:
- `prow.viewer` - Auditing `roles/viewer` from `k8s-infra-e2e-prow-viewers@` made me realize they don't have view access to buckets in most e2e projects.  Now they will.

Adds the following previously unmanaged org roles:
- `CustomRole` - not a good name, but held off on migrating to better name until I get some historical context
- `StorageBucketLister` - this can go away once audit.viewer is live

Adds the following new managed org roles:
- `audit.viewer` - replace the sundry bindings used by `k8s-infra-gcp-auditors@`
  - I ran the script this far already to vet the effective permissions before and after look good (https://gist.github.com/spiffxp/7d2bfaa9c510de89a7f2ce1c31ec1a00)
- `organization.admin` - replace the sundry non-`roles/owner` bindings used by `k8s-infra-gcp-org-admins@`
  - not running the script to deploy this until I've gotten review

There is a little more followup I would like to do to close out https://github.com/kubernetes/k8s.io/issues/1659, but not in this PR:
- pull the `ServiceAccountLister` role out of projects and up into the org (but keep the binding(s) project level)
- remove `StorageBucketLister` unless there's a reason to keep it around
- remove the detritus cleanup code marked with TODO's
- setup a periodic job to refresh the custom roles to pick up new permissions that are added over time

PR #1722, opened 2021-02-25
audit-followup: remove stale iam bindings for e2e projects

spiffxp:
Should address the following:
- `group:k8s-infra-prow-viewers@kubernetes.io` should not have `roles/viewer` (ref: https://github.com/kubernetes/k8s.io/issues/1661)

Should partially address the following:
- `user:*` should not have `roles/owner` for projects (ref: https://github.com/kubernetes/k8s.io/issues/299)

PR #1737, opened 2021-03-01
ensure-organization.sh followup

spiffxp:
Followup to https://github.com/kubernetes/k8s.io/pull/1726, intended to fix https://github.com/kubernetes/k8s.io/issues/1659

Basically:
- drop StorageBucketLister
- replace per-project ServiceAccountLister with org-level iam.serviceAccountLister (though still not actually sure why this role is necessary)
- add functions to lib_iam to allow removal of stale bindings
- lib_iam will only support creating custom roles from file
- try diffing results of binding changes (adding `yq` to the mix to do this)
- remove the stale e2e project binding removal calls introduced in https://github.com/kubernetes/k8s.io/pull/1722

Order is going to matter for rolling this out:
- ensure-organization.sh to create the org-level roles
- terraform apply in prow-build / prow-build-trusted
- check to see if terraform removed the project-level role
- ensure-main-project.sh
- followup PR to remove ensure_removed calls

There is one more custom role I am interested in creating, for the "cluster-admins" group, but I can do that in followup.


infra/gcp/lib_iam.sh
```diff
@@ -184,32 +235,139 @@ function ensure_removed_project_role_binding() {
...
-    local principal="${1}"
-    local role="${2}"
+    local scope="${1}"
+    local id="${2}"
+    local name="${3}"
+    local file="${4}"
+    case "${scope}" in
+      organization | project ) ;;
+      * )
+        echo >/dev/stderr "_ensure_custom_iam_role_from_file(scope, id, name, file) scope must 'organization' or 'project'"
```
thockin, line 257: I forget the details, but there was something that caused me to switch everything to 

```
echo "_ensure_custom_iam_role_from_file(scope, id, name, file) scope must 'organization' or 'project'" >&2
```

syntax

infra/gcp/lib_iam.sh
```diff
@@ -184,32 +235,139 @@ function ensure_removed_project_role_binding() {
...
+    # deleted roles can be undeleted within 7 days; after that must wait 30 days to create a role with same id
+    # ref: https://cloud.google.com/iam/docs/creating-custom-roles#deleting-custom-role
+    if <"${before}" grep -q "^deleted: true"; then
+        gcloud iam roles undelete ${scope_flag} "${name}"
+    fi
+
+    # name is foo.bar, but gcloud wants scopes/id/role/foo.bar in the file
+    local full_name="${scope}s/${id}/roles/${name}"
+    <"${file}" sed -e "s|^name: ${name}|name: ${full_name}|" >"${ready}"
+    gcloud iam roles "${verb}" ${scope_flag} "${name}" --file "${ready}" | yq -Y 'del(.etag)' > "${after}"
```
thockin, line 283: is it kosher to depend on yq?  It's not installed on my desktop, for example :)  What package is it in?

No option for JSON?

infra/gcp/lib_util.sh
```diff
@@ -35,7 +38,7 @@ function _nocolor() {
 function color() {
     _color "$1"
     shift
-    echo "$@$(_nocolor)"
+    echo "$*$(_nocolor)"
```
thockin, line 41: I am curious why this?

infra/gcp/roles/iam.serviceAccountLister.yaml
```diff
@@ -0,0 +1,18 @@
...
+#   permissions:
+#   - iam.serviceAccounts.list
+# 
+#
+description: Can list ServiceAccounts
+includedPermissions:
+  - iam.serviceAccounts.list
+name: iam.serviceAccountLister
+stage: GA
+title: Secret Account Lister
```
thockin, line 18: Secret or Service?

thockin: Also ServiceAccount is probably a proper noun (one word?)

infra/gcp/roles/specs/iam.serviceAccountLister.yaml
```diff
@@ -0,0 +1,9 @@
+# formerly a project-level custom role for any project with GKE clusters,
+# not actually sure why this account is necessary but here we are
+title: Secret Account Lister
```
thockin, line 3: "Secret Account" or "ServiceAccount"?

infra/gcp/roles/specs/iam.serviceAccountLister.yaml
```diff
@@ -0,0 +1,9 @@
...
+description: Can list ServiceAccounts
+name: iam.serviceAccountLister
```
thockin, line 5: This is is identical to the Role defined above in infra/gcp/roles/iam.serviceAccountLister.yaml  ??  Why do we have both?

infra/gcp/lib_iam.sh
thockin: Oops, I saw the ensure.  NM

infra/gcp/lib_iam.sh
spiffxp: done

infra/gcp/lib_util.sh
spiffxp: I don't feel strongly about this, but my editor is setup to nag me about shellcheck errors (vs. warnings), and this came up as: https://github.com/koalaman/shellcheck/wiki/SC2145

infra/gcp/roles/specs/iam.serviceAccountLister.yaml
spiffxp: updated to `ServiceAccount List`

infra/gcp/roles/specs/iam.serviceAccountLister.yaml
spiffxp: so I agree, it's silly in this case to have two files that look identical

but the idea is to consistently follow the pattern of: 
- custom roles are defined via a (custom) spec
- specs generate concrete roles
- custom role definitions are checked in to git (so we know _when_ custom roles get new permissions)
- then _those_ get applied

this lets us apply common defaults, and moves us toward a model of "write a spec, have something else apply it"

we can skip the "check in concrete roles" step and go straight to applying, but at the moment I find it useful to have them checked in for diffing when regeneration picks up new p

infra/gcp/lib_util.sh
thockin: nteresting read

infra/gcp/roles/specs/iam.serviceAccountLister.yaml
thockin: So the non-spec is the readback?  I'm looking at the script and trying to figure out what it's doing.  It mixes the spec and the result of `gcloud iam roles describe`.  It's not clear to me when that script is run or who runs it or when.

spiffxp: It is not currently run except by humans.  Now that @hh has proven pr-creator to be reusable out of test-infra, the last followup item I have for https://github.com/kubernetes/k8s.io/issues/1659 is to setup prowjobs that run this on postsubmit and periodically.

The reason for mixing the spec into comments of the generated role was to be 100% clear about what spec the role was generated from (vs. "this file", "this time", "this repo", "this hash" etc... but perhaps those would be a better / more concise approach to aim for)

spiffxp:
>replace per-project ServiceAccountLister with org-level iam.serviceAccountLister (though still not actually sure why this role is necessary)

btw @thockin do you have the context for why we needed to make this role in the first place?  I haven't plumbed the depths of `aaa`'s creation to figure that out yet

thockin:
I do not recall exactly why it exists - I think it was to allow audits to
list SA's without modifying them and there was no fine-grained role for
that.

spiffxp:
If that's the case, maybe we can drop it entirely (should be subsumed by custom `audit.viewer` role now)

infra/gcp/roles/specs/iam.serviceAccountLister.yaml
thockin: What I am missing is why we check the readback in, and how that is different from the audit.

IOW, why is it more than 
* Here's a spec
* Here's a script to apply the spec
* Here's a periodic script to audit the results
* Here's the last known-good audit state

If the periodic audit shows new perms, alert.  I'm missing the role of the intermediate?

I trust you so I am totally willing to OK this and just grow my comprehension async to this PR

spiffxp:
>If that's the case, maybe we can drop it entirely (should be subsumed by custom audit.viewer role now)

I don't have an account that's unprivileged enough to be affected by this so I don't want to try it as part of this PR, opened https://github.com/kubernetes/k8s.io/issues/1753

infra/gcp/roles/specs/iam.serviceAccountLister.yaml
spiffxp: You may be right, this could be spurious and an unnecessary extra step.

I sorta want to keep this in as an experiment and remove later if we feel it's too annoying.  Reasons why:
- this might make it easier to vet audit output "here's what I absolutely expect it to be", and regenerating from primitive roles at time of audit might pick up new permissions I didn't know about last time I said "sure, refresh the role"
- it may be easier for contributors to find out "ok but what _exact_ permissions does this role have" even if they don't have iam permissions or gcloud handy

I guess I sort of think of this as the equivalent to a terraform state file.  What was the state of the world the last time I managed it.  Compare that to what the state of the world is now (did something else change things?) + the changes I intend to make to the state of the world.

spiffxp:
/hold
ok, wait, the latest audit pr does not line up with what I expect the state of the world to be

infra/gcp/lib_util.sh
spiffxp: FWIW, I'm now seeing colorized output on my mac (running bash 4.4 installed by homebrew).  I suspect this was the change that did it.

hh: Nice.

spiffxp:
Related to refactoring infra/gcp, ref: https://github.com/kubernetes/k8s.io/issues/516

PR #1739, opened 2021-03-01
Add cert monitor manifest

rikatz:
Cert-monitor will read the metrics from cert-manager prometheus exporter, transform it into days and push to stackdriver.

TODO:
* Turn the GCP Project ID configurable
* Improve the Prometheus and GKE exporter security to drop privileges


cert-manager/cert-monitor.yaml
```diff
@@ -0,0 +1,94 @@
...
+        volumeMounts:
+        - mountPath: /etc/prometheus
+          name: exporterconfig
+        - mountPath: /data
+          name: promdata
+      - name: sidecar
+        image: gcr.io/stackdriver-prometheus/stackdriver-prometheus-sidecar:0.8.2
+        imagePullPolicy: Always
+        args:
+        - --stackdriver.project-id=fluent-plate-304118
```
ameukam, line 65: Do you mean `kubernetes-public` ?

cert-manager/cert-monitor.yaml
```diff
@@ -0,0 +1,94 @@
...
+        - --stackdriver.generic.location=us-central1-c
```
ameukam, line 66: Can we use `us-central` instead ?

ameukam:
/assign @spiffxp @thockin 
/hold
If others people want to comment

cert-manager/cert-monitor.yaml
rikatz: @ameukam thanks! Yes, it needs to be the project id, so I'll change here.

cert-manager/cert-monitor.yaml
rikatz: Let me test and I'll let you know :)

rikatz: Corrected, it works fine :)

cert-manager/cert-monitor.yaml
```diff
@@ -0,0 +1,94 @@
...
+      rules:
+      - record: certificate_expire_remaining_days 
+        expr: (certmanager_certificate_expiration_timestamp_seconds - time())/86400
+  sdexporter.yaml: |
+    static_metadata:
+    - metric: certificate_expire_remaining_days
+      type: gauge 
+      value_type: double 
+      help: Seconds until a certificate days
+kind: ConfigMap
```
thockin, line 28: If humans are going to manage this, maybe move all these to the top of each block? :)

cert-manager/cert-monitor.yaml
```diff
@@ -0,0 +1,94 @@
...
+  replicas: 1
+  selector:
+    matchLabels:
+      app: prom-certmanager
+  template:
+    metadata:
+      labels:
+        app: prom-certmanager
+    spec:
+      containers:
```
thockin, line 50: You want to specify resources for these, right?

cert-manager/cert-monitor.yaml
```diff
@@ -0,0 +1,94 @@
...
+      type: gauge 
+      value_type: double 
+      help: Seconds until a certificate days
+kind: ConfigMap
+metadata:
+  name: exporterconfig
+  namespace: cert-manager
+---
+apiVersion: apps/v1
+kind: Deployment
```
thockin, line 34: I would love some comments (yay YAML, you beat JSON at something) to explain what's going on.  Or a README-ish file?

cert-manager/cert-monitor.yaml
rikatz: what if those are humans that deal with lisp yamls, and read from the end to the beginning? :P

Kidding, corrected here!

cert-manager/cert-monitor.yaml
rikatz: Yes! Also I do want to make this image run as non-root, just didn't had enough time to test it, but I'll test right now and just push the manifest corrected

cert-manager/cert-monitor.yaml
rikatz: I'll make both :) 

Put some comments on the YAML, at least about the prometheus and sdexporter configuration, and write on the README (I've already added a README on the another PR)

rikatz:
@thockin added some limits, some comments, some security context :)

Tested in my cluster and worked fine, hopefully this is the first step to renew the certificates before anything blows.

Next step: I'll check a way to automate the dashboard and alerting (there's an issue assigned to me anyway) so we can somehow automate everything.

spiffxp:
Checking in... do we still need this?  I admit I have kinda lost track at where we are with using cert-manager vs. GKE managed certificates these days

rikatz:
/close

As we discussed, per migration to Google managed certificates this is not necessary anymore :)

PR #1744, opened 2021-03-02
Don't audit the node count in a cluster

thockin:
Those things change over time, not worth alerting on.


spiffxp (approved):
/approve
/lgtm
fine by me

if we want to _really_ audit these more fully, we should dump `--format=yaml` and strip out things we know will change automatically

but that can be a later task when we find the bandwidth or need

Issue #1476, opened 2020-12-10
cert-manager cannot renew k8s-io-prod certificate due to second IPv6 ingress


munnerz:
As initially reported and discussed [on Slack](https://kubernetes.slack.com/archives/CCK68P2Q2/p1607546569252400), the `k8s-io-prod` certificate (used for the redirector service) is failing to renew.

After some debugging, there are two issues at play here:

* #1374 adding a second IPv6 only Ingress resource and the AAAA record configured to point to this - cert-manager only knows to update a _single_ Ingress (named via the `edit-in-place` annotation) to inject `path` entries for the HTTP01 challenge solvers. As per https://letsencrypt.org/docs/ipv6-support/, if an AAAA record is returned then Let's Encrypt will *prefer* that and utilise it first. If we update cert-manager to *only* update the IPv6 Ingress resource, Let's Encrypt will quite likely pass validation (as it won't check IPv4) *however*, because cert-manager performs a 'self check' to ensure all routes are serving traffic correctly, and because our *Pods* do not utilise IPv6, the self check performed by cert-manager will never pass either. Ultimately, we need to ensure *both* Ingress resources contain the challenge path entries (which is not something that cert-manager supports today).

* When running `kubectl describe` on our Ingress resources, the follow error is shown:

```
  Warning  Sync    6m38s (x29 over 85m)  loadbalancer-controller  Error during sync: error running backend syncing routine: googleapi: Error 403: Exceeded limit 'MAX_DISTINCT_NAMED_PORTS' on resource 'k8s-ig--ea949c440a044527'. Limit: 1000.0, limitExceeded
```

It appears that ingress-gce does *not* clean up 'unused named ports' - it _was_ originally added in https://github.com/kubernetes/ingress-gce/pull/430, but was later reverted in https://github.com/kubernetes/ingress-gce/pull/585.

We can see that there are a _lot_ of named ports associated with the 3 'unmanaged instance groups' that ingress-gce creates:

```
gcloud compute instance-groups get-named-ports k8s-ig--ea949c440a044527 --project kubernetes-public --zone us-central1-c | wc -l
1001
```

As you can see here, there are far fewer than 1000 nodePorts in our aaa cluster:

```
kubectl get services -A -o yaml | grep nodePort
      nodePort: 30062
      nodePort: 32044
      nodePort: 30404
      nodePort: 31694
      nodePort: 32072
      nodePort: 32212
      nodePort: 32382
      nodePort: 30980
[...]

Issue #1745, opened 2021-03-02
Wildcard and migration of cert-manager solver from HTTP01 to DNS01


rikatz:
Per Slack conversation with @munnerz , we could migrate the cert-manager solver from HTTP01 to DNS01, and issue a *.kubernetes.io certificate instead of relying on each ingress object.

This will solve two problems:
* You can use any ingress or any IP as long as you own that DNS name
* Using wildcard certificate reduces significantly the amount of managed certificates

This have some caveats:
* cert-manager needs to have permissions managing the DNS that owns kubernetes.io domain, to write the right challenges on _acme-challenge.kubernetes.io TXT (and other desired domains)
* This may represent a permission that we don't want to give for a workload running inside our public cluster

The alternative approach for this is to separate these privileges [delegating the solving challenges](https://www.eff.org/deeplinks/2018/02/technical-deep-dive-securing-automation-acme-dns-challenge-validation) to another non sensitive domain.

So the solution might be:
```
_acme-challenge.kubernetes.io IN CNAME cert-solver.my-other-domain.com
```
and  cert-solver.my-other-domain.com being managed by cert-manager

Another solution that can be discussed is to create a granular permission within IAM allowing only TXT record changes, but this doesn't seems to be supported yet.

This issue was opened so we can start drafting the migration to DNS Challenges (if desired) and how to securely implement this

/assign

rikatz:
@munnerz per meeting today, we're going to move forward with the controller until cert-manager gets the proper patch.

Issuing a wildcard can bring to us some risk we don't want to have right now, about if someone can use that certificate the person can impersonate any kubernetes.io certificate, so for now let's move with per-ingress certificate :)

I'll try to review the controller this week, but I guess everything is fine with the controller that's already deployed right now! Thanks @munnerz 

cc @spiffxp @thockin 

/close

PR #1746, opened 2021-03-02
Add Ingress rules syncer controller to resolve dual-stack k8s.io cert-manager renewal

munnerz:
fixes https://github.com/kubernetes/k8s.io/issues/1476

This PR adds a controller, written with kubebuilder, that syncs the `rules` from one Ingress to another. This does *not* add any of the machinery needed to *push* the docker image for this controller, nor does it add deployment manifests to actually run it in the AAA cluster.

An alternative resolution to this issue is to switch to DNS-01 for challenges, as per https://github.com/kubernetes/k8s.io/issues/1745. That may be a better option as it'll also enable us to use a wildcard certificate for `*.k8s.io` and `*.kubernetes.io`, rather than re-validating ~50 domains every 3 months (and avoids having to maintain another binary/controller).

cc @rikatz @thockin


cert-manager/ingress-rules-syncer/config/prometheus/monitor.yaml
```diff
@@ -0,0 +1,16 @@
+
+# Prometheus Monitor Service (Metrics)
+apiVersion: monitoring.coreos.com/v1
+kind: ServiceMonitor
```
ameukam, line 4: IIRC, we still don't have a prometheus-operator instance deployed in `aaa`.

cert-manager/ingress-rules-syncer/controllers/ingress_controller.go
```diff
@@ -0,0 +1,139 @@
...
+func (r *IngressReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
+	log := r.Log.WithValues("ingress", req.NamespacedName)
+
+	var ing networkingv1beta1.Ingress
+	if err := r.Client.Get(ctx, req.NamespacedName, &ing); err != nil {
+		return ctrl.Result{}, client.IgnoreNotFound(err)
+	}
+
+	// Base case where an Ingress does NOT have a 'sync-from' label
+	if ing.Labels == nil || ing.Labels[syncFromLabelKey] == "" {
```
rikatz, line 62: I guess you don't need to verify if Labels is nil. 

When you call an empty Map, with an unexistant value it will work. -> https://play.golang.org/p/0eLKBSSyyZJ

cert-manager/ingress-rules-syncer/controllers/ingress_controller.go
```diff
@@ -0,0 +1,139 @@
...
+		log.V(4).Info("no sync-from label - skipping syncing")
+		return ctrl.Result{}, nil
+	}
+
+	syncFromName := ing.Labels[syncFromLabelKey]
```
rikatz, line 67: Maybe with the comment above, and to simplify, move `syncFromName := ing.Labels[syncFromLabelKey]` to earlier assignment, verify if the value is "", and then use it later.

```golang
syncFromName := ing.Labels[syncFromLabelKey]
if syncFromName == "" {
      log.V(4).Info("no sync-from label - skipping syncing")
      return ctrl.Result{}, nil
}

// Explicitly avoid creating infinite loops if an ingress references itself
if syncFromName == req.Name {
	log.V(4).Info("ingress has sync-from label referencing itself - skipping syncing")
	return ctrl.Result{}, nil
}
```

rikatz (commented):
@munnerz 

So far I could understand things, just left some minor reviews :)

Anyway, I guess it would be good also to get rid of all the extra things generated by kubebuilder (webhooks, unnecessary RBAC files for proxy, the cert-manager certificate files) as this operator mainly acts reconciling the objects, right?

Thank you for your effort on this!

cert-manager/ingress-rules-syncer/main.go
```diff
@@ -0,0 +1,102 @@
...
+	if err != nil {
+		setupLog.Error(err, "unable to start manager")
+		os.Exit(1)
+	}
+
+	if err = (&controllers.IngressReconciler{
+		Client: mgr.GetClient(),
+		Log:    ctrl.Log.WithName("controllers").WithName("Ingress"),
+	}).SetupWithManager(mgr); err != nil {
+		setupLog.Error(err, "unable to create controller", "controller", "Ingress")
```
rikatz, line 83: maybe we should use a different name for the controller, to not make confusion with Ingress Controllers :)

spiffxp:
/close
I'm going to close this per https://github.com/kubernetes/k8s.io/issues/1943 but please /reopen if I've missed something

Issue #1597, opened 2021-01-29
Rename default branch for kubernetes/k8s.io


spiffxp:
Part of umbrella issue https://github.com/kubernetes/org/issues/2222

Issue to spec out and then run through renaming the default branch for this repository

- update pre/postsubmits to trigger on old branch name and new branch name
- determine whether periodics use default branch or have name hardcoded in git commands
  - if hardcoded, does attempting to checkout old name break?
- determine whether branch name is part of any job names
- determine if any docs need to be updated
- send notice
- rename branch (ref: https://github.com/github/renaming#renaming-existing-branches)
- if necessary, update periodics to refer to new branch name
- if necessary, rename jobs
- update branch protection config
- update jobs to only trigger on new branch name
- make sure things still work
  - keep an eye on ci and prs for a bit
  - trial the local development experience with pre-rename clone
- send notice

/sig contributor-experience
/area github-management
trialing a repo for this subproject
/wg k8s-infra
owns this repo
/sig testing
to discover any prow issues
/milestone v1.21
/priority important-soon
/assign @spiffxp 
I plan on speccing this out, whether/when to execute depends on how kubernetes/org goes

spiffxp:
PR up for job config changes + prow config changes: https://github.com/kubernetes/test-infra/pull/20675 (holding merge until next week, and completion of the below review)

Still need to:
- review the code in k8s.io to confirm whether it explicitly references the master branch...
- review docs to see whether master is explicitly references (beyond uris that github will redirect for us)

spiffxp:
Merged https://github.com/kubernetes/test-infra/pull/20675

spiffxp:
Renamed the branch

spiffxp:
Will leave as-is and cleanup tomorrow

Known issues:
- go.k8s.io/redirects 404's for me (but visiting https://github.com/kubernetes/k8s.io/tree/master/k8s.io#redirections is properly redirected to main by github)
https://github.com/kubernetes/k8s.io/blob/a8e0b654626c77de30220097cf8253d32363bcb8/k8s.io/configmap-nginx.yaml#L242

spiffxp:
looking at jobs in our test grid:
- [ ] https://testgrid.k8s.io/wg-k8s-infra-k8sio#ci-k8sio-groups
- [x] https://testgrid.k8s.io/wg-k8s-infra-k8sio#ci-k8sio-image-promo
[...]

PR #1764, opened 2021-03-07
Replace master branch references with main

nikhita:
Ref: https://github.com/kubernetes/k8s.io/issues/1597#issuecomment-790810435

/assign @spiffxp 

/hold
hold should be removed by someone who can deploy the k8s.io changes

### DIFF
--- a/dns/octodns-docker/cloudbuild.yaml
+++ b/dns/octodns-docker/cloudbuild.yaml
@@ -5,7 +5,7 @@ steps:
           '.']
 substitutions:
   _GIT_TAG: "12345"
-  _PULL_BASE_REF: 'master'
+  _PULL_BASE_REF: 'main'
 images:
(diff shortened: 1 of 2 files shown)


nikhita:
>  hold should be removed by someone who can deploy the k8s.io changes

It looks like @thockin has been deploying recently

/assign @thockin

dims:
i can do this

/approve
/lgtm
/hold cancel

thockin:
@dims - when you push these nginx changes to prod you also need to bounce
the nginx instances - they don't self-restart.

dims:
@thockin aha! could we add any extra step to the README please? usually i just read that and execute step by step :)

thockin:
Which README?  The one that says to `make deploy` (which doesn't exist, LOL).

dims:
LOL

PR #1765, opened 2021-03-07
hack: add verify script for boilerplate and yamllint check

nikhita:
Ref: "ideas of follow up" in https://github.com/kubernetes/test-infra/pull/20956

This PR:

- Updates existing `verify-yamllint.sh` script to install yamllint if
it does not exist.

- Adds `verify-boilerplate.sh` script for verifying boilerplate.
This is derived from the [script](https://github.com/kubernetes/enhancements/blob/master/hack/verify-boilerplate.sh) in the enhancements repo.

- Adds `verify.sh` script to test all verify scripts at once. This would
be useful for adding future verify scripts.

- Fixes boilerplate headers to make the script pass

/assign @spiffxp

### DIFF
--- a/hack/verify-yamllint.sh
+++ b/hack/verify-yamllint.sh
@@ -37,8 +37,8 @@ if [ $# != 0 ]; then
 fi
 
 if ! command -v yamllint >/dev/null 2>&1; then
-  echo >/dev/stderr "ERROR: yamllint not found - please install with: ${pip} install -r ${pip_requirements}"
-  exit 1
+  echo "yamllint not found - installing with: ${pip} install -r ${pip_requirements}"
+  ${pip} install -r ${pip_requirements}
 fi
 
 version=$(yamllint --version | awk '{ print $2 }')
--- a/hack/verify.sh
+++ b/hack/verify.sh
@@ -1,0 +1,120 @@
+#!/usr/bin/env bash
+
+# Copyright 2021 The Kubernetes Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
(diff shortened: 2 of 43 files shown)


hack/verify-yamllint.sh
```diff
@@ -37,8 +37,8 @@ if [ $# != 0 ]; then
 fi
 
 if ! command -v yamllint >/dev/null 2>&1; then
-  echo >/dev/stderr "ERROR: yamllint not found - please install with: ${pip} install -r ${pip_requirements}"
-  exit 1
+  echo "yamllint not found - installing with: ${pip} install -r ${pip_requirements}"
+  ${pip} install -r ${pip_requirements}
```
ameukam, line 41: It would be great to have a comment saying we assume `pip3` is already installed.

spiffxp: Fair point, I think we can address in followup. Other scripts in this repo assume the presence of python3, which implies pip3. Not sure if they verify their assumptions either

hack/verify.sh
```diff
@@ -0,0 +1,120 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+set -o errexit
+set -o nounset
+set -o pipefail
+
+KUBE_ROOT=$(dirname "${BASH_SOURCE}")/..
```
spiffxp, line 21: Nit for followup: I think I would rather use REPO_ROOT or something not so tied to k/k script conventions. Repo root is what I saw in some test-infra scripts. 

I don't feel strongly about this but it would be good to standardize across the scripts we have in this repo. This should at least use `BASH_SOURCE[0]` as a followup

spiffxp (approved):
/approve
/lgtm

PR #1770, opened 2021-03-09
Exclude verify-terraform.sh

ameukam:
`verify-terraform.sh` needs a path as argument when it's executed.
However the script is invoked through `verify.sh` and will most likely
fail.


nikhita (approved):
This can go in with https://github.com/kubernetes/k8s.io/pull/1732 but it doesn't harm us if we exclude it right now.

also see https://github.com/kubernetes/test-infra/pull/21214#issuecomment-793380112

/lgtm
/approve

PR #1781, opened 2021-03-11
Promote v0.1.7 image in scl-image-builder

SanikaGawhane:
/cc @codenrhoden


k8s.gcr.io/images/k8s-staging-scl-image-builder/images.yaml
```diff
@@ -1 +1,3 @@
-# No images yet
+- name: cluster-api-aws-controller
```
codenrhoden, line 1: This should be `cluster-node-image-builder-amd64`

PR #1791, opened 2021-03-16
Simplify go.k8s.io/owners/$ghuser

nikhita:
With the switch to a new version of hound, we can now specify
filepaths (like `vendor/`) that must be excluded. This means that we no
longer need to use the nastry regexp.

Ref: https://kubernetes.slack.com/archives/CCK68P2Q2/p1615493343163600

Example: https://cs.k8s.io/?q=spiffxp&i=fosho&files=OWNERS&excludeFiles=vendor%2F&repos=

/assign @spiffxp @mrbobbytables 

/hold
should be approved by someone who can deploy this

PR #1789, opened 2021-03-15
feat: add shortlink pointing to k8s logo page

RinkiyaKeDad:
Signed-off-by: Arsh Sharma <arshsharma461@gmail.com>

Fixes kubernetes/community#5360

PR #1792, opened 2021-03-16
k8s.io: add no_parent_owners: true for OWNERS

nikhita:
Changes to this directory need to be deployed manually and should not be approved by someone who cannot deploy.

Previous cases where `/hold` is explicitly used to ensure this - https://github.com/kubernetes/k8s.io/pull/1791, https://github.com/kubernetes/k8s.io/pull/1789, https://github.com/kubernetes/k8s.io/pull/1764

Adding `no_parent_owners: true` removes the need to add an explict hold and enforces this restriction.

/assign @thockin 
/cc @spiffxp @ameukam

### DIFF
--- a/k8s.io/OWNERS
+++ b/k8s.io/OWNERS
@@ -1,5 +1,7 @@
 # See the OWNERS docs at https://go.k8s.io/owners
 
+options:
+  no_parent_owners: true
 approvers:
 - ixdy
 - thockin


k8s.io/OWNERS
```diff
@@ -1,5 +1,7 @@
 # See the OWNERS docs at https://go.k8s.io/owners
 
+options:
+  no_parent_owners: true
 approvers:
 - ixdy
```
BenTheElder, line 6: @ixdy are you still active here?

also happy to help here @nikhita

ixdy: no, I should probably be removed.

spiffxp: I can do this as followup

spiffxp (approved):
/approve
/lgtm

PR #1796, opened 2021-03-16
Document and streamline k8s.io site deploy

thockin:
1) Retool the canary.sh script into deploy.sh.
2) Use `rollout restart` instead of scale to 0.
3) Document how to do a rollout.

PR #1799, opened 2021-03-17
Adding dims as approver for k8s.io

dims:
@thockin unblocked me from being able to do things with https://github.com/kubernetes/k8s.io/pull/1796

thanks Tim! 

Signed-off-by: Davanum Srinivas <davanum@gmail.com>

### DIFF
--- a/k8s.io/OWNERS
+++ b/k8s.io/OWNERS
@@ -5,7 +5,11 @@ options:
 approvers:
 - ixdy
 - thockin
+- aojea
+- bentheelder
+- cblecker
 - dims
+- spiffxp
 
 # limiting factor from being an approver is inability to deploy changes once PRs merge
 reviewers:
--- a/k8s.io/OWNERS
+++ b/k8s.io/OWNERS
@@ -5,6 +5,7 @@ options:
 approvers:
 - ixdy
 - thockin
+- dims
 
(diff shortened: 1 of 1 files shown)


spiffxp (approved):
/approve
/lgtm

k8s.io/OWNERS
```diff
@@ -5,6 +5,7 @@ options:
 approvers:
 - ixdy
 - thockin
+- dims
```
spiffxp, line 8: We should get parity between approvers and this group https://github.com/kubernetes/k8s.io/blob/390481ba911059c23a2739cb56bccf8be3ea183b/groups/groups.yaml#L543-L554

```suggestion
- aojea
- bentheelder
- cblecker
- dims
- spiffxp
```

(you'll also need to remove ixdy)

Can you update this to add yourself to the group? Add thockin and myself (@google.com) if it's not too much trouble?

spiffxp (changes requested):
since I don't have approve, let's make this followup I never got around to today mentioned in https://github.com/kubernetes/k8s.io/pull/1792#discussion_r595494784

I'm also fine if we decide to prune this list of people but I feel like parity first, prune followup

dims:
@spiffxp this is ready now. thanks!

spiffxp (approved):
/lgtm

Group updates are missing but that can be a followup

PR #1829, opened 2021-03-24
Add iscsi csi driver

huffmanca:
This PR attempts to follow the instructions at https://github.com/kubernetes/k8s.io/tree/main/k8s.gcr.io for including the iSCSI CSI Driver staging repo. Once merged, we can begin configuring this driver for automated image builds.

### DIFF
--- a/k8s.gcr.io/images/k8s-staging-sig-storage/generate.sh
+++ b/k8s.gcr.io/images/k8s-staging-sig-storage/generate.sh
@@ -24,6 +24,7 @@ csi-snapshotter
 csi-external-health-monitor-agent
 csi-external-health-monitor-controller
 hostpathplugin
+iscsiplugin
 livenessprobe
 local-volume-provisioner
 mock-driver


k8s.gcr.io/images/k8s-staging-iscsi/OWNERS
```diff
@@ -0,0 +1,11 @@
+# See the OWNERS docs at https://go.k8s.io/owners
+# Should always mirror the maintainers/reviewers in https://github.com/kubernetes-csi/csi-driver-iscsi/blob/master/OWNERS
+
+approvers:
```
spiffxp, line 4: /assign @msau42 @saad-ali @j-griffith 
Looking for an lgtm to confirm on behalf of sig/subproject

spiffxp (commented):
/hold
I'll approve once I see lgtm

groups/groups.yaml
```diff
@@ -1513,3 +1513,14 @@ groups:
       - natashasarkar@google.com
       - extaeger@google.com
       - samwronski@google.com
+
+  - email-id: k8s-infra-staging-iscsi@kubernetes.io
+    name: k8s-infra-staging-iscsi
```
msau42, line 1518: We actually have k8s-infra-staging-sig-storage where we've been publishing all our sig projects under. I think we can use that instead.

j-griffith: @huffmanca seems like it just needs the changeto infra-staging as mentioned about and then should be good to go?

huffmanca: I added the iscsiplugin to the `k8s.gcr.io/images/k8s-staging-sig-storage/generate.sh` script. Checking the other repos here, I didn't see specific information for them under `k8s.gcr.io/images` or `k8s.gcr.io/manifests`, so I've removed these references.

k8s.gcr.io/images/k8s-staging-iscsi/OWNERS
huffmanca: @spiffxp ,

Michelle lgtm'd this PR (in its much smaller state) about a month ago. What are your thoughts on removing the hold?

msau42:
/hold cancel

The updated pr doesn't require creating new groups.

PR #1831, opened 2021-03-25
slack-moderator-words: fix missing port definition

cpanato:
was missing the port configuration

/assign @ameukam @nikhita


nikhita:
THANK YOU!!

PR #1833, opened 2021-03-25
Give some love to gcsweb

ameukam:
Switch the ingress to `networking.k8s.io/v1beta1`
Add more instructions to the docs

### DIFF
--- a/gcsweb.k8s.io/README.md
+++ b/gcsweb.k8s.io/README.md
@@ -32,3 +32,28 @@ https://github.com/kubernetes/test-infra/tree/master/gcsweb into Docker
 container, which is then uploaded to private container storage at
 https://k8s.gcr.io/gcsweb-amd64 and fetched during processing
 of `deployment.yaml` by `kubectl apply`.
+
+##### How to deploy
+
+Ensure you have [access to the cluster]
+
+Ensure you are a member of both:
+
+- k8s-infra-cluster-admins@kubernetes.io
+- k8s-infra-rbac-gcsweb@kubernetes.io
+
+Connecting to cloud-shell:
+
+```shell
+gcloud alpha cloud-shell ssh --authorize-session --project kubernetes-public
+```
+
+Deploy gcsweb:
+
+```shell
+git clone https://github.com/kubernetes/k8s.io
+cd k8s.io
+kubectl apply -Rf gcsweb/
+```
+
+[access to the cluster]: https://github.com/kubernetes/k8s.io/blob/main/running-in-community-clusters.md#access-the-cluster
--- a/gcsweb.k8s.io/deployment.yaml
+++ b/gcsweb.k8s.io/deployment.yaml
@@ -2,6 +2,7 @@ apiVersion: apps/v1
 kind: Deployment
 metadata:
   name: gcsweb
+  namespace: gcsweb
   labels:
     app: gcsweb
 spec:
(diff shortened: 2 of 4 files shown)


gcsweb.k8s.io/README.md
```diff
@@ -32,3 +32,28 @@ https://github.com/kubernetes/test-infra/tree/master/gcsweb into Docker
...
 https://k8s.gcr.io/gcsweb-amd64 and fetched during processing
 of `deployment.yaml` by `kubectl apply`.
+
+##### How to deploy
+
+Ensure you have [access to the cluster]
+
+Ensure you are a member of both:
+
+- k8s-infra-cluster-admins@kubernetes.io
```
thockin, line 42: You should only need to be in `k8s-infra-rbac-gcsweb@kubernetes.io` to work within that single namespace.  If that is not sufficient, something is wrong.

`k8s-infra-cluster-admins` is basically root.

gcsweb.k8s.io/deployment.yaml
```diff
@@ -2,6 +2,7 @@ apiVersion: apps/v1
 kind: Deployment
 metadata:
   name: gcsweb
+  namespace: gcsweb
```
thockin, line 5: Historically I left the NS out of the YAML so that we could do canaries to another namespace (as in k8s.io). 

Obviously this is a tradeoff, and I am happy to discuss :)

gcsweb.k8s.io/ingress.yaml
```diff
@@ -1,7 +1,8 @@
-apiVersion: extensions/v1beta1
+apiVersion: networking.k8s.io/v1beta1
 kind: Ingress
 metadata:
   name: gcsweb
+  namespace: gcsweb
```
thockin, line 5: same as above

gcsweb.k8s.io/service.yaml
```diff
@@ -2,6 +2,7 @@ apiVersion: v1
 kind: Service
 metadata:
   name: gcsweb
+  namespace: gcsweb
```
thockin, line 5: same as above

gcsweb.k8s.io/deployment.yaml
spiffxp: We can override this at deploy time. I personally like the idea of not deploying to default if someone forgets

gcsweb.k8s.io/README.md
spiffxp: If deployment involves updating secrets then this is needed. If not, it isn't

I want us to revisit that constraint at some point, to allow namespace owners to admin their own secrets if possible

spiffxp (approved):
/approve
/lgtm
/hold
If we need to resolve comments, and to remove when ready to deploy

gcsweb.k8s.io/deployment.yaml
thockin: If the YAML and the `-n` disagree it fails (last I tried).  I'm happy to cross that bridge later

gcsweb.k8s.io/README.md
thockin: Yeah, happy to revisit.  The main thinking was to defend .k8s.io certs.

I think we should NOT document this as a group leaf app-operators should join.

ameukam: I'll update the docs with a following PR.

PR #1844, opened 2021-03-30
Setup sippy ns, add spiffxp to debug

spiffxp:
I think there's a mismatch between group name and ns for rbac.

I had commented earlier that the convention to follow is naming the namespace/group after the app in question, instead of the community group responsible for it.

Let's start down that path (and add my non-privileged account to the existing group for troubleshooting)

### DIFF
--- a/groups/groups.yaml
+++ b/groups/groups.yaml
@@ -192,6 +192,7 @@ groups:
       - k8s-infra-rbac-prow@kubernetes.io
       - k8s-infra-rbac-publishing-bot@kubernetes.io
       - k8s-infra-rbac-triageparty-release@kubernetes.io
+      - k8s-infra-rbac-sippy@kubernetes.io
       - k8s-infra-rbac-slack-infra@kubernetes.io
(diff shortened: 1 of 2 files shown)


spiffxp:
Applied namespace changes
```
spiffxp@cloudshell:~/k8s.io/infra/gcp/namespaces (kubernetes-public)$ ./ensure-namespaces.sh gke_kubernetes-public_us-central1_aaa
```

spiffxp:
Yeah, mismatch between group name and namespace is the problem.  This group doesn't exist.
```
$ k describe rolebindings -n wg-reliability-sippy namespace-user
Name:         namespace-user
Labels:       <none>
Annotations:  <none>
Role:
  Kind:  Role
  Name:  namespace-user
Subjects:
  Kind   Name                                               Namespace
  ----   ----                                               ---------
  Group  k8s-infra-rbac-wg-reliability-sippy@kubernetes.io
```

I would prefer to correct and rectify with the `sippy` NS I just created

spiffxp:

PR #1846, opened 2021-03-30
groups: cleanup old wg-reliability group/ns

spiffxp:
the rbac group didn't match the ns, and I had asked for this to be
changed to "sippy" anyway; these are no longer needed now that sippy
exists

followup to https://github.com/kubernetes/k8s.io/pull/1844


spiffxp:

PR #1862, opened 2021-04-01
Add initial groups reconciliation for sig-cluster-lifecycle

randomvariable:
Creates new replacement group for lost Google Group for testgrid/prow
for Kubernetes Cluster API.

Signed-off-by: Naadir Jeewa <jeewan@vmware.com>

/hold

As I'm expecting to add a few more folk.


randomvariable:
Ah, I see,. Need the group created in the kubernetes.io gsuite.

Issue #1838, opened 2021-03-27
REMINDER: clean up thockin-test certs mess after June 26, 2021


thockin:
Cert expires Fri, 25 Jun 2021 16:43:40 UTC.  I want to make sure it refreshes properly.

* Delete namespace `thockin-test-certs` and make sure ingress LBs are cleaned up
* Delete static IPs `thockin-test-v{4,6}`
* Delete DNS names (A and AAAA) `thockin-test{1,2,3}`

thockin:
It looks like these did refresh properly, so I am going to clean up early.

thockin:
Done

PR #1877, opened 2021-04-06
Add initial terraform manifests for monitoring

rikatz:
Add initial manifest for gitops monitoring.

Some points of attention for this PR:

* This is not repeatable. Although I think adding too much modules for terraform can turn things more complicated, maybe we can define how and what we monitor in URLs. Assuming we always use SSL/TLS we could create a module and have this simpler in a future, like:
```
module "thockin_test1_monitoring" {
  source = "../../../modules/monitoring"
  project_name       = "kubernetes-public"
  url       = "thockin-test1.k8s.io"
  cert_expiration_threshold = 15
  // Channel needs to be manually created in GCP interface with the following display name
  channel_notification = "Kubernetes.io Cert Alert"
}
```
* We should probably want to define the structure for this. I'm proposing inside infra/monitoring just because got no better idea :D
* We can add more parameters, and more notification channel/options (like, we can add a list of emails that can turn into Email Notification Channel on GCP) 
* I'm using all the GCP locations, but getting only metrics from usa-iowa as valid to cert_expiration. This can be changed later as well, depending on what we want to do (like, properly monitoring and alerting in case of some downtime/increased latency)

### DIFF
--- a/infra/gcp/monitoring/k8s.io/00-provider.tf
+++ b/infra/gcp/monitoring/k8s.io/00-provider.tf
@@ -1,0 +1,23 @@
+
+/*
+This file defines:
+- Required provider versions
+- Storage backend details
+*/
+
+terraform {
+
+  backend "gcs" {
+    bucket = "k8s-infra-clusters-terraform"
+    prefix = "monitoring/state" // $project_name/$cluster_name
+  }
+
+  required_providers {
+    google = {
+      source  = "hashicorp/google"
+      version = "~> 3.63.0"
+    }
+  }
+}
+
+
--- a/infra/gcp/monitoring/k8s.io/10-thockin-k8s-io.tf
+++ b/infra/gcp/monitoring/k8s.io/10-thockin-k8s-io.tf
@@ -1,0 +1,75 @@
+locals {
+  project_id       = "kubernetes-public"
+  monitored_domain = "thockin-test1.k8s.io"
+}
+
+// Needed to import the notification channel
+provider "google" {
+  project = local.project_id
+}
+
+// Manual step: Create a StackDriver alert channel pointing to a channel in Slack
+// It will select the channel here by its display name
+data "google_monitoring_notification_channel" "alertchannel" {
+  display_name = "Kubernetes.io Cert Alert"
+}
+
+
+// We can turn this into a module and then add then standardize the resource display names
+resource "google_monitoring_uptime_check_config" "uptime_check" {
+  display_name = "${local.monitored_domain} https"
+  timeout      = "5s"
+  period       = "300s"
+
+  http_check {
+    path         = "/"
+    port         = "443"
+    use_ssl      = true
+    validate_ssl = true
+  }
+
+  monitored_resource {
+    type = "uptime_url"
+    labels = {
+      project_id = local.project_id
+      // Host to be verified (1)
+      host = local.monitored_domain
+    }
+  }
+}
+
+
+resource "google_monitoring_alert_policy" "cert_expiration_alert" {
+  combiner              = "OR"
+  display_name          = "${local.monitored_domain} certificate monitor"
+  enabled               = true
+  notification_channels = [data.google_monitoring_notification_channel.alertchannel.name]
+  project               = local.project_id
+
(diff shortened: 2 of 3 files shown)


infra/gcp/monitoring/k8s.io/versions.tf
```diff
@@ -0,0 +1,8 @@
+/*
+This file defines:
+- Required Terraform version
+*/
+
+terraform {
+  required_version = "~> 0.14.0"
```
ameukam, line 7: We still use the 0.13 version

infra/gcp/monitoring/k8s.io/10-thockin-k8s-io.tf
```diff
@@ -0,0 +1,75 @@
...
+
+// Needed to import the notification channel
+provider "google" {
+  project = local.project_id
+}
+
+// Manual step: Create a StackDriver alert channel pointing to a channel in Slack
+// It will select the channel here by its display name
+data "google_monitoring_notification_channel" "alertchannel" {
+  display_name = "Kubernetes.io Cert Alert"
```
ameukam, line 14: We can use the existing channel : `#k8s-infra-alerts`. Can you mention the type ?

rikatz:
@ameukam thanks!

I've changed the channel searcher to use the channel name (label) and the type slack, and did changed the terraform to >= 0.13.0 :)

rikatz:
@ameukam anything else on this PR? I wanted to check if monitoring is working :)

infra/gcp/monitoring/k8s.io/00-provider.tf
```diff
@@ -0,0 +1,23 @@
...
+/*
+This file defines:
+- Required provider versions
+- Storage backend details
+*/
+
+terraform {
+
+  backend "gcs" {
+    bucket = "k8s-infra-clusters-terraform"
```
ameukam, line 11: @rikatz the terraform states are now in different buckets : https://github.com/kubernetes/k8s.io/pull/1952
Feel free to add a new one.

infra/gcp/monitoring/k8s.io/00-provider.tf
```diff
@@ -0,0 +1,23 @@
...
+    prefix = "monitoring/state" // $project_name/$cluster_name
+  }
+
+  required_providers {
+    google = {
+      source  = "hashicorp/google"
+      version = "~> 3.63.0"
```
ameukam, line 18:
```suggestion
      version = "~> 3.66.0"
```

infra/gcp/monitoring/k8s.io/versions.tf
```diff
@@ -0,0 +1,8 @@
+/*
+This file defines:
+- Required Terraform version
+*/
+
+terraform {
+  required_version = ">= 0.13.0"
```
ameukam, line 7:
```suggestion
  required_version = ">= 0.14.0"
```
See: https://github.com/kubernetes/k8s.io/pull/2019
Sorry for make you revert it. :-D

ameukam:
@rikatz I dropped some comments.

spiffxp:
I have a meta-level nit which is that right now I think @ameukam has preferred to keep all of our terraform stuff in infra/gcp/clusters 

I feel like we need to chat a bit about what we want organization of files to look like, but until then I think having them all homed under one directory (even if its name is misleading) makes more sense than allowing sprawl

So move to something like `infra/gcp/clusters/projects/kubernetes-public/monitoring`

spiffxp:
I'm ok with merge and iterate if we immediately follow up with that move

ameukam:
@rikatz What do you think about closing this PR now the domains are gone : https://github.com/kubernetes/k8s.io/issues/1838 ?

infra/gcp/monitoring/k8s.io/versions.tf
sftim: This will need a rebase / update I think now, should be:
```suggestion
  required_version = "~> 1.0.0"
```

rikatz:
Let's do this :)
/close

PR #1878, opened 2021-04-06
groups: try posting to some staging groups

spiffxp:
I would like to see if it's possible to treat the staging groups as
mailing lists for announcements pertinent to each project.

Since we use these as ACLs I'm hoping to avoid "someone needs to be a member
of all groups".

I don't want to set "ANYONE_CAN_POST" because that's a spam magnet unless
moderation is setup, which is a burden I wouldn't want to place on these
groups.


hasheddan (approved):
/lgtm

spiffxp:
Looks like membership in a group does not mean I'm in the organization domain.  I could post a message to a group I'm a member of, but not to a group I'm not a member of.

Issue #1882, opened 2021-04-08
CRI-O artifact staging


saschagrunert:
We require a staging location for image and binary artifacts, which is modifiable by us (without a retention policy).

I'll request a staging project for that purpose.

cc @haircommander @mrunalp

spiffxp:
... so actually, cri-o is not actually part of kubernetes anymore (as of kubernetes/community#3668).

It's now a CNCF incubating project: https://cri-o.io/

If it's not part of kubernetes, I don't think k8s-infra is the right place to host its artifacts.

https://github.com/kubernetes/k8s.io/issues/915 - we still need to document our policies on what is/not fair game for a staging project

saschagrunert:
Thank you for the clarification, I'll check if I can request a project directly from the CNCF
/close

PR #1859, opened 2021-03-31
Refresh infra scripts to address audit followups, reduce output noise

spiffxp:
I've run these a few times locally and the results are showing up in audit PRs, so I should probably share what I'm working on.

Will update this description to link to issues these fixes resolve once I've taken care of the WIP commits

Should allow us to address https://github.com/kubernetes/k8s.io/issues/1675
- Running `K8S_INFRA_ENSURE_ONLY_SERVICES_WILL_FORCE_DISABLE=true ./infra/gcp/ensure-staging-storage.sh` should disable all of the unexpected services.  I plan on doing a first pass without that set to get a log of which projects would have which services disabled first.

Fixes https://github.com/kubernetes/k8s.io/issues/299
- Any new projects going forward will have the `user:foo roles/owner` binding removed immediately after the project is created.  I plan on running ./infra/gcp/*.sh to ensure the `roles/owner` binding is removed for all existing projects.

See individual commits for details

PR #1883, opened 2021-04-08
Add cri-o staging project

saschagrunert:
Per https://github.com/kubernetes/k8s.io/tree/main/k8s.gcr.io#creating-staging-repos

Fixes https://github.com/kubernetes/k8s.io/issues/1882

cc @mrunalp @haircommander

### DIFF
--- a/groups/groups.yaml
+++ b/groups/groups.yaml
@@ -1630,3 +1630,14 @@ groups:
       - natashasarkar@google.com
       - extaeger@google.com
       - samwronski@google.com
+
+  - email-id: k8s-infra-staging-cri-o@kubernetes.io
+    name: k8s-infra-staging-cri-o
+    description: |-
+      ACL for staging CRI-O artifacts
+    settings:
+      ReconcileMembers: "true"
+    members:
+      - mpatel@redhat.com
+      - pehunt@redhat.com
+      - saschagrunert@gmail.com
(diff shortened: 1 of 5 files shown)


spiffxp (approved):
/approve
/lgtm

spiffxp:
I'll post back here once I've run scripts, I'm trying to land https://github.com/kubernetes/k8s.io/pull/1859 first

spiffxp:
... so actually, cri-o is not actually part of kubernetes anymore (as of https://github.com/kubernetes/community/pull/3668).

It's now a CNCF incubating project: https://cri-o.io/

If it's not part of kubernetes, I don't think k8s-infra is the right place to host its artifacts.

/assign @thockin @dims @ameukam 
for k8s-infra input
/assign @derekwaynecarr @dawnchen
for sig-node input

I think I'd like to revert this.  WDYT new assignees?

dims:
@saschagrunert the best way is to open a servicedesk ticket with CNCF asking for a GCP project for use by CRI-O with a list of names who are owners. then you will have full control over the resources you need for CRI-O

thockin:
A maybe important clarification:  To serve binary artifacts that are not Kubernetes, you need a project (with a name that is acceptable as part of your image pull path) under a different GCP organization with a different billing account.

I don't like being exclusionary, but the funding we have is explicitly for kubernetes, and we have to draw a line somewhere.

saschagrunert:
Sounds good, thank you again for the clarification.

PR #1884, opened 2021-04-08
Add sig-cluster-lifecycle-cluster-api-alerts google group via SIG Cluster Lifecycle delegated groups management.

### LINKED ISSUE #1863: Request for sig-cluster-lifecycle-cluster-api-alerts@kubernetes.io Google Group
We were using kubernetes-sig-cluster-lifecycle-cluster-api-alerts@googlegroups.com for Testgrid/Prow alerts for Cluster API. Unfortunately, after a Google Workspace migration, all the owners were removed and the group has disappeared into the ether.

Would it be possible to create a Google Group within the kubernetes.io domain, so we can manage it via this repository, e.g. in this example PR https://github.com/kubernetes/k8s.io/pull/1862 where I attempted to do it using a normal googlegroups.com email address?

Thanks.

randomvariable:
Fixes #1863 

Signed-off-by: Naadir Jeewa <jeewan@vmware.com>

/assign @ameukam

### DIFF
--- a/groups/sig-cluster-lifecycle/groups.yaml
+++ b/groups/sig-cluster-lifecycle/groups.yaml
@@ -1,0 +1,15 @@
+groups:
+  - email-id: sig-cluster-lifecycle-cluster-api-alerts@kubernetes.io
+    name: sig-cluster-lifecycle-cluster-api-alerts
+    description: |-
+
+    settings:
+      WhoCanPostMessage: "ANYONE_CAN_POST"
+      ReconcileMembers: "true"
+    owners:
+      - detiber@gmail.com
+      - jdetiberus@equinix.com
+      - jeewan@vmware.com
+      - sbueringer@gmail.com
+    members:
+      - naadir@randomvariable.co.uk
--- a/groups/sig-cluster-lifecycle/OWNERS
+++ b/groups/sig-cluster-lifecycle/OWNERS
@@ -1,0 +1,8 @@
+# See the OWNERS docs at https://go.k8s.io/owners
+
+reviewers:
+  - sig-cluster-lifecycle-leads
+approvers:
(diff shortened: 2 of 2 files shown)


groups/sig-cluster-lifecycle/groups.yaml
```diff
@@ -0,0 +1,15 @@
...
+    description: |-
+
+    settings:
+      WhoCanPostMessage: "ANYONE_CAN_POST"
+      ReconcileMembers: "true"
+    owners:
+      - detiber@gmail.com
+      - jdetiberus@equinix.com
+      - jeewan@vmware.com
+      - sbueringer@gmail.com
```
spiffxp, line 13: I'll merge to allow us to iterate, but we need to stop making people owners of groups unless there are specific reasons to do so.

It's the same reason we disallow GitHub team owners.  If people can make manual changes, they will, and then get confused when our automation reverts their changes.

spiffxp: Where would you want to see such a policy documented?  I plan on using tests to enforce, but that's the friendliest way to find out.

spiffxp (approved):
/approve
/lgtm
But with caveats

groups/sig-cluster-lifecycle/groups.yaml
ameukam: 🤔 We can start with an example of how a mailing list should be created in https://github.com/kubernetes/k8s.io/tree/master/groups and also add a link to https://www.kubernetes.dev/docs/comms/mailing-lists/ (I just found out).

randomvariable: Oh, I was just copying the community one that had some folk in OWNERS. Happy to move everyone to members

PR #1898, opened 2021-04-09
Fix sippy manifests to work on 1.16 cluster

wojtek-t:
/assign @deads2k


spiffxp:

PR #1904, opened 2021-04-09
infra/gcp/prow: use two default ssh keys

spiffxp:
This is followup to https://github.com/kubernetes/k8s.io/pull/1894

Fixes https://github.com/kubernetes/k8s.io/issues/1751

I would rather just use the one, but I no longer have time to untangle whatever is going on between prow, kubetest2, cluster/log-dump.sh and `gcloud ssh` (ref: https://github.com/kubernetes/k8s.io/issues/1751#issuecomment-816835097)

Since the steady state we've converged on is most projects having both keys, make that the default / expected state

As with https://github.com/kubernetes/k8s.io/pull/1894 I've added a [wip] commit to this PR to make running this script faster.  Will remove commit and [wip] from PR once actuated.

PR #1907, opened 2021-04-09
infra/gcp/prow: fix typos

spiffxp:
Good catch by @ameukam https://github.com/kubernetes/k8s.io/pull/1904#pullrequestreview-632743991

Found another one while I was testing


cpanato (commented):
late /lgtm

PR #1909, opened 2021-04-09
Ensure gcs bucket for k8s-infra prow instance

ameukam:
Following prow [documentation](https://github.com/kubernetes/test-infra/blob/master/prow/getting_started_deploy.md#configure-a-gcs-buckethttps://github.com/kubernetes/test-infra/blob/master/prow/getting_started_deploy.md#configure-a-gcs-bucket) guidance :
Create a GCS bucket for tide history and build logs.
Create a service account and grant admin access to the bucket.
Create a service account key and add the generated key to Secret
Manager.

### DIFF
--- a/infra/gcp/ensure-main-project.sh
+++ b/infra/gcp/ensure-main-project.sh
@@ -78,6 +78,12 @@ readonly TERRAFORM_STATE_BUCKET_ENTRIES=(
     k8s-infra-tf-sandbox-ii:k8s-infra-ii-coop@kubernetes.io
 )
 
+
+#GCS buckets for k8s-infra-prow
+readonly PROW_BUCKETS=(
+    k8s-infra-prow-results
+)
+
 # The services we explicitly want enabled for the main project
 #
 # NOTE: Expected services include dependencies of these services, which may be
@@ -166,6 +172,48 @@ function ensure_terraform_state_buckets() {
             # TODO(spiffxp): figure out a way to do this per-bucket
             ensure_project_role_binding "${project}" "group:${owners}" "roles/viewer"
         ) 2>&1 | indent
+    done
+}
+
+function ensure_prow_buckets() {
+    if [ $# -ne 1 ] || [ -z "$1" ]; then
+        echo "${FUNCNAME[0]}(gcp_project) requires 1 argument" >&2
+        return 1
+    fi
+
+    local project="${1}"
+
+    for bucket in "${PROW_BUCKETS[@]}"; do
+        local svc_acct_name="${bucket}-sa"
+        local svc_acct_email="$(svc_acct_email "${project}" \
+        "${svc_acct_name}")"
+        local SECRET_ID="${svc_acct_name}-key"
+
+        color 6 "Ensuring bucket ${bucket} exists and is only word-readable"
+        ensure_public_gcs_bucket "${project}" "gs://${bucket}"
+
+
+        color 6 "Creating service account: ${svc_acct_name}"
+        ensure_service_account \
+            "${project}" \
+            "${svc_acct_name}" \
+            "${svc_acct_name}"
+
+        color 6 "Empowering service account: ${svc_acct_name}"
+        empower_svcacct_to_write_gcs_bucket "${svc_acct_email}" "gs://${bucket}"
+
+        color 6 "Ensure secret ${SECRET_ID} exists in project ${PROJECT}"
+        ensure_secret "${project}" "${SECRET_ID}"
+
+        color "Ensure ${SECRET_ID} contains secret key for ${svc_acct_name}"
+        ensure_serviceaccount_key_secret "${project}" "${SECRET_ID}" "${svc_acct_email}"
+
+        color 6 "Empowering k8s-infra-prow-oncall@kubernetes.io to read secret ${SECRET_ID}"
+        ensure_secrets_role_binding \
+            "projects/${project}/secrets/${SECRET_ID}" \
+            "group:k8s-infra-prow-oncall@kubernetes.io" \
+            "roles/secretmanager.secretAccessor"
+
     done
 }
 
@@ -378,6 +426,9 @@ function ensure_main_project() {
 
     color 6 "Ensuring terraform state buckets exist with correct permissions in: ${project}"
     ensure_terraform_state_buckets "${project}" 2>&1 | indent
+
+    color 6 "Ensuring prow buckets exist in: ${project}"
+    ensure_prow_buckets "${project}" 2>&1 | indent
 
     color 6 "Empowering cluster users and admins for clusters in: ${project}"
     empower_cluster_admins_and_users "${project}" 2>&1 | indent
--- a/infra/gcp/lib_gsm.sh
+++ b/infra/gcp/lib_gsm.sh
@@ -102,8 +102,13 @@ function ensure_secret_with_admins() {
 #   $2: The secret name (e.g. "my-secret")
 #   $3: The service-account (e.g. "foo@k8s-infra.iam.gserviceaccount.com")
 function ensure_serviceaccount_key_secret() {
+<<<<<<< HEAD
     if [ ! $# -eq 3 ] || [ -z "$1" ] || [ -z "$2" ] || [ -z "$3" ]; then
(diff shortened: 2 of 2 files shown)


ameukam:
/retitle Ensure gcs bucket for k8s-infra prow instance

cpanato (approved):
/lgtm

infra/gcp/ensure-main-project.sh
```diff
@@ -57,6 +57,11 @@ CLUSTER_USERS_GROUP="gke-security-groups@kubernetes.io"
 # The DNS admins group.
 DNS_GROUP="k8s-infra-dns-admins@kubernetes.io"
 
+# Buckets for the logs of prow
+PROW_BUCKETS=(
+    k8s-prow-infra-logs
```
spiffxp, line 62: Assuming this is supposed to be the community equivalent of `kubernetes-jenkins`?
```suggestion
    k8s-infra-prow-results
```

Naming is hard when I overthink things.
- First thought: `k8s-infra-prow-logs` to get the consistent `k8s-infra` prefix
- But...
  - "prow logs" sounds like logs of the prow components (plank, sinker, crier, etc)
  - there are more than "logs" that end up `kubernetes-jenkins`, there are also "artifacts" like junit.xml files, coverage files, metadata, etc.
- OK, so `k8s-infra-prow-artifacts` for "artifacts" produced by prow, including job results
- But...
  - "artifacts" clashes with the sig-release meaning of "binaries and container images that represent a given release of a given subproject"
  - makes me thing it's going to be the binaries/images for prow itself, a.k.a. the equivalent to `gcr.io/k8s-prow`
- OK, so how about just `k8s-infra-prow` and be done with it
- But...
  - If we're doing things over, one of the problems with `kubernetes-jenkins` is that it allows per-object ACLs. They allowed that single bucket to be used for multiple purposes, including occasionally storing secrets or credentials.  
  - We don't want per-object ACLs in kubernetes.io, as they are painful and laborious to maintain, audit and enforce.  We are planning on setting an org policy to explicitly require per-bucket IAM only (aka uniform bucket level access).  
  - Thus we should name this bucket for the main purpose everyone expects of `kubernetes-jenkins`, public read access to things about and created by prow jobs for the kubernetes project.  If other purposes arise, they should result in other buckets named/IAM'ed as such.
- OK, so how about `k8s-infra-prow-results`, as in the "results" of prow job runs
  - I looked around in prow's codebase and docs and couldn't find a consistent term for "things that are uploaded to GCS or S3 by the pod-utilities `sidecar` component and `crier`".  Results isn't just "test results" but from the concept of running a prow job and then looking at the "resulting" files in GCS/S3.
  - I considered "data" but that sounds like "training data" or other metadata that is loaded in to help run a job, where I believe this bucket's intended purpose is storage of things about and created by prow jobs

WDYT about `k8s-infra-prow-results`?

infra/gcp/ensure-main-project.sh
```diff
@@ -79,6 +84,40 @@ ensure_private_gcs_bucket \
     "${PROJECT}" \
     "gs://${CLUSTER_TERRAFORM_BUCKET}"
 
+
+color 6 "Ensuring all the prow buckets exist"
```
spiffxp, line 88: /hold
The rest of this I would like to wait to merge until I land a refactor of this file (https://github.com/kubernetes/k8s.io/pull/1974)

I am happy to take up redoing this part to fit in with the refactor as followup, or leaving it to you to ensure someone other than me understands the structure I tried to put in place

spiffxp (changes requested):

infra/gcp/ensure-main-project.sh
ameukam: > "artifacts" clashes with the sig-release meaning of "binaries and container images that represent a given release of a given subproject"

The buckets are prefixed (`k8s-infra-prow`) with a reference. I don't think we should have confusion with other projects. 

Let's go with `k8s-infra-prow-results`, it's more aligned with the needs of this instance.

infra/gcp/ensure-main-project.sh
ameukam: @spiffxp made a refactor now #1974 is merged.

infra/gcp/ensure-main-project.sh
```diff
@@ -167,6 +173,48 @@ function ensure_terraform_state_buckets() {
...
+function ensure_prow_buckets() {
+    if [ $# -ne 1 ] || [ -z "$1" ]; then
+        echo "${FUNCNAME[0]}(gcp_project) requires 1 argument" >&2
+        return 1
+    fi
+
+    local project="${1}"
+
+    for bucket in "${PROW_BUCKETS[@]}"; do
+        local svc_acct_name="${bucket}-sa"
```
spiffxp, line 185: I think you are going to want to reuse the existing service account(s) that prow jobs run as instead of creating a new service account

spiffxp (changes requested):
I'm not sure provisioning new service accounts is required here. For the existing prow build clusters their service accounts are setup via terraform

infra/gcp/ensure-main-project.sh
spiffxp: So tl;dr I would create a `k8s-infra-prow@kubernetes-public` service account for the prow control plane, since it may require more than just write access to this one bucket.

Since you're doing this for the cluster in kubernetes-public, you could do all this in https://github.com/kubernetes/k8s.io/blob/main/infra/gcp/clusters/projects/kubernetes-public/aaa

But no objections to iterating in bash if you need to.

infra/gcp/ensure-main-project.sh
```diff
@@ -167,6 +173,48 @@ function ensure_terraform_state_buckets() {
...
+        ensure_secret "${project}" "${SECRET_ID}"
+
+        color "Ensure ${SECRET_ID} contains secret key for ${svc_acct_name}"
+        ensure_serviceaccount_key_secret "${project}" "${SECRET_ID}" "${svc_acct_email}"
+
+        color 6 "Empowering k8s-infra-prow-oncall@kubernetes.io to read secret ${SECRET_ID}"
+        ensure_secrets_role_binding \
+            "projects/${project}/secrets/${SECRET_ID}" \
+            "group:k8s-infra-prow-oncall@kubernetes.io" \
+            "roles/secretmanager.secretAccessor"
```
spiffxp, line 213: This is basically https://github.com/kubernetes/k8s.io/blob/4d98e1c294c6e6158eca1efb9e6a4f8c3003a572/infra/gcp/ensure-conformance-storage.sh#L104-L131

You could parameterize it on project and pull into lib_iam.sh for re-use.  Can be done as followup though

infra/gcp/ensure-main-project.sh
ameukam: @spiffxp I'll add a second commit with a HCL version.

ameukam: @spiffxp The Terraform version is done in [`03d55ef`](https://github.com/kubernetes/k8s.io/pull/1909/commits/03d55efa685fda4718553202c171b378029223dc).

infra/gcp/lib_gsm.sh
```diff
@@ -102,8 +102,13 @@ function ensure_secret_with_admins() {
 #   $2: The secret name (e.g. "my-secret")
 #   $3: The service-account (e.g. "foo@k8s-infra.iam.gserviceaccount.com")
 function ensure_serviceaccount_key_secret() {
+<<<<<<< HEAD
     if [ ! $# -eq 3 ] || [ -z "$1" ] || [ -z "$2" ] || [ -z "$3" ]; then
         echo "${FUNCNAME[0]}(project, secret, serviceaccountt) requires 3 arguments" >&2
+=======
+    if [ ! $# -eq 3 -o -z "$1" -o -z "$2" -o -z "$3" ]; then
+        echo "ensure_serviceaccount_key_secret(project, secret, serviceaccount) requires 3 arguments" >&2
+>>>>>>> 0b821977 (Add gcs public bucket for k8s-infra-prow logs.)
```
spiffxp, line 111: merge conflict, you want to keep HEAD

spiffxp (changes requested):

spiffxp:
/close
closing this in favor of https://github.com/kubernetes/k8s.io/pull/2181

PR #1913, opened 2021-04-12
Add ii.coop to k8s-infra-gcp-{auditors,accounting}

hh:
These are the @ii folks helping with #k8s-infra-wg


spiffxp (approved):
/approve
/lgtm
For followup: make an ii group

Issue #1900, opened 2021-04-09
Get sippy.k8s.io up and running


spiffxp:
/assign @deads2k @wojtek-t 
Umbrella issue to link relevant PRs to

PRS related to getting sippy up and running that I'm aware of so far
- https://github.com/kubernetes/k8s.io/pull/1899
- https://github.com/kubernetes/k8s.io/pull/1898
- https://github.com/kubernetes/k8s.io/pull/1846
- https://github.com/kubernetes/k8s.io/pull/1844
- https://github.com/kubernetes/k8s.io/pull/1615
- https://github.com/kubernetes/k8s.io/pull/1614

/wg reliability
/area cluster-infra
/priority important-soon
/milestone v1.22

spiffxp:
I think the reason sippy is returning 500's is because testgrid is returning 500's

```
$ k logs -n sippy -l app=sippy -c sippy
Refresh complete
Refreshing data
Error loading job details for ci-kubernetes-unit: could not read local data file /data/https---testgrid.k8s.io-sig-release-master-blocking-table-show-stale-tests=&tab=ci-kubernetes-unit: open /data/https---testgrid.k8s.io-sig-release-master-blocking-table-show-stale-tests=&tab=ci-kubernetes-unit: no such file or directory
Refresh complete
```

```
$ k logs -n sippy -l app=sippy -c fetchdata --since=1h --tail=-1
Fetching new testgrid data
Job gce-cos-master-alpha-features has bad status PASSING
Job kind-ipv6-master-parallel has bad status PASSING
Job node-kubelet-master has bad status PASSING
Job Conformance - GCE - master - kubetest2 has bad status PASSING
Job gce-cos-master-reboot has bad status PASSING
Job skew-cluster-latest-kubectl-stable1-gce has bad status PASSING
Job build-master-fast has bad status PASSING
Job conformance-ga-only has bad status PASSING
Job gce-cos-master-default has bad status FLAKY
Job gce-cos-master-scalability-100 has bad status PASSING
Job gce-device-plugin-gpu-master has bad status FLAKY
Job gce-ubuntu-master-containerd has bad status FLAKY
Job integration-master has bad status PASSING
Job build-master has bad status PASSING
Job ci-kubernetes-unit has bad status PASSING
response dump
HTTP/2.0 500 Internal Server Error
Content-Length: 323
Content-Type: text/html; charset=UTF-8
Date: Wed, 30 Jun 2021 18:03:48 GMT
[...]

PR #1915, opened 2021-04-12
Fix sippy ingress by changing underlying service to NodePort

wojtek-t:
Ref #1900

Currently ingress reconciliation is failing with
```
2m13s       Normal    Sync        ingress/sippy   Scheduled for sync
61s         Warning   Translate   ingress/sippy   Translation failed: invalid ingress spec: service "sippy/sippy" is type "ClusterIP", expected "NodePort" or "LoadBalancer"
```

/assign @deads2k

### DIFF
--- a/sippy/sippy/service.yaml
+++ b/sippy/sippy/service.yaml
@@ -13,5 +13,5 @@ spec:
   selector:
     app: sippy
   sessionAffinity: None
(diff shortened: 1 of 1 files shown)


ameukam:
/lgtm
/hold
Remove the hold when you're ready to deploy.

Issue #1528, opened 2021-01-13
Migrate away from google.com gcp project authenticated-image-pulling


spiffxp:
Part of umbrella issue to migrate away from google.com gcp projects: https://github.com/kubernetes/k8s.io/issues/1469

Part of umbrella to migrate kubernetes e2e test images/registries to community-owned infrastructure: https://github.com/kubernetes/k8s.io/issues/1458

This project is referenced in kubernetes/kubernetes (ref: https://cs.k8s.io/?q=authenticated-image-pulling&i=nope&files=&repos=)

At a minimum it appears to host:
- the image-pulling@authenticated-image-pulling.iam.gserviceaccount.com service account (hardcoded in some tests)
- the gcr.io/authenticate-image-pulling registry

We should:
- determine if we want to keep this test as-is or rewrite to be more portable
- if as-is
  - identify or create an equivalent CNCF-owned GCP project
  - migrate tests to use new project
  - announce deprecation window
  - flag project for deletion post-deprecation window

/wg k8s-infra
/sig testing
/sig release
/area release-eng
/sig node

spiffxp:
/sig windows
since @jeremyje pinged about needing a more recent version of windows-nanoserver in gcr.io/authenticated-image-pulling

spiffxp:
/remove-lifecycle stale
/milestone v1.23
Would like to push https://github.com/kubernetes/k8s.io/issues/1458 over the line in v1.23

ameukam:
/remove-lifecycle rotten
We need to make about this during 1.26. Migration to Artifact Registry will happen during 1.26 milestone.

/milestone v1.26

ameukam:
/remove-lifecycle rotten
We need to make about this during 1.26. Migration to Artifact Registry will happen during 1.26 milestone.

/milestone v1.26

spiffxp:
kubernetes e2e tests that use `GcAuthenticatedRegistry` use this (ref: https://github.com/kubernetes/kubernetes/blob/38bc7131b37f8299809d8e61fdc8a776ebf40bdc/test/utils/image/manifest.go#L125)

So tests that use `AuthenticatedAlpine` or `AuthenticatedWindowsNanoServer` (ref: https://github.com/kubernetes/kubernetes/blob/38bc7131b37f8299809d8e61fdc8a776ebf40bdc/test/utils/image/manifest.go#L236-L237)

[...]

Issue #1459, opened 2020-12-02
Setup community-owned equivalent to gcr.io/k8s-authenticated-test registry


spiffxp:
Part of umbrella to kubernetes e2e test images/registries to community-owned infrastructure: https://github.com/kubernetes/k8s.io/issues/1458

The registry is used by the following kubernetes e2e tests:
- [sig-apps] ReplicaSet should serve a basic image on each replica with a private image
- [sig-apps] ReplicationController should serve a basic image on each replica with a private image

The k8s-authenticated-test project was accidentally earlier today, and has caused these tests to fail (ref: ref: https://github.com/kubernetes/kubernetes/issues/97002#issuecomment-737435131)

We should:
- ensure the project is restored
- determine what permissions are setup on that project
- determine how the tests get credentials needed to use the private registry
- determine whether k8s.gcr.io will allow us to setup private paths (I doubt this)
- if not, setup a new project e.g. k8s-infra-private-test
- if we can re-use same credentials, set those up
- if we can't, determine whether we'll need to cherry pick changes back to older release branches

spiffxp:
For reference, here's the output of `gsutil iam get gs://artifacts.k8s-authenticated-test.appspot.com/`
```json
{
  "bindings": [
    {
      "members": [
        "projectEditor:k8s-authenticated-test",
        "projectOwner:k8s-authenticated-test"
      ],
      "role": "roles/storage.legacyBucketOwner"
    },
    {
      "members": [
        "allAuthenticatedUsers",
        "projectViewer:k8s-authenticated-test"
      ],
      "role": "roles/storage.legacyBucketReader"
    }
  ],
  "etag": "CAk="
}
```

To keep the behavior of these tests as-is using a new registry, the key part is `allAuthenticatedUsers` instead of `allUsers`.

That said, I question whether we should keep these tests at all, ref: https://github.com/kubernetes/kubernetes/issues/97026#issuecomment-738500525

spiffxp:
/miletone v1.21
/sig apps
test owner
/sig node
I think this is a more appropriate test owner for this functionality

spiffxp:
/remove-lifecycle rotten
/milestone v1.23
Push to close https://github.com/kubernetes/k8s.io/issues/1458 for v1.23

BenTheElder:
copying from https://github.com/kubernetes/kubernetes/issues/113925#issuecomment-2304834317

[...]

PR #1929, opened 2021-04-14
tests: Adds e2e-test-auth-img project

claudiubelu:
The ``gcr.io/k8s-staging-e2e-test-auth-img`` and ``k8s.gcr.io/kubernetes-e2e-test-auth-img`` registries are meant to be private, to be used in a few E2E test scenarios, and it is meant to replace the ``gcr.io/authenticated-image-pulling``.

An auth token with read-only access will have to be generated afterwards, and then baked into the E2E tests.

Related: https://github.com/kubernetes/k8s.io/issues/1459
Related: https://github.com/kubernetes/k8s.io/issues/1528

### DIFF
--- a/groups/sig-testing/groups.yaml
+++ b/groups/sig-testing/groups.yaml
@@ -47,6 +47,20 @@ groups:
     - jgrafton@google.com
     - skuznets@redhat.com
 
+  - email-id: k8s-infra-staging-e2e-test-auth-img@kubernetes.io
+    name: k8s-infra-staging-e2e-test-auth-img
+    description: |-
+      Group that owns pushing rights to the gcr.io/k8s-staging-e2e-test-auth-img staging registry
+    settings:
+      ReconcileMembers: "true"
+    members:
+      - bentheelder@google.com
+      - davanum@gmail.com
+      - linusa@google.com
+      - mkumatag@in.ibm.com
+      - spiffxp@google.com
+      - spiffxp@gmail.com
+
   - email-id: k8s-infra-staging-e2e-test-images@kubernetes.io
     name: k8s-infra-staging-e2e-test-images
     description: |-
--- a/infra/gcp/infra.yaml
+++ b/infra/gcp/infra.yaml
@@ -280,6 +280,7 @@ infra:
       k8s-staging-csi-secrets-store:
       k8s-staging-descheduler:
       k8s-staging-dns:
+      k8s-staging-e2e-test-auth-img:
       k8s-staging-e2e-test-images:
       k8s-staging-etcd:
       k8s-staging-etcdadm:
--- a/k8s.gcr.io/images/k8s-staging-e2e-test-auth-img/OWNERS
+++ b/k8s.gcr.io/images/k8s-staging-e2e-test-auth-img/OWNERS
@@ -1,0 +1,16 @@
+# See the OWNERS docs at https://go.k8s.io/owners
+
+approvers:
+- BenTheElder
+- dims
+- listx
+- mkumatag
+- spiffxp
+reviewers:
+- BenTheElder
+- dims
+- listx
+- mkumatag
+- spiffxp
+labels:
+- sig/testing
--- a/k8s.gcr.io/images/k8s-staging-e2e-test-auth-img/images.yaml
+++ b/k8s.gcr.io/images/k8s-staging-e2e-test-auth-img/images.yaml
@@ -1,0 +1,1 @@
+# No images yet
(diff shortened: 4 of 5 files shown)


ameukam:
@claudiubelu Please add a comment in the `images.yaml` file as suggested: https://github.com/kubernetes/k8s.io/tree/main/k8s.gcr.io#creating-staging-repos.

claudiubelu:
> 
> 
> @claudiubelu Please add a comment in the `images.yaml` file as suggested: https://github.com/kubernetes/k8s.io/tree/main/k8s.gcr.io#creating-staging-repos.

Ty!

spiffxp:
I have this sneaking suspicion we can't get away with `k8s.gcr.io/kubernetes-e2e-test-auth-img` being private, while the rest of `k8s.gcr.io` is public

spiffxp:
I also feel like "should the tests that use this even exist?" remains unanswered: https://github.com/kubernetes/kubernetes/issues/97026#issuecomment-738500525

infra/gcp/ensure-staging-storage.sh
```diff
@@ -126,6 +127,10 @@ readonly STAGING_PROJECT_SERVICES=(
     storage-component.googleapis.com
 )
 
+PRIVATE_PROJECTS=(
+    e2e-test-auth-img
+)
+
```
spiffxp, line 133: I'm not sure this is a pattern we want to encourage more of?

spiffxp (changes requested):
>An auth token with read-only access will have to be generated afterwards

Can you describe that a little bit more?  I feel like that should be part of this PR.  I suspect we can get away with either copying or refactoring what we use in ensure-conformance.sh to at least ensure the service account that will be authed is present:

https://github.com/kubernetes/k8s.io/blob/9e2220ce246c59efeca43c4d45be08aa7874f0b1/infra/gcp/ensure-conformance-storage.sh#L104-L131

spiffxp:
/cc @thockin @listx @hh @saschagrunert @jeremyrickard 
Given that we can't use `k8s.gcr.io` for this, I have a couple questions/thoughts:
- should we be thinking of promoting to a private repo, or should we just use a one-off private repo? if the latter, should we drop the "k8s-staging" from the name for this one-off repo?
- should we be thinking about cross-cloud private artifacts for registry.k8s.io? or do we just decide to take the hit and have the world hit this one private repo in GCR whenever they run their kubernetes e2e tests?

hh:
> should we drop the "k8s-staging" from the name for this one-off repo?
If we stay grc only for e2e tests yes, as we'll need a source to promote from. (CIP would likely be updated to push not just to prod, but other registries as well)

If we try to cross-cloud here, how many types of private registry logins do we test? Testing only authenticating to google may not provide the cover we need. Possibly testing pulling from two different styles of authentication?

spiffxp:
/reopen
This was closed in error

ameukam:
/lgtm
gentle ping @BenTheElder @spiffxp

groups/sig-testing/groups.yaml
```diff
@@ -47,6 +47,20 @@ groups:
     - jgrafton@google.com
     - skuznets@redhat.com
 
+  - email-id: k8s-infra-staging-e2e-test-auth-img@kubernetes.io
+    name: k8s-infra-staging-e2e-test-auth-img
+    description: |-
+      Group that owns pushing rights to the gcr.io/k8s-staging-e2e-test-auth-img staging registry
```
BenTheElder, line 53: possibly should be aligned with the OWNERS file?

BenTheElder (commented):
conceptually seems OK but not comfortable approving the whole thing myself given prior debate about how this should be done / named / ...

groups/sig-testing/groups.yaml
claudiubelu: Done

claudiubelu:
So, the question remains: can this be made private? If so, how? It seems that kubernetes/k8s.io changed since this was initially proposed, so this just simply creates a new project, it's no longer trying to make it private at the moment.

BenTheElder:
This was always done on a separate registry, and it was always a GCR registry.

I don't think the purpose of the test is to test cloud providers at all, just the flow of an image that requires credentials.

I don't see a reasonable better alternative proposed here, other than perhaps giving up on the test ...

I'm going to drop a link in the SIG slack, I think this PR deserved a follow-up answering @claudiubelu's questions.

BenTheElder:
> should we be thinking of promoting to a private repo, or should we just use a one-off private repo? if the latter, should we drop the "k8s-staging" from the name for this one-off repo?

IMO: yes. we should give it a one-off, more descriptive name if we do this.

> should we be thinking about cross-cloud private artifacts for registry.k8s.io? or do we just decide to take the hit and have the world hit this one private repo in GCR whenever they run their kubernetes e2e tests?

Proposing one alternative (maybe) ... can we rely on running a simple authenticated registry *in-cluster* as part of the test? And just not serve a public endpoint for this purpose at all. We could probably bake up a docker image that is an authenticated registry with some image pre-pushed to it, the question is mostly of if we can run an in-cluster registry sufficiently portably for test purposes. I think the answer might be "no, we can't" (TLS? CRI <-> pod reachability?), but perhaps worth exploring.

PR #1941, opened 2021-04-16
Remove deprecated interpolation syntax.

ameukam:
Since Terraform 0.12, interpolation syntax is deprecated.

```
Terraform 0.11 and earlier required all non-constant expressions to be
provided via interpolation syntax, but this pattern is now deprecated. To
silence this warning, remove the "${ sequence from the start and the }"
sequence from the end of this expression, leaving just the inner expression.
```

The change will not impact GKE Cluster aaa.


rikatz:
/lgtm

Thanks!

PR #1947, opened 2021-04-19
groups/sig-cluster-lifecycle: add kubeadm test alerts ML

neolit123:
Add a new mailing list to serve as the location to send
kubeadm e2e test failures to.

fixes https://github.com/kubernetes/kubeadm/issues/2451

/sig cluster-lifecycle
/kind feature


spiffxp (approved):
/approve
/lgtm

Issue #1914, opened 2021-04-12
Cloud provider ASNs store


BobyMCbobs:
I'd like to create a folder in this repo to manage the known ASNs for each CNCF-supported cloud-provider. This will mean for query-able IP blocks that will end up being used with a 302-redirect or split-horizon DNS for {artifacts,registry}.k8s.io.

Examples:
```yaml
#infra/meta/asns/microsoft.yaml
name: microsoft
redirectsTo: 
  registry: kubernetes.acr.io
  artifacts: k8s-artifacts.acr.io
asns:                                                                                                                               
- 12076                                                                                                                           
- 8075                                                                                                                            
- 8068                                                                                                                            
- 8069                                                                                                                            
- 16550
```
```yaml
#infra/meta/asns/google.yaml
name: google
redirectsTo:
  registry: kubernetes.gcr.io
  artifacts: kubernetes-release.storage.googleapis.com                                                                                                                        
asns:                                                                                                                               
- 39190                                                                                                                           
- 139070                                                                                                                          
- 45566                                                                                                                           
- 15169                                                                                                                           
- 19527                                                                                                                           
- 36040                                                                                                                           
- 43515                                                                                                                           
- 16550
```
```yaml
#infra/meta/asns/amazon.yaml
name: amazon
redirectsTo:
  registry: kubernetes.ecr.io
  artifacts: kubernetes.s3.amazon.com
asns:
  - 16509
```
(so forth...)

It would be lovely to also have a representative from each cloud-provider to :ballot_box_with_check: verify these ASNs represent their networks.

hh:
I wonder if we could prioritise vendors from this list?
https://docs.google.com/spreadsheets/d/1uF9BoDzzisHSQemXHIKegMhuythuq_GL3N1mlUUK2h0/edit#gid=0

ameukam:
@BobyMCbobs Feel free to open a PR about this. I would suggest you to centralize those ASNs in a dedicated folder (maybe `registry.k8s.io`) at the base of this repo.

BobyMCbobs:
Vendor ASN PRs:
- Google: https://github.com/kubernetes/k8s.io/pull/1949
- Microsoft: https://github.com/kubernetes/k8s.io/pull/1950
- Amazon: https://github.com/kubernetes/k8s.io/pull/1951

ameukam:
> Vendor ASN PRs:
> 
>     * Google: #1949
> 
>     * Microsoft: #1950
> 
[...]

PR #1951, opened 2021-04-20
Add infra/meta/asns/amazon.yaml

BobyMCbobs:
Umbrella issue: #1914


BobyMCbobs:
I believe that the domains are wrong for both artifacts and container registry

PR #1952, opened 2021-04-20
setup separate gcs buckets for different sets of terraform resources

spiffxp:
Allow groups less privileged than k8s-infra-gcp-org-admins to use terraform to manage resources over which they have ownership.

Terraform state can include potentially include sensitive values. Since we have terraform setup to store state in GCS, we need to ensure visibility and access to state matches ownership of (privileges to modify) the resources it describes.

We're using uniform bucket-level access on our GCS buckets to avoid the complexity introduced by per-object ACLs. This means if we want different groups with different privilege levels using terraform to manage different sets of resources, we need to provision a GCS bucket for each group.

The new bucket schema is `k8s-infra-tf-{folder}[-{suffix}]` where:
- `{folder}` is the intended GCP folder for GCP projects managed by this group, access level should be ~owners of folder
- `{suffix}` is subset of resources contained somewhere underneath folder, access level should  ~editors of those resources
- `k8s-infra-tf-`: is 14 chars, leaving 49 before max length for bucket name

So for example I was tempted to create `gs://k8s-infra-tf-public` and move `aaa` cluster state there. But we may want to use terraform to manage a whole bunch of other resources beside clusters in `kubernetes-public`, and we might not want to grant `cluster-admins` the privileges to acccidentally delete everything in `kubernetes-public` (e.g. DNS)

The GCP folders don't actually exist yet, but the plan is:
- `public`: kubernetes-public (potentially release related projects too)
- `prow`: prow-build clusters and e2e projects
- `aws`: if there are gcp projects being used to manage aws resources
- `sandbox`: temporary projects

The buckets being added are:
- `k8s-infra-tf-aws`: to manage aws resources
- `k8s-infra-tf-prow-clusters`: to manage prow-build, prow-build-trusted
- `k8s-infra-tf-public-clusters`: to manage aaa
- `k8s-infra-tf-sandbox-ii`: for the ii team to manage things in sandbox

Organization admins are given `storage.admin` privileges to all `k8s-infra-tf-` buckets for break-glass purposes.

Terraform modules were migrated by running `terraform init` and `terraform refresh` for each module

/cc @ameukam @thockin

### DIFF
--- a/infra/gcp/clusters/projects/k8s-infra-ii-sandbox/provider.tf
+++ b/infra/gcp/clusters/projects/k8s-infra-ii-sandbox/provider.tf
@@ -7,7 +7,9 @@ This file defines:
 terraform {
 
   backend "gcs" {
-    bucket = "k8s-infra-clusters-terraform"
+    bucket = "k8s-infra-tf-sandbox-ii"
+    // TODO(spiffxp): the names not matching weirds me out a bit, it would be
+    //                nice to rename the project at some point
     prefix = "k8s-infra-ii-sandbox"
   }
 
--- a/infra/gcp/clusters/projects/k8s-infra-prow-build-trusted/prow-build-trusted/00-provider.tf
+++ b/infra/gcp/clusters/projects/k8s-infra-prow-build-trusted/prow-build-trusted/00-provider.tf
@@ -7,7 +7,7 @@ This file defines:
 terraform {
 
   backend "gcs" {
-    bucket = "k8s-infra-clusters-terraform"
+    bucket = "k8s-infra-tf-prow-clusters"
     prefix = "k8s-infra-prow-build-trusted/prow-build-trusted" // $project_name/$cluster_name
   }
 
--- a/infra/gcp/clusters/projects/k8s-infra-prow-build/prow-build/00-provider.tf
+++ b/infra/gcp/clusters/projects/k8s-infra-prow-build/prow-build/00-provider.tf
@@ -7,7 +7,7 @@ This file defines:
 terraform {
 
   backend "gcs" {
-    bucket = "k8s-infra-clusters-terraform"
+    bucket = "k8s-infra-tf-prow-clusters"
     prefix = "k8s-infra-prow-build/prow-build" // $project_name/$cluster_name
   }
 
--- a/infra/gcp/clusters/projects/kubernetes-public/aaa/00-inputs.tf
+++ b/infra/gcp/clusters/projects/kubernetes-public/aaa/00-inputs.tf
@@ -10,13 +10,17 @@ terraform {
   required_version = ">= 0.12.8"
 
   backend "gcs" {
-    bucket = "k8s-infra-clusters-terraform"
+    bucket = "k8s-infra-tf-public-clusters"
     prefix = "kubernetes-public/aaa" // $project_name/$cluster_name
   }
 
   required_providers {
-    google      = "~> 2.14"
-    google-beta = "~> 2.14"
+    google      = {
+      version = "~> 2.14"
+    }
+    google-beta = {
+      version = "~> 2.14"
+    }
   }
 }
 
--- a/infra/gcp/ensure-main-project.sh
+++ b/infra/gcp/ensure-main-project.sh
@@ -49,7 +49,7 @@ CLUSTER_ADMINS_GROUP="k8s-infra-cluster-admins@kubernetes.io"
 ACCOUNTING_GROUP="k8s-infra-gcp-accounting@kubernetes.io"
 
 # The GCS bucket which hold terraform state for clusters
-CLUSTER_TERRAFORM_BUCKET="k8s-infra-clusters-terraform"
+LEGACY_CLUSTER_TERRAFORM_BUCKET="k8s-infra-clusters-terraform"
 
 # The GKE security groups group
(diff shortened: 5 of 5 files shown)


hack/migrate-tf-buckets.sh
```diff
@@ -0,0 +1,30 @@
+#!/usr/bin/env bash
+
+# one-off script to migrate terraforms state files to the appropriate buckets
+set -x
+
+function tfstate_cp() {
```
ameukam, line 6: This function has a similar logic in Terraform. Once you change the bucket in the provider field, TF will copy the state to this new destination. An example: https://www.terraform.io/docs/cloud/migrate/index.html#step-6-run-terraform-init-to-migrate-the-workspace.

spiffxp: I'll remove this script from the PR and update terraform then, once a few in-flight PRs have landed

Still using this as a test run to ensure we org admins can view the different buckets.  Still having problems copying aws resources over atm due to use of hmac keys

spiffxp:
In running `terraform apply` to ensure bucket changes took, I discovered/applied some undeployed changes from https://github.com/kubernetes/k8s.io/pull/1737 for k8s-infra-prow-build/prow-build

```
An execution plan has been generated and is shown below.
Resource actions are indicated with the following symbols:
  - destroy
-/+ destroy and then create replacement

Terraform will perform the following actions:

  # module.project.google_project_iam_custom_role.service_account_lister will be destroyed
  - resource "google_project_iam_custom_role" "service_account_lister" {
      - deleted     = false -> null
      - description = "Can list ServiceAccounts." -> null
      - id          = "projects/k8s-infra-prow-build/roles/ServiceAccountLister" -> null
      - name        = "projects/k8s-infra-prow-build/roles/ServiceAccountLister" -> null
      - permissions = [
          - "iam.serviceAccounts.list",
        ] -> null
      - project     = "k8s-infra-prow-build" -> null
      - role_id     = "ServiceAccountLister" -> null
      - stage       = "GA" -> null
      - title       = "Service Account Lister" -> null
    }

  # module.project.google_project_iam_member.cluster_admins_as_service_account_lister must be replaced
-/+ resource "google_project_iam_member" "cluster_admins_as_service_account_lister" {
      ~ etag    = "BwWsUJQxxEw=" -> (known after apply)
      ~ id      = "k8s-infra-prow-build/projects/k8s-infra-prow-build/roles/ServiceAccountLister/group:k8s-infra-cluster-admins@kubernetes.io" -> (known after apply)
        member  = "group:k8s-infra-cluster-admins@kubernetes.io"
        project = "k8s-infra-prow-build"
      ~ role    = "projects/k8s-infra-prow-build/roles/ServiceAccountLister" -> "organizations/758905017065/roles/iam.serviceAccountLister" # forces replacement
    }

Plan: 1 to add, 0 to change, 2 to destroy.
```

spiffxp:
Bumping into this while trying to `terraform plan` for `k8s-infra-prow-build-trusted/prow-build-trusted`
```
Error: Provider configuration not present

To work with
module.project.google_project_iam_custom_role.service_account_lister its
original provider configuration at provider["registry.terraform.io/-/google"]
is required, but it has been removed. This occurs when a provider
configuration is removed while objects created by that provider still exist in
the state. Re-add the provider configuration to destroy
module.project.google_project_iam_custom_role.service_account_lister, after
which you can remove the provider configuration again.
```

spiffxp:
Bumped into https://github.com/kubernetes/k8s.io/issues/2000 while trying to `terraform init` for `kubernetes-public`
```
An execution plan has been generated and is shown below.
Resource actions are indicated with the following symbols:
  + create
  ~ update in-place

Terraform will perform the following actions:

  # google_project_iam_binding.readonlymonitoringbinding will be updated in-place
  ~ resource "google_project_iam_binding" "readonlymonitoringbinding" {
        etag    = "BwXAZra+ioU="
        id      = "kubernetes-public/roles/monitoring.viewer"
      ~ members = [
            "group:gke-security-groups@kubernetes.io",
          - "serviceAccount:k8s-infra-monitoring-viewer@kubernetes-public.iam.gserviceaccount.com",
        ]
        project = "kubernetes-public"
        role    = "roles/monitoring.viewer"
    }

  # google_project_iam_member.cluster_node_sa_monitoring_viewer will be created
  + resource "google_project_iam_member" "cluster_node_sa_monitoring_viewer" {
      + etag    = (known after apply)
      + id      = (known after apply)
      + member  = "serviceAccount:gke-nodes-aaa@kubernetes-public.iam.gserviceaccount.com"
      + project = "kubernetes-public"
      + role    = "roles/monitoring.viewer"
    }

Plan: 1 to add, 1 to change, 0 to destroy
```

Went ahead and ran `terraform apply` to deploy these changes, and bumped into
```
Error: Provider produced inconsistent result after apply

When applying changes to
google_project_iam_member.cluster_node_sa_monitoring_viewer, provider
"registry.terraform.io/hashicorp/google" produced an unexpected new value:
Root resource was present, but now absent.

This is a bug in the provider, which should be reported in the provider's own
issue tracker.
```

ameukam:
@spiffxp It's possible we may no longer need this resource. GKE Monitoring changed a lot since `aaa` creation. Also a bump of the terraform provider could fix this. I plan to do it this week.

spiffxp:
Related to refactoring infra/gcp, ref: https://github.com/kubernetes/k8s.io/issues/516

Issue #1945, opened 2021-04-19
Requesting access to GCP & GCS for ii.coop


Riaankl:
# Reason for the issue:

We created the PR's
- Create group for access to gcs logs #1871
- Add ii.coop to k8s-infra-gcp-{auditors,accounting} #1913

We are still unable to access the logs.
In K8s Infra meeting on 14 April 2021 Arnaud Meukam offered to help unblock.

## The purpose of the access is to determine:
1. artifacts logs per artifact and location
2. Originating IP for artifact requests.

 With local testing we found the user need these permissions to access these logs
- storage.buckets.get
- storage.objects.get
- storage.objects.list
- logging.viewer

## Questions:
- What else needs to be done from ii's side to successfully get access to the logs?
- What is the best practice for the analysis of these logs? (BigQuery, DataStudio, SQL)

ameukam:
@Riiankl For better transparency, can you mention why you need access to those logs ? Thanks!

Riaankl:
As part of the K8s Infra WG's mandate to move K8s Infra to community own infra, ii will investigate the use of artifacts by volume of traffic and ASN data related to the traffic.
CNCF & Google legal agreed on the terms and the ii team agreed to all the legalities around it.

Riaankl:
Access seems to work as expected.
Thank you @ameukam

Issue #1730, opened 2021-02-26
audit followup: restore k8s-infra-gcr-auditor service account


spiffxp:
I accidentally the account, noticed in https://github.com/kubernetes/k8s.io/pull/1718#discussion_r582441736

We need to:
- restore the account?
- or create a new one and rotate service account keys wherever necessary
- ensure everything that depended on this account is... back up? for whatever that looked like before?

/sig release
/release eng
/priority important-soon
AFAIK nobody has noticed this is down yet.  I remember Tim mentioning something about silencing slack alerts a while back due to general nosie level
/assign @spiffxp 
I did the oops, I feel pretty responsible for addressing it.  That said, this might be a good opportunity for someone from @kubernetes/release-engineering to refresh knowledge on how this is supposed to be wired up and why.

spiffxp:
What is supposed to create this?
- The name but not the project is declared here: https://github.com/kubernetes/k8s.io/blob/9216eda1065496c8a51a9e215a857122296695e2/infra/gcp/lib.sh#L51
- The account is created for a given project by `empower_artifact_auditor(project)` here: https://github.com/kubernetes/k8s.io/blob/9216eda1065496c8a51a9e215a857122296695e2/infra/gcp/lib.sh#L399-L404
  - This block should be changed to a call to `ensure_service_account(project, name,display_name)` from `lib_iam.sh`
- It looks like `ensure-prod-storage.sh` should be creating this account: https://github.com/kubernetes/k8s.io/blob/163acfdb52366228acbc425fccb384da4ae138c9/infra/gcp/ensure-prod-storage.sh#L323
  - But this script crashed right after "Empowering artifact-admins to release prod auditor" which is the line before
- So it's probably a matter of swapping the order so `empower_artifact_auditor` is called before `empower_group_to_admin_artifact_auditor`

spiffxp:
Opened https://github.com/kubernetes/k8s.io/pull/1998 to hopefully get us as far as creating the service account in question

Rotating credentials and getting things up and running would be the remainder of the work here

spiffxp:
Deployed and the service account now exists again https://github.com/kubernetes/k8s.io/pull/1998#issuecomment-832256462

spiffxp:
I think I started on the next step of restoring this by running `infra/gcp/ensure-env-cip-auditor.sh` while deploying https://github.com/kubernetes/k8s.io/pull/2016
```
$ ./ensure-env-cip-auditor.sh
Enabling services
[...]

Issue #904, opened 2020-05-27
Enable GCS logs


thockin:
At least for gs://k8s-artifacts-prod, it would be nice to break down the details.

https://cloud.google.com/storage/docs/access-logs

spiffxp:
Going to repurpose this as umbrella issue for:
- setting up group restricting access to these logs: https://github.com/kubernetes/k8s.io/pull/1871
- setting up IAM such that only this group has access to the logs (https://github.com/kubernetes/k8s.io/pull/1881 tried creating a custom role but I don't think this is necessary)
- enabling the logs

spiffxp:
I'm guessing gs://k8s-artifacts-gcslogs is what was used when we trialed access logs and then disabled them
```console
$ gsutil ls -p k8s-artifacts-prod
gs://asia.artifacts.k8s-artifacts-prod.appspot.com/
gs://eu.artifacts.k8s-artifacts-prod.appspot.com/
gs://k8s-artifacts-cni/
gs://k8s-artifacts-cri-tools/
gs://k8s-artifacts-csi/
gs://k8s-artifacts-gcslogs/
gs://k8s-artifacts-kind/
gs://k8s-artifacts-prod/
gs://us.artifacts.k8s-artifacts-prod.appspot.com/
```

It's currently empty
```console
$ gsutil ls gs://k8s-artifacts-gcslogs
$
```

I can't find anything in git history that shows this bucket being created via our scripts

spiffxp:
Let's look at what IAM is today

For the project containing the bucket
```yaml
$ gcloud projects get-iam-policy k8s-artifacts-prod --format="yaml(bindings)"
bindings:
- members:
  - serviceAccount:service-388270116193@compute-system.iam.gserviceaccount.com
  role: roles/compute.serviceAgent
- members:
  - serviceAccount:service-388270116193@container-analysis.iam.gserviceaccount.com
  role: roles/containeranalysis.ServiceAgent
- members:
  - group:k8s-infra-artifact-security@kubernetes.io
  - serviceAccount:k8s-infra-gcr-vuln-dashboard@k8s-artifacts-prod.iam.gserviceaccount.com
  role: roles/containeranalysis.occurrences.viewer
- members:
  - serviceAccount:service-388270116193@gcp-sa-containerscanning.iam.gserviceaccount.com
  role: roles/containerscanning.ServiceAgent
- members:
  - serviceAccount:388270116193-compute@developer.gserviceaccount.com
[...]

PR #1966, opened 2021-04-22
Read access to read access k8s-artifacts-gcslogs

ameukam:
Ref: https://github.com/kubernetes/k8s.io/issues/1945

Initial request:
https://groups.google.com/g/kubernetes-wg-k8s-infra/c/Wkw0uyMKSXk/m/QLVIAMZzAAAJ.

### DIFF
--- a/infra/gcp/ensure-prod-storage.sh
+++ b/infra/gcp/ensure-prod-storage.sh
@@ -359,6 +359,12 @@ color 6 "Handling special cases"
             $(svc_acct_email "${GCR_BACKUP_TEST_PRODBAK_PROJECT}" "${PROMOTER_SVCACCT}")
     done
 
+    # Special case: empower k8s-infra-gcs-access-logs@kubernetes.io to read k8s-artifacts-gcslogs
+    # k8s-artifacts-gcslogs receive and store Cloud Audit logs for k8s-artificats-prod.
+    ensure_gcs_role_binding "gs://k8s-artifacts-gcslogs" \
+        "group:k8s-infra-gcs-access-logs@kubernetes.io" \
+        "objectViewer"
+
     color 6 "Ensuring prod promoter vuln scanning svcacct exists"
     ensure_service_account \
         "${PROD_PROJECT}" \


ameukam:
/hold
Need approval from chairs and TLs.
/assign @dims @spiffxp @thockin 
cc @hh @Riaankl

hh:
Thanks @ameukam

infra/gcp/ensure-prod-storage.sh
```diff
@@ -359,6 +359,12 @@ color 6 "Handling special cases"
             $(svc_acct_email "${GCR_BACKUP_TEST_PRODBAK_PROJECT}" "${PROMOTER_SVCACCT}")
     done
 
+    # Special case: empower II Coop folks to have read access to GCS bucket k8s-artifcats-gcs-logs
```
dims, line 362: typo @ameukam ... `k8s-artifacts-gcslogs`

spiffxp:
```suggestion
    # Special case: empower II Coop folks to have read access to GCS bucket k8s-artifacts-gcs-logs
```

spiffxp (commented):
One nit otherwise LGTM

spiffxp:
I still feel like moving the bucket to its own project might better accomplish the access rules laid out in https://github.com/kubernetes/k8s.io/issues/904#issuecomment-816484483 but if this gets us started for now so be it

infra/gcp/ensure-prod-storage.sh
hh: 👍

ameukam: @spiffxp @dims Fixed.

infra/gcp/ensure-prod-storage.sh
```diff
@@ -359,6 +359,12 @@ color 6 "Handling special cases"
             $(svc_acct_email "${GCR_BACKUP_TEST_PRODBAK_PROJECT}" "${PROMOTER_SVCACCT}")
     done
 
+    # Special case: empower k8s-infra-gcs-access-logs@kubernetes.io to read k8s-artifacts-gcslogs
+    # k8s-artifacts-gcslogs receive and store Cloud Audit logs for k8s-artificats-prod.
+    ensure_gcs_role_binding "gs://k8s-artifacts-gcslogs" \
```
thockin, line 364: Is this sufficient?  Don't we also want to import into BQ ?

thockin (commented):
No objections to this alone, but I suspect it will just be the start

infra/gcp/ensure-prod-storage.sh
ameukam: @thockin I prefer to leave them the choice of the analysis tool. I would take care of it if there is a need to use BQ.

hh:
@ii will need bq.

spiffxp (approved):
/approve
/lgtm
I'll merge and deploy this as-is

Can migrate to a different bucket as followup, which I will track under the umbrella issue for GCS access logs: https://github.com/kubernetes/k8s.io/issues/904

spiffxp:
Ran `./infra/gcp/ensure-prod-storage.sh` but it stopped with this error before getting to the new binding

```  ERROR: (gcloud.iam.service-accounts.add-iam-policy-binding) NOT_FOUND: Not found; Gaia id not found for email k8s-infra-gcr-auditor@k8s-artifacts-prod.iam.gserviceaccount.com```

spiffxp:
OK this is https://github.com/kubernetes/k8s.io/issues/1730 rearing its head.  Time to tackle that to unblock us here

spiffxp:
A rerun appears to have successfully deployed these changes: https://github.com/kubernetes/k8s.io/pull/1998#issuecomment-832256462

PR #1969, opened 2021-04-26
Promote Addon Resizer 1.8.12

bskiba:


bskiba:
@krzysied would you approve? Seems I missed promoting images for #3833

PR #1973, opened 2021-04-26
groups: unblock group reconciliation by fixing security groups

spiffxp:
The ci-k8sio-groups and post-k8sio-groups testgrid tabs were showing continuous failure for the past N days

I noticed this when trying to deploy changes that added k8s-infra-ii-coop to IAM bindings, and got back a "group does not exist error"

These commits tell part of the story.  I had to do some manual reconciliation runs to make this work

### DIFF
--- a/groups/committee-product-security/groups.yaml
+++ b/groups/committee-product-security/groups.yaml
@@ -43,7 +43,6 @@ groups:
       - security@giantswarm.io
       - security@gravitational.com
       - security@kinvolk.io
-      - security@kubernetes.io
       - security@loodse.com
       - security@platform9.com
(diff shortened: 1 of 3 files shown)


spiffxp:
Hm.  The groups postsubmit didn't trigger when this merged.  Wonder if the trigger only looks at groups.yaml

spiffxp:
Yup, that was it.  I ran the reconciler manually before I opened this PR so we're all caught up for now.

Opened https://github.com/kubernetes/test-infra/pull/21954 to make sure the postsubmit will trigger for _all_ changes in groups/

PR #1974, opened 2021-04-26
infra/gcp: clean up main project

### LINKED ISSUE #2000: Terraform state is not fully applied
I got when I deployed #1604 for `aaa`:

```shell 
  # google_project_iam_binding.readonlymonitoringbinding will be updated in-place
  ~ resource "google_project_iam_binding" "readonlymonitoringbinding" {
        etag    = "BwXAZra+ioU="
        id      = "kubernetes-public/roles/monitoring.viewer"
      ~ members = [
            "group:gke-security-groups@kubernetes.io",
          - "serviceAccount:k8s-infra-monitoring-viewer@kubernetes-public.iam.gserviceaccount.com",
        ]
        project = "kubernetes-public"
        role    = "roles/monitoring.viewer"
    }

  # google_project_iam_member.cluster_node_sa_monitoring_viewer will be created
  + resource "google_project_iam_member" "cluster_node_sa_monitoring_viewer" {
      + etag    = (known after apply)
      + id      = (known after apply)
      + member  = "serviceAccount:gke-nodes-aaa@kubernetes-public.iam.gserviceaccount.com"
      + project = "kubernetes-public"
      + role    = "roles/monitoring.viewer"
    }
```

Resources introduced in https://github.com/kubernetes/k8s.io/pull/365

spiffxp:
Briefly:
- remove `projectEditor:kubernetes-public` bindings from `k8s-infra-tf-*` buckets
  - this required redoing `ensure_removed_gcs_role_binding` to use `gsutil iam set` instead of `gsutil iam ch`
- fix color output in logs
- remove service accounts from groups and bind them directly to their roles, since we discovered this didn't work for the prow serviceaccounts
- refactor ensure-main-project.sh and ensure-organization.sh into functions for better testing and separation of code/data (it is not a far leap from bash arrays to yaml specs)
  - added new `empower_gke_for_serviceaccount` functions to lib
  - handled a dependency cycle between ensure-main-project and ensure-organization for the `k8s-infra-gcp-auditor` serviceaccount
- prune services specified for kubernetes-public down to what we directly use, and comment why we use them
  - we use `container` which uses `compute` which uses `oslogin`; drop the latter two since we only need to specify the former
  - `bigquery-json` is not a service, use `bigquery` instead
- fix `ensure_serviceaccount_role_binding` either never working or hitting a recent API change

See individual commits for details

Fixes #2000


spiffxp:
I ran `./ensure-main-project.sh` and `./ensure-organization.sh` as of https://github.com/kubernetes/k8s.io/commit/d28844979893b4f4bca1bc7f888cf5cb2b28bc47

I've run these scripts a few times before while working on this PR, so some changes were already picked up in https://github.com/kubernetes/k8s.io/pull/2001

spiffxp:
The most recent commit should address https://github.com/kubernetes/k8s.io/issues/2000

spiffxp:
If you want to see the outcomes of these changes, please take a look at my comments on the latest audit PR https://github.com/kubernetes/k8s.io/pull/2001#pullrequestreview-652536676

dims:
/lgtm

( since "I've run these scripts a few times before while working on this PR" :) )

infra/gcp/ensure-main-project.sh
```diff
@@ -34,249 +34,347 @@ if [ $# != 0 ]; then
...
+
+#
+# main functions
+#
 
-# Merge existing permissions with the ones we need to exist.  We merge
-# permissions because:
-#   * The full list is large and has stuff that is inherited listed in it
-#   * All of our other IAM binding logic calls are additive
+function ensure_terraform_state_buckets() {
```
ameukam, line 149: Non-blocking comment: IMHO we should empower the owners of each bucket with `roles/viewer`so they can easily run `terraform plan`.

spiffxp: Thank you once again for the good catch.  Apparently `roles/viewer` is necessary to get `storage.buckets.list`.

I didn't catch this when testing with my less-privileged account, because it has `audit.viewer` at the org level and thus `roles/viewer` on everything implicitly.

They only other off-the-shelf roles that have this are:
- `roles/storage.admin`
- `roles/viewer`
- `roles/editor`
- `roles/owner`

I'm tempted to suggest `storage.admin`, let me look real quick

spiffxp: So it looks like we can't set any of the primitive roles just on the bucket
```
$ gsutil iam ch "user:spiffxp@google.com:roles/viewer" gs://k8s-infra-clusters-terraform
BadRequestException: 400 Role (roles/viewer) does not exist in the resource's hierarchy.
```

Which leaves use with `roles/viewer` on the project, or `roles/storage.admin` on the bucket as the quick fix here.  I'll add to the project for now

spiffxp: Opened https://github.com/kubernetes/k8s.io/pull/2005 to address

PR #1994, opened 2021-05-03
Promote agnhost 2.32

mtaufen:
This promotes the version of agnhost with the conformance test fix from
https://github.com/kubernetes/kubernetes/pull/101589

/assign @liggitt 

/sig auth


liggitt:
not sure how to verify the sha, but I agree with the title of the PR

mtaufen:
`docker pull gcr.io/k8s-staging-e2e-test-images/agnhost:2.32` is how I got it.

PR #2003, opened 2021-05-05
Increase NAP limits in aaa cluster

mborsz:
https://github.com/kubernetes/k8s.io/pull/1604 incorrectly sets resource limits to 16 cores and 64GiB of memory, while the existing cluster has already 26 cores and 94GiB of memory (and the limits are for the entire clusters, not for a single node as I thought)

/assign @ameukam

### DIFF
--- a/infra/gcp/clusters/projects/kubernetes-public/aaa/10-cluster-configuration.tf
+++ b/infra/gcp/clusters/projects/kubernetes-public/aaa/10-cluster-configuration.tf
@@ -158,7 +158,7 @@ resource "google_container_cluster" "cluster" {
     }
     resource_limits {
       resource_type = "memory"
-      maximum = 500
+      maximum = 256
     }
   }
 
--- a/infra/gcp/clusters/projects/kubernetes-public/aaa/10-cluster-configuration.tf
+++ b/infra/gcp/clusters/projects/kubernetes-public/aaa/10-cluster-configuration.tf
@@ -154,7 +154,7 @@ resource "google_container_cluster" "cluster" {
     resource_limits {
       resource_type = "cpu"
       minimum = 2
-      maximum = 200
+      maximum = 64
     }
     resource_limits {
       resource_type = "memory"
--- a/infra/gcp/clusters/projects/kubernetes-public/aaa/10-cluster-configuration.tf
+++ b/infra/gcp/clusters/projects/kubernetes-public/aaa/10-cluster-configuration.tf
@@ -154,11 +154,11 @@ resource "google_container_cluster" "cluster" {
     resource_limits {
       resource_type = "cpu"
       minimum = 2
-      maximum = 16
+      maximum = 200
     }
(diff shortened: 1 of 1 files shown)


infra/gcp/clusters/projects/kubernetes-public/aaa/10-cluster-configuration.tf
```diff
@@ -154,11 +154,11 @@ resource "google_container_cluster" "cluster" {
     resource_limits {
       resource_type = "cpu"
       minimum = 2
-      maximum = 16
+      maximum = 200
```
ameukam, line 157: is it possible to have :
```suggestion
      maximum = 64
```

infra/gcp/clusters/projects/kubernetes-public/aaa/10-cluster-configuration.tf
```diff
@@ -154,11 +154,11 @@ resource "google_container_cluster" "cluster" {
...
     }
     resource_limits {
       resource_type = "memory"
-      maximum = 64
+      maximum = 500
```
ameukam, line 161: is it possible to have :
```suggestion
      maximum = 256
```

infra/gcp/clusters/projects/kubernetes-public/aaa/10-cluster-configuration.tf
mborsz: Arnaud, the limit is for the entire cluster, not a single machine. Technically we can use 64, but this is relatively close to the cluster size we use today (~26 cores). Why do we want to put this limit low?

ameukam: From my perspective, we increase the limits for only 1 workload. Most of the workloads actually running don't need a lot of CPU. I think we should for the moment double the current limit and adjust again when needed. 
If the limits I suggest is not enough for your workloads I'm happy to remove my suggestions.

mborsz:
Commited both suggestions. The limits should be fine for our workload.

ameukam:
/lgtm
/approve
Thanks!

ameukam:
Running `./terraform apply -target google_container_cluster.cluster`

ameukam:
Change successfully applied. 

We have now a new nodepool :

ameukam:
perf-dash is running on this node. (main reason why we did this change)

```
kubectl get pods -n perfdash -o wide
NAME                       READY   STATUS    RESTARTS   AGE     IP          NODE                                              NOMINATED NODE   READINESS GATES
perfdash-7c8b6c784-4vx2w   1/1     Running   0          4h20m   10.40.4.2   gke-aaa-nap-n1-highmem-8-1epnwu3d-1d17f798-3t49   <none>           <none>
```

cc @spiffxp

mborsz:
Thank you! :)

Issue #2007, opened 2021-05-06
Try to simplify storage permissions


spiffxp:
Followup based on https://github.com/kubernetes/k8s.io/pull/2005

For whatever reason, I am still confused by exactly which permissions are available to whom on which GCS buckets.  I've started consulting [a spreadsheet I put together](https://docs.google.com/spreadsheets/d/1_N5C5MLfoGfVNuNXlE1ylnUDR9sX8Di6ISMonNtN5Ls/edit#gid=281633490) to help figure out what makes sense from a policy perspective.  I've been trying keep [GCS IAM best practices](https://cloud.google.com/storage/docs/access-control/iam#best_practices) in mind, as well as the guidance that [basic roles](https://cloud.google.com/iam/docs/understanding-roles#basic) aren't a great idea for production environments.

Three things currently bug me:
1. whether uniform bucket-level access is enabled at bucket-creation time determines whether convenience roles like `roles/viewer` are granted `storage.objects.get` (ref: https://cloud.google.com/storage/docs/access-control/iam-roles#basic-roles-modifiable)
2. there's no predefined role that grants `storage.buckets.list` on a per-bucket basis except `roles/storage.admin`
3. some projects have buckets that are more sensitive than others (e.g. terraform resources for disparate privilege levels, logs containing PII), and it's unclear how to ensure basic roles are denied permissions to these buckets, while retaining admin/read privileges for other buckets

What I think we can do to address:
1.  Set `constraints/storage.uniformBucketLevelAccess` at the org level (ref: https://cloud.google.com/resource-manager/docs/organization-policy/org-policy-constraints). This wouldn't suddenly wipe out existing per-object ACLs.  The only per-object ACLs I am currently aware of are related to billing data in kubernetes-public, and I'm optimistic they can be removed in favor of the bucket's IAM policy.
2. research if this is feasible or worthwhile by creating a custom `storage.bucketLister` role and trying to apply it on a per-bucket basis (we do this today with `secretmanager.secretLister`)
3. based on 1 and 2, determine if it makes sense to prune convenience bindings like `projectOwner:roles/legacyBucketOwner`, and replace with direct `roles/storage.admin` or a custom storage admin role that remove `storage.buckets.delete`

References:
- https://cloud.google.com/storage/docs/access-control/iam#best_practices
- https://cloud.google.com/storage/docs/access-control/iam#project-level_roles_vs_bucket-level_roles
[...]

Issue #1981, opened 2021-04-29
GCP gcloud bulk-export as an audit trail


bernokl:
Terraformer  to dump files from existing infrastructure to support K8s Infra Migration
[Terraformer](https://github.com/GoogleCloudPlatform/terraformer) can be used to "reverse Terraform" existing infrastructure as part of the Migration of K8s Infrastructure.

To achieve this a person with full access to the relevant infrastructure is needed to run it for us

spiffxp:
Is this not runnable by members of the auditor group? What resources does this support? How is this different than gcloud supporting tf as an output format?

hh:
Looks mostly possible. I tried with DNS and get quite few errors:


```
```
Error when reading or editing DNSResourceDnsRecordSet "k8s-dev/k8s.dev./SOA": googleapi: Error 403: Forbidden, forbidden
WARN: Fail read resource from provider, wait 300ms before retry
```

```
terraformer import google --projects=kubernetes-public --resources=backendServices,bigQuery,firewall,gcs,gke,globalAddresses,globalForwardingRules,healthChecks,iam,instanceTemplates,monitoring,networks,project,pubsub,routes,sslCertificates,targetHttpProxies,targetHttpsProxies,urlMaps
```

**Comment by hh** [comment]
List of supported Resources: https://github.com/GoogleCloudPlatform/terraformer/blob/master/docs/gcp.md

**Comment by ameukam** [comment]
@hh You can check the bulk export with gcloud here : https://cloud.google.com/sdk/gcloud/reference/alpha/resource-config/bulk-export.

**Comment by bernokl** [comment]
Thank you @ameukam  
It looks like I am still missing some permissions?
```
gcloud alpha resource-config bulk-export --resource-format=terraform --organization=840466421052 --path=~/foo/dump1/ 

ERROR: (gcloud.alpha.resource-config.bulk-export) Error executing export:: [error in 'config-connector' version '1.45.0': error exporting asset inventory: error response from exportassets request: googleapi: Error 403: Request denied by Cloud IAM., forbidden

hh:
```
$ config-connector bulk-export --output=./ --organization=840466421052        
error in 'config-connector' version '1.46.0': error exporting asset inventory: error response from exportassets request: googleapi: Error 403: Request denied by Cloud IAM., forbidden
$ gcloud auth list
         Credentialed Accounts
ACTIVE             ACCOUNT
*                  hh@ii.coop
```

hh:
[...]

PR #1975, opened 2021-04-26
Add budget to k8s-infra-ii-sandbox

ameukam:
Add a defined budget to `k8s-infra-ii-sandbox` but also use the project
to experiment GCP budgets.

Ref: https://github.com/kubernetes/k8s.io/issues/1375

Signed-off-by: Arnaud Meukam <ameukam@gmail.com>

PR #2010, opened 2021-05-06
infra/gcp/roles: update custom org roles

spiffxp:
See individual commits for details, but the tl;dr is this should help with:
- https://github.com/kubernetes/k8s.io/pull/1975
- https://github.com/kubernetes/k8s.io/issues/1981#issuecomment-830926733
- part of https://github.com/kubernetes/k8s.io/issues/2007

The groups change also wraps up part of https://github.com/kubernetes/k8s.io/pull/1974, specifically the part about service accounts being bound directly to their roles rather than indirectly through group membership

### DIFF
--- a/groups/groups.yaml
+++ b/groups/groups.yaml
@@ -250,7 +250,6 @@ groups:
       - james@munnelly.eu
       - spiffxp@google.com
       - thockin@google.com
-      - k8s-infra-dns-updater@kubernetes-public.iam.gserviceaccount.com
 
   #
   # Push groups: k8s-infra-push-*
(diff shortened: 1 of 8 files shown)


spiffxp:
/cc @ameukam @hh
I have run `./infra/gcp/ensure-organization.sh` as part of this, so the next audit PR should pick up results

hasheddan (approved):
/lgtm

spiffxp:
https://github.com/kubernetes/k8s.io/pull/2006 confirms all of the changes here, I reviewed it linking back to commits in this PR

PR #2017, opened 2021-05-07
infra/gcp/clusters: bump terraform google provider versions

spiffxp:
We have been pinned at 3.46.0 which was released 2020-11-02 ([3.46.0 changelog]).  The latest is 3.66.1 which was released 2020-04-29 ([3.66.1 changelog]).

Pin to `~> 3.66.0` which will pick up the patch release (and any subsequent patch releases)

While I was here I also removed terraform version constraints that pinned to a specific patch release of terraform. In other words, everything is `~> 0.13.0`

None of these should generate a diff using `terraform plan`.  For extra confidence, I ran `terraform apply` for `k8s-infra-prow-build-trusted/prow-build-trusted`

I'm going to take a quick look at bumping terraform itself next, https://www.terraform.io/upgrade-guides/0-14.html doesnt seem like it has anything that breaks us

[3.46.0 changelog]: https://github.com/hashicorp/terraform-provider-google/blob/master/CHANGELOG.md#3460-november-02-2020
[3.66.1 changelog]: https://github.com/hashicorp/terraform-provider-google/blob/master/CHANGELOG.md#3661-april-29-2021


ameukam:
/lgtm
Already test in #2018. 😂 

> I'm going to take a quick look at bumping terraform itself next, terraform.io/upgrade-guides/0-14.html doesnt seem like it has anything that breaks us
I'm a little concerned by the provider lock mechanism. I'll look into it deeper.

aojea:
👏

spiffxp:

PR #2019, opened 2021-05-07
infra/clusters: use teraform 0.14

spiffxp:
Part of issue to upgrade the version of terraform used by our modules: https://github.com/kubernetes/k8s.io/issues/2020

The upgrade guide looked significantly less scary, given that we aren't doing anything too complex: https://www.terraform.io/upgrade-guides/0-14.html

> You will need to successfully complete a terraform apply at least once under Terraform v0.13 before upgrading an existing configuration to Terraform v0.14.

This was the only thing that jumped out at me, and we have done this by way of deploying the last few PRs

I'm interested in this bump for three reasons:
- to get us one step closer to the latest release
- to be able to use dependency lock files
- to be able to use sensitive values in plan output (e.g. if we wanted something to run plan as a part of a presubmit...)

To verify this:
```
terraform init
terraform plan
# 0 changes expected
```

I have _not_ run `terraform apply` for any of these.  Up to you whether this should happen prior to merge, or post-merge

### DIFF
--- a/infra/gcp/clusters/OWNERS
+++ b/infra/gcp/clusters/OWNERS
@@ -10,4 +10,4 @@ reviewers:
 - thockin
 
 labels:
-- area/terraform
+- area/infra/terraform
--- a/infra/gcp/clusters/modules/gke-cluster/versions.tf
+++ b/infra/gcp/clusters/modules/gke-cluster/versions.tf
@@ -15,7 +15,7 @@
  */
 
 terraform {
-  required_version = "~> 0.13.0"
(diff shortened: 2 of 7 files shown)


ameukam:
Ran `terraform apply -auto-approve -refresh` 
TF states are converted to `0.14.0` 

No changes as expected. `.terraform.lock.hcl` is created locally.

/lgtm

ameukam (approved):

sftim:
Followed by #2217

PR #2043, opened 2021-05-13
Initial version of repo tag dumper

justinsb:
Dumps tags to json for later grepping / import.

Issue #2037

Issue #2037, opened 2021-05-12
Need help to resolve manifest hashes into tag names? 


Riaankl:
For log analysis, we need the names for these images.
And also a way to resolve all names for images in the logs

```
Most downloaded:
39fafc05754f195f134ca11ecdb1c9a691ab0848c697fffeb5a85f900caaf6e1
ee1b208fd75f17fa8aa049033394640e789fb4fdb73e23a19248e585b16e6692
0602e62c0683ecf24585bcc82a301797065e06c34183292046bfb3cb4c30c68a
4ba180b702c8b29f4d24cb24eeb91a26a5924e797a806bbca1b73d26b5fe8b1e
96d50afe6a7da017ce837645fa039a144f4c91c0c664799877bd1f02525b8563
ff98489bfe3dfce426d3f7ef1cfea4e5a53fdf4d7837032563573f442a32943d
8f60c1d0b400a98e9fcefb67c1a00aff0f44346d4f440326696a3f879d7d771a
faa3eb9a46ce716cb5bf0cbc9cee299956dacc94350021ac5ab34798a854e3e7
85b604bcc41af6fc75a72fe80441a116e48db8afc97fb8ed8c768d371abc43d3
d60490a768b56275fa3b9792b7363fcaab1ac0b2de82dd7e440c89dfdcd94b1d
Highest number of downloads:
80d28bedfe5dec59da9ebf8e6260224ac9008ab5c11dbbe16ee3ba3e4439ac2c
c74f8866df097496217c9f15efe8f8d3db05d19d678a02d01cc7eaed520bb136
da86e6ba6ca197bf6bc5e9d900febd906b133eaa4750e6bed647b0fbe50ed43e
67ddbfb20a22d7c0ea0df568069e7ffc42378467402d04f28ecfa244e78c5eb8
357edb12c3fcd174330823335009f29abe0d6d0aae8e84882bca61161e4839a8
96d50afe6a7da017ce837645fa039a144f4c91c0c664799877bd1f02525b8563
ee1b208fd75f17fa8aa049033394640e789fb4fdb73e23a19248e585b16e6692
0602e62c0683ecf24585bcc82a301797065e06c34183292046bfb3cb4c30c68a
bbba60e5ece3d902e14f41aac1bdc5dd010954e9865b0ce2bbf6798d75b4314b
39fafc05754f195f134ca11ecdb1c9a691ab0848c697fffeb5a85f900caaf6e1
```

[...]

Issue #1834, opened 2021-03-25
Umbrella issue: k8s.gcr.io => registry.k8s.io solution


BobyMCbobs:
# Umbrella issue: k8s.gcr.io => registry.k8s.io solution #1834

This markdown is synced from https://hackmd.io/gN-1GeSpSgyNSvmjKSULbg?edit to https://github.com/kubernetes/k8s.io/issues/1834#issue-841372237 manually by @BobyMCBobs

Scope: https://github.com/kubernetes/k8s.io/wiki/New-Registry-url-for-Kubernetes-(registry.k8s.io)

Design Doc: https://docs.google.com/document/d/1yNQ7DaDE5LbDJf9ku82YtlKZK0tcg5Wpk9L72-x2S2k/edit (shared w/ dev@kubernetes.io and SIG mailing list)

Board: https://github.com/orgs/kubernetes/projects/77

DRAFT AIs that need filled turned into tickets:
https://github.com/orgs/kubernetes/projects/77/views/2?filterQuery=is%3Adraft

# What exactly are you doing? (and how?)
- [x] We are setting up an AWS account with an IAM role and s3 buckets in AWS regions where we see a large percentage of source image pull traffic
  - [x] IAM Role + User for write - [k/k8s.io#3568](https://github.com/kubernetes/k8s.io/issues/3568) 
  - [x] List of Regions with 80% traffic - [oci-proxy#39](https://github.com/kubernetes-sigs/oci-proxy/issues/39)
    - [x] Determine which regions should serve the image layers - [oci-proxy/issues#38](https://github.com/kubernetes-sigs/oci-proxy/issues/38)
      - kubernetes/k8s.io#3568
      - kubernetes/k8s.io#3620
      - kubernetes-sigs/promo-tools#533
- [x] Create Buckets per Region Names
  - [x]  Add Terraform for registry-k8s-io s3 buckets [k/k8s.io#PR3605](https://github.com/kubernetes/k8s.io/pull/3605)
  - [x]  Create s3 Buckets in AWS Regions [k/k8s.io#3595](https://github.com/kubernetes/k8s.io/issues/3595)
- [x] We will iterate on a sandbox url (registry.sandbox.k8s.io) for our experiments and ONLY promote things to (registry.k8s.io) when we have complete confidence
  - [x] standup sandbox deployment [oci-proxy#13](https://github.com/kubernetes-sigs/oci-proxy/issues/13)
[...]

PR #2044, opened 2021-05-13
[WIP] Add poc scripts (+data dump) for pulling manifests and matching hashes

stp-ip:
/hold as it contains ~30k data files!

Poc for pulling all manifests and storing them per tag and per repo using manifester.sh, which can be matched to a hash or multiple hashes with the matcher.sh.

Usage for manifester.sh. Note: Heavy operation and will cleanup old data and start from scratch.
```
sh manifester.sh
```

Usage for matcher.sh.
```
sh matcher.sh <layer-hash> <layer-hash>
```


spiffxp:
Seems like this is an alternate take on https://github.com/kubernetes/k8s.io/pull/2043 to solve https://github.com/kubernetes/k8s.io/issues/2037?

I actually don't have a strong preference away from bash anymore (uh oh) but I'm going to guess folks would prefer the golang version

My other hot take is: I don't think we need to check these files into git.  Sure, we check in audit results, but those resources are expected to change over time; we want to be able walk back in time on our audit trail.  I don't see why we would want to walk back in time with these.

What about storing results in GCS and then downloading them for use?

stp-ip:
The PR was only as a way to share this similar to #2043.
Did not know #2043 was implemented and only had the notion of making a mapping possible for hashes to images so I created the bash version.
Expected this to be closed without merge especially without the data files.
The idea seems to be, that the resulting mapping gets pushed into big query for analysis and any interim files would be only temporary.

spiffxp:
/close
Closing this per https://github.com/kubernetes/k8s.io/issues/1834#issuecomment-912426368

Please /reopen if you feel this was done in error

PR #2080, opened 2021-05-21
Convert gcsweb to VPC native LB

thockin:
I did this the scary way.  Apparently there's a safer way when we do others.  Anyway, this much is done already :)

### DIFF
--- a/gcsweb.k8s.io/service.yaml
+++ b/gcsweb.k8s.io/service.yaml
@@ -3,12 +3,12 @@ kind: Service
 metadata:
   name: gcsweb
   namespace: gcsweb
-  labels:
-    app: gcsweb
+  annotations:
+    cloud.google.com/neg: '{"ingress": true}'
 spec:
   selector:
     app: gcsweb
-  type: NodePort
+  type: ClusterIP
   ports:
     - name: http
       port: 80


spiffxp (approved):
>I did this the scary way. Apparently there's a safer way when we do others.

A few non-blocking questions, since this is live:
- why?
- was the scary way applying these changes all once? or were there some steps involved?
- is there a link to the safer way?

spiffxp:
Verified via
```yaml
# $ k --context=aaa get services -n gcsweb -o=yaml \
# | yq -y '.items[] | {annotations: (.metadata.annotations | with_entries(select(.key | contains("cloud.google.com")))), type: .spec.type, status: .status}'
annotations:
  cloud.google.com/neg: '{"ingress": true}'
  cloud.google.com/neg-status: '{"network_endpoint_groups":{"80":"k8s1-ea949c44-gcsweb-gcsweb-80-15e79f37"},"zones":["us-central1-a","us-central1-c","us-central1-f"]}'
type: ClusterIP
status:
  loadBalancer: {}
```

thockin:
> why?

Why make this change: because we have VPC-native and I thought that was required to enable GCP certificates.  It turns out that it is NOT required.

Why I did it the scary way: because it seemed obvious and gcsweb seemed like a low-risk target.  Since then I learned that there's a safer way, which I will pay attention to :)

> was the scary way applying these changes all once? or were there some steps involved?

Step 1: Add annotation and wait for the NEG to be created.
Step 2: Verify the LB was pointing to the NEG
Step 3: Change `type` to "ClusterIP"

> is there a link to the safer way?

It's pending docs review internally :(  As I look at the next one, I will write down the steps I take.

thockin:
The followup to this will be to switch to GCP's certs, to prove we can and
to race with the expiry of the cert on our main sites :)

PR #2112, opened 2021-05-28
Add kep.k8s.io redirect to enhacements

thockin:


dims:
if no number is specified, then it just goes to issues/ url right?

/lgtm
/hold please feel free to unhold when you wish!

thockin:
Correct.  I will canary it before unholding.

PR #2061, opened 2021-05-19
image-builder/capg: add more permissions to the service account

cpanato:
after we merge: https://github.com/kubernetes/k8s.io/pull/2057 and https://github.com/kubernetes/k8s.io/pull/2025

the periodic job ran https://prow.k8s.io/view/gs/kubernetes-jenkins/logs/periodic-image-builder-gcp-all-nightly/1394944554532081664 but we got an error

```
Error creating instance: googleapi: Error 403: Required 'compute.zones.get' permission for 'projects/k8s-staging-cluster-api-gcp/zones/us-central1-a', forbidden
```

So looks like we need to give a bit more permissions to the svcacc in order to work. 
doing a search saw the `role/editor` have the necessary permissions. However, I don't know if this the correct way to enable it.

/assign @spiffxp @ameukam

PR #2124, opened 2021-06-03
capg/image-builder: add extra role to be able to create gcp vms

cpanato:
Add extra role `"roles/iam.serviceAccountUser"` for the service account

more context: https://github.com/kubernetes-sigs/image-builder/pull/625#issuecomment-853808094

/assign @ameukam @spiffxp @dims

### DIFF
--- a/infra/gcp/ensure-staging-storage.sh
+++ b/infra/gcp/ensure-staging-storage.sh
@@ -421,6 +421,7 @@ function staging_special_case__k8s_staging_cluster_api_gcp() {
 
     ensure_services "${STAGING_PROJECT}" compute.googleapis.com
     ensure_project_role_binding "${STAGING_PROJECT}" "serviceAccount:${serviceaccount}" "roles/compute.instanceAdmin.v1"
+    ensure_project_role_binding "${STAGING_PROJECT}" "serviceAccount:${serviceaccount}" "roles/iam.serviceAccountUser"
     ensure_staging_gcb_builder_service_account "cluster-api-gcp" "k8s-infra-prow-build-trusted"
 }
 


ameukam:
Little experimentation done here : https://github.com/kubernetes-sigs/image-builder/pull/625#issuecomment-853808094

ameukam:
Running: `./infra/gcp/ensure-staging-storage.sh`

infra/gcp/ensure-staging-storage.sh
```diff
@@ -421,6 +421,7 @@ function staging_special_case__k8s_staging_cluster_api_gcp() {
 
     ensure_services "${STAGING_PROJECT}" compute.googleapis.com
     ensure_project_role_binding "${STAGING_PROJECT}" "serviceAccount:${serviceaccount}" "roles/compute.instanceAdmin.v1"
+    ensure_project_role_binding "${STAGING_PROJECT}" "serviceAccount:${serviceaccount}" "roles/iam.serviceAccountUser"
```
spiffxp, line 424: @cpanato @ameukam  So the issue I have with this is that it allows this service account to act as _any_ other service account in this project. That's privilege escalation waiting to happen. Can we not constrain this binding to the actual service account that needs to be used?

spiffxp (commented):
Post merge review

infra/gcp/ensure-staging-storage.sh
spiffxp: This is the same comment I had here https://github.com/kubernetes/k8s.io/pull/2061#discussion_r635436384

ameukam: @spiffxp we could bind the service account to the default SA for the compute service.

```shell
gcloud iam service-accounts list --project k8s-staging-cluster-api-gcp | grep 'compute'
Compute Engine default service account                                         606075400249-compute@developer.gserviceaccount.com                               False
```

I'll investigate this week-end.

cpanato: sorry for the delay to reply, i tried to be off this weekend

How can we fix that? I will try to create the same environment again and apply other roles to check.
but the comment from @ameukam is something that we can do?

spiffxp: Basically it would help to know what service account packer is trying to use, so we could apply this role just to that service account

cpanato: I thought the service account it is using is the one we set in the Job Config : https://github.com/kubernetes/test-infra/blob/master/config/jobs/kubernetes/wg-k8s-infra/trusted/image-builder/image-builder-periodics.yaml#L12

it is available in the container and packer uses that one.
This is my understanding, now I don't know if that is correct. :/

spiffxp: The service account that the job (and thus packer) runs as is what we set in the job config, correct.  However packer's instructions [1] indicate that whatever is running packer needs the `Service Account User` role, meaning they expect packer to be able to impersonate a different service account [2]

I'm going to take a wild guess that this is to allow packer to attach the service account it's currently running as to any other instances it happens to create [4].  So probably narrow the binding to itself.  I would hope that logs or the code could confirm the guess and allow us to narrow the binding's scope.

References:
[1]: https://www.packer.io/docs/builders/googlecompute#running-on-google-cloud
[2]: https://cloud.google.com/iam/docs/impersonating-service-accounts
[3]: https://cloud.google.com/iam/docs/impersonating-service-accounts#attaching-to-resources
[4]: https://cloud.google.com/compute/docs/access/service-accounts#associating_a_service_account_to_an_instance

ameukam:
@spiffxp made a revert : https://github.com/kubernetes/k8s.io/pull/2147.

PR #2134, opened 2021-06-04
images: add k8s-infra image

spiffxp:
This is intended to be used for running scripts and tests in CI, but may also serve as a development environment if someone were to mount things in appropriately.

I'm trying an approach of not copy-pasting the same set of default values around, so until we need variants, the Dockerfile is the place to change them.

Setup a `make cloudbuild` target that doesn't push `:latest` for dev/test purposes

Verified by running `make -C images/k8s-infra build run TAG=spiffxp WHAT=...`

- ./audit/audit-gcp.sh
- ./audit/check-audit-pr.sh
- ./audit/create-or-update-audit-pr.sh
- ./hack/verify.sh
- ./hack/verify-terraform.sh
- ./infra/gcp/ensure-organization.sh
- ./infra/gcp/prow/ensure-e2e-projects.sh

Verified cloudbuild.yaml via `make -C images/k8s-infra cloudbuild`

### DIFF
--- a/images/k8s-infra/Dockerfile
+++ b/images/k8s-infra/Dockerfile
@@ -1,0 +1,157 @@
+# Copyright 2021 The Kubernetes Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# Intended to hold all dependencies needed to build, test and run all tools
+# used by wg-k8s-infra. With sufficient credentials mounted in, should be
+# capable of running those tools to test and deploy all kubernetes project
+# infrastructure managed by wg-k8s-infra.
+
+# base image
+FROM debian:buster
+
+# build args
+
+# unless otherwise stated, versions were arbitrarily chosen to whatever was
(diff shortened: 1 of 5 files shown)


images/k8s-infra/Dockerfile
```diff
@@ -0,0 +1,149 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# Intended to hold all dependencies needed to build, test and run all tools
+# used by wg-k8s-infra. With sufficient credentials mounted in, should be
+# capable of running those tools to test and deploy all kubernetes project
+# infrastructure managed by wg-k8s-infra.
+
+# base image
+FROM debian:buster
```
ameukam, line 21: Why not use `golang:1.16` or `golang:1.16.4` ? both images are based on it.

images/k8s-infra/Dockerfile
```diff
@@ -0,0 +1,149 @@
...
+
+# build args
+ARG CONFTEST_VERSION=0.25.0
+ARG GCLOUD_CHANNEL=rapid
+ARG GO_VERSION=1.16.4
+ARG JQ_VERSION=1.6
+ARG K8S_VERSION=1.20.0
```
ameukam, line 28: Any issue with 1.21.0 or 1.21.z?

spiffxp: Good point, no reason it needs to be .0

I think I chose 1.20 because our clusters are still at 1.18 and I wanted to stay within 2 versions.  I'll add a comment explaining that.

spiffxp: (I ended up double-checking and moved back to 1.19.x because official version skew is +/- 1 minor version)

spiffxp:
/remove-area infra/auditing
/sig testing
/hold
for merge of job that will build this image https://github.com/kubernetes/test-infra/pull/22458

spiffxp:
/hold cancel
Now that https://github.com/kubernetes/test-infra/pull/22458 has merged

PR #2133, opened 2021-06-04
audit: service export changes

spiffxp:
Audit changes based on last audit review:
- don't export logging metrics if they're empty
- don't export logging logs for k8s-infra-e2e-* projects due to
  churn/noise

Newly handled services:
- cloudfunctions: functions
- cloudresourcemanager: org-policies (project-level only)
- pubsub: topics, subscriptions

Newly ignored services (due to lack of gcloud-based resources to export
unless otherwise noted):
- bigquerystorage
- cloudshell
- datastore (ignored until permissions are sorted out)
- iam (ignored until we export workload-identity-pools
- iamcredentials
- oslogin
- serviceusage

Other changes:
- add comment with purpose, requirements, usage
- add support for K8S_INFRA_AUDIT_SERVICES for selective service export
- add TODOs for remaining unhandled services
- break up into sections: config, utils, main
- format file output on same indent level as gcloud/bq command

PR #2138, opened 2021-06-05
roles: add orgpolicy to audit.viewer

spiffxp:
Sigh.  I failed to test out org policy export in https://github.com/kubernetes/k8s.io/pull/2133 and now the audit job is broken.

This picked up some other permissions while refreshing the role, which I opted not to split out into a separate commit out of laziness


spiffxp:
Ran using image from https://github.com/kubernetes/k8s.io/pull/2134

`make -C images/k8s-infra run WHAT=./infra/gcp/ensure-organization.sh`

PR #2139, opened 2021-06-06
k8s.gcr.io/images/k8s-staging-kube-state-metrics: Promote all v2.1.0 images

mrueg:
Apologies looks like my previous commit was incomplete and the staging images were not fully uploaded yet when I run the command. 

Fixes: https://github.com/kubernetes/kube-state-metrics/issues/1498

/assign @lilic


lilic (commented):
hah when I tested it worked fine, but was the staging arch that just happened to work :D 

/lgtm

mrueg:
@lilic yes same happened here as well. :D  

tide says it need an approved label as well for some reason. Can you approve additionally? Thanks!

PR #2143, opened 2021-06-07
infra/gcp/main: enable cloudasset service

spiffxp:
The audit service account is in this project, and the long term intent
is to use cloud asset inventory (or a  tool that consumes it) to more
efficiently export / audit our GCP resources

This should unblock the audit job


spiffxp:
I have already run `./infra/gcp/ensure-main-project.sh` to enable this service, so as to unblock the audit job

PR #2153, opened 2021-06-07
prow: add deploy.sh scripts

spiffxp:
These are basically extracts of the existing embedded bash in the
post-k8sio-*prow*deploy jobs, refactored into functions with some
standard boilerplate. Before attempting to do any further refactoring,
get these in place so the jobs will use them.


spiffxp:
/hold
just since it happens to be open, I'm going to use this to test out presubmits once https://github.com/kubernetes/test-infra/pull/22463 merges

spiffxp:
/hold cancel
Done using this to test out random presubmits

PR #2154, opened 2021-06-07
images/k8s-infra: include _PULL_BASE_REF

spiffxp:
one more try at unblocking https://github.com/kubernetes/test-infra/pull/22463

image-builder is always going to pass this substitution along, but we
are using the default substitutionOptions setting of MUST_MATCH to
catch typos or unused variables as early as possible, which causes
cloudbuild to error unless we use _PULL_BASE_REF somewhere in the build
template


spiffxp:
/cc @cblecker @ameukam 
third time's a charm

PR #2148, opened 2021-06-07
prow-build-trusted: add kubernetes-external-secrets

spiffxp:
Followup to https://github.com/kubernetes/k8s.io/pull/2078

In support of https://github.com/kubernetes/test-infra/pull/22298 and https://github.com/kubernetes/test-infra/pull/22293

PR #2156, opened 2021-06-08
roles: add custom org role container.deployer

spiffxp:
The intent is to bind this to prow-deployer@k8s-infra-prow-build-trusted to give prow the ability to CRUD all resources in a given GKE cluster without granting it permission to CRUD the cluster itself.

Motivated by issues deploying kubernetes-external-secrets via post-k8sio-deploy-prow-build-trusted-resources: https://github.com/kubernetes/k8s.io/pull/2148#issuecomment-856385138

See individual commits for details


BenTheElder (approved):
👍 
/hold

infra/gcp/roles/audit.viewer.yaml
```diff
@@ -610,6 +612,12 @@ includedPermissions:
   - consumerprocurement.entitlements.list
   - consumerprocurement.freeTrials.list
   - consumerprocurement.orders.list
+  - contactcenterinsights.analyses.list
+  - contactcenterinsights.conversations.list
+  - contactcenterinsights.issueModels.list
+  - contactcenterinsights.issues.list
+  - contactcenterinsights.operations.list
+  - contactcenterinsights.phraseMatchers.list
```
ameukam, line 620: @spiffxp out of curiosity, why do we need this ? Do we have a GCP API use this as a dependency ?

spiffxp:
Ran 'make -C images/k8s-infra TAG=latest run ./infra/gcp/ensure-organization.sh'

infra/gcp/roles/audit.viewer.yaml
ameukam: I'm happy with the current state of laziness just want to make sure I understand correctly.  
Looking at https://github.com/darkbitio/gcp-iam-role-permissions/blob/master/roles/viewer, I think we should stick at custom roles.

PR #2181, opened 2021-06-10
infra/gcp/kubernetes-public/aaa : Add GCS bucket for k8s-infra-prow

ameukam:
Create a GCS bucket and ensure member of
k8s-infra-prow-oncall@kubernetes have admin acess to it.

Create GCP SA for prow control-plane and ensure it has write access to
the bucket.

### DIFF
--- a/infra/gcp/clusters/projects/kubernetes-public/aaa/k8s-infra-prow.tf
+++ b/infra/gcp/clusters/projects/kubernetes-public/aaa/k8s-infra-prow.tf
@@ -1,0 +1,110 @@
+/*
+This file defines:
+- a bucket for k8s-infra-prow
+- a GCP service account for the bucket
+- IAM bindings for the bucket
+- a Secret Manager secret for the service account key
+*/
+
+locals {
+  bucket_name               = "k8s-infra-prow-results"
+  bucket_location           = "us-central1"
+  prow_service_account      = "k8s-infra-prow"
+  test_pods_namespace       = "k8s-infra-prow-pods"
+  test_pods_service_account = "default"
+}
+
+
+// Create GCP Service Account for prow control plane
+resource "google_service_account" "k8s_infra_prow" {
+  project      = data.google_project.project.project_id
+  account_id   = local.prow_service_account
+  display_name = local.prow_service_account
+}
+
+// Create a key for GCP Service Account k8s-infra-prow
+resource "google_service_account_key" "k8s_infra_prow" {
+  service_account_id = google_service_account.k8s_infra_prow.name
+  private_key_type   = "TYPE_GOOGLE_CREDENTIALS_FILE"
+}
+
+// Allow pods using the build cluster KSA to use the GCP SA k8s-infra-prow via workload identity
+resource "google_service_account_iam_member" "prow_build_cluster_sa_iam" {
+  role               = "roles/iam.workloadIdentityUser"
+  service_account_id = google_service_account.k8s_infra_prow.name
+  member             = format("serviceAccount:%s.svc.id.goog[%s/%s]", "k8s-infra-prow-build", local.test_pods_namespace, local.test_pods_service_account)
+}
+
+// Allow pods using the build cluster KSA to use the GCP SA k8s-infra-prow via workload identity
+resource "google_service_account_iam_member" "prow_build_trusted_cluster_sa_iam" {
+  role               = "roles/iam.workloadIdentityUser"
+  service_account_id = google_service_account.k8s_infra_prow.name
+  member             = format("serviceAccount:%s.svc.id.goog[%s/%s]", "k8s-infra-prow-build-trusted", local.test_pods_namespace, local.test_pods_service_account)
+}
+
+// Create a GCS bucket for ProwJobs logs and tide history
+resource "google_storage_bucket" "k8s_infra_prow_bucket" {
+  name          = local.bucket_name
+  project       = data.google_project.project.project_id
+  storage_class = "REGIONAL"
+  location      = local.bucket_location
+
+  lifecycle_rule {
+    condition {
+      age = 90
+    }
+    action {
+      type = "Delete"
+    }
+  }
+
+  uniform_bucket_level_access = true
+}
+
+// Allow GCP SA k8s-infra-prow admin for the objects in the bucket k8s-infra-prow-results
+resource "google_storage_bucket_iam_member" "k8s_infra_prow_admin" {
+  bucket = google_storage_bucket.k8s_infra_prow_bucket.name
+  role   = "roles/storage.objectAdmin"
+  member = "serviceAccount:${google_service_account.k8s_infra_prow.email}"
+}
+
+// Allow GCP SA k8s-infra-prow admin (legacy role) for the objects in the bucket k8s-infra-prow-results
+resource "google_storage_bucket_iam_member" "k8s_infra_prow_admin_legacy" {
+  bucket = google_storage_bucket.k8s_infra_prow_bucket.name
+  role   = "roles/storage.legacyBucketWriter"
+  member = "serviceAccount:${google_service_account.k8s_infra_prow.email}"
+}
+
+// Allow read access to members of k8s-infra-prow-oncall@kubernetes.io
+resource "google_storage_bucket_iam_member" "k8s_infra_prow_oncall" {
+  bucket = google_storage_bucket.k8s_infra_prow_bucket.name
+  role   = "roles/storage.objectViewer"
+  //TODO(ameukam): switch to allUsers when https://github.com/kubernetes/k8s.io/issues/752 is closed.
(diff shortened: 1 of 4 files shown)


ameukam:
Aka Terraform version of #1909. 

/assign @spiffxp

infra/gcp/clusters/projects/kubernetes-public/aaa/k8s-infra-prow.tf
```diff
@@ -0,0 +1,98 @@
...
+- a GCP service account for the bucket
+- IAM bindings for the bucket
+- a Secret Manager secret for the service account key
+*/
+
+locals {
+  bucket_name               = "k8s-infra-prow-results"
+  bucket_location           = "us-central1"
+  prow_service_account      = "k8s-infra-prow"
+  test_pods_namespace       = "k8s-infra-prow-pods"
```
spiffxp, line 13: I'm assuming this is intentionally different than "test-pods" that's used by prow.k8s.io (which I am OK with, I think dueling prows in the same namespace would be really tricky to troubleshoot)

I think this means if you were to schedule pods to this namespace on e.g. k8s-infra-prow-build, I don't think they'd get access to boskos or greenhouse, but I may be wrong.

infra/gcp/clusters/projects/kubernetes-public/aaa/k8s-infra-prow.tf
```diff
@@ -0,0 +1,98 @@
...
+  test_pods_service_account = "default"
+}
+
+
+// Create GCP Service Account for prow control plane
+resource "google_service_account" "k8s_infra_prow" {
+  project      = data.google_project.project.project_id
```
spiffxp, line 20: assuming this comes from a different .tf file

infra/gcp/clusters/projects/kubernetes-public/aaa/k8s-infra-prow.tf
```diff
@@ -0,0 +1,98 @@
...
+  role               = "roles/iam.workloadIdentityUser"
+  service_account_id = google_service_account.k8s_infra_prow.name
+  member             = format("serviceAccount:%s.svc.id.goog[%s/%s]", "k8s-infra-prow-build-trusted", local.test_pods_namespace, local.test_pods_service_account)
+}
+
+// Create a GCS bucket for ProwJobs logs and tide history
+resource "google_storage_bucket" "k8s_infra_prow_bucket" {
+  name          = local.bucket_name
+  project       = data.google_project.project.project_id
+  storage_class = "REGIONAL"
```
spiffxp, line 49: Storage class is `STANDARD` for gs://kubernetes-jenkins.  We default to STANDARD in this organization, ref: https://github.com/kubernetes/k8s.io/blob/798d7a3cae7617865c35d00d81928012604ccd5d/audit/check-audit-pr.sh#L90 

There's only one bucket that doesn't have STANDARD and I feel like that's an accident, ref: https://github.com/kubernetes/k8s.io/blob/798d7a3cae7617865c35d00d81928012604ccd5d/audit/check-audit-pr.sh#L223-L225

I want to double check what the pricing implications are for this before allowing it through.  Or we just use STANDARD.

infra/gcp/clusters/projects/kubernetes-public/aaa/k8s-infra-prow.tf
```diff
@@ -0,0 +1,98 @@
...
+// Allow read access to members of k8s-infra-prow-oncall@kubernetes.io
+resource "google_storage_bucket_iam_member" "k8s_infra_prow_oncall" {
+  bucket = google_storage_bucket.k8s_infra_prow_bucket.name
+  role   = "roles/storage.objectViewer"
+  member = "group:k8s-infra-prow-oncall@kubernetes.io"
+}
+
+// Create a secret for GCP Service Account key of k8s-infra-prow
+resource "google_secret_manager_secret" "k8s_infra_prow_key" {
+  secret_id = "k8s-infra-prow-results-sa-key"
```
spiffxp, line 76: To match `prow_service_account` (could also construct it with `local` or `google_service_account.k8s_infra_prow` but maybe this is easier to grep)
```suggestion
  secret_id = "k8s-infra-prow-sa-key"
```

infra/gcp/clusters/projects/kubernetes-public/aaa/k8s-infra-prow.tf
```diff
@@ -0,0 +1,98 @@
...
+  role   = "roles/storage.legacyBucketWriter"
+  member = "serviceAccount:${google_service_account.k8s_infra_prow.email}"
+}
+
+// Allow read access to members of k8s-infra-prow-oncall@kubernetes.io
+resource "google_storage_bucket_iam_member" "k8s_infra_prow_oncall" {
+  bucket = google_storage_bucket.k8s_infra_prow_bucket.name
+  role   = "roles/storage.objectViewer"
+  member = "group:k8s-infra-prow-oncall@kubernetes.io"
+}
```
spiffxp, line 72: If this bucket is for viewing results, it should be for `allUsers` I think? Unless you're planning to start narrow and then replace with something broad later

infra/gcp/clusters/projects/kubernetes-public/aaa/k8s-infra-prow.tf
```diff
@@ -0,0 +1,98 @@
...
+  member             = format("serviceAccount:%s.svc.id.goog[%s/%s]", "k8s-infra-prow-build-trusted", local.test_pods_namespace, local.test_pods_service_account)
+}
+
+// Create a GCS bucket for ProwJobs logs and tide history
+resource "google_storage_bucket" "k8s_infra_prow_bucket" {
+  name          = local.bucket_name
+  project       = data.google_project.project.project_id
+  storage_class = "REGIONAL"
+  location      = local.bucket_location
+}
```
spiffxp, line 51:
```suggestion
  lifecycle_rule {
      condition {
        age = 90
      }
      action {
        type = "Delete"
      }
    }
  }
  uniform_bucket_level_access = true
```
Match parity with lifecycle in gs://kubernetes-jenkins.  

Also _definitely_ want UBLA enabled on this.  It's basically impossible to untangle the ACL's in gs://kubernetes-jenkins, and I don't even want to start with them on here.

spiffxp (changes requested):
Blocking comments mostly on the bucket itself, and a nit on the key

infra/gcp/clusters/projects/kubernetes-public/aaa/k8s-infra-prow.tf
spiffxp: https://cloud.google.com/storage/docs/storage-classes#legacy OK `REGIONAL` ~= `STANDARD`

infra/gcp/clusters/projects/kubernetes-public/aaa/k8s-infra-prow.tf
ameukam: Not exactly. I use : https://github.com/kubernetes/k8s.io/blob/main/infra/gcp/clusters/projects/kubernetes-public/aaa/00-inputs.tf#L27-L30

infra/gcp/clusters/projects/kubernetes-public/aaa/k8s-infra-prow.tf
ameukam: I prefer to start with a restricted group of people until we decide we want to use this bucket definitively. Let me add a TODO.

infra/gcp/clusters/projects/kubernetes-public/aaa/k8s-infra-prow.tf
ameukam: For the beginning do we need access to boskos or greenhouse ? Is prow need boskos to run successfully even we don't run jobs using it ?

infra/gcp/clusters/projects/kubernetes-public/aaa/k8s-infra-prow.tf
ameukam: Lifecyle added.

infra/gcp/clusters/projects/kubernetes-public/aaa/k8s-infra-prow.tf
ameukam: Usd the attribute `name`.

spiffxp (approved):
/approve
/lgtm
/hold
when ready to deploy

ameukam:
Huh got : 

```shell
Error: Error creating Secret: googleapi: Error 400: The provided Secret ID [projects/kubernetes-public/serviceAccounts/k8s-infra-prow@kubernetes-public.iam.gserviceaccount.com] does not match the expected format [[a-zA-Z_0-9]+]
```

ameukam:
Fixed with #2206.

PR #2190, opened 2021-06-10
prow: use container.deployer role for deployer SA

spiffxp:
prow-deployer was using roles/container.developer which is insufficient to automatically deploy resources related to RBAC and webhooks, so we've created a custom role to allow full control of all in-cluster resources without granting access to control the cluster itself

This is followup to https://github.com/kubernetes/k8s.io/pull/2156 which created the role

And https://github.com/kubernetes/k8s.io/pull/2148#issuecomment-856385138 which is where we discovered the insufficiency of `roles/container.developer` for auto-deploying kubernetes-external-secrets

### DIFF
--- a/infra/gcp/clusters/projects/k8s-infra-prow-build-trusted/prow-build-trusted/main.tf
+++ b/infra/gcp/clusters/projects/k8s-infra-prow-build-trusted/prow-build-trusted/main.tf
@@ -35,6 +35,10 @@ locals {
   prow_deployer_sa_name        = "prow-deployer"                // Allowed to deploy to prow build clusters
 }
 
+data "google_organization" "org" {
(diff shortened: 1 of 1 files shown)


spiffxp:
/cc @BenTheElder @ameukam 
I dropped the manual binding
```
gcloud projects remove-iam-policy-binding \
  k8s-infra-prow-build-trusted \
  --member="serviceAccount:prow-deployer@k8s-infra-prow-build-trusted.iam.gserviceaccount.com" \
  --role="roles/container.admin"
```
And have already run `terraform apply` for these changes

BenTheElder (approved):
lgtm

spiffxp:
tracking issue ref: https://github.com/kubernetes/k8s.io/issues/2218

PR #2204, opened 2021-06-11
hack: add verify-executable.sh

spiffxp:
Expect that any file that ends in .sh or .py should be executable,
except lib*.sh, which should be sourced. Ignore the same files as git
when searching for files to verify.

This was honestly more bash than I was expecting for this, but I
found I couldn't quite figure out the right `find` incantation to
do everything there... and bash doesn't have a logical xor operator.

(I definitely blame @BenTheElder for sending me down this one)

### DIFF
--- a/hack/verify-executable.sh
+++ b/hack/verify-executable.sh
@@ -1,0 +1,89 @@
+#!/usr/bin/env bash
+
+# Copyright 2021 The Kubernetes Authors.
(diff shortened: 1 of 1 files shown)


ameukam:
/lgtm
/hold

BenTheElder (approved):
/hold cancel

PR #2207, opened 2021-06-13
Remove jeefy from sig-ui leads

jeefy:
Stepping away from sig-ui leadership. :)


spiffxp (approved):
/approve
/lgtm
ref: https://github.com/kubernetes/community/pull/5828

PR #2209, opened 2021-06-14
groups/sig-testing: Add Christian Glombek to k8s-infra-staging-test-infra

LorbusChris:
This commit adds myself to the k8s-infra-staging-test-infra group.

I'm currently working on getting `arm64` images built for Prow, and this will give me access to the staging infra so I can test the changes myself.
Adding myself here per @spiffxp's invitation on Slack.
/assign @spiffxp 

(I'm not (yet) a member of the `kubernetes` org, is that a prerequisite for this?)


spiffxp (approved):
/approve
/lgtm

Issue #2020, opened 2021-05-07
Get our terraform modules to the latest version of terraform


spiffxp:
Mainly to avoid falling stale

There are features in terraform 0.14 that are interesting:
- masking `sensitive` values from output (e.g. to allow for a `terraform plan` presubmit)
- dependency lock files (so we could maybe stop having to bump the same version in 6 different files)

It's not yet clear to me whether there is anything super critical or interesting in 0.15, it mainly looks like the removal of deprecated things: https://www.terraform.io/upgrade-guides/0-15.html

/wg k8s-infra
/area terraform

spiffxp:
/close
We're up to terraform 1.0 and I think that's good enough to call this done.

Auto-bumping our terraform would be the next logical step, however we need to get better testing and validation in place before forcing that churn upon us IMO

PR #2211, opened 2021-06-14
images/k8s-infra: add terraform 1.0

spiffxp:
not removing v0.14 until we've migrated k8s-infra-public-pii to at least
v0.15


spiffxp:

PR #2217, opened 2021-06-15
infra/clusters: upgrade to terraform 1.0

spiffxp:
All changes have been deployed for the following in infra/gcp/projects:

- k8s-infra-ii-sandbox
- k8s-infra-prow-build/prow-build
- k8s-infra-prow-build-trusted/prow-build-trusted
- k8s-infra-public-pii
- kubernetes-public/aaa

k8s-infra-public-pii was upgraded to v0.15 before upgrading to 1.0

All changes were no-ops; terraform suggested `apply -refresh-only`
but I ran a full `terraform apply` anyway

### DIFF
--- a/infra/gcp/clusters/modules/gke-cluster/versions.tf
+++ b/infra/gcp/clusters/modules/gke-cluster/versions.tf
@@ -15,7 +15,7 @@
  */
 
 terraform {
-  required_version = "~> 0.15.0"
+  required_version = "~> 1.0.0"
   required_providers {
     google = {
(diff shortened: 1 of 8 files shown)


spiffxp:
/hold cancel
All changes have been deployed

sftim (commented):
_Just an aside_

Now that Terraform is generally available (and in other projects outside Kubernetes work) I've switched the version pin to `~> 1.0` because I am more confident in the semantic versioning and guarantees.

PR #2219, opened 2021-06-15
prow: switch cncf-ci-github-token to ExternalSecret

spiffxp:
Migrate the cncf-ci-github-token Kubernetes `Secret` used by the ci-k8sio-audit job to an `ExternalSecret`.

This let me uncover a few things:
- we didn't have `prow-build-trusted` setup the same way as `prow-build` for access
- we didn't have a `kubernetes-external-secrets` service account setup
- move to a `{namespace}-{resource}.yaml` convention for `serviceaccounts` and `externalsecrets`

Everything here has already been deployed

I exercised by changing the secret value to "monkeys" and seeing the change reflected in-cluster, before changing back.


spiffxp:
Migrating secrets ref: https://github.com/kubernetes/k8s.io/issues/2220

PR #2237, opened 2021-06-17
hack: verify-terraform no args == all *.tf paths

spiffxp:
Not opting to have this run by default as part of verify.sh because:

- Errors when run against paths that had terraform init run with a
  backend the current user can't access (see below)
- Running no-args takes too long IMO; 15s locally for no changes

Working around "already init'ed with a bad backend" is hard because:

- terrafom init -from-source into a tmp dir fixes this (no backend to
  begin with), but breaks on loading modules from local relative paths
- terraform has a git_repository data source, but it needs a manually
  specified root (either in source, or via the GIT_DIR env var)
- so... not yet found a way to avoid having to hardcode a path or always
  invoking terraform with GIT_DIR

### DIFF
--- a/hack/verify-terraform.sh
+++ b/hack/verify-terraform.sh
@@ -20,46 +20,83 @@ set -o pipefail
 REPO_ROOT=$(cd "$(dirname "${BASH_SOURCE[0]}")/.." && pwd -P)
 
 function usage() {
-  echo >&2 "Usage: $0 <PATH>"
-  exit 1
+    echo >&2 "Usage: $0 [path...]"
+    exit 1
+}
+
+function ensure_dependencies() {
+    if [ -z "$(which curl)" ]; then
+        echo "Please install curl"
+        exit 1
+    fi
+
+    if [ -z "$(which tfswitch)" ]; then
+        echo "Please install tfswitch"
+        exit 1
+    fi
 }
 
 function check_terraform() {
-  if [ $# != 1 ]; then
-    echo "check_terraform(path) requires 1 argument" >&2
-    exit 1
+  local path="${1}"
+
+  if [ ! -d "${path}" ]; then
+      echo "check_terraform: path '${path}' is not a directory (or does not exist)"
+      return 1
   fi
 
-  if [ ! -d "$1" ]; then
-    echo "path not found"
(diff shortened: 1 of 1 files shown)


hack/verify-terraform.sh
```diff
@@ -20,46 +20,77 @@ set -o pipefail
...
-
-cd "$REPO_ROOT"
-check_terraform "$1"
+    ensure_dependencies
+
+    # if no paths specified, default to all paths that contain *.tf
+    pushd "$REPO_ROOT"
+    if [ "${#paths[@]}" == 0 ]; then
+        mapfile -t paths < <(
+          find . -name '*.tf' -print0 | xargs -0 -n1 dirname | sort | uniq
```
ameukam, line 69: The results include the terraform modules(/infra/gcp/clusters/modules). Do we need to validate them knowing `check_terraform ${path}` already does this part ?

spiffxp: I'll update to exclude them.  My pedantic concern was that would allow us to checkin dead / broken code, since we have nothing to verify/enforce that all modules are used at least once.  But I think you're right, it's probably overkill

spiffxp: That saves some runtime, but it still feels not quite ready to run by default all the time to me.  Most PR's in this repo don't touch terraform.  But we could probably go PR test-infra to have a pull-k8sio-verify-terraform job that triggers for `.tf$` instead of the per-cluster jobs we have now.   That's what motivated this in the first place, I realized we were missing some of those jobs for the new projects of late.

spiffxp:
Followup for anyone who wants to: go PR test-infra to have a pull-k8sio-verify-terraform job that calls `./hack/verify-terraform.sh` with no args, and triggers for `.tf$` instead of the per-cluster jobs we have now.  

That's what motivated this in the first place.  I realized we were missing some of those jobs for the new projects of late, and rather than add them, felt like allowing for one job for all terraform would avoid that toil.

spiffxp:
@ameukam PTAL

ameukam:
_Betrayed by brain. Saw it as an issue_
/lgtm
/hold
Remove hold when ready.

PR #2229, opened 2021-06-15
infra/gcp/prow-build: Add bucket for scalability prowjobs

ameukam:
Ref:
https://github.com/kubernetes/k8s.io/issues/1310#issuecomment-856692948.

Part of https://github.com/kubernetes/k8s.io/issues/1469.

Signed-off-by: Arnaud Meukam <ameukam@gmail.com>

Issue #1310, opened 2020-10-06
Migrate away from google.com gcp project kubernetes-jenkins


spiffxp:
Part of umbrella issue to migrate away from google.com gcp projects: https://github.com/kubernetes/k8s.io/issues/1469

Umbrella issue for migrating or removing dependence on all of the kubernetes project infra that lives under kubernetes-jenkins:

**GCS buckets:**
```gsutil ls -p kubernetes-jenkins | grep -v kubernetes-staging- | sed -e 's/.*/- [ ] `&`/'```
- [x] `gs://artifacts-upload-test/`: TODO: ???
- [x] `gs://artifacts.kubernetes-jenkins.appspot.com/`: TODO: sincerely doubt we need to migrate GCR buckets
- [x] `gs://gcf-sources-661044217466-us-central1/`: TODO: going to assume we can ignore this
- [x] `gs://k8s-bazel-cache/`: May be used by kubernetes/kubernetes bazel, may be defunct? Last updated files 2019-03-11
- [x] `gs://k8s-kops-gce/`: TODO: ???
- [x] `gs://kubernetes-e2e-soak-configs/`: TODO: ???
- [x] `gs://kubernetes-federation-release/`: TODO: ??? for this and the other federation buckets, surely these aren't used
- [x] `gs://kubernetes-federation-release-1-4/`
- [x] `gs://kubernetes-federation-release-1-5/`
- [x] `gs://kubernetes-federation-release-1-6/`
- [x] `gs://kubernetes-federation-release-1-7/`
- [x] `gs://kubernetes-federation-release-1-8/`
- [x] `gs://kubernetes-federation-release-jenkins/`: TODO: ???
- [x] `gs://kubernetes-federation-release-serial/`: TODO: ???
- [ ] `gs://kubernetes-jenkins/`: hosts all logs/artifacts from prow.k8s.io, will need a plan/proposal
- [x] `gs://kubernetes-jenkins-gcslogs/`: definitely do not migrate
- [x] `gs://kubernetes-jenkins-old/`: TODO: : we probably don't care if this is migrated
- [x] `gs://kubernetes-release-dev-jenkins/`: TODO: this probably shouldn't be migrated
- [x] `gs://kubernetes-test-history/`: we can ignore this
- [x] `gs://sig-scalability-logs/`: should happen as part of https://github.com/kubernetes/k8s.io/issues/2241
[...]

PR #2239, opened 2021-06-17
fixup scalability test logs

spiffxp:
Part of https://github.com/kubernetes/k8s.io/issues/1310#issuecomment-856692948

Followup to https://github.com/kubernetes/k8s.io/pull/2229

Started by noticing an improper binding in an audit PR https://github.com/kubernetes/k8s.io/pull/2234#discussion_r653556889

Then I noticed we were using a terraform resource that overwrites existing policy, so I added back my best guess for existing policy

Now I'm beginning to think this bucket should _actually_ go live over in kubernetes-public since this is effectively another results bucket

### DIFF
--- a/infra/gcp/clusters/projects/k8s-infra-prow-build/prow-build/README.md
+++ b/infra/gcp/clusters/projects/k8s-infra-prow-build/prow-build/README.md
@@ -4,7 +4,7 @@ These terraform resources define a GCP project containing a GKE cluster
 intended to serve as a "build cluster" for prow.k8s.io. There are also
 some service accounts defined for use by pods within the cluster.
 
-## Accessing the control plane
+## Access
 
 Access to the [k8s-infra-prow-build project][k8s-infra-prow-build-console] hosting the cluster is granted by membership in one of two @kubernetes.io groups:
 - [k8s-infra-prow-oncall@kubernetes.io][k8s-infra-prow-oncall@]: grants [`roles/owner`][roles/owner] access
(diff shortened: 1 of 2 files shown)


spiffxp:
What the bucket looks like
```yaml
# $ gsutil iam get gs://sig-scalability-logs | yq -y
bindings:
  - members:
      - projectEditor:kubernetes-jenkins
      - projectOwner:kubernetes-jenkins
    role: roles/storage.legacyBucketOwner
  - members:
      - allUsers
    role: roles/storage.legacyBucketReader
  - members:
      - serviceAccount:prow-build@k8s-infra-prow-build.iam.gserviceaccount.com
    role: roles/storage.objectAdmin
etag: CAU
```
How it's used in kubernetes/test-infra
```
$ rg sig-scalability-logs
config/jobs/kubernetes/sig-testing/kubetest-canaries.yaml
180:      - --logexporter-gcs-path=gs://sig-scalability-logs/$(JOB_NAME)/$(BUILD_ID)
281:      - --logexporter-gcs-path=gs://sig-scalability-logs/$(JOB_NAME)/$(BUILD_ID)

config/jobs/kubernetes/sig-scalability/sig-scalability-release-blocking-jobs.yaml
46:      - --logexporter-gcs-path=gs://sig-scalability-logs/$(JOB_NAME)/$(BUILD_ID)
129:      - --logexporter-gcs-path=gs://sig-scalability-logs/$(JOB_NAME)/$(BUILD_ID)
215:      - --logexporter-gcs-path=gs://sig-scalability-logs/$(JOB_NAME)/$(BUILD_ID)
```

So, thoughts:
- yes, I think this should live in kubernetes-public (until/unless we decide there should be a different GCP project for ~ci generated logs and artifacts across the whole kubernetes project)
- IMO this should have been hooked up to gcsweb.k8s.io, so we'll do that for both buckets

spiffxp:
I've deployed the changes in https://github.com/kubernetes/k8s.io/pull/2239/commits/adca45189b1a6da70423e8bf487573b3be692550 via `terraform apply` already, so I won't rebase that commit away.  I plan on marking this as ready once I've moved over to kubernetes-public.

spiffxp:
/hold
So it's looking like I might not finish this by today, in which case it probably makes more sense to have what is deployed get merged.  Removing the WIP label, I'll leave the /hold for @ameukam to decide if it's better to leave this open or not.

But I would rather see a PR to change to `gs://k8s-infra-sig-scalability-logs` in `kubernetes-public` before anything else is changed to rely on this bucket as-is.

ameukam:
> /hold
> So it's looking like I might not finish this by today, in which case it probably makes more sense to have what is deployed get merged. Removing the WIP label, I'll leave the /hold for @ameukam to decide if it's better to leave this open or not.
> 
> But I would rather see a PR to change to `gs://k8s-infra-sig-scalability-logs` in `kubernetes-public` before anything else is changed to rely on this bucket as-is.

The bucket is empty and unused for the moment. It can be deleted and recreated in kubernetes-public. I'll take care of that in a followup PR.

Issue #2150, opened 2021-06-07
Reorganize apps running on aaa under /apps


spiffxp:
This will make it easier to setup a prowjob that auto-deploys these apps on postsubmit

We can also auto-label PRs against these apps and enforce ownership conventions:
- group: `k8s-infra-rbac-{app}@kubernetes.io`
- namespace on aaa: `{app}`
- label on k8s resources: `app: {app}`
- label on gcloud secrets: `app: {app}`
- OWNERS: `approvers: {app}-approvers`
- etc...

Going off of the list of namespaces in `infra/gcp/namespaces/ensure-namespaces.sh` and some manual inspection
- [x] cert-manager (@jimdaga, https://github.com/kubernetes/k8s.io/pull/2374)
- [x] gcsweb (@jimdaga, https://github.com/kubernetes/k8s.io/pull/2284)
- [x] k8s.io (@spiffxp, https://github.com/kubernetes/k8s.io/pull/2383)
- [X] kubernetes-external-secrets (@spiffxp, https://github.com/kubernetes/k8s.io/pull/2290)
- [x] node-perf-dash (@spiffxp, https://github.com/kubernetes/k8s.io/pull/2290)
- [x] perfdash (@jimdaga, https://github.com/kubernetes/k8s.io/pull/2317)
- [X] prow (@ameukam, https://github.com/kubernetes/k8s.io/pull/2330)
- [X] publishing-bot (@jimdaga, https://github.com/kubernetes/k8s.io/pull/2315)
- [x] sippy (@jimdaga, https://github.com/kubernetes/k8s.io/pull/2261)
- [x] slack-infra (@jimdaga, https://github.com/kubernetes/k8s.io/pull/2304)
- [x] triageparty-release (@jimdaga, https://github.com/kubernetes/k8s.io/pull/2303)

/priority important-longterm
/kind cleanup
/wg k8s-infra
/area cluster-infra

jimdaga:
@spiffxp I could maybe give this a try; 

Just to make sure I follow; you're looking to re-org this `k8s.io` repo so that the folders containing app code get moved under `/apps` - and the various labels, OWNER files, etc get updated to match the `{app}` name - correct?

Also; since you mentioned prow - I take it there is no existing deploy automation I'd have to consider, right?

jimdaga:
Sorry things have been slow, I'm hoping to find some time soon to sit down and finish a bunch of PRs at once.

spiffxp:
[...]

PR #2261, opened 2021-06-23
Move sippy under /apps

jimdaga:
## Resolves:
* https://github.com/kubernetes/k8s.io/issues/2150 (first PR of a few)

## Changes:
* Move `sippy` directory under the new `/apps` path. 
* Move yaml out of the secondary `sippy` dir, into the root `sippy` dir (aka: `apps/sippy` vs `apps/sippy/sippy`)
* Add `app:` label to resources that didn't already have it. 
* Re-order content under `metadata:` to have a consistent order 

## Notes:
* Checked https://cs.k8s.io/ and didn't find any references to these files, so no external dependencies require updates. 
* This "app" looks to be an easy starting point while I get feedback from others. I plan to move the other "apps" per the ISSUE request, but will handle that in multiple smaller PRs.


ameukam:
@jimdaga Thank you so much for doing this!

Can you move those YAML files to `apps/sippy` instead of `apps/sippy/sippy` ?

jimdaga:
> @jimdaga Thank you so much for doing this!
> 
> Can you move those YAML files to `apps/sippy` instead of `apps/sippy/sippy` ?

@ameukam no problem, happy to be involved.

This app was actually the reason I asked in #wg-k8s-infra about the idea of storing yaml under a `config` dir. I agree that having `apps/sippy/sippy` feels odd (I left the folder layout the way it was originally). But if just moving the yaml to the `apps/sippy` root is fine, I can make that change.

ameukam:
> > @jimdaga Thank you so much for doing this!
> > Can you move those YAML files to `apps/sippy` instead of `apps/sippy/sippy` ?
> 
> @ameukam no problem, happy to be involved.
> 
> This app was actually the reason I asked in #wg-k8s-infra about the idea of storing yaml under a `config` dir. I agree that having `apps/sippy/sippy` feels odd (I left the folder layout the way it was originally). But if just moving the yaml to the `apps/sippy` root is fine, I can make that change.

I think you can proceed. Once the WIP is removed, we'll leave the approval to the OWNERS. 
cc @spiffxp

ameukam:
/ok-to-test
/lgtm
/hold 

/assign @wojtek-t @deads2k 
cc @stevekuznetsov

apps/sippy/deployment.yaml
```diff
@@ -1,10 +1,10 @@
 apiVersion: apps/v1
 kind: Deployment
 metadata:
-  labels:
-    app: sippy
   name: sippy
   namespace: sippy
+  labels:
+    app: sippy
```
spiffxp, line 7: I think this ordering makes more sense to a human, but it will mean automated enforcement of yaml formatting/ordering can't be done with jq/yq.

I see the type is ordered this way in https://git.k8s.io/apimachinery/pkg/apis/meta/v1/types.go#L116-L288 so presumably we could pass through `kubectl --dry-run` before something that enforced spacing/indenting standards

spiffxp (approved):
/approve
/lgtm
I may `/hold cancel` to move this forward once I'm done watching over some other deploys.

We ought to spend some time fleshing out what the standard layout should be in https://github.com/kubernetes/k8s.io/issues/2150, though I'm fine iterating our way there with a few apps at first

apps/sippy/deployment.yaml
jimdaga: Fair point. I'll admit I usually base things on how `helm create` generates their default yaml, but I'm totally fine with reverting the order if it helps our automation. 

```
$ cat testchart/templates/deployment.yaml | head -6
apiVersion: apps/v1
kind: Deployment
metadata:
  name: {{ include "testchart.fullname" . }}
  labels:
    {{- include "testchart.labels" . | nindent 4 }}
```

Let me know if I should re-order, and I can get another commit added today. 

Thanks!

Issue #1968, opened 2021-04-22
Moving k8s-artifacts-gcslogs bucket to its own project 


Riaankl:
ii.coop request access to the k8s-artifacts-gcslogs to work on K8s Infra WG projects
Granting access with the correct level of permission proof to be difficult.

If the bucket is in its own project it need not pick up `projectOwners` 
Bindings is a cleaner solution to accomplish

Fixes issue: #1945 
Refers to PR: 
Read access to read access k8s-artifacts-gcslogs #1966
Enable GCS logs [#904](https://github.com/kubernetes/k8s.io/issues/904#issuecomment-816484483)

Riaankl:
k8s-artifacts-gcslogs bucket access available to ii
Thank you @spiffxp @ameukam

ameukam:
/reopen
I'll consider this done when #2031 is merged. I plan to do the transition to the dedicated project next week.

spiffxp:
/assign @hh 
I'm assigning @hh for awareness, but I think this is something that @ameukam and I can handle.  We just might someone from ii to confirm that they are capable of polling/regenerating their reports at the end of this.

My sketch of a plan:
- switch all access log settings to point to the new bucket
- copy the old access logs to the proper location in the new bucket
- ask ii to ensure polling/reporting works with the new bucket location
- drop old data, IAM resources (service accounts and/or bindings)
- verify that all PII data goes solely into the one project, and that only members authorized to handle PII have access

ameukam:
We will need to migrate the BigQuery datasets used for images pulling reports.
@Riaankl @bernokl Can you provide us the list of datasets that need to be migrated ?

Riaankl:
@ameukam I cleaned up the `k8s-infra-ii-sandbox` with only 2 data sets remaining:
- `riaan_data_store`
- `etl_script_generated_set_prod`
These we need for reports

ameukam:
@Riaankl Do you need those buckets to be migrated :

```shell
cloud alpha storage ls --project k8s-infra-ii-sandbox
gs://artifacts.k8s-infra-ii-sandbox.appspot.com/
gs://export-c2e4nmc5jmg9n5nacc60/
gs://ii_bq_scratch_dump/
gs://k8s-infra-ii-sandbox-bb-test/
```

Apart from all this, is there anything else that needs to be migrated ?
cc @bernokl

Riaankl:
> @Riaankl Do you need those buckets to be migrated :
> 
[...]

PR #2262, opened 2021-06-23
Add service account to k8s-infra-ii-coop

bernokl:
I think this should also use ensure_service_account from lib_iam.sh, but I am unsure where to call it from


groups/groups.yaml
```diff
@@ -1278,6 +1277,7 @@ groups:
       - riaan@ii.coop
       - stephen@ii.coop
       - zz@ii.coop
+      - asn-etl@k8s-infra-ii-sandbox.iam.gserviceaccount.com
```
ameukam, line 1280: @bernokl I'm not sure to understand why you need to add this SA to this google group. What is the particularity of this SA ?

ameukam:
> I think this should also use ensure_service_account from lib_iam.sh, but I am unsure where to call it from

You can directly create the Service Account with Terraform here : https://github.com/kubernetes/k8s.io/blob/main/infra/gcp/clusters/projects/k8s-infra-ii-sandbox/main.tf

groups/groups.yaml
bernokl: @ameukam it will be our reporting user that is intended to run the ash-ip-lookup job on prow. Is there a more appropriate way for a job runner to get permissions to the logs?

spiffxp: I would rather we keep service accounts out of groups.  I can't recall the specific example right now, but it doesn't always work.  I removed all previous SA memberships in https://github.com/kubernetes/k8s.io/pull/2010, specifically https://github.com/kubernetes/k8s.io/pull/2010/commits/9ebc221608f429d41db4d18bf8dd6d9f9ec2a21a

Bind the service account directly to what is needed

ameukam: @bernokl My suggestion would be look at https://cloud.google.com/iam/docs/understanding-roles to identity what permissions you need for this SA and add the binding as a Terraform code in https://github.com/kubernetes/k8s.io/tree/main/infra/gcp/clusters/projects/k8s-infra-ii-sandbox. 

Ultimately, every resource used to do PII analysis will be moved to https://github.com/kubernetes/k8s.io/tree/main/infra/gcp/clusters/projects/k8s-infra-public-pii. See : https://github.com/kubernetes/k8s.io/issues/1968.

bernokl:
I changed the pr to create the SA and bind-role in permissions.tf, if needed I can move it to main.tf as suggested above.

The project for the role will potentially need to be kubernetes.io but need some guidance. 
The logs this SA would currently consume will be in gs://k8s-artifacts-gcslogs/ ideally it pulls from k8s-infra-public-pii, but I think that is only staging data right now.

infra/gcp/clusters/projects/k8s-infra-ii-sandbox/permissions.tf
```diff
@@ -0,0 +1,41 @@
...
+}
+
+resource "google_project" "project" {
+  name       = local.project_id
+  project_id = local.project_id
+}
+
+#Create service account
+resource "google_service_account" "ii-logs-sa@k8s-infra-ii-sandbox.iam.gserviceaccount.com" {
+  account_id   = "ii-logs-sa-id"
```
spiffxp, line 29:
```suggestion
resource "google_service_account" "ii_logs_sa" {
  account_id   = "ii-logs"
```

infra/gcp/clusters/projects/k8s-infra-ii-sandbox/permissions.tf
```diff
@@ -0,0 +1,41 @@
...
+  display_name = "ii logs service account"
+  description  = "service-account-to-facilitate-ii-log-analysis"
```
spiffxp, line 31:
```suggestion
  description  = "service account to facilitate log analysis by ii"
```

infra/gcp/clusters/projects/k8s-infra-ii-sandbox/permissions.tf
```diff
@@ -0,0 +1,41 @@
...
+  project      = google_project.project.id
+}
+
+resource "google_project_iam_binding" "k8s-infra-ii-sandbox" {
+  project = google_project.project.id
+  role    = "roles/storage.objectViewer"
+  members = [
+    "user:ii-logs-sa@k8s-infra-ii-sandbox.iam.gserviceaccount.com",
```
spiffxp, line 39:
```suggestion
    "user:ii-logs@k8s-infra-ii-sandbox.iam.gserviceaccount.com",
```

infra/gcp/clusters/projects/k8s-infra-ii-sandbox/permissions.tf
```diff
@@ -0,0 +1,41 @@
...
+
+#Create service account
+resource "google_service_account" "ii-logs-sa@k8s-infra-ii-sandbox.iam.gserviceaccount.com" {
+  account_id   = "ii-logs-sa-id"
+  display_name = "ii logs service account"
+  description  = "service-account-to-facilitate-ii-log-analysis"
+  project      = google_project.project.id
+}
+
+resource "google_project_iam_binding" "k8s-infra-ii-sandbox" {
```
spiffxp, line 35: From https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/google_project_iam

>`google_project_iam_binding`:  Authoritative for a given role. Updates the IAM policy to grant a role to a list of members. Other roles within the IAM policy for the project are preserved.

Are you sure you want this behavior? It means nobody else gets `roles/storage.objectViewer` in this project except the service account.

What you might want instead want is to say "this service account should get this role, regardless of who else has the role"

Which would be
```suggestion
resource "google_project_iam_member" "k8s-infra-ii-sandbox" {
```

spiffxp (commented):
question and some naming nits

spiffxp (changes requested):
(earlier review was meant to be "request changes")

PR #2231, opened 2021-06-16
infra/gcp: enable GCS access logs for staging project GCS and GCR

bernokl:
Addresses:
https://github.com/kubernetes/k8s.io/issues/904#issuecomment-859932423
Intent:                                                                                                                                                                                                                                                       
- Ensure a bucket exist for the logs inside each project                                                                                                                                                                                                        
- Enable logging for k8s-staging-* buckets                                                                                                                                                                                                                      
- Ensure access is gated to limit access per #2031     
Research:
https://github.com/ii/org/blob/main/research/k8sio-ensure-staging-pii-gcs.org

PR #2283, opened 2021-07-01
infra/gcp/lib_gcs: fix ensure_gcs_bucket_logging

spiffxp:
Followup to https://github.com/kubernetes/k8s.io/pull/2231

I missed the copy-paste in three rounds of review, it took actually running the script for me to notice.

I also quieted down the noise from diff

### DIFF
--- a/infra/gcp/ensure-staging-storage.sh
+++ b/infra/gcp/ensure-staging-storage.sh
@@ -146,7 +146,7 @@ function ensure_staging_project() {
     # Ensure staging project GCS
 
     color 3 "Ensuring staging GCS bucket: ${staging_bucket}"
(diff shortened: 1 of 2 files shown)


spiffxp:
/hold
I also notice logging is being enabled for `foo-gcb` which is unnecessary, so this might change to a fix to ensure-staging-storage.sh as well

spiffxp:
/hold cancel
These changes have already been deployed, ref: https://github.com/kubernetes/k8s.io/pull/2231#issuecomment-871839272

spiffxp:
ref: https://github.com/kubernetes/k8s.io/issues/904#issuecomment-859932423

PR #1857, opened 2021-03-31
dl.k8s.io: Redirect CI URIs to Kubernetes Community infra

justaugustus:
- dl.k8s.io: Redirect CI URIs to Kubernetes Community infra

  krel ci-build has been successfully building and pushing CI Kubernetes
  builds for a time now.
  
  Updating this redirect (dl.k8s.io/ci) to Kubernetes Community infra
  moves us closer to no longer relying on the kubernetes-release-dev
  (Google-owned) bucket.

  ref: https://github.com/kubernetes/test-infra/pull/20663, https://github.com/kubernetes/test-infra/pull/20964

- dl.k8s.io: Remove ci-cross redirects
  
  All Kubernetes builds that land in the 'ci' directory of k8s-release-dev
  are cross builds, so this reference is no longer required.

  ref: https://github.com/kubernetes/test-infra/pull/18517

Signed-off-by: Stephen Augustus <foo@auggie.dev>

Part of https://github.com/kubernetes/release/issues/1711 and https://github.com/kubernetes/k8s.io/issues/1569

PR #2291, opened 2021-07-02
hack: add k8s-infra-diff-buckets.sh

spiffxp:
Diff the contents of two GCS bucket paths via `gsutil ls`. This is
motivated by trying to monitor the delta between google-owned and
community-owned CI artifacts.

The original plan was to rewrite this to use `gsutil rsync --dry-run`
and then remove the `--dry-run` to actually sync the buckets
periodically via a prow job. It turns out the files that motivated this
are not bit-for-bit identical, nor do they have accurate enough
timestamps to be used for comparison, since they are produced by
different jobs running on different clusters at slightly different
times.

So for now, since I've been reporting results from this publicly, get
the script off of my laptop.

Sample usage:

    # diff a single-level listing, discard intermediate files
    ./hack/k8s-infra-diff-buckets.sh \
        gs://kubernetes-release-dev/ci/ \
        gs://k8s-release-dev/ci/

    # diff a recursive listing, leave intermediate files behind
    DEBUG=true ./hack/k8s-infra-diff-buckets.sh \
        gs://kubernetes-release-dev/ci/** \
        gs://k8s-release-dev/ci/**

Comments where I've referenced output from this:
- https://github.com/kubernetes/k8s.io/pull/1857#issuecomment-825386815
- https://github.com/kubernetes/k8s.io/pull/1857#issuecomment-858037795

ref: https://github.com/kubernetes/k8s.io/issues/846#issuecomment-801278162

### DIFF
--- a/hack/k8s-infra-diff-buckets.sh
+++ b/hack/k8s-infra-diff-buckets.sh
@@ -1,0 +1,126 @@
+#!/usr/bin/env bash
+
+# Copyright 2021 The Kubernetes Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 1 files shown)


justaugustus (approved):
TIL `comm`!
/lgtm
/approve

PR #2295, opened 2021-07-06
Promote Addon Resizer 1.14 images

jbartosik:
Add Addon Resizer images. It contains fixes necessary for Addon resizer to work on K8s 1.22 (https://github.com/kubernetes/autoscaler/pull/4105, https://github.com/kubernetes/autoscaler/pull/4176).

I tested that amd64 image works by:
- Creating cluster,
- Creating deployment (configured to use metrics),
- Resizing cluster (from 2 to 5 nodes),
- Observing that deployments request changed (from cpu: 340m, memory: 220Mi to cpu: 400m, memory: 250Mi)


jbartosik:
I also checked with Addon Resizer configured not to use metrics

@bskiba please take a look

PR #2290, opened 2021-07-02
apps: mv kubernetes-external-secrets and node-perf-dash, update existing, add README

spiffxp:
Part of umbrella issue https://github.com/kubernetes/k8s.io/issues/2150

I spiraled around the existing apps migrated (gcsweb, sippy) and moved over two more (kubernetes-exteral-secrets, node-perf-dash)

In sum, I tried to converge on and document a few conventions:
- ensure `apps/{app}` matches the namespace declared in `infra/gcp/namespaces/ensure-namespaces.sh`
- ensure `apps/{app}/deploy.sh` exists and is referenced in `apps/{app}/README.md` - I opted for a copy pasted script, see below
- ensure `apps/{app}/OWNERS` has a `sig/foo` label in it
- ensure `apps/{app}/*.yaml` k8s resources have appropriate `labels: { app: {app} }` stanzas
- document these and other intended conventions in `apps/README.md`
- add known issues to `apps/README.md` and warning that conventions are subject to change as we iterate
- refresh `/README.md` to reference `apps` and link to https://github.com/kubernetes/k8s.io/issues/2150

The `deploy.sh` script:
- The script is copy-pasted to make it clear we could likely reduce down to a single parameterized script
- But I didn't reduce down yet, in case we find there are apps that don't fit this model and need to be customized as one-offs

A nice-to-have I did while I was here: sort the buckets used by gcsweb

PR #2303, opened 2021-07-08
Move triage party under /apps

jimdaga:
## Resolves:
* https://github.com/kubernetes/k8s.io/issues/2150 (Part of umbrella issue)

## Changes:
* Move `triage-party` directory under the `apps/triageparty-release` path.
* Move yaml out of the secondary `release-team` dir, into the root `triageparty-release` path.
* Add the `app:` label on resources missing such label.
* Add `deploy.sh` script, and update README with install steps

## Notes:
* Checked https://cs.k8s.io/ and didn't find any references to these files for automation use.
* For now we will preserve the `app: triage-party` label until a final decision is made about standards.
* The storage.yaml file used a slightly different `app:` label (`app: triageparty` without a `-`) that I'll be leaving in place until we finish the convo about naming.


jimdaga:
Note: 
After https://github.com/kubernetes/k8s.io/pull/2290 is merged, we'll need to get the apps/README.md updated to reflect this app moved.

apps/triageparty-release/deployment.yaml
```diff
@@ -4,16 +4,16 @@ metadata:
   name: triage-party
   namespace: triageparty-release
   labels:
-    app: triage-party
+    app: triageparty-release
```
ameukam, line 7: I would have preferred that we keep the current labels and add new one referring the community group:

```yaml
labels:
	app: triage-party
	group: sig-release
```

@spiffxp WDYT ?

spiffxp: I was looking for a term we can use to enforce same-name conventions for:
1. a namespace on aaa
2. a directory under apps/ that holds (or generates) all k8s resources in that namespace
3. an rbac group that is authorized to deploy k8s resources to that namespace
4. (maybe overkill?) a label on all resources deployed from the aforementioned apps subdirectory (vs. resources dynamically created by the k8s resources or its users) <-- this is why the `app:` value is being renamed

I think I latched onto "app" because it's what I feel like we've been calling them, but it's actually mixing two discrete meanings, since they've been the same for many of the, uh, "things" we're running in aaa.

To borrow from OO terminology, I think your objection is that the "app:" label better fits the "class" and not the "instance" of app.  In Helm parlance "app" would be the 1 Chart, not the N Releases.  In kpt terms, I think "app" would be the 1 Package, not the N ResourceGroups.  

Since naming is hard I think we should iterate on how to resolve this on a PR against https://github.com/kubernetes/k8s.io/blob/main/apps/README.md

In the interest of unblocking, does the label change break things now?  If it does, let's revert.  If it doesn't, are you OK if we iterate on this post-merge?

jimdaga: I know the question was directed at @ameukam - but I'm thinking I would be most comfortable reverting the `app:` label back to the original, and waiting to figure out the "story" before making a change. We can always come back and update.

spiffxp (approved):
/approve
/lgtm

apps/triageparty-release/deployment.yaml
ameukam: +1 for to figure out this later.

spiffxp:
Had to edit deploy.sh to remove dry-run (I forgot to remove this when checking it in, will fix in followup PR shortly)
```
spiffxp@cloudshell:~/k8s.io/apps/triageparty-release (kubernetes-public)$ ./deploy.sh
managedcertificate.networking.gke.io/release-triage-k8s-io configured
configmap/triage-party-config-release configured
deployment.apps/triage-party configured
Warning: networking.k8s.io/v1beta1 Ingress is deprecated in v1.19+, unavailable in v1.22+; use networking.k8s.io/v1 Ingress
ingress.networking.k8s.io/triage-party-release unchanged
externalsecret.kubernetes-client.io/github-token created
service/triage-party-release configured
persistentvolumeclaim/triage-party-persistence unchanged
```

spiffxp:
Followup PR: https://github.com/kubernetes/k8s.io/pull/2314

PR #2308, opened 2021-07-08
Bump terraform provider

ameukam:
https://github.com/hashicorp/terraform-provider-google/blob/master/CHANGELOG.md#3740-june-28-2021


spiffxp (approved):
/approve
/lgtm
Scanned CHANGELOG to confirm no breaking changes that would impact us.  

We do use bigquery resources, but we don't change their ids, so the breaking change in 3.67.0 doesn't seem relevant to us.

PR #2317, opened 2021-07-09
Move perfdash under /apps

jimdaga:
## Resolves:
* https://github.com/kubernetes/k8s.io/issues/2150 (Part of umbrella issue)

## Changes:
* Move `perf-dash.k8s.io` directory under the `apps/perfdash` path.
* Add `deploy.sh` with some "custom" updates to apply the external yaml (based on existing README notes)
* Update various README notes
* Add `app:` label as needed

## Notes:
- [x] Looking for feedback - should we drop the FQDN format (perf-dash.k8s.io) of the directory name to make it match the `app:` name (perfdash)?


spiffxp:
>should we drop the FQDN format (perf-dash.k8s.io) of the directory name to make it match the app: name (perfdash)?

yes please, I did the same thing for node-perf-dash

apps/perfdash/deploy.sh
```diff
@@ -0,0 +1,55 @@
...
+# ensure we have a context to talk to the target cluster
+if ! kubectl config get-contexts "${context}" >/dev/null 2>&1; then
+  gcloud container clusters get-credentials "${cluster_name}" --project="${cluster_project}" --region="${cluster_region}"
+  context="${gke_context}"
+fi
+
+# deploy kubernetes resources
+pushd "${SCRIPT_ROOT}" >/dev/null
+kubectl --context="${context}" --namespace="${namespace}" apply -f "${perfdash_repo_url}/deployment.yaml"
+kubectl --context="${context}" --namespace="${namespace}" apply -f "${perfdash_repo_url}/perfdash-service.yaml"
```
spiffxp, line 54: Uh, TIL this is how it's always been for this app, so I won't blow your PR's scope.  But these files should live here (I was thinking in the direction of a postsubmit job that could run the relevant deploy.sh if any app's files changed).  Not a fan of deploy scripts that will deploy something different depending on the time of day they're run.

spiffxp (approved):
/approve
/lgtm

apps/perfdash/deploy.sh
spiffxp: FYI @mborsz @mm4tt @jprzychodzen

PR #2327, opened 2021-07-12
Update sippy deployment commandline to adjust to newer image

wojtek-t:
Update to pick https://github.com/openshift/sippy/pull/241

/assign @deads2k


spiffxp (approved):
/approve
/lgtm

Issue #1734, opened 2021-02-27
Validate configuration changes


ameukam:
if possible, any configuration change introducted should be validated with presubmits prowjobs.

There are different and tools we can explore : 

- Globally, define policies with [rego language](https://www.openpolicyagent.org/docs/latest/policy-language/) and enforce them with [Open Policy Agent](https://www.openpolicyagent.org)
- units tests and e2e tests with [terratest](https://github.com/gruntwork-io/terratest)
	- Doesn't cover Kubernetes resources
- Kubernetes resources validation with [conftest](https://github.com/open-policy-agent/conftest) (also based on rego)

/priority backlog

ameukam:
cc @spiffxp

spiffxp:
Was literally just about to file this issue, thank you!

/remove-priority backlog
/priority important-longterm
/sig testing
/area access
/area cluster-mgmt
/area cluster-infra

spiffxp:
Presubmits we have thus far:
- sig-release:
  - if `/k8s.gcr.io/**/{images,promoter-manifest}.yaml` changes
    - `pull-k8sio-cip`: dry-run container image promoter
    - `pull-k8sio-cip-vuln`: (for `vuln-check-test` branch) dry-run container image promoter with vuln-severity-threshold=1
  - if `/infra/gcp/backup_tools` changes
    - run `infra/gcp/backup_tools/backup_test`
- sig-contribex:
  - if `/groups` changes
    - `pull-k8sio-groups-test`: run `cd groups && go test` (some policy enforcement)
- sig-testing:
  - `pull-k8sio-yamllint`: verifies all yaml in this repo is valid yaml (could enforce style conventions)

spiffxp:
>- Globally, define policies with rego language and enforce them with Open Policy Agent
>- Kubernetes resources validation with conftest (also based on rego)

Really, really good idea.  Way more interested in `conftest` as a starting point, seems like basically any YAML is fair game.

I am pretty interested in how much this could allow us to share enforcement logic at different phases in our deployment lifecycle, eg: presubmit, admission, etc.

ameukam:
Also, since 0.13 Terraform has embebbed mechanism for variable validation : https://www.hashicorp.com/blog/custom-variable-validation-in-terraform-0-13.

spiffxp:
An update on where we're at.

[...]

PR #2328, opened 2021-07-12
hack: enforce conftest and yamllint failures

spiffxp:
Add more enforcement to hack/verify.sh via:
- failing on `yamllint` rules that were previously warnings (`new-line-at-end-of-file` and `trailing-spaces`) to bring us to parity with the `yamllint` enforcment used for kubernetes/test-infra
- failing on `conftest` failures based on Open Policy Agent policies defined in policies/ 

Also tuned down the output noise from `verify-shellcheck.sh`

Part of https://github.com/kubernetes/k8s.io/issues/1734


hack/verify-conftest.sh
```diff
@@ -0,0 +1,45 @@
...
+
+SCRIPT_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+REPO_ROOT="$(cd "${SCRIPT_DIR}/.." && pwd )"
+
+function usage() {
+    echo >&2 "Usage: $0"
+    exit 1
+}
+
+function ensure_dependencies() {
```
aojea, line 28: you are not calling this function neither usage

spiffxp: Great catch thank you. I'll also try and fold in a commit that adds a failing policy and then fixes next commit

hack/verify-conftest.sh
```diff
@@ -0,0 +1,45 @@
...
+    pushd "${REPO_ROOT}" >/dev/null
+    local k8s_yaml_paths=(
+        apps/
+        infra/gcp/clusters/projects/*/*/resources/*.yaml
+    )
+    conftest test --policy policies/ "${k8s_yaml_paths[@]}"
+
+}
+
+main
```
cpanato, line 45: nit: new line

spiffxp:
I could not get as far as "writing a failing policy and then fix it".  What we have now is not quite as copy-paste friendly as a thought, and I'd rather not have this hang out until I more properly learn Rego.

I did at least add some warnings for v1.22 removals we'll want to address

PR #2332, opened 2021-07-13
Add sa/role-binding for log viewer

bernokl:
I think the project for the role will need to be kubernetes.io, the logs this SA would currently consume will be in gs://k8s-artifacts-gcslogs/ ideally it pulls from k8s-infra-public-pii, but I think that is only staging data right now.


bernokl:
will add work to https://github.com/kubernetes/k8s.io/pull/2262/

Issue #2345, opened 2021-07-15
Google groups description limit < 300 chars not checked in tests


PushkarJ:
Google groups description max. limit is 300 characters: https://developers.google.com/admin-sdk/groups-settings/v1/reference/groups#:~:text=entry/-,apps%3Adescription,-object

This should be tested in `groups_test.go` otherwise it can lead to failed jobs like this: https://storage.googleapis.com/kubernetes-jenkins/logs/post-k8sio-groups/1415792053220020224/build-log.txt 

Will open a PR to fix this shortly. There is also an existing group: `k8s-infra-staging-cip-test` that seem to have gone through with a higher description character length (maybe because it was created before this limit was imposed). Can add the test to ignore the check for this group. 

/kind bug
/sig security
/area access

spiffxp:
/priority critical-urgent
Since this is breaking group sync right now
/milestone v1.22

PR #2342, opened 2021-07-15
Request new group for sig-security-tooling

PushkarJ:
- Added `OWNERS` file and `groups.yaml` file for a security tooling group that is used for private triage of vulnerabilities identified via automated scans
- Removed a line in README, that does not reflect the current conventions tested
  
 Related PRs: https://github.com/kubernetes/community/pull/5853 and https://github.com/kubernetes/test-infra/pull/22833
  
 Relevant Slack threads: 
 [Question](https://kubernetes.slack.com/archives/C1TU9EB9S/p1626375603426800) 
 [Answer 1](https://kubernetes.slack.com/archives/C1TU9EB9S/p1626378312427400)
 [Answer 2](https://kubernetes.slack.com/archives/C1TU9EB9S/p1626379613431900)
  
/cc @IanColdwater @tabbysable @justaugustus

PR #2346, opened 2021-07-15
Check max. length for groups description

PushkarJ:
- Google groups API imposes group description length to be no greater than 300
- New test checks the length and fails when description greater than 300 is attempted
- It ignores the description of an existing group for backwards compatability even though it exceeds this limit

Fixes https://github.com/kubernetes/k8s.io/issues/2345

Related: https://github.com/kubernetes/k8s.io/pull/2342

/sig security
/area access
/kind bug 
/cc @spiffxp

### DIFF
--- a/groups/groups_test.go
+++ b/groups/groups_test.go
@@ -115,6 +115,32 @@ func TestStagingEmailLength(t *testing.T) {
 	}
 }
 
+// TestDescriptionLength tests that the number of characters in the
+// google groups description does not exceed 300.
+//
+// This validation is needed because gcloud allows apps:description
+// with length no greater than 300
+func TestDescriptionLength(t *testing.T) {
+	var errs []error
+	for _, g := range cfg.Groups {
+		description := g.Description
+
+		len := utf8.RuneCountInString(description)
+		//Ref: https://developers.google.com/admin-sdk/groups-settings/v1/reference/groups
+		if len > 300 {
+			errs = append(errs,
+				fmt.Errorf("Number of characters in description \"%s\" for group name \"%s\" "+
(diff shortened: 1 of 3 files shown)


groups/groups_test.go
```diff
@@ -115,6 +115,33 @@ func TestStagingEmailLength(t *testing.T) {
...
+// with length no greater than 300
+func TestDescriptionLength(t *testing.T) {
+	var errs []error
+	for _, g := range cfg.Groups {
+		description := g.Description
+
+		len := utf8.RuneCountInString(description)
+		//Ignoring this group's description for backwards compatibility because it seems at
+		//the time of that group's creation, this description limit was not imposed
+		if len > 300 && g.Name != "k8s-infra-staging-cip-test" {
```
spiffxp, line 131: I would rather we truncate this group's description and avoid the special case

Looking at https://github.com/kubernetes/k8s.io/blob/a7d10e5664779ec6c47be2b91db64f00e496617b/groups/sig-release/groups.yaml

I would suggest changing the description to:
```
ACL for staging cip-test buckets, the test-only backup, and auditing
jobs for the promoter. See infra/gcp/ensure-prod-storage.sh for more
details.
```

spiffxp (changes requested):
Thank you for doing this!

Let's make this simpler

groups/groups_test.go
PushkarJ: This should be fixed
```
pjoglekar@pjoglekar-a01 groups % make test
go test
PASS
ok  	k8s.io/k8s.io/groups	0.229s

```

spiffxp (approved):
/approve
/lgtm

PR #2349, opened 2021-07-15
infra/gcp: mv k8s-releng-prod to its own folder

spiffxp:
k8s-releng-prod may be a release-related project, but it should not be handled by `ensure-release-projects.sh`

Move it to its own folder/group in infra.yaml to prevent https://github.com/kubernetes/k8s.io/pull/2322#discussion_r670713886 from happening again

### DIFF
--- a/infra/gcp/ensure-releng.sh
+++ b/infra/gcp/ensure-releng.sh
@@ -36,10 +36,8 @@ function usage() {
     echo > /dev/stderr
 }
 
-# NB: Please keep this sorted.
-PROJECTS=(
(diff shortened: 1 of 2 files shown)


spiffxp:
/cc @puerco @cpanato @justaugustus 
FYI @kubernetes/release-engineering

spiffxp:
I've already re-run `./infra/gcp/ensure-releng.sh` and `./infra/gcp/ensure-release-projects.sh` to ensure no delta from the current, correct state

PR #2353, opened 2021-07-16
Adding tjferrara to k8s-infra-gcp-accounting

tylerferrara:
This change is an attempt to gain access to k8s-artifacts-prod's `cip-auditor` logs within cloud run.

Such access will allow me to continue working on CIP's [auditor scaling issue](https://github.com/kubernetes-sigs/k8s-container-image-promoter/issues/259). This is NOT a blocker at the moment, so please let me know if there is a better group to join in order to gain such access.

cc: @listx @amwat @justaugustus @kubernetes-sigs/release-engineering

### DIFF
--- a/groups/wg-k8s-infra/groups.yaml
+++ b/groups/wg-k8s-infra/groups.yaml
@@ -112,6 +112,7 @@ groups:
       - spiffxp@gmail.com
       - spiffxp@google.com
       - thockin@google.com
+      - tjferrara@google.com
 
   - email-id: k8s-infra-gcp-auditors@kubernetes.io
     name: k8s-infra-gcp-auditors


spiffxp:
/hold
I don't think accounting is the right group for this

I had a TODO about giving the auditors group read-only access to Cloud Run anyway, let's try giving that a shot and see if it gets you the right level of access

Opened https://github.com/kubernetes/k8s.io/pull/2355

spiffxp:
/close
I think the auditor change gives you what you need.  If you get to the point of needing to deploy changes, it looks like k8s-infra-artifact-admins@ is the only group currently authorized:

https://github.com/kubernetes/k8s.io/blob/f5e083900d472280cc86e568f978552a39312e75/audit/projects/k8s-artifacts-prod/iam.json#L60-L65

I don't think it has to be that way, but let's chat with releng before you open your next PR

tylerferrara:
> /hold
> I don't think accounting is the right group for this
> 
> I had a TODO about giving the auditors group read-only access to Cloud Run anyway, let's try giving that a shot and see if it gets you the right level of access
> 
> Opened #2355

Much appreciated!

PR #2362, opened 2021-07-19
[Scalability] Add new oncall member

jupblb:
/cc @wojtek-t
/cc @marseel 

I am a new member of sig-scalability oncall. :)


groups/groups.yaml
```diff
@@ -387,6 +387,7 @@ groups:
       - janluk@google.com
       - jkaniuk@google.com
       - jprzychodzen@google.com
+      - jupblb@google.com
```
wojtek-t, line 390: Since you touch this file - can you please remove janluk@ ?

Please also add marcelzieba@google.com

jupblb: Done.

Issue #2318, opened 2021-07-09
Deprecate and migrate away from gs://kubernetes-release-dev and gcr.io/kubernetes-ci-images


spiffxp:
Part of umbrella issue to migrate the kubernetes project away from use of GCP project google-containers: https://github.com/kubernetes/k8s.io/issues/1571

This issue covers the deprecation of and migration away from two google.com assets:
- the google.com-owned GCS bucket `gs://kubernetes-release-dev` living in GCP project `google-containers`, in favor of the community-owned GCS bucket `gs://k8s-release-dev` living in GCP project `k8s-release`
- the google.com-owned GCR repo `gcr.io/kubernetes-ci-images`, in favor of the community-owned GCR repo `gcr.io/k8s-staging-ci-images`

#### Step 1: Set and announce deprecation window

The first part is setting and announcing a deprecation window for CI artifacts placed in gs://kubernetes-release-dev, and their companion images in gcr.io/kubernetes-ci-images e.g.

- v1.17 was the first release that had CI builds of kubernetes land in `gs://k8s-release-dev/ci` and `gs://kubernetes-release-dev/ci` simultaneously
- v1.18 was the first release that cluster-api and kubeadm started defaulting to `gcr.io/k8s-staging-ci-images` instead of `gcr.io/kubernetes-ci-images`
- v1.20 was the last release that had CI builds land in `gs://kubernetes-release-dev/ci-cross`
- v1.22 is the last release to have CI builds of kubernetes land in `gs://kubernetes-release-dev` and `gcr.io/kubernetes-ci-images`
- we will stop using `gs://kubernetes-release-dev` and `gcr.io/kubernetes-ci-images` completely once older CI jobs have aged out according to the kubernetes version support policy, or when we see no remaining traffic to `gs://kubernetes-release-dev` for a period of N weeks

#### Step 2: Setup community-owned alternatives, and use them for release-blocking / merge-blocking CI

This is a rough breakdown of the work necessary to complete the heavy lifting before we can shard out work to all repos across the project

- [x] Add k8s-release GCP project and gs://k8s-release-dev buckets - https://github.com/kubernetes/k8s.io/pull/1110 (@spiffxp)
- [x] Ensure gs://k8s-release-dev visible via https://gcsweb.k8s.io - https://github.com/kubernetes/k8s.io/pull/1195 (@spiffxp), https://github.com/kubernetes/k8s.io/pull/1196 (@spiffxp) 
[...]

PR #2363, opened 2021-07-19
remove references to kubernetes-release-dev

spiffxp:
The one exception is gcsweb.k8s.io, which should continue serving the
bucket for as long as we have deprecated builds living in it. It should
stop serving once v1.22 has aged out of support, or when there are no
more builds being written to the bucket, whichever comes first

Part of https://github.com/kubernetes/k8s.io/issues/2318


ameukam:
/lgtm
/hold

PR #2371, opened 2021-07-20
prow: dir-per-namespace resources deployed to cluster

spiffxp:
Reorganize prow build cluster resources to they fit the following pattern:

    # for namespace-scoped resources
    infra/gcp/clusters/projects/{project}/{cluster}/resources/{namespace}/*.yaml

    # for cluster-scoped resources
    infra/gcp/clusters/projects/{project}/{cluster}/resources/*.yaml

I had a good reason for this back when I noodled on it, but I no longer remember exactly what it is. At the very least, if we're going to have two different control planes pointed at the same build clusters with different namespaces, this should make it extra clear which set of service accounts etc. are intended for what


spiffxp:
Opening issue to add tests to enforce this as followup, I would like to get this merged ahead of adding some resources to avoid merge-conflict rebases

PR #2373, opened 2021-07-21
Define min and max instances for cip-auditor

tylerferrara:
This change prevents Cloud Run from reducing the number of instances for the cip-auditor. Naturally, when cloud run notices low CPU usage, it scaled down the number of instances. This behavior occurs too often, since the auditor should always be running.

Here is a view of the current cip-auditor running in the GCP project: `k8s-artifacts-prod`.
pantheon corp google com_run_detail_us-central1_cip-auditor_metrics_project=k8s-artifacts-prod (1)

However, by explicitly setting `--min-instances` and `--max-instances` during deployment, we can be sure that the auditor will always be running.

Related to CIP issue: [#353](https://github.com/kubernetes-sigs/k8s-container-image-promoter/issues/353)

cc: @listx @amwat @justaugustus @kubernetes-sigs/release-engineering

### DIFF
--- a/infra/gcp/cip-auditor/deploy.sh
+++ b/infra/gcp/cip-auditor/deploy.sh
@@ -51,7 +51,9 @@ deploy_cip_auditor()
         --no-allow-unauthenticated \
         --region=us-central1 \
         --project="${PROJECT_ID}" \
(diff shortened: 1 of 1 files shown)


dims:
/hold

someone needs to run this as well!

cpanato (approved):
/lgtm

Issue #2316, opened 2021-07-09
Convert Ingresses and services on k8s-infra cluster aaa


ameukam:
The GKE `aaa` cluster have been [upgraded](https://github.com/kubernetes/k8s.io/pull/2294/files) to GKE 1.19. We should move the existing ingresses to the GA version `networking.k8s.io/v1`. 

We currently use the  `networking.k8s.io/v1beta1`:

```shell:
$grep 'networking.k8s.io/v1beta1'

perf-dash.k8s.io/ingress.yaml
1:apiVersion: networking.k8s.io/v1beta1

k8s.io/ingress-canary.yaml
1:apiVersion: networking.k8s.io/v1beta1

k8s.io/ingress-canary-v6.yaml
1:apiVersion: networking.k8s.io/v1beta1

k8s.io/ingress-prod-v6.yaml
1:apiVersion: networking.k8s.io/v1beta1

k8s.io/ingress-prod.yaml
1:apiVersion: networking.k8s.io/v1beta1

apps/triageparty-release/ingress.yaml
1:apiVersion: networking.k8s.io/v1beta1

apps/slack-infra/resources/ingress.yaml
1:apiVersion: networking.k8s.io/v1beta1

apps/node-perf-dash/ingress.yaml
1:apiVersion: networking.k8s.io/v1beta1

apps/gcsweb/ingress.yaml
1:apiVersion: networking.k8s.io/v1beta1

apps/sippy/ingress.yaml
1:apiVersion: networking.k8s.io/v1beta1
```

And possibly convert Kubernetes services of type NodePort to the ClusterIP type with [GCP Networking Endpoint](https://cloud.google.com/kubernetes-engine/docs/how-to/standalone-neg).

References: 
https://cloud.google.com/kubernetes-engine/docs/concepts/ingress
https://kubernetes.io/docs/concepts/services-networking/ingress/
https://cloud.google.com/kubernetes-engine/docs/how-to/container-native-load-balancing

jimdaga:
First PR submitted: https://github.com/kubernetes/k8s.io/pull/2375

spiffxp:
/close
Closed out by:
- https://github.com/kubernetes/k8s.io/pull/2397 (@spiffxp)
- https://github.com/kubernetes/k8s.io/pull/2425 (@jimdaga)

```
$ ./hack/verify-conftest.sh | grep networking || echo all migrated
all migrated
```

ameukam:
Still want to do : 

[...]

PR #2375, opened 2021-07-21
Migrate to networking.k8s.io/v1

jimdaga:
## Resolves:
* https://github.com/kubernetes/k8s.io/issues/2316 (First PR, more to come, maybe)

## Notes:
- Starting with the ingress version change as a starting point.
- I'll do some review on the NEG part of the request, and follow up with another PR if possible.
- I only have read-only access on the `aaa` cluster, so I'll need help getting the `deploy.sh` scripts run. 

## Changes:
* Update all k8s-infra `aaa` ingress objects to `networking.k8s.io/v1`

### DIFF
--- a/apps/gcsweb/ingress.yaml
+++ b/apps/gcsweb/ingress.yaml
@@ -1,4 +1,4 @@
-apiVersion: networking.k8s.io/v1beta1
+apiVersion: networking.k8s.io/v1
 kind: Ingress
 metadata:
   name: gcsweb
--- a/apps/node-perf-dash/ingress.yaml
+++ b/apps/node-perf-dash/ingress.yaml
@@ -1,4 +1,4 @@
-apiVersion: networking.k8s.io/v1beta1
+apiVersion: networking.k8s.io/v1
 kind: Ingress
 metadata:
   name: node-perf-dash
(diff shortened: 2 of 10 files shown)


ameukam:
> I only have read-only access on the aaa cluster, so I'll need help getting the deploy.sh scripts run.

@jimdaga spiffxp or me can take care of this once it's merged. 

> I'll do some review on the NEG part of the request, and follow up with another PR if possible.

You can check what's it's already done for gcsweb: https://github.com/kubernetes/k8s.io/pull/2080

apps/gcsweb/ingress.yaml
```diff
@@ -1,4 +1,4 @@
-apiVersion: networking.k8s.io/v1beta1
+apiVersion: networking.k8s.io/v1
```
ameukam, line 1: With the Ingress API in GA, new fields have been introduced to manage the backends. https://kubernetes.io/docs/concepts/services-networking/ingress/#the-ingress-resource. 
Can be done as followup but would great to convert the ingresses in the PR. :-D

jimdaga: Makes sense! :) 
I'm unfortunately not going to have time this week (heading out for PTO) but can do that next week if we're ok holding off until then.

ameukam: No rush about  this. Enjoy your PTO! :-)

PR #2382, opened 2021-07-21
apps/cert-manager: move ixdy to emeritus_approvers

ixdy:
Not sure whether we should find other folks to add to the approvers list, or just rely on the top-level approvers.


ameukam:
@ixdy Thank you for everything you did!

/lgtm

PR #2383, opened 2021-07-21
apps: mv k8s.io to apps/

spiffxp:
The last app to migrate as part of https://github.com/kubernetes/k8s.io/issues/2150

This was made possible by making `deploy.sh` accept zero arguments, ref: https://github.com/kubernetes/k8s.io/pull/2378

I had to fix `hack/verify-contest.sh` as part of this because it was trying to parse `/k8s.io/Makefile` as yaml.  This has always been broken, we just didn't have it pointed at all apps not-yet-in-`apps/`

I suspect there are cheatsheets and things that link to to what used to be `/k8s.io/README.md`, so I've dropped a tombstone / placeholder there for now to avoid dead links.  I fixed the references in this repo, and may come back to remove the tombstone if I feel like https://cs.k8s.io has let me find everything else.


spiffxp:
```
Verifying verify-conftest.sh
Error: running test: parse configurations: parser unmarshal: unmarshal yaml: error converting YAML to JSON: yaml: line 19: found character that cannot start any token
FAILED   verify-conftest.sh	0s
```

Well that's a super helpful error message.  Line 19 of what?  And yamllint is happy.

spiffxp:
...because `yamllint` doesn't try to parse `Makefile` as yaml, that's why

spiffxp:
/pony party

and it deployed too! https://prow.k8s.io/view/gs/kubernetes-jenkins/logs/post-k8sio-deploy-app-k8s-io/1417990743854682112

PR #2395, opened 2021-07-26
[WIP] Validate kubernetes resources changes at presubmit level

ameukam:
The script will connect a GKE cluster and validate kubernetes on the server
side.

WIP because : 
- I want it to be more generic. (not sure if it's the best approach)
- possibly move to dedicated bash lib and split for aaa and the build clusters.

### DIFF
--- a/hack/verify-k8s-resources.sh
+++ b/hack/verify-k8s-resources.sh
@@ -1,0 +1,67 @@
+#!/usr/bin/env bash
+# Copyright 2021 The Kubernetes Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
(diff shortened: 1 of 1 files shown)


ameukam:
Is it supposed to trigger the script even if it's not add to the main branch ? 🤔

spiffxp:
mmmmm I am not sure I want something talking to our clusters via presubmit

is it possible to do this with an account that only has read-only access?

ameukam:
> mmmmm I am not sure I want something talking to our clusters via presubmit
> 
> is it possible to do this with an account that only has read-only access?

🤔  I don't know if read-only access is enough for CRD resources (like resources of the gke.io API) validation on the server-side. We could do that client-side but we can't validate things like `ManagedCertificate`. I'll follow up with more investigation.

PR #2401, opened 2021-07-28
groups: allow restricting creation of new groups in sub-directories 

nikhita:
For https://github.com/kubernetes/k8s.io/issues/460#issuecomment-849958721

This PR allows restricting creation of new groups in sub-directories
by specifying the restrictions in a config file called `restrictions.yaml`.

It follows a similar model to the one we use in slack-infra today:
https://github.com/kubernetes/community/blob/master/communication/slack-config/restrictions.yaml

This is done to ensure that approvers in groups/sig-foo/OWNERS don't
accidentally create random official-sounding groups without sufficient review/oversight.

After this PR:
- Creation of new groups will require changes to `restrictions.yaml` and
will require approval from groups/OWNERS.
- Updates to existing groups can be approved by groups/sig-foo/OWNERS,
if appropriate restrictions are specified in `restrictions.yaml`.

If a new group is specified in a sub-directory but `restrictions.yaml` is not updated, an error will be displayed:

```
Could not load groups config: couldn't merge groups: cannot define group "bad-group@kubernetes.io" in "sig-contributor-experience/groups.yaml"
```

/assign @spiffxp


nikhita:
From https://github.com/kubernetes/k8s.io/issues/460#issuecomment-849958721:

> The former can probably handled by properly sharding groups, and enumerating which use cases don't need all the extra paranoia.

FYI I haven't added the `approvers` field in groups/sig-foo/OWNERS in this PR because we still need to add more docs around this.

spiffxp (approved):
/approve
/lgtm
Thank you so much!

I don't have my paranoia engine spun up enough to decide if we want to add something like disallowed members so I'll save that for followup on whether we're comfortable enough to add approvers to subdirs

spiffxp:
This isn't quite working as expected.  Will take a look in hopefully an hour
```
2021/07/28 11:57:35 couldn't merge groups: cannot define group "conduct@kubernetes.io" in "**/*"
```

spiffxp:
Took longer for me to get back here, and somehow it was just _now_ that I realized how untested this thing is.  It's a pile of funcs.

https://github.com/kubernetes/k8s.io/pull/2413 will get us back up and running

I'll have another PR that includes other stuff I did along the way to find that, but let's get back to green first

PR #2410, opened 2021-07-28
audit: use k8s-infra-ci-robot for audit PRs

spiffxp:
Followup to https://github.com/kubernetes/test-infra/pull/23048

I thought I had this all setup so config changes were only needed in one place (test-infra), but may as well make the defaults match the secret I just set the job to use.

### DIFF
--- a/audit/create-or-update-audit-pr.sh
+++ b/audit/create-or-update-audit-pr.sh
@@ -30,6 +30,7 @@
 set -o errexit
 set -o nounset
 set -o pipefail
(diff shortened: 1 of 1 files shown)


spiffxp:
/cc @ameukam @dims 
FYI @hh cncf-ci-robot can still be used for stuff but I figured it was time to start using a k8s account for the audit PR job now that we have one setup

ameukam:
/lgtm
/hold

PR #2414, opened 2021-07-29
groups: fix groups reconciliation, misc cleanup

spiffxp:
Related:
- Part of: https://github.com/kubernetes/k8s.io/issues/460
- Followup to: https://github.com/kubernetes/k8s.io/pull/2401
- Possibly supercedes: https://github.com/kubernetes/k8s.io/pull/2413

This is https://github.com/kubernetes/k8s.io/pull/2413 but with all the extra stuff I was doing along the way before I decided... wait... just the few-line fix first.

Split into distinct commit explaining what I was doing, but I'm happy to squash after review

### DIFF
--- a/groups/reconcile.go
+++ b/groups/reconcile.go
@@ -120,13 +120,14 @@ var (
 
 	verbose = flag.Bool("v", false, "log extra information")
 
+	defaultConfigFile       = "config.yaml"
 	defaultRestrictionsFile = "restrictions.yaml"
 	emptyRegexp             = regexp.MustCompile("")
 	defaultRestriction      = Restriction{Path: "*", AllowedGroupsRe: []*regexp.Regexp{emptyRegexp}}
 )
 
 func main() {
-	configFilePath := flag.String("config", "config.yaml", "the config file in yaml format")
+	configFilePath := flag.String("config", defaultConfigFile, "the config file in yaml format")
 	confirmChanges := flag.Bool("confirm", false, "false by default means that we do not push anything to google groups")
 	printConfig := flag.Bool("print", false, "print the existing group information")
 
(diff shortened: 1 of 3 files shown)


groups/reconcile.go
```diff
@@ -246,40 +246,48 @@ func main() {
 	}
 }
 
-func readConfig(configFilePath string, confirmChanges bool) error {
-	if *verbose {
-		log.Printf("reading config file %s", configFilePath)
-	}
+func (c *Config) Load(configFilePath string, confirmChanges bool) error {
+	log.Printf("reading config file: %s", configFilePath)
```
nikhita, line 250: nit: these logs are enabled at other places only if `verbose` is true. Is there a reason why we remove that check here?

nikhita (approved):
One nit but we can probably get to that in a follow-up.

/lgtm

/hold
please cancel the hold when ready

groups/reconcile.go
spiffxp: I don't know that I accomplished this cleanly, but what I tried to do was have top-level config type stuff output by default, and leave verbose for individual operations.  The reason being that I, personally, find a lot more comfort in getting a sense of what this is going to do and where it's reading from before it does things... but then I only want to know when it's making changes (vs. walking though things)

I realize in running the tests this makes things a little noisy for non-CLI-invocation use-cases.  But the tests are it, for now, and there aren't many.

Were someone to keep going on refactoring this for testability, I would suggest using an actual logging package with verbosity levels, such that the really granular stuff could be dropped to a higher verbosity level (debug or v=2 or something), and configuration of that logging to be done by the main entrypoint vs. a much quieter default used by tests.

PR #2425, opened 2021-07-30
Migrate to networking.k8s.io/v1

jimdaga:
## Resolves:
* https://github.com/kubernetes/k8s.io/issues/2316

## Changes:
* Update all k8s-infra `aaa` ingress objects to `networking.k8s.io/v1`
* Move backend configs to newer format

## Notes
* I ended up closing my first PR and staring over since there was a mess of merge conflicts. This one should be cleaner since I will not be sitting on it.
* I was reading over https://github.com/kubernetes/k8s.io/pull/2080#issuecomment-847243627 and Tim hinted there may be a "safer" way to migrate to NEGs. I think it would be safest to get this change done first, then we can move a few of the smaller risk apps over to NEGs as a starting point, then do everything else. I may test a few things out in my own clusters to get comfortable with the migration process first.


spiffxp (approved):
/approve
/lgtm

spiffxp (approved):
/lgtm

spiffxp:
Spot check of apps post-deploy, they're all working fine.  Thanks @jimdaga !

Issue #2183, opened 2021-06-10
API Groups registry


aojea:
Following up on https://github.com/kubernetes-sigs/gateway-api/issues/690

> After some discussion with @thockin, @liggitt, and an earlier Slack thread with @aojea, it sounds like we should be moving to the k8s.io API group as part of the v1alpha2 release.

API groups on k8s.io are special, they require API review and they may be used for internal APIs or CRDs.

Since the project is growing and there are more subprojects using or wanting to use the k8s.io API group, having a central registry for all them can avoid possible conflicts or naming issues and provide a single source of truth for users and developers.

Should we create a registry with all the official api groups?
How can we "enforce" using it?

thockin:
This has been on my TODO list forever.  We do something internally like
this.  In fact, let me copy from the internal doc.  I was hoping to
institute something like this


Kubernetes allows “fully qualified” names for various things, to allow
3rd-party extension with lower risks of collisions (e.g. “
foo.example.com/bar” vs “bar”).  In particular this is used for labels and
annotations (applicable to all API objects), resources (for nodes and
pods), and API groups (all extensions via CRD or EAS).  We have
historically had few conventions for how to name these things, which has
led to myriad forms of names in use.

## Goal
This doc describes a convention for managing Kubernetes identifiers.

## The Standard

### Domain: k8s.io

`<explain what this is for>`

### Domain: kubernetes.io

`<explain what this is for>`

### Domain: x-k8s.io

`<explain what this is for>`

### Style: embrace dashes

Both the prefix portion of label/annotation names and API group names must
be a DNS subdomain (lower-case alphanumeric characters, '-' or '.', and
must start and end with an alphanumeric character).

The suffix portion of label/annotation names is more free-form.  We use
lower-case characters and digits, with logical words separated by dashes.
We do not use periods, upper-case characters, or underscores.

`<examples>`

### Curation: source control

All identifiers will be curated via revision-controlled files in git,
hosted under `<TBD>`  The directory structure is a Java-style (reverse-DNS)
hierarchy representing the identifier prefix.  The files are named
identifiers.yaml in each directory, and each file contains a single record
of the following schema (expressed here in Go syntax):

```
struct {
   Description string   // short justification for this prefix
    Names []struct {
        Key string // exact name being claimed
        Type string // data type of the value (int, string, JSON-encoded
list, etc)
        Context string // where it is used (Node annotation, API group, etc)
        Description string // short description of what it is for
[...]

Issue #320, opened 2019-07-24
Setup DNS and redirect for x-k8s.io


dims:
To support the new efforts around API/CRDs https://github.com/kubernetes/community/pull/3875

dims:
```
dims 1:36 PM
@thockin another one i missed mentioning this morning. CNCF now has the x-k8s.io domain we requested. what do we do next with it? Could we setup a redirect or something to a page that we own?
thockin 1:37 PM
uggh, are we actually proceeding with that proposal
do we want it to resolve to somethng?  To what?  We can own the domain for purposes of CRD scoping without actually DOING anything with it
stp 1:38 PM
Damn wanted to use it as testing ground for redirector. Side benefit it would by default redirect to kubernetes.io ^^
jdetiber:vmware: 1:39 PM
Netlify redirects for api documentation for CRDs defined under a *.x-k8s.io api group? (edited) 
:+1:
1

dims 1:41 PM
@thockin we could just point that to k/enhancements repo for now
munnerz 1:43 PM
Netlify redirects for api documentation for CRDs defined under a *.x-k8s.io api group? (edited)
I know this of interest for kind - we currently have kind.sigs.k8s.io/v1alpha1 redirect to our API docs, and I know the intent is to have kind.x-k8s.io/v1alpha1 redirect too… :see_no_evil: (cc @bentheelder)
stp 1:43 PM
This would be something I could do with the redirector. Could also be something that's non breaking, while we test redirect ideas?
Could bounce of needs and ideas with munnerz, if that makes sense.
dims 1:44 PM
works for me @stp
@stp whoa! multiple stp`s :slightly_smiling_face:
jdetiber:vmware: 1:44 PM
It would be nice from my perspective for cluster-api to be able to have API docs published under the api group URL and user docs under *.sigs.k8s.io
stp 1:45 PM
Haha multiple of us. Wouldn't have guesses.
dims 1:45 PM
@jdetiber agree, but that’s a longer discussion. i wanted something quick to take ownership of that domain
:+1:
1

stp 1:46 PM
OK so proposal: Use TXTDirect DNS records to redirect the cluster-api to the api docs under the new domain with the hosted service to test out the DNS side of things and do the self-hosting and actual go.k8s.io in parallel. Aka test with new non breaking stuff and do the rest on the side/next step.
bentheelder:keanu: 1:46 PM
[catching up]
yeah, that proposal went forward AFAIK.
[...]

PR #2427, opened 2021-07-30
dns: point x-k8s.io to k8s.io redirector

spiffxp:
Related
- Part of: https://github.com/kubernetes/k8s.io/issues/320
- Related to: https://github.com/kubernetes/k8s.io/issues/2183

Not using the fixes keyword so I can confirm this works as expected when live

### DIFF
--- a/dns/zone-configs/x-k8s.io._0_base.yaml
+++ b/dns/zone-configs/x-k8s.io._0_base.yaml
@@ -1,3 +1,13 @@
 # The top-level x-k8s.io zone
+# point x-k8s.io at redirector to get sent to kubernetes.io
+'':
+  - type: A
+    value: 34.107.204.206 # Our vanity redirector, hosted in GKE
+  - type: AAAA
+    value: "2600:1901:0:26f3::" # Our vanity redirector, hosted in GKE
 
-# Nothing here.
+# there are no subdomains and no plans for any fancy redirect logic,
+# this is a placeholder domain to allow the use of x-k8s.io as
+# a namespace for "SIG sponsored CRD based APIs outside of the core"
+#
+# ref: https://git.k8s.io/community/sig-architecture/api-review-process.md#voluntary
--- a/apps/k8s-io/configmap-nginx.yaml
+++ b/apps/k8s-io/configmap-nginx.yaml
@@ -59,6 +59,19 @@ data:
         }
       }
 
+      # Redirect x-k8s.io to main site as proof of domain ownership to allow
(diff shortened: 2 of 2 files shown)


aojea:
I'm not very familiar with the structure, but the others domains in the folder have a canary, is it needed here?

thockin:
I don't hard object to it, but I don't know if there's a goal beyond "it
resolves to something" ?

aojea:
lgtm then

spiffxp:
Discussed offline.  Yeah, the goal is "it resolves to something".  All I really want is for it to end up at kubernetes.io as a hint that this is officially project-owned.  I'll update PR to make that clearer.

>I'm not very familiar with the structure, but the others domains in the folder have a canary, is it needed here?

Canary zone already exists, this PR doesn't need to touch it (it is identical to the config for `k8s.io._1_canary.yaml`)

dns/zone-configs/x-k8s.io._0_base.yaml
```diff
@@ -1,3 +1,4 @@
 # The top-level x-k8s.io zone
-
-# Nothing here.
+'':
```
aojea, line 2: heh, it was a long time since I touch DNS, do we want a wildcard here? 
https://github.com/octodns/octodns/blob/master/docs/records.md#names

thockin:
We don't usually do wildcards - we don't want random names to resolve.

dns/zone-configs/x-k8s.io._0_base.yaml
```diff
@@ -1,3 +1,4 @@
...
+  type: CNAME
```
aojea, line 3: I think that cnames are not allowed at root level, 
the octodns code seems to confirm that
https://github.com/octodns/octodns/blob/e087a3b26f0272a994d3efba2a62ead0f4617a16/octodns/record/__init__.py#L886

aojea: the first approach with the A seems the simplest/safest, no?

thockin:
Yeah, it has to be A.

spiffxp:
I guess I had hopes of doing this without leaving room open for someone to get clever with the redirector in the future.  Hopefully comments and a link to what `x-k8s.io` is supposed to used for will be sufficient.

So back to A, and an explicit block for it in the redirector

spiffxp:
Followup: https://github.com/kubernetes/k8s.io/pull/2437 is what is actually deployed (to canary), merging will deploy this to canary (again) and then prod if tests pass

PR #2441, opened 2021-08-02
prow: use pre-provisioned ghproxy-metrics address

spiffxp:
Prep for the transfer of more jobs that consume GitHub tokens by exposing the metrics IP of ghproxy to be scraped by something.  Likely monitoring.prow.k8s.io

We'll want to do something similar for the ghproxy running in aaa, to help get a sense of things like triageparty github token usage.


spiffxp:
/cc @ameukam @dims 
The terraform changes were already deployed in order to get the IP address for the `loadBalancerIP` field.  The ghproxy service will be updated automatically on PR merge

PR #2452, opened 2021-08-03
Allow Service Account admin access to dataset

ameukam:
permissions added to bq-data-transfer were not enough to make the data
transfer job successful.
Use a predefined role instead of a custom role should fix it.

Got: 

```shell
1_etl_test_distinct_ip_org with error PERMISSION_DENIED: 
Access Denied: Project 226195303281: User does not have bigquery.jobs.create permission in project 226195303281.;
JobID: 226195303281:bqts_6208b112-0000-23fe-91dd-94eb2c0569f4
```


spiffxp (approved):
/approve
/lgtm

PR #2479, opened 2021-08-06
prow: match prow-build-trusted nodepool to prow-build

spiffxp:
the motivating factor here is trying to run the triage job on this
cluster, which asks for more memory than an n1-standard-8 can provide

but in general I can't think of a good reason to restrict this cluster
to less resources than the regular prow-build cluster, while having it
match the cluster would create less surprises when transferring a job
from one to the other for whatever reason

PR #2480, opened 2021-08-06
infra/gcp/clusters: fix gke-nodepool ignore_changes

spiffxp:
Followup to https://github.com/kubernetes/k8s.io/pull/2479

I suspect the syntax to ignore node taint changes was incorrect before,
causing terraform to ignore _all_ node_config changes

(confirmed `terraform plan` doesn't want to change k8s-infra-prow-build nodes, but does want to change k8s-infra-prow-build-trusted)


spiffxp:
This is what I was able to deploy https://github.com/kubernetes/k8s.io/pull/2479 with

Issue #2218, opened 2021-06-15
Prow: allow deployment of kubernetes-external-secrets without roles/container.admin


spiffxp:
Should be able to CRUD kubernetes resources without CRUD rights for the cluster

Some work has already been done toward this, to link in later

spiffxp:
https://github.com/kubernetes/k8s.io/pull/2190 setup a custom `container.deployer` role

But that apparently still isn't sufficient: https://github.com/kubernetes/k8s.io/pull/2190#issuecomment-859750534

spiffxp:
It seems like creating RBAC-related resources is the sticking point here.  I suspect `container.admin` is by default given a Kubernetes role that allows it to do so, but custom roles maybe not so much?

spiffxp:
/milestone v1.22
If I can't figure out what's preventing the custom role I'm going to add back container.admin for now to unblock auto-deploys

spiffxp:
Went back to `roles/container.admin` for now

PR #2483, opened 2021-08-06
infra/gcp: prow-deployer as container.admin

spiffxp:
The custom container.deployer role isn't getting privileges to deploy RBAC and I'd rather not leave our postsubmit deploy process broken while I continue to not have time to figure out why.

Related:
- part of https://github.com/kubernetes/k8s.io/issues/2218
- should unblock auto deploy of https://github.com/kubernetes/k8s.io/pull/2235

### DIFF
--- a/infra/gcp/ensure-main-project.sh
+++ b/infra/gcp/ensure-main-project.sh
@@ -212,8 +212,11 @@ function empower_cluster_admins_and_users() {
     # TODO(spiffxp): consider replacing this with a service account per namespace
     local prow_deployer_acct prow_deployer_role
     prow_deployer_acct="prow-deployer@k8s-infra-prow-build-trusted.iam.gserviceaccount.com"
+    # TODO(spiffxp): use container.deployer once we figure out why it isn't getting RBAC privileges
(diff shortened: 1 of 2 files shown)


infra/gcp/ensure-main-project.sh
```diff
@@ -212,8 +212,11 @@ function empower_cluster_admins_and_users() {
     # TODO(spiffxp): consider replacing this with a service account per namespace
     local prow_deployer_acct prow_deployer_role
     prow_deployer_acct="prow-deployer@k8s-infra-prow-build-trusted.iam.gserviceaccount.com"
+    # TODO(spiffxp): use container.deployer once we figure out why it isn't getting RBAC privileges
+    old_prow_deployer_role=$(custom_org_role_name "container.deployer")
```
spiffxp, line 216: 🤦🏻 I somehow saved but then undid and saved the whole point of this change, which was to add `roles/container.admin` and remove `container.deployer`.  This removed/added container.deployer

Fix in https://github.com/kubernetes/k8s.io/pull/2484

PR #2482, opened 2021-08-06
infra/gcp: mv clusters to terraform

spiffxp:
This changes a few things within infra/gcp
```
clusters                           -> terraform
terraform/projects/{project}       -> terraform/{project}
terraform/{project}/{cluster}/*.tf -> terraform/{project}/*.tf
```

The intent is:
- the name "terraform" better fits what this folder is... terraform modules used to manage (some of) our infrastructure
- less nesting means less directory traversal to get to where I need to `terraform apply`
- make the cluster folder just about cluster resources (possible readmes/scripts for cluster)
- make the project folder the site for all .tf files for the project

I can remove the script I used to generate this PR, but thought I'd check it in, assuming I'd have to do a fair amount of rebasing

PR #2485, opened 2021-08-06
cleanup infra/gcp/terraform mv detritus

spiffxp:
Related:
- the move we're cleaning up from: https://github.com/kubernetes/k8s.io/pull/2482
- the corresponding jobs move: https://github.com/kubernetes/test-infra/pull/23163

specifically
- remove the script that made the move
- go over the scripted substitutions since some of them were clearly
  wrong

/cc @ameukam

/cc @MushuEE
Thank you for the catches, that would have been a pain to discover after
the fact


ameukam:
Ref: https://github.com/kubernetes/k8s.io/issues/516

Issue #2497, opened 2021-08-09
k8s-infra-prow: allow deck to access to gs://k8s-testgrid


ameukam:
Followup issue of #2235 

deck (component of prow) need access (Read-Only ?) to `gs://k8s-tesgrid` so it can display Testgrid links in the SpyGlass.

```
{"component":"deck","error":"Get \"https://storage.googleapis.com/k8s-testgrid/config\": compute: Received 403 `Unable to generate access token; IAM returned 403 Forbidden: The caller does not have permission\nThis error could be caused by a missing IAM policy binding on the target IAM service account.\nFor more information, refer to the Workload Identity documentation:\n\thttps://cloud.google.com/kubernetes-engine/docs/how-to/workload-identity#authenticating_to\n\n`","file":"prow/spyglass/testgrid.go:45","func":"k8s.io/test-infra/prow/spyglass.(*TestGrid).Start","level":"error","msg":"Couldn't fetch TestGrid config.","severity":"error","time":"2021-08-09T17:19:55Z"}
{"component":"deck","file":"prow/kube/config.go:55","func":"k8s.io/test-infra/prow/kube.kubeConfigs","level":"info","msg":"Parsed kubeconfig context: prow-build-trusted","severity":"info","time":"2021-08-09T17:20:49Z"}
```

I believe this bucket in a Google GCP organization. 

/assign @spiffxp 
cc @BenTheElder 
/area prow
/area apps/prow
/sig testing
/priority important-soon
/milestone v1.23

fejta:
Who is this job trying to authenticate as? Google rejecting deck's attempt to authenticate as whatever identity it is trying to use.

```
Unable to generate access token; IAM returned 403 Forbidden: The caller does not have permission\
```

ameukam:
> Who is this job trying to authenticate as? Google rejecting deck's attempt to authenticate as whatever identity it is trying to use.
> 
> ```
> Unable to generate access token; IAM returned 403 Forbidden: The caller does not have permission\
> ```

This a deck instance running on a public GKE cluster (see #1394). The attempt is triggered by the configuration of this instance : https://github.com/kubernetes/k8s.io/blob/main/apps/prow/config.yaml#L74-L75.

spiffxp:
[...]

PR #2512, opened 2021-08-10
apps/prow: Allow deck service account to act as a GCP SA

ameukam:
Allow deck to act as GCP ServiceAccount
k8s-infra-prow@kubernetes-public.iam.gserviceaccount.com

Also make `gs://k8s-infra-prow-results` publicly accessible in read-only.


ameukam:
This is already deployed.
We can now see https://gcsweb.k8s.io/gcs/k8s-infra-prow-results/. 

cc @spiffxp @dims 

ameukam:
Ref: #2497

PR #2520, opened 2021-08-11
apps/prow: Fix postsubmit failures.

ameukam:
The namespace `k8s-infra-test-pods` is now part of the resources
deployed by the `deploy.sh` script.
We pass the incorrect namespace for this resource.
Drop the flag `--namespace` should fix the failures we get:
https://storage.googleapis.com/kubernetes-jenkins/logs/post-k8sio-deploy-app-prow/1425519080953090048/build-log.txt

Also use FQDN for prow services consumed by deck.

### DIFF
--- a/apps/prow/deploy.sh
+++ b/apps/prow/deploy.sh
@@ -25,15 +25,17 @@ set -o pipefail
 
 SCRIPT_ROOT=$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd -P)
 
-app=$(basename "${SCRIPT_ROOT}")
+#TODO(ameukam): drop usage of it.
+#app=$(basename "${SCRIPT_ROOT}")
 
 # coordinates to locate the target cluster in gke
 cluster_name="aaa"
 cluster_project="kubernetes-public"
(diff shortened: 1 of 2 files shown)


apps/prow/deploy.sh
```diff
@@ -57,4 +59,4 @@ make update-plugins
 echo "Update k8s-infra-prow prowjobs"
 make update-prowjobs
 
-kubectl --context="${context}" --namespace="${namespace}" apply -Rf cluster/
+kubectl --context="${context}" apply -Rf cluster/
```
spiffxp, line 62: I'm wary of this because the thing applying these changes has rights to more than just the namespaces relevant to prow.  I don't want namespaces to magically get created from files that sneak in here.  Is it possible to segment resources by namespace into separate dirs?

spiffxp (approved):
/lgtm
/approve
A nit for followup

apps/prow/deploy.sh
ameukam: I'll move the namespace to a separate dir.

PR #2523, opened 2021-08-12
Add k8s-infra-google-build-admins to ci buckets writers

puerco:
This commit modifies the ensure-release-project script to add the k8s-infra-google-build-admins@kubernetes.io group as writers of the CI Buckets.

The goal is enabling `google-build-admins` to stage packages in the -dev buckets as the group has no permissions in the bucket now.

Ref: https://kubernetes.slack.com/archives/C2C40FMNF/p1628725183223700

Signed-off-by: Adolfo García Veytia (Puerco) <adolfo.garcia@uservers.net>


puerco:
We are going to try skipping pushing the .debs to the bucket, I will close this one for now and will reopen if we need it.
/close

spiffxp:
FWIW I'm OK with this change if you want it in place.  Chatting on slack it seems like folks were able to workaround without this change though.

PR #2535, opened 2021-08-13
updating co-chairs

jimangel:
Updating SIG Docs chairs.

Ref: https://github.com/kubernetes/website/issues/29354

/cc @parispittman @sftim


sftim (commented):
/lgtm

PR #2461, opened 2021-08-04
Migrate k8s-triage, k8s-metrics to terraform, setup k8s-triage dataset

spiffxp:
Related:
- Part of: https://github.com/kubernetes/k8s.io/issues/1305
- Followup to: https://github.com/kubernetes/k8s.io/pull/2454
- Depends on: https://github.com/kubernetes/test-infra/pull/23126

Since converting go.k8s.io/triage to triage.k8s.io is going to be more involved than I thought, setup to migrate gs://k8s-gubernator/triage to gs://k8s-triage.  The bucket is currently populated with whatever it contained a while ago (which is basically as fresh as it's going to get since kettle has been down this whole time)

The dependency test-infra PR sets up a job that runs in k8s-infra-prow-build-trusted to update gs://k8s-triage

While I initially did all the provisioning for this via ensure-main-project.sh, I have since updated the PR to do all of this via terraform.  I migrated all of the setup code for gs://k8s-metrics while doing so.

This also adds a bigquery dataset `kubernetes-public:k8s-triage` for the ci-test-infra-triage-canary job to be able to write to.  A future test-infra PR will configure the job to write to it instead of currently failing to write to `k8s-gubernator:temp.triage`

In setting up the dataset, I'm trying out use of IAM roles instead of basic access (ref: https://cloud.google.com/bigquery/docs/access-control-basic-roles)

PR #2539, opened 2021-08-13
Adding missing copyright boilerplate in infra tf files

deobieta:
Added missing copyright boilerplate to terraform source files as requested by @spiffxp .

Ref: https://github.com/kubernetes/k8s.io/pull/2461/files#r688071188


spiffxp:
Thank you for doing this!

Can you `cp hack/boilerplate/boilerplate.{go,tf}.txt` and add the resulting file? That should let `verify-boilerplate` enforce that it's been add to all .tf files

spiffxp (approved):
/approve
/lgtm

PR #2551, opened 2021-08-16
prow: add k8s-infra-test-pods ns to prow-build-trusted

spiffxp:
Should stop deploys from failing

Fixes: https://github.com/kubernetes/k8s.io/issues/2552

### DIFF
--- a/infra/gcp/terraform/k8s-infra-prow-build-trusted/prow-build-trusted/resources/k8s-infra-test-pods/build-serviceaccounts.yaml
+++ b/infra/gcp/terraform/k8s-infra-prow-build-trusted/prow-build-trusted/resources/k8s-infra-test-pods/build-serviceaccounts.yaml
@@ -3,6 +3,14 @@ apiVersion: v1
 kind: ServiceAccount
 metadata:
   annotations:
+    iam.gke.io/gcp-service-account: prow-build-trusted@k8s-infra-prow-build-trusted.iam.gserviceaccount.com
+  name: prow-build-trusted
+  namespace: k8s-infra-test-pods
+---
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  annotations:
     iam.gke.io/gcp-service-account: k8s-infra-prow@kubernetes-public.iam.gserviceaccount.com
   name: prowjob-default-sa
   namespace: k8s-infra-test-pods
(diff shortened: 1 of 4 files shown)


infra/gcp/terraform/k8s-infra-prow-build-trusted/prow-build-trusted/resources/k8s-infra-test-pods/build-serviceaccounts.yaml
```diff
@@ -1,6 +1,14 @@
 ---
 apiVersion: v1
 kind: ServiceAccount
+metadata:
+  annotations:
+    iam.gke.io/gcp-service-account: prow-build@k8s-infra-prow-build.iam.gserviceaccount.com
```
spiffxp, line 6: /hold
whoops, this isn't right

spiffxp: /hold cancel
fixed

infra/gcp/terraform/k8s-infra-prow-build-trusted/prow-build-trusted/resources/k8s-infra-test-pods/build-serviceaccounts.yaml
```diff
@@ -1,6 +1,14 @@
 ---
 apiVersion: v1
 kind: ServiceAccount
+metadata:
+  annotations:
+    iam.gke.io/gcp-service-account: prow-build-trusted@k8s-infra-prow-build-trusted.iam.gserviceaccount.com
```
ameukam, line 6: @spiffxp out of curiosity, why we allowing this SA to use this namespace ?

ameukam:
/lgtm
/hold

infra/gcp/terraform/k8s-infra-prow-build-trusted/prow-build-trusted/resources/k8s-infra-test-pods/build-serviceaccounts.yaml
spiffxp: hmmm you know what, that's a very good question, I was just copy-pasting what was in prow-build

This was intended to be the default "is running on the trusted cluster" service account.  Like I think it corresponds to the `service-account` secret that gets mounted in to pods using podutils.  This allows them to write to gs://kubernetes-jenkins.  But I'm not sure that we actually have anything run as this explicitly via workload identity.  And you're probably right that we wouldn't want k8s-infra-prow to be able to trigger jobs that could write to `gs://kubernetes-jenkins`

I'm gonna remove this from the k8s-infra-test-pods NS across both clusters as a followup PR

The `k8s-infra-prow` account is, I'm guessing, intended to be the default "was triggered by k8s-infra prow" service account.  In terms of the GCP SA binding, I'm wondering if it's confusing to have jobs running potentially as the same SA on trusted and untrusted clusters.

spiffxp:
/hold cancel
Want to confirm this fixes the deploy

PR #2574, opened 2021-08-18
terraform: add and use workload-identity-service-account module

spiffxp:
Refactor out at least part of a common pattern, where the following three resources get created:
- a GCP SA
- a policy to bind the GCP SA to a K8s SA
- an IAM policy binding using the policy

This required a fair amount of `terraform state mv` commands which in retrospect I should have scripted

Point being I have not run `terraform apply` but the state changes are live

### DIFF
--- a/infra/gcp/terraform/modules/workload-identity-service-account/outputs.tf
+++ b/infra/gcp/terraform/modules/workload-identity-service-account/outputs.tf
@@ -1,0 +1,25 @@
+/*
+Copyright 2021 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+output "service_account" {
+  description = "The serviceaccount that was created"
+  value       = google_service_account.serviceaccount
+}
+
+output "iam_policy" {
+  description = "The serviceaccount iam_policy"
+  value       = google_service_account_iam_policy.serviceaccount_iam
+}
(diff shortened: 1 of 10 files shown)


infra/gcp/terraform/modules/workload-identity-service-account/outputs.tf
```diff
@@ -0,0 +1,25 @@
...
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+output "service_account" {
+  description = "The serviceaccount that was created"
+  value       = google_service_account.serviceaccount
```
ameukam, line 19:
```suggestion
  value       = google_service_account.serviceaccount.email
```

Or do we want to get a boolean value ? (the description suggest something like that ?

infra/gcp/terraform/modules/workload-identity-service-account/outputs.tf
```diff
@@ -0,0 +1,25 @@
...
+}
+
+output "iam_policy" {
+  description = "The serviceaccount iam_policy"
+  value       = google_service_account_iam_policy.serviceaccount_iam
```
ameukam, line 24: If you want to get the policy created : 

```suggestion
  value       = data.google_iam_policy.workload_identity.policy_data

infra/gcp/terraform/modules/workload-identity-service-account/outputs.tf
spiffxp: I thought we might want access to the full service account resource vs. just its email

So far I've only been using email though, so I can refactor to just that

infra/gcp/terraform/modules/workload-identity-service-account/outputs.tf
spiffxp: Again, I thought we might want the actual iam resource here, not just its policy data.  I'll rename the resource to signify the full iam resource, or bind to the policy data.  My concern with the latter is if we find a case where we need to add the policy bindings vs. set them

ameukam:
/lgtm
/hold
I'll rebase https://github.com/kubernetes/k8s.io/pull/2632. 
Remove hold when ready.

spiffxp:
Ran `terraform apply` for `k8s-infra-prow-build` and `k8s-infra-prow-build-trusted`

Issue #2575, opened 2021-08-18
Setting up Elekto on Kubernetes infra


jberkus:
For the 2021 Steering Committee election, we need to set up [Elekto](https://elekto.dev) running on the Kubernetes infra cluster.  Note that, once running, Elekto will keep running indefinitely, to host any elections within the community that we need to run.

I'm currently setting up container image auto-build according to feedback I got from @ameukam and will update this issue when that image is ready.  One question regarding the image, though: are there specific requirements for what base images I should use for Python applications?  Currently we're using Python:3.7.11-stretch.

Beyond that, we will need the following:

- one running pod hosting the application
- route/ingress to HTTPS
- DBAAS for the database for Elekto, either PostgreSQL or MySQL (could be self-hosted but I've been told we prefer dbaas)
- one configmap OR just hand-coded YAML
- a couple of secrets
- a GitHub token we can use for Oauth and repository checkout

Getting this started now so that we can assemble the requirements.  Thanks!

/sig contributor-experience
/commitee steering

ameukam:
cc @spiffxp @dims

ameukam:
> DBAAS for the database for Elekto, either PostgreSQL or MySQL (could be self-hosted but I've been told we prefer dbaas)

The current GCP DBaaS offer is [Cloud SQL](cloud.google.com/sql/). Can you give us more details about what's the needed configuration for elekto based of the existing features :  https://cloud.google.com/sql/docs/features ?
More detailed list : https://cloud.google.com/sql/docs/feature_support. 

> a GitHub token we can use for Oauth and repository checkout

Github tokens are responsibility of SIG Contribex (Github Administration Team). You can request it here : https://github.com/kubernetes/org/issues/new/choose

jberkus:
CloudSQL requirements are pretty simple: just a Cloud SQL instance, current MySQL or PostgreSQL, configured with name/password login from our cluster.  One database.  Storage requirement: 5GB max.

ameukam:
Opened https://github.com/kubernetes/k8s.io/pull/2587. cc @jberkus

jberkus:
Here's what we need for secrets and config for Elekto.  All of these need to be translated into the same-name ENV variables in the container (please remove {comments})

```
configmap:
  APP_NAME=k8s.elections
  APP_ENV=production
  APP_KEY=
  APP_DEBUG=False
  APP_URL=http://localhost
  APP_PORT=5000 {or whatever you want ingress to connect to}
[...]

PR #2587, opened 2021-08-20
Deploy infrastructure for k8s elections platform

ameukam:
Ref: https://github.com/kubernetes/k8s.io/issues/2575.

We will deploy those resources with terraform:
- Cloud SQL instance with PostreSQL internal to the default network

### DIFF
--- a/infra/gcp/bash/ensure-main-project.sh
+++ b/infra/gcp/bash/ensure-main-project.sh
@@ -107,6 +107,8 @@ readonly MAIN_PROJECT_SERVICES=(
     monitoring.googleapis.com
     # We host secrets in this project for use by prow and other apps
     secretmanager.googleapis.com
+    # We store elections results in Cloud SQL databases
+    sql-component.googleapis.com
     # We host public-facing and private GCS buckets in this project
     storage-api.googleapis.com
     # TODO: do we really need the legacy XML API enabled for them though?
--- a/infra/gcp/terraform/kubernetes-public/external-ips.tf
+++ b/infra/gcp/terraform/kubernetes-public/external-ips.tf
@@ -13,7 +13,7 @@ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 See the License for the specific language governing permissions and
 limitations under the License.
 */
- 
+
 /*
 This file defines all the external IP addresses in kubernetes-public
 */
@@ -29,6 +29,10 @@ locals {
     canary-v6 = {
       name = "k8s-io-ingress-canary-v6",
       ipv6 = true
+    },
+    elections = {
+      name        = "k8s-io-elections",
+      description = "Used for elections.k8s.io"
     },
     infra-prow = {
       name = "k8s-infra-prow",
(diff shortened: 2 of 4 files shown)


ameukam:
/hold
cc @spiffxp @dims 
Need to be deployed post-merge

infra/gcp/terraform/k8s-infra-elections/database.tf
```diff
@@ -0,0 +1,95 @@
...
+      enabled    = true
+      location   = var.cloudsql_backup_location
+      start_time = "02:13"
+    }
+
+    maintenance_window {
+      day          = 7
+      hour         = 2
+      update_track = "stable"
+    }
```
spiffxp, line 73: Would it be wise to have backups happen sometime that doesn't overlap with (possibly precedes) the maintenance window?

infra/gcp/terraform/k8s-infra-elections/main.tf
```diff
@@ -0,0 +1,64 @@
...
+  org_id          = data.google_organization.org.org_id
+  billing_account = data.google_billing_account.account.id
+}
+
+resource "google_project_service" "services" {
+  project = google_project.project.id
+
+  for_each = toset([
+    "cloudresourcemanager.googleapis.com",
+    "container.googleapis.com",
```
spiffxp, line 48: This is going to cause more services to get activated, like `containerregistry`,`compute`, etc.

I've been trying to list these out in our bash.  OTOH I'm pretty sure terraform gave up trying to be authoritative about the list of services enabled for a given project

infra/gcp/terraform/k8s-infra-elections/variables.tf
```diff
@@ -0,0 +1,119 @@
...
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+variable "region" {
+  type        = string
+  default     = "us-central1"
+  description = "The default region for resources in the project"
+}
```
spiffxp, line 21: Out of curiosity why variables vs. locals for all these?

infra/gcp/terraform/k8s-infra-elections/main.tf
```diff
@@ -0,0 +1,64 @@
...
+
+/*
+This file defines:
+- Google Project k8s-infra-elections to host the elections platform
+- Google services enabled for the project
+- GCP internal IP for the database k8s-infra-elections-db
+*/
+
+locals {
+  project_id = "k8s-infra-elections"
```
spiffxp, line 25: Why put this in its own project / module vs. adding more to the public project / module? Ostensibly you're provisioning a resource / service that an app on the public cluster is going to connect to.

spiffxp (commented):
My one big hangup is why not put this in kubernetes-public?

I also want an answer to the backup question before moving forward.

Otherwise the other questions are curiosity and the rest LGTM

infra/gcp/terraform/k8s-infra-elections/variables.tf
ameukam: Input validation is only available for variables : https://www.terraform.io/docs/language/values/variables.html#custom-validation-rules. Also I developed the habit to use locals for programmatically-derived values based on variables.

infra/gcp/terraform/k8s-infra-elections/main.tf
ameukam: This is a great question. I think I got triggered by the term "elections" and I wanted to apply some specific security measures because I don't think `elekto` should be trusted to run `aaa` but I realize now that the other workloads are at the same level of trust.

infra/gcp/terraform/k8s-infra-elections/database.tf
ameukam: Moved the maintenance window to Friday, 04:00 UTC.

ameukam:
10 GB is required for the DB 

```shell
╷
│ Error: Error, failed to create instance terraform-20210920204501028800000001: googleapi: Error 400: Invalid request: Invalid flag for instance role: disk size should be at least 10 GB.., invalid
│ 
│   with google_sql_database_instance.db_instance,
│   on k8s-elections-database.tf line 23, in resource "google_sql_database_instance" "db_instance":
│   23: resource "google_sql_database_instance" "db_instance" {
│ 
```

ameukam:
Opened https://github.com/kubernetes/k8s.io/pull/2758

Issue #1469, opened 2020-12-08
Umbrella issue: migrate away from google.com gcp projects


spiffxp:
An umbrella issue for wg-k8s-infra's mission: migrate project infrastructure away from the google.com GCP org, to the CNCF-owned/community-managed kubernetes.io GCP org.

Projects that host infrastructure (services, images, gcs buckets, service accounts, clusters, etc)
- k8s-federated-conformance
- k8s-gubernator
- k8s-mungegithub
- k8s-prow
- k8s-prow-builds
- k8s-tempelis
- k8s-testimages
- kubernetes-jenkins
- kubernetes-jenkins-pull
- kubernetes-site
- kubernetes-tools

Projects referenced by kubernetes/kubernetes CI
- authenticated-image-pulling
- cos-cloud
- gke-release
- k8s-authenticated-test
- kubernetes-ci-images
- kubernetes-e2e-test-images
- kubernetes-node-e2e-images
- TODO: identify the rest

Projects that host release or CI artifacts
- google-containers

Projects that are used for e2e testing (pinned to via `--gcp-project=foo`)
- TODO

Projects that are used for e2e testing (those managed by prow.k8s.io's "default" cluster's boskos)
- ref: https://github.com/kubernetes/test-infra/blob/master/config/prow/cluster/boskos-resources.yaml

/wg k8s-infra
/sig testing
/sig release
/area release-eng

Riaankl:
**Migrate away from google.com GCP project google-containers #1571** is related to this issue, but not hyperlinked in the description.
Adding this comment just to help to track related issues.

ameukam:
I found a bunch of GCP buckets that can potentially need be migrated:

- gs://kops-ci
- gs://cloud-sdk-testing
- gs://k8s-testimages-scratch
- gs://sig-scalability-logs
- gs://kubernetes-e2e-soak-configs
- gs://cri-containerd-staging

I couldn't find how they were created and which group(s) have admin access to them.

BenTheElder:
Also not sure k8s should be staging cri-containerd at all?

BenTheElder:
https://github.com/kubernetes/k8s.io/issues/1311 is done.

leilajal:
hey we just realized kubebuildr uses a google-internal project

ameukam:
/milestone v1.26
I'll try to do a few things before EOY.

BenTheElder:
Updated the list, most of these are done.

[...]

PR #2635, opened 2021-08-30
add initial pass at infra diagram

spiffxp:
It's likely incomplete in a number of places, some of which I marked
with TODO. This is loosely supposed to represent all of the
infrastructure we manage, and the infrastructure we have yet to migrate.

This was generated from hand written source, but using a google-internal
tool, so I'm not sure if I can share the source from which it was generated

The connections are laid out poorly by the tool, so I wouldn't pay a ton
of attention to them. Roughly speaking solid lines are the path code
flows as it turns into a kubernetes/kubernetes release artifact. The
dotted lines are asynchronous or polling checks. The sample use case I had
in my head was the first cut of an in-development release off of a release
branch, thus including some testgrid dashboard changes.

One thing missing off the top of my head is things that run as Prow Jobs,
so some detail are missing like:
- how k8s-triage is populated
- how k8s-metrics is populated
- how artifact promotion happens
- github automation like label_sync or peribolos

My preference would be to merge and iterate with this, because this is about
as far as I have time to push this for now. I'm open to feedback though,
and interested most specifically in things that are missing from this
diagram, including external services not directly managed by us.


spiffxp:
Rendered as an SVG for ease of review
k8s-infra-as-of-2021-q3

spiffxp:
/cc @listx 

FYI @kubernetes/release-engineering I didn't get around to enumerating all of the releng-related projects in this pass, I'll cc in more folks specifically the next time I've got bandwidth to dig into that

ameukam:
This is an amazing representation of what we are doing right now. We still discovering resources that need to be migrated (e.g. https://github.com/kubernetes/k8s.io/issues/1469#issuecomment-887041038). 

/lgtm

cpanato (commented):
nice
late /lgtm

PR #2641, opened 2021-08-31
Promote existing images

nckturner:
* Promote the existing alpha image releases.


wongma7:
/lgtm

plz paste the list-tags output in future like `$ gcloud container images list-tags gcr.io/k8s-staging-provider-aws/cloud-controller-manager --format='get(tags, digest)'` so it easier to verify digests

PR #2656, opened 2021-09-03
Fix the readme

mkumatag:
Entry in the `restrictions.yaml` file mandatory so this README.md doc needs to be updated with the same information.


ameukam:
/lgtm
/approve
Thanks!

Issue #2671, opened 2021-09-08
Auditors should not be able to bill arbitrary projects org-wide


spiffxp:
Pulling this out of a TODO from https://github.com/kubernetes/k8s.io/blob/fc60c860f1eac3c4eb2c33d78767a6eba27601ac/infra/gcp/bash/roles/specs/audit.viewer.yaml#L50-L53

Addressing this may break existing use cases so opening this to allow for tracking of any followup issues:
- things that should never have been allowed and any cleanup that needs to happen
- things that should be allowed and should explicitly use a billing project

/kind bug
/priority critical-urgent
/wg k8s-infra
/area access
/area audit
/assign

spiffxp:
/close
Audit job still works, and I haven't heard of anyone's workflow breaking.  Please /reopen and comment here if something broke for you

Issue #1590, opened 2021-01-28
Investigate community-accessible Data Studio reports and data sources


spiffxp:
The Kubernetes community should be able to create publicly viewable Data Studio reports and data sources that are billed to the kubernetes.io GCP org.

Data Studio ownership / billing is not completely intuitive to me.  AFAICT:
- there is some some tie-in with GSuite that allows org admins to restrict sharing outside the org (this is how google.com is setup)
- billing for bigquery usage by a data source can be assigned to a billing project
- access to a report doesn't guarantee access to the data sources used by the report
- creating a report with my gmail.com account allows me to share a report publicly

References:
- share data studio assets with viewers and editors: https://support.google.com/datastudio/answer/7459147?hl=en
- transfer data studio asset to new owner: https://support.google.com/datastudio/answer/9003614
- control access to data: https://support.google.com/datastudio/answer/6371135?hl=en
- connect to bigquery: https://support.google.com/datastudio/answer/6370296?hl=en

My proposal is that we try to set things up such that kubernetes-public is billed for whatever usage is necessary to generate/display reports.  WDYT @thockin @cblecker @dims @bartsmykla ?

/assign @thockin 
to put on your radar to answer where usage should be billed

/assign @MushuEE @spiffxp 
I'm happy to help wield credentials and enable whatever API's are needed, but not sure I can guarantee dedicated time to work on this.  @MushuEE has been experimenting with a Data Studio report to replace the dashboards we used to provide via velodrome.k8s.io, but is currently held back by the assets being google.com-only

/wg k8s-infra
/sig testing
/area billing
/area access
/priority important-longterm

MushuEE:
Much appreciated. I will continue investigation

MushuEE:
```
For example, you can turn link sharing on to share Data Studio reports with non-Google accounts, include them in emails, blogs or social posts, or embed them in your website.
```
From [HERE](https://support.google.com/datastudio/answer/7459147?hl=en)

It also looks like you can specify a Project ID for billing within a specific report. 
```
Advanced options
Override Billing Project ID

Specify a different Billing Project ID for this custom query.
```
[HERE](https://support.google.com/datastudio/answer/6370296?hl=en)

spiffxp:
/milestone v1.23
/unassign @MushuEE @thockin @spiffxp 
All of our bandwidth has been elsewhere. While I'm interested in this, I'm not currently working on it; I'll /assign myself if that ever changes

[...]

PR #2672, opened 2021-09-08
roles: drop serviceusage.services.com from audit.viewer role

spiffxp:
Related:
- Part of: https://github.com/kubernetes/k8s.io/issues/2671 
- Noticed this while investigating data studio reports: https://github.com/kubernetes/k8s.io/issues/1590

 This permission was effectively letting auditors use any project for 
 billing purposes, we should stop that


spiffxp:
I have already run `./infra/gcp/bash/ensure-organization.sh` to get this deployed

cpanato (approved):
/lgtm

PR #2682, opened 2021-09-09
terraform/kubernetes-public-pii: fix permissions for gs://k8s-infra-artifacts-logs

spiffxp:
Related:
- Part of: https://github.com/kubernetes/k8s.io/issues/904
- Part of: https://github.com/kubernetes/k8s.io/issues/1968
- Noticed via: https://github.com/kubernetes/k8s.io/issues/904#issuecomment-891443463

See commits for details.  I'm not sure the permissions have ever been quite correct on this bucket.  This should get them to a place where folks authorized for PII can read the bucket, and org admins can write to the bucket (at least for long enough to complete migration from the old gcslogs bucket)


spiffxp:
Found a typo.  Ran `terraform apply` using https://github.com/kubernetes/k8s.io/pull/2686

PR #2685, opened 2021-09-09
images/k8s-infra: bump dependencies

ameukam:
conftest: https://github.com/open-policy-agent/conftest/releases/tag/v0.27.0
github CLI: https://github.com/cli/cli/releases/tag/v2.0.0
golang 1.17: https://golang.org/doc/go1.17
gcloud: https://cloud.google.com/sdk/docs/release-notes#35500_2021-08-31
OPA: https://github.com/open-policy-agent/opa/blob/main/CHANGELOG.md#0320
kubectl : https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG/CHANGELOG-1.21.md
 - All the GKE clusters are now running Kubernetes 1.20.8

Signed-off-by: Arnaud Meukam <ameukam@gmail.com>

PR #2687, opened 2021-09-09
groups: update to go1.17

spiffxp:
Related:
- Followup to: https://github.com/kubernetes/k8s.io/pull/2685

### DIFF
--- a/groups/go.mod
+++ b/groups/go.mod
@@ -9,7 +9,7 @@ require (
 	golang.org/x/oauth2 v0.0.0-20200107190931-bf48bf16ab8d
 	google.golang.org/api v0.20.0
 	google.golang.org/genproto v0.0.0-20200429120912-1f37eeb960b2
-	gopkg.in/yaml.v2 v2.2.2
+	gopkg.in/yaml.v3 v3.0.0-20190709130402-674ba3eaed22
 	k8s.io/apimachinery v0.0.0-20190817020851-f2f3a405f61d
 	k8s.io/test-infra v0.0.0-20191024183346-202cefeb6ff5
 )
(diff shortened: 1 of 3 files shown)


ameukam:
/lgtm
/hold
Remove hold when ready.

groups/go.mod
```diff
@@ -14,4 +14,18 @@ require (
...
+	github.com/golang/groupcache v0.0.0-20200121045136-8c9f03a8e57e // indirect
+	github.com/golang/protobuf v1.3.5 // indirect
+	github.com/google/go-cmp v0.4.0 // indirect
+	github.com/googleapis/gax-go/v2 v2.0.5 // indirect
+	go.opencensus.io v0.22.3 // indirect
+	golang.org/x/sys v0.0.0-20200331124033-c3d80250170d // indirect
+	golang.org/x/text v0.3.2 // indirect
+	google.golang.org/appengine v1.6.5 // indirect
+	google.golang.org/grpc v1.28.0 // indirect
+	gopkg.in/yaml.v3 v3.0.0-20190709130402-674ba3eaed22 // indirect
```
ameukam, line 28: We need to have 2 versions of this package ? (I feel like the answer is yes because go modules)

spiffxp:
```
$ go mod why gopkg.in/yaml.v3
# gopkg.in/yaml.v3
k8s.io/k8s.io/groups
k8s.io/test-infra/pkg/genyaml
gopkg.in/yaml.v3
```

spiffxp: Which is used for `go run reconcile.go --print` although to be fair the only thing is getting us at the moment is a preamble of 
```
# This file has the list of groups in kubernetes.io gsuite org that we use
# for granting permissions to various community resources.
```
before the groups are dumped out.  This is because the `GoogleGroup` type doesn't have any commented fields.

spiffxp: I opted to depend directly on v3 in the reconciler, there weren't any breaking changes in our usage.

```
$ go mod why gopkg.in/yaml.v3
# gopkg.in/yaml.v3
k8s.io/k8s.io/groups
gopkg.in/yaml.v3
```

Issue #1675, opened 2021-02-17
audit followup: k8s-staging-* projects should not have misc services enabled


spiffxp:
ref: https://github.com/kubernetes/k8s.io/pull/1534/commits/cf5f5e3360c76e660e0af8be25059e0b9d732339 and https://github.com/kubernetes/k8s.io/pull/1534#discussion_r577242644

Staging projects run GCB, they shouldn't need compute, bigquery, iam, monitoring, oslogin, etc...

We should consider explicitly disabling:
- compute
- dns
- bigquery
- etc.

We could go bigger and explicitly disable anything not intended to be enabled.

How/when to do this? I see a few options:
- A) run this script prior to an audit (but now the job running audit needs a lot more permissions)
- B) make this part of ensure-staging-storage.sh (so humans will cleanup when they run the script)
- C) run ensure-staging-storage.sh periodically (means setting up an SA to do this, and deciding we're ready to trust CI runs of untested bash)

I prefer option B, and then option C.  And not "going big" until we've gained confidence that we're not going to foot-gun ourselves.  

Would also be curious what a terraform or crossplane approach to "ensure only these services are enabled and no others" looks like.

/wg k8s-infra
/sig release
/area release-eng
/priority important-soon
/milestone v1.21

spiffxp:
/assign
https://github.com/kubernetes/k8s.io/pull/1859 should allow me to address, just want to be careful about rolling it out

spiffxp:
gathered output related to extra services from https://github.com/kubernetes/k8s.io/pull/1859#issuecomment-814430035 into https://gist.githubusercontent.com/spiffxp/1cbf779d7dc1c025a445b91909f55bf7/raw/49971cd6699c4dace4f47fcba8c068c212e0ad2e/k8s-staging-service-disable-plan.yaml

```yaml
# curl -s https://gist.githubusercontent.com/spiffxp/1cbf779d7dc1c025a445b91909f55bf7/raw/49971cd6699c4dace4f47fcba8c068c212e0ad2e/k8s-staging-service-disable-plan.yaml | yq -y 'with_entries(.value |= .to_disable)'
k8s-staging-artifact-promoter:
  - compute.googleapis.com
  - oslogin.googleapis.com
k8s-staging-cip-test:
  - bigquery.googleapis.com
  - bigquerystorage.googleapis.com
  - cloudapis.googleapis.com
  - clouddebugger.googleapis.com
  - cloudtrace.googleapis.com
[...]

PR #2688, opened 2021-09-09
bash/staging: update expected services

spiffxp:
Related:
- part of: https://github.com/kubernetes/k8s.io/issues/1675

At some point in the past few months, `gcloud services` stopped returning dependency information. This means our scripts don't implicitly expect those services anymore, and list them under `to_disable`.

I manually discovered and encoded the dependencies by trying to remove a service our scripts log as unexpected, and seeing what the error returned.


spiffxp:
I'm running `ensure-staging-storage.sh` for everything to see what other services show up in the `to_disable` stanza

Issue #1963, opened 2021-04-22
Disable Vulnerability Scanning


dims:
Just got a nice note from the friendly GCP folks:
```
[Action Required] Free Container Scanning ends May 19, 2021; Normal billing begins July 1, 2021
```

I was able to click on disable for `k8s-artifacts-prod`, but looks like we need to do this across the board automatically.

dims:
cc @thockin @spiffxp

spiffxp:
It's now May 19 and I haven't been able to land #2016 yet.  I will reprioritize this.

In the meantime, let's manually disable it for the project that generates the majority of costs
```
$ gcloud services disable containerscanning.googleapis.com --project=k8s-staging-ci-images
Provide the --force flag if you wish to force disable services.
ERROR: (gcloud.services.disable) FAILED_PRECONDITION: The service containerscanning.googleapis.com has usage in the last 30 days. Please specify check_if_service_has_usage=SKIP if you want to proceed with disabling the service.
```

OK, let's use `--force` as suggested
```
$ gcloud services disable --force containerscanning.googleapis.com --project=k8s-staging-ci-images
Operation "operations/acf.p17-731599680865-1c03d2a4-d620-4b90-b319-65f44b3ac2c6" finished successfully
```

justaugustus:
Thanks for the [heads-up](https://kubernetes.slack.com/archives/CJH2GBF7Y/p1621470584037700), Aaron!
@kubernetes/release-engineering -- We should be fine here, but please keep an eye out for anything strange in our GCP projects.

spiffxp:
https://github.com/kubernetes/k8s.io/pull/2016 disabled `containerscanning.googleapis.com` for all projects

there is leftover detritus in the form of a container scanning service agent IAM binding

spiffxp:
/priority awaiting-more-evidence
/remove-priority important-soon
/milestone v1.23
to remove the service agent IAM binding

spiffxp:
Based on a survey of all staging projects, I'm also going ahead and disabling the `containeranalysis` API
```bash
#!/usr/bin/env bash

gcloud services disable --force containeranalysis.googleapis.com --project=k8s-staging-addon-manager
gcloud services disable --force containeranalysis.googleapis.com --project=k8s-staging-apisnoop
gcloud services disable --force containeranalysis.googleapis.com --project=k8s-staging-artifact-promoter
[...]

PR #2697, opened 2021-09-10
bash/staging: remove containeranalysis service agent binding

spiffxp:
Related:
- part of: https://github.com/kubernetes/k8s.io/issues/1963

Now that the service has been disabled, let's clean up the IAM binding for the servie agent


spiffxp:
Ran `infra/gcp/bash/ensure-staging-storage.sh` modified to call just the containeranalaysis removal

PR #2710, opened 2021-09-13
[WIP] Add Public Log Vendor Matcher

BobyMCbobs:
Adds the (pii) pipeline that ii runs to generate the data used for discovering the cloud-providers/vendors/companies requesting Kubernetes public artifacts, based on the Kubernetes Public PII logs


images/public-log-vendor-matcher/Dockerfile
```diff
@@ -0,0 +1,33 @@
+FROM golang:1.16.5-stretch AS godeps
+RUN go get -u github.com/mikefarah/yq/v4 && \
+  test -f /go/bin/yq
```
spiffxp, line 3: Is there any reason this can't use the same version of yq used by the k8s-infra image?

images/public-log-vendor-matcher/Dockerfile
```diff
@@ -0,0 +1,33 @@
...
+  gettext-base \
+  procps \
+  google-cloud-sdk && \
+  rm -rf /var/lib/apt/lists/*
+RUN pip3 install pyasn
+WORKDIR /app
+COPY --from=godeps /go/bin/yq /usr/local/bin/yq
+COPY ./pg-init.d /docker-entrypoint-initdb.d
+COPY ./app .
+ENV POSTGRES_PASSWORD=postgres
```
spiffxp, line 33: Why?

images/public-log-vendor-matcher/Dockerfile
```diff
@@ -0,0 +1,33 @@
...
+  libc6-dev \
+  gettext-base \
+  procps \
+  google-cloud-sdk && \
+  rm -rf /var/lib/apt/lists/*
+RUN pip3 install pyasn
+WORKDIR /app
+COPY --from=godeps /go/bin/yq /usr/local/bin/yq
+COPY ./pg-init.d /docker-entrypoint-initdb.d
+COPY ./app .
```
spiffxp, line 32: What is this and how is it built?

images/public-log-vendor-matcher/pg-init.d/lib-06-bq-load-logs.sh
```diff
@@ -0,0 +1,25 @@
+#!/bin/bash
+
+# Load the k8s-infra-artifacts-gcslogs logs into the usage_all_raw dataset
+if [ -z "${GCP_BIGQUERY_DATASET_LOGS:-}" ]; then
+    echo "Using dataset logs, since \$GCP_BIGQUERY_DATASET_LOGS was provided and set to '$GCP_BIGQUERY_DATASET_LOGS'"
+    BUCKETS=(
```
spiffxp, line 6: We should be including all of the staging buckets as well

images/public-log-vendor-matcher/pg-init.d/lib-15-stop-postgres.sh
```diff
@@ -0,0 +1,23 @@
+#!/bin/bash
+
+# Stop Postgres
```
spiffxp, line 3: So it doesn't seem like postgres is actually used?  Can we use something other than pg-init.d as our entrypoint?  Like this appears at a glance to be a set of scripts that are invoked in order.  

I would rather see something like `/app/entrypoint.sh` that runs the scripts in the expected order

images/public-log-vendor-matcher/pg-init.d/lib-00-prepare.sh
```diff
@@ -0,0 +1,63 @@
...
+project_id = ${GCP_PROJECT}
+EOF
+
+gcloud config set project "${GCP_PROJECT}"
+
+## This is just to continue testing wile I wait for permissions for the service account
+## Use the activate-service-account live once it has permissions
+## The container is being run it so it should let me manually do the auth
+# gcloud auth login
+gcloud auth activate-service-account "${GCP_SERVICEACCOUNT}" --key-file="${GOOGLE_APPLICATION_CREDENTIALS}"
```
spiffxp, line 18: You shouldn't need to specify the name of the service account, e.g. https://github.com/kubernetes/test-infra/blob/3896f0bc47a2fa00e20f7c6ef3640681314a78c6/metrics/bigquery.py#L154-L157
```suggestion
gcloud auth activate-service-account  --key-file="${GOOGLE_APPLICATION_CREDENTIALS}"
```

images/public-log-vendor-matcher/README.md
```diff
@@ -0,0 +1,44 @@
...
+sudo chmod 0777 "${TMP_DIR_ETL}"
+sudo chown 999 /tmp/asn-etl-pipeline-gcp-sa.json
+docker run \
+    -it \
+    --rm \
+    -e TZ=$TZ \
+    -e POSTGRES_PASSWORD="postgres" \
+    -e GOOGLE_APPLICATION_CREDENTIALS=/tmp/asn-etl-pipeline-gcp-sa.json \
+    -e GCP_PROJECT=k8s-infra-ii-sandbox \
+    -e GCP_SERVICEACCOUNT=asn-etl@k8s-infra-ii-sandbox.iam.gserviceaccount.com \
```
spiffxp, line 38:
```suggestion
```

images/public-log-vendor-matcher/app/ext-ip-asn.sql
```diff
@@ -0,0 +1,9 @@
+SELECT
+    asn as asn,
+    ip as cidr_ip,
+    ip_start as start_ip,
+    ip_end as end_ip,
+    NET.IPV4_TO_INT64(NET.IP_FROM_STRING(ip_start)) AS start_ip_int,
+    NET.IPV4_TO_INT64(NET.IP_FROM_STRING(ip_end)) AS end_ip_int
+    FROM `k8s-infra-ii-sandbox.${GCP_BIGQUERY_DATASET_WITH_DATE}.pyasn_ip_asn_extended`
```
spiffxp, line 8: Not clear the explicit gcp project needs to be specified here, but it shouldn't be a hardcode
```suggestion
    FROM `${GCP_PROJECT}.${GCP_BIGQUERY_DATASET_WITH_DATE}.pyasn_ip_asn_extended`
```

images/public-log-vendor-matcher/README.md
```diff
@@ -0,0 +1,44 @@
...
+
+## Environment variables
+
+| Name                             | Default                                                | Description                                                                           |
+| -------------------------------- | ------------------------------------------------------ | ------------------------------------------------------------------------------------- |
+| `TZ`                             | ``                                                     | Container time zone                                                                   |
+| `GOOGLE_APPLICATION_CREDENTIALS` | ``                                                     | The path to the GCP service account json key                                          |
+| `GCP_PROJECT`                    | `k8s-infra-ii-sandbox`                                 | The project to target                                                                 |
+| `GCP_SERVICEACCOUNT`             | `asn-etl@k8s-infra-ii-sandbox.iam.gserviceaccount.com` | The GCP service account name                                                          |
```
spiffxp, line 12:
```suggestion
```

images/public-log-vendor-matcher/README.md
```diff
@@ -0,0 +1,44 @@
...
+
+A Postgres-based k8s-infra data pipeline that produces BigQuery tables for reviewing Kubernetes Public artifact traffic in DataStudio.
+
+## Environment variables
+
+| Name                             | Default                                                | Description                                                                           |
+| -------------------------------- | ------------------------------------------------------ | ------------------------------------------------------------------------------------- |
+| `TZ`                             | ``                                                     | Container time zone                                                                   |
+| `GOOGLE_APPLICATION_CREDENTIALS` | ``                                                     | The path to the GCP service account json key                                          |
+| `GCP_PROJECT`                    | `k8s-infra-ii-sandbox`                                 | The project to target                                                                 |
```
spiffxp, line 11: This feels ambiguous...
- the project that hosts `GCP_BIGQUERY_DATASET`?
- the project which is being billed for writing this data?

images/public-log-vendor-matcher/README.md
```diff
@@ -0,0 +1,44 @@
+# Public log vendor matcher
+
+A Postgres-based k8s-infra data pipeline that produces BigQuery tables for reviewing Kubernetes Public artifact traffic in DataStudio.
```
spiffxp, line 3: Need more of a description of the set of steps this takes.

(totally making this up)
1. scrapes logs from $place
2. anonymizes the following fields:
  - $foo by doing $x
  - $bar by doing $y
3. write the resulting anonymized data to $other_place
4. aggregates the data by doing $z and writes results to $place_that_could_be_promoted

the $promoted_place houses A B and C and is intended to used by E to display F

images/public-log-vendor-matcher/README.md
```diff
@@ -0,0 +1,44 @@
...
+| `NO_PROMOTE`                     | ``                                                     | Disable the promotion of `${GCP_BIGQUERY_DATASET}_${DATE}` to ${GCP_BIGQUERY_DATASET} |
+| `ASN_DATA_PIPELINE_RETAIN`       | ``                                                     | Keeps Postgres running after the job has completed                                    |
+
+## Running the Pipeline Manually
+
+Generate a key file for ServiceAccount auth
+
+```
+gcloud iam service-accounts keys create /tmp/asn-etl-pipeline-gcp-sa.json --iam-account=asn-etl@k8s-infra-ii-sandbox.iam.gserviceaccount.com
+```
```
spiffxp, line 23: We need to not do this.  I do not want these keys laying around.

Realistically if you're running this as a developer, it should be running with _your_ credentials, not as a service account.

As a super last resort if you truly need to repro acting as the service account, we could investigate using `--impersonate-service-account` (ref: https://cloud.google.com/sdk/gcloud/reference/#--impersonate-service-account)

images/public-log-vendor-matcher/pg-init.d/lib-00-prepare.sh
```diff
@@ -0,0 +1,63 @@
...
+bq rm -r -f "${GCP_BIGQUERY_DATASET}_${PIPELINE_DATE}" || true
+
+# initalise a new data set with the given name
+bq mk \
+    --dataset \
+    --description "Kubernetes Public artifact traffic, related to CNCF supporting vendors of k8s infrastructure (${PIPELINE_DATE})" \
+    "${GCP_PROJECT}:${GCP_BIGQUERY_DATASET}_${PIPELINE_DATE}"
+
+if [ ! -f "/tmp/potaroo_data.csv" ]; then
+    gsutil cp gs://ii_bq_scratch_dump/potaroo_company_asn.csv  /tmp/potaroo_data.csv
```
spiffxp, line 37: Seems like a hardcode we want to avoid.  Would it make sense to bake this in the image?

images/public-log-vendor-matcher/pg-init.d/lib-00-prepare.sh
```diff
@@ -0,0 +1,63 @@
...
+# curl -s  https://bgp.potaroo.net/cidr/autnums.html | sed -nre '/AS[0-9]/s/.*as=([^&]+)&.*">([^<]+)<\/a> ([^,]+), (.*)/"\1", "\3", "\4"/p'  | head
+
+# Remove the previous data set
+bq rm -r -f "${GCP_BIGQUERY_DATASET}_${PIPELINE_DATE}" || true
+
+# initalise a new data set with the given name
+bq mk \
+    --dataset \
+    --description "Kubernetes Public artifact traffic, related to CNCF supporting vendors of k8s infrastructure (${PIPELINE_DATE})" \
+    "${GCP_PROJECT}:${GCP_BIGQUERY_DATASET}_${PIPELINE_DATE}"
```
spiffxp, line 34: Seems like all the ASN stuff below could be in its own file, thus leaving this and everything above to be "dataset init" related

images/public-log-vendor-matcher/pg-init.d/lib-01-migrate-schemas.sql
```diff
@@ -0,0 +1,37 @@
+begin;
+
+create table if not exists cust_ip (
+  c_ip bigint not null
+);
```
spiffxp, line 5: So here's my confusion: is this happening in BQ? Or in postgres?  I am very anti do-it-in-postgres when it comes to handling PII, because I don't want a stray container lying around with PII in it

spiffxp (changes requested):
I need a better understanding of what this is trying to do.  As-written this needs more explanation.

I can set some time aside my thursday (NZ friday) to go over this in person with someone.

images/public-log-vendor-matcher/Dockerfile
BobyMCbobs: I like single binaries, but that would be easy enough to swap it out.

images/public-log-vendor-matcher/Dockerfile
BobyMCbobs: Postgres fails to launch without this variable set. Since it's an ephemeral job, it don't believe it matters so much

images/public-log-vendor-matcher/Dockerfile
BobyMCbobs: What do you mean by this sorry?

images/public-log-vendor-matcher/pg-init.d/lib-06-bq-load-logs.sh
BobyMCbobs: Where is a list of the buckets?

images/public-log-vendor-matcher/pg-init.d/lib-15-stop-postgres.sh
BobyMCbobs: Postgres is used for preparing the data from the ASN providers and trimming some data down, such as
- https://github.com/kubernetes/k8s.io/pull/2710/files#diff-66e7079932b4bc08f013e6db34207fdce92d36623328555e1362bc5ad167087eR1
- https://github.com/kubernetes/k8s.io/pull/2710/files#diff-338e9f84ded10fcacd04517dc8b9c9581dd05a17889263e520d666753f3381f3R1
- https://github.com/kubernetes/k8s.io/pull/2710/files#diff-30b2418800710c115d14c538080097c6a6f78ea0eaf9816f6fd006e11a983b0bR1

I like the idea of using the primitives that Postgres provides, with the init scripts. It works quite well

images/public-log-vendor-matcher/pg-init.d/lib-00-prepare.sh
BobyMCbobs: Removed in ad9fe407

images/public-log-vendor-matcher/README.md
BobyMCbobs: Removed in ad9fe407

images/public-log-vendor-matcher/README.md
BobyMCbobs: Removed in ad9fe407

images/public-log-vendor-matcher/app/ext-ip-asn.sql
BobyMCbobs: Ah, whoopses that shouldn't be there. Updated in d4e73426

images/public-log-vendor-matcher/Dockerfile
spiffxp: There's an app directory, this is looked at by postgres

images/public-log-vendor-matcher/README.md
spiffxp: so e.g. describe the pg-init.d stuff here, and what gets invokled when, and how the app directory gets invoked, etc.

images/public-log-vendor-matcher/pg-init.d/lib-03-load-into-a-bigquery-dataset.sh
```diff
@@ -0,0 +1,67 @@
...
+VENDORS=(
+    microsoft
+    google
+    amazon
+    alibabagroup
+    baidu
+    digitalocean
+    equinixmetal
+    huawei
+    tencentcloud
```
spiffxp, line 25: alpha sort

images/public-log-vendor-matcher/pg-init.d/lib-06-bq-load-logs.sh
spiffxp: `ls -1 audit/projects/*/buckets/ | grep ^k8s-staging`

spiffxp: that misses the gcr buckets but close enough to start

images/public-log-vendor-matcher/pg-init.d/lib-08-download-c-ip-int.sh
```diff
@@ -0,0 +1,19 @@
+#!/bin/bash
+
+## Set a timestamp to work with
+TIMESTAMP=$(date +%Y%m%d%H%M)
+echo "$TIMESTAMP" > /tmp/my-timestamp.txt
+## Dump the entire table to gcs
+bq extract \
+  --destination_format CSV \
+  "${GCP_BIGQUERY_DATASET}_${PIPELINE_DATE}.2a_ip_int" \
+  "gs://ii_bq_scratch_dump/2a_ip_inti-$TIMESTAMP-*.csv" > "${BQ_OUTPUT:-/dev/null}" 2>&1
```
spiffxp, line 10: found another hardcode, and below

images/public-log-vendor-matcher/pg-init.d/lib-15-stop-postgres.sh
spiffxp: as long as you can explain what this does, and as long as we run this someplace where an ephemeral container won't be accessible to others and potentially leak data, then I'm ok relying on what postgres provides out of the box

spiffxp:
We discussed offline and I'm ok with using an in-image Postgres if triggered as a cloud build

images/public-log-vendor-matcher/pg-init.d/lib-03-load-into-a-bigquery-dataset.sh
BobyMCbobs: Now it's all sorted :wink:

images/public-log-vendor-matcher/README.md
```diff
@@ -0,0 +1,49 @@
+# Public log vendor matcher
+
+A Postgres-based k8s-infra data pipeline that produces BigQuery tables for reviewing Kubernetes Public artifact traffic in DataStudio.
+
+## Flow (WIP)
+
+1. Kubernetes Public GCS buckets for artifacts are configured to write public access logs to a GCS bucket called _k8s-infra-artifacts-gcslogs_ in the _k8s-infra-public-pii_ project
```
spiffxp, line 7:
```suggestion
1. Kubernetes Public GCS buckets for artifacts are configured to write public access logs to a GCS bucket called `k8s-infra-artifacts-gcslogs` in the `k8s-infra-public-pii` project
```

images/public-log-vendor-matcher/README.md
spiffxp: pg-init.d still isn't described in sufficient detail, I'm looking something on the level of

- when the container starts
- something launches postgres with the password baked in the container
- something kicks off pg-init.d
- pg-init.d then does the following
  - runs through all files in /foo/bar in alphabetical order
    - if the file is executable, runs it
    - if the file is sql, executes it against the local postgres instance
 - if all files successfully executed, runs /app/foo/bar

etc. (I made all that up, maybe that's not how it actually works)

images/public-log-vendor-matcher/pg-init.d/lib-07-bq-load-logs.sh
```diff
@@ -0,0 +1,25 @@
+#!/bin/bash
+
+# Load the k8s-infra-artifacts-gcslogs logs into the usage_all_raw dataset
+if [ -z "${GCP_BIGQUERY_DATASET_LOGS:-}" ]; then
+    echo "Using dataset logs, since \$GCP_BIGQUERY_DATASET_LOGS was provided and set to '${GCP_BIGQUERY_DATASET_LOGS:-}'"
+    BUCKETS=(
```
spiffxp, line 6: still missing staging buckets here but could be done as followup

images/public-log-vendor-matcher/pg-init.d/lib-16-stop-postgres.sh
```diff
@@ -0,0 +1,23 @@
+#!/bin/bash
+
+# Stop Postgres
+
+[ ! "${ASN_DATA_PIPELINE_RETAIN:-}" = true ] && \
+# in the Postgres container image,
+# the command run changes to "postgres" once it's completed loading up
+# and is in a ready state and all of the init scripts have run
+#
+# here we wait for that state and attempt to exit cleanly, without error
```
spiffxp, line 10: ... and we do that because ___ ?  The answer to this also probably belongs in the README

images/public-log-vendor-matcher/Dockerfile
BobyMCbobs: If a password isn't set, the container won't start

images/public-log-vendor-matcher/pg-init.d/lib-00-prepare.sh
BobyMCbobs: This has been swapped out for fetching the data from Potaroo instead of a bucket

images/public-log-vendor-matcher/pg-init.d/lib-07-bq-load-logs.sh
BobyMCbobs: There's now a script to manually generate the list of buckets to pull usage from (as `${BUCKET}_usage`)

https://github.com/kubernetes/k8s.io/pull/2710/commits/98bd4dba50c0613eb7304e20fb6413cc93e7efc2

images/public-log-vendor-matcher/pg-init.d/lib-16-stop-postgres.sh
BobyMCbobs: Updated https://github.com/kubernetes/k8s.io/pull/2710/commits/c50490d7defdf48e79f1020326d3884fce9a9210

images/public-log-vendor-matcher/Dockerfile
BobyMCbobs: This has been swapped out in cf5172cc47e96bd721a4e983e0a194b28759f83f

images/public-log-vendor-matcher/README.md
BobyMCbobs: Updated in https://github.com/kubernetes/k8s.io/commit/b767950c205f03cc8345447d31032a6cd2f10b36

images/public-log-vendor-matcher/pg-init.d/lib-08-download-c-ip-int.sh
BobyMCbobs: This has been swapped out in https://github.com/kubernetes/k8s.io/commit/81f93b365f24ff638486c6cc2ae9b3c6d7e26e56

images/public-log-vendor-matcher/README.md
BobyMCbobs: Updated

BobyMCbobs:
@spiffxp, I think I've managed to update per all comments.
If you think this is ready, I can do a squash on all the commits.

images/public-log-vendor-matcher/README.md
spiffxp: https://github.com/kubernetes/k8s.io/pull/2710/files#r718612992 still isn't addressed AFAICT, I don't see any references to pg-init.d in this README

images/public-log-vendor-matcher/pg-init.d/lib-16-stop-postgres.sh
spiffxp: I think I'm looking for answer like "postgres now stays up and running, but we're running this container as a one-shot job and using pg-init.d as the driver for all of the logic in this container. Once the command changes to postgres, we know pg-init.d has finished executing all of our scripts and sql, so we're ready to exit this container cleanly to signify to prow or whatever CI system runs this that that the import/transform has finished" but cleaned up to be less rambling

spiffxp (changes requested):
I still feel like pg-init.d remains undocumented magic here

BobyMCbobs:
> https://github.com/kubernetes/k8s.io/pull/2710/files#r718612992 still isn't addressed AFAICT, I don't see any references to pg-init.d in this README

@spiffxp, added comments in 24f40e2

BobyMCbobs:
> I think I'm looking for answer like "postgres now stays up and running, but we're running this container as a one-shot job and using pg-init.d as the driver for all of the logic in this container. Once the command changes to postgres, we know pg-init.d has finished executing all of our scripts and sql, so we're ready to exit this container cleanly to signify to prow or whatever CI system runs this that that the import/transform has finished" but cleaned up to be less rambling

Added some hopefully clearer description for the stop-postgres step in 55f09dbc

spiffxp:
Alternatively, I've opened up https://github.com/kubernetes/k8s.io/pull/2906 to fix the "vendor" issue, though I still wonder if a  new name would be better

BobyMCbobs:
> > I just realized... all of the python and bash in here should be failing our boilerplate check, and it's not?
> 
> So this tripped a subtle bug in verify_boilerplate.py in repo-infra. If "vendor" is anywhere in the path name of a file, it will exclude it from being checked.
> 
> Can we rename this to something else like `images/public-log-asn-matcher`? Seems a better fit anyway, since we're trying to map ips to asns

That sounds fine.

Currently getting errors saying 
```
17 files have incorrect boilerplate headers:
images/public-log-asn-matcher/Dockerfile
images/public-log-asn-matcher/app/asn-autnums-extractor.py
images/public-log-asn-matcher/app/ip-from-pyasn.py
images/public-log-asn-matcher/generate-buckets.sh
images/public-log-asn-matcher/pg-init.d/lib-00-prepare.sh
images/public-log-asn-matcher/pg-init.d/lib-01-prepare-company-asn-data.sh
images/public-log-asn-matcher/pg-init.d/lib-04-load-into-a-bigquery-dataset.sh
images/public-log-asn-matcher/pg-init.d/lib-06-bq-load-metadata.sh
images/public-log-asn-matcher/pg-init.d/lib-07-bq-load-logs.sh
images/public-log-asn-matcher/pg-init.d/lib-08-bq-usage-data-transformation.sh
images/public-log-asn-matcher/pg-init.d/lib-09-download-c-ip-int.sh
images/public-log-asn-matcher/pg-init.d/lib-10-download-expanded-ips.sh
images/public-log-asn-matcher/pg-init.d/lib-12-upload-ip-range-2-ip.sh
images/public-log-asn-matcher/pg-init.d/lib-13-add-c-ip-int-to-usage-all.sh
images/public-log-asn-matcher/pg-init.d/lib-14-prepare-final-table.sh
images/public-log-asn-matcher/pg-init.d/lib-15-promote-bq-dataset-as-prod.sh
images/public-log-asn-matcher/pg-init.d/lib-16-stop-postgres.sh
```

might you have an idea @spiffxp?

spiffxp:
>might you have an idea @spiffxp?

Looks like you figured it out, this was the missing boilerplate I had expected the job to error on.

Issue #420, opened 2019-10-25
groups: needs unit tests


cblecker:
The groups management scripts in `groups/` need unit tests.

sgmac:
I was looking into this, but I don't know how to mock the calls to Admin Directory, do you have any suggestion? For instance, I wanted to work on the one below. 

```Go
func removeMembersFromGroup(service *admin.Service, groupEmailId string, members []string) error ```

**Comment by cblecker** [comment]
It probably needs a refactoring of the code to make this viable. Things like the client services refactored into an interface, so a fake client can be mocked.

**Comment by sgmac** [comment]
I would like to work on this, do I require some sort of access to Admin API? At least to test I don't break anything while refactoring. Let me know what you think.

**Comment by cblecker** [comment]
@sgmac I'd love the help! Unfortunately, access to the Admin API is highly privileged. If you open a PR I can work to test any changes with you and confirm they work as expected.

**Comment by sgmac** [comment]
blind coding such a challenge !! The only issue I see here is that 

```Go
srv *admin.Service
```
I'm not very experienced in Go but I took a look  *admin.Service, and has not single method definition but concrete types. 

https://github.com/googleapis/google-api-go-client/blob/master/admin/directory/v1/admin-gen.go#L249

mayankkapoor:
@sgmac  I'm starting to give the refactoring a shot. Let me know if you want to collaborate on this.
@cblecker Any advice is appreciated.

sgmac:
@mayankkapoor looking forward to help, please go ahead and share your branch so I can take a look how you are approaching this.

mayankkapoor:
@sgmac I've opened a PR here: https://github.com/kubernetes/k8s.io/pull/466, and branch is https://github.com/mayankkapoor/k8s.io/tree/refactor-client-services.

mayankkapoor:
@cblecker I'm having trouble creating a fake client for admin.Service. In order to mock/stub admin.Service, I need to create an interface with a function that admin.Service implements. However, there are no exported functions that admin.Service implements. Only `userAgent()` which is not exported. 

E.g. In order to test `func printGroupMembersAndSettings(srv *admin.Service, srv2 *groupssettings.Service)`, I did:

```
type IService interface {
	userAgent() string
}
...
func printGroupMembersAndSettings(srv IService, srv2 *groupssettings.Service) error {
```

I get following error:
```
# k8s.io/k8s.io/groups [k8s.io/k8s.io/groups.test]
[...]

PR #2713, opened 2021-09-13
Refactor code in groups/ to make it mockable

MadhavJivrajani:
- Introduce mockable client interfaces that represent interaction with the admins API.
- Introduce mockable service interfaces that represent high level operations that are
  performed during reconcilation.
- Misc: fix a few warnings and linter errors

Part of #420

### DIFF
--- a/groups/reconcile.go
+++ b/groups/reconcile.go
@@ -17,15 +17,12 @@ limitations under the License.
 package main
 
 import (
-	"encoding/json"
 	"flag"
 	"fmt"
 	"io/ioutil"
 	"log"
-	"net/http"
 	"os"
 	"path/filepath"
-	"reflect"
 	"regexp"
 	"strings"
 
@@ -34,13 +31,18 @@ import (
 	"golang.org/x/net/context"
 	"golang.org/x/oauth2/google"
 	admin "google.golang.org/api/admin/directory/v1"
-	"google.golang.org/api/googleapi"
 	"google.golang.org/api/groupssettings/v1"
 	"google.golang.org/api/option"
 	secretmanagerpb "google.golang.org/genproto/googleapis/cloud/secretmanager/v1"
 	"gopkg.in/yaml.v3"
 
 	"k8s.io/test-infra/pkg/genyaml"
+)
+
+const (
+	ownerRole   = "OWNER"
+	managerRole = "MANAGER"
+	memberRole  = "MEMBER"
 )
 
 type Config struct {
@@ -93,6 +95,7 @@ type GoogleGroup struct {
 type RestrictionsConfig struct {
 	Restrictions []Restriction `yaml:"restrictions,omitempty" json:"restrictions,omitempty"`
 }
+
 type Restriction struct {
 	// Path is the relative path of a sub-directory to the groups-path.
 	Path string `yaml:"path" json:"path"`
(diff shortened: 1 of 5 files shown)


MadhavJivrajani:
/cc @nikhita
fyi

groups/reconcile.go
```diff
@@ -181,69 +184,176 @@ func main() {
...
+	err := r.adminService.DeleteGroupsIfNecessary()
 	if err != nil {
-		log.Fatal(err)
+		return err
+	}
+
+	return nil
+}
+
+func (r *reconciler) printGroupMembersAndSettings() error {
```
spiffxp, line 283: What would make this easier to review is to leave this function where it was instead of reordering

Make whatever changes you're going to make to the function

Then reorder in a new commit

groups/reconcile.go
```diff
@@ -311,12 +421,19 @@ func (gc *GroupsConfig) Load(rootDir string, restrictions *RestrictionsConfig) e
 	log.Printf("reading groups.yaml files recursively at %s", rootDir)
 
 	return filepath.Walk(rootDir, func(path string, info os.FileInfo, err error) error {
+		// check if the current path is accessible or not.
+		if err != nil {
+			return err
+		}
```
spiffxp, line 427: This looks like new logic.  Ideally we'd see this in a separate commit from a refactor, with a test case that proves why it's needed

groups/reconcile.go
```diff
@@ -311,12 +421,19 @@ func (gc *GroupsConfig) Load(rootDir string, restrictions *RestrictionsConfig) e
...
 			cleanPath := strings.Trim(strings.TrimPrefix(path, rootDir), string(filepath.Separator))
 			log.Printf("groups: %s", cleanPath)
 
-			var groupsConfigAtPath GroupsConfig
-			var content []byte
+			var (
+				groupsConfigAtPath GroupsConfig
+				content            []byte
+				err                error
+			)
```
spiffxp, line 436: I am not used to seeing this style of var declaration outside of globals but maybe that's just me

groups/reconcile.go
```diff
@@ -181,69 +184,176 @@ func main() {
...
+	for _, g := range groups {
 		if g.EmailId == "" {
-			log.Fatal(fmt.Sprintf("Group has no email-id: %#v", g))
+			return fmt.Errorf("group has no email-id: %#v", g)
 		}
 
-		err = createOrUpdateGroupIfNecessary(adminService, g.EmailId, g.Name, g.Description)
+		// update the group that is currently being considered.
+		r.adminService.SetGroup(g)
+		r.groupService.SetGroup(g)
```
spiffxp, line 237: This context-aware API seems prone to human error.  I would rather either:
- make group an explicit parameter to function calls (e.g. `adminService.CreateOrUpdateGroupIfNescessary(g)`)
- return and use an instance of something that is configured to use group (e.g. `a = adminServiceFactory.AdminServiceForGroup(g)` but those are terrible names I'm sure you can think of something better)

groups/service.go
```diff
@@ -0,0 +1,415 @@
...
+
+	admin "google.golang.org/api/admin/directory/v1"
+	"google.golang.org/api/googleapi"
+	groupssettings "google.golang.org/api/groupssettings/v1"
+	"google.golang.org/api/option"
+)
+
+type Service interface {
+	SetGroup(GoogleGroup)
+}
```
spiffxp, line 35: See above re: prone to human error

groups/service.go
```diff
@@ -0,0 +1,415 @@
...
+		}
+	}
+
+	return nil
+}
+
+func (as *adminService) CreateOrUpdateGroupIfNescessary() error {
+	if *verbose {
+		log.Printf("createOrUpdateGroupIfNecessary %q", as.group.EmailId)
+	}
```
spiffxp, line 142: Logging style nit: I was trying to roughly follow the function signature before, and it's changed.  I'm not sure I like the way this suggestion looks, but moving without changing also means the log statement doesn't really line up with anything so makes me wonder what it should be instead
```suggestion
	if *verbose {
		log.Printf("adminService(%v).CreateOrUpdateGroupIfNecessary", as.group.EmailId)
	}
```

groups/service.go
```diff
@@ -0,0 +1,415 @@
...
+
+	group, err := as.client.GetGroup(as.group.EmailId)
+	if err != nil {
+		if apierr, ok := err.(*googleapi.Error); ok && apierr.Code == http.StatusNotFound {
+			if !config.ConfirmChanges {
```
spiffxp, line 147: This is maybe a nice to have.  Keeping it like this is making it easier for me to diff what was moved, but I also think it'd be preferable to push config options like "dry run" and "verbose" into the service itself, so we're not tying back to globals.

groups/service.go
```diff
@@ -0,0 +1,415 @@
...
+				}
+				if as.group.Name != "" {
+					g.Name = as.group.Name
+				}
+				if as.group.Description != "" {
+					g.Description = as.group.Description
+				}
+				g4, err := as.client.InsertGroup(&g)
+				if err != nil {
+					return fmt.Errorf("unable to add new group %q: %v", as.group.EmailId, err)
```
spiffxp, line 162: Out of scope for this PR, but for followup: it would be ideal if we used `%w` instead of `%v` for errs

groups/service.go
```diff
@@ -0,0 +1,415 @@
...
+			log.Printf("Added %s to %q as a %s\n", memberEmailId, as.group.EmailId, role)
+		} else {
+			log.Printf("dry-run: would add %s to %q as %s\n", memberEmailId, as.group.EmailId, role)
+		}
+	}
+
+	return nil
+}
+
+func (as *adminService) CreateOrUpdateGroupIfNescessary() error {
```
spiffxp, line 139: Could be done here, could be done for followup: would like to see comments describing what these functions do

groups/service.go
```diff
@@ -0,0 +1,415 @@
...
+					return fmt.Errorf("unable to update group %q: %v", as.group.EmailId, err)
+				}
+				log.Printf("> Successfully updated group %s\n", g4.Email)
+			}
+		}
+	}
+	return nil
+}
+
+func (as *adminService) DeleteGroupsIfNecessary() error {
```
spiffxp, line 196: To keep harping on the oddity of the `SetGroup` thing... while other functions behave different depending on whether that was called, this does not.  The inconsistency feels off to me.

groups/service.go
```diff
@@ -0,0 +1,415 @@
...
+	g, err := as.client.ListGroups()
+	if err != nil {
+		return fmt.Errorf("unable to retrieve users in domain: %v", err)
+	}
+	if len(g.Groups) == 0 {
+		log.Println("No groups found.")
+		return nil
+	}
```
spiffxp, line 204: For followup since this is a straight refactor: we don't actually need this, since we'd skip by iterating over nothing below (and I don't think there's value in this logging statement)

groups/service.go
```diff
@@ -0,0 +1,415 @@
...
+
+	for _, m := range l.Members {
+		found := false
+		for _, m2 := range members {
+			if m2 == m.Email {
+				found = true
+				break
+			}
+		}
+		if found || m.Role == "MEMBER" {
```
spiffxp, line 258: I know this is how you found it, but I scratched my head over this for a while.  Do you think you can add a comment explaining why this is needed? It's the only thing that makes `RemoveOwnersOrManagersFromGroup` different than `RemoveMembersFromGroup`

It would be nice to refactor this away as a followup.  I think this is to handle someone moving from manager to member, and not dropping them.  But making two passes over all group members seems like an inefficiency we'll want to remove

groups/service.go
```diff
@@ -0,0 +1,415 @@
...
+type adminService struct {
+	client AdminServiceClient
+	group  GoogleGroup
+}
+
+func (as *adminService) SetGroup(group GoogleGroup) {
+	as.group = group
+}
+
+func (as *adminService) AddOrUpdateGroupMembers(role string, members []string) error {
```
spiffxp, line 80: Same deal here re: returning `[]error` as a followup PR

spiffxp (changes requested):
It's been a while since I looked at all the code here holistically, so I called out a few things I would suggest as followups

Since this looks straightforward enough I'm OK if you get to the point of landing this before landing additional test coverage.  But it needs to run first ;)

```
go run reconcile.go
# command-line-arguments
./reconcile.go:211:15: undefined: AdminService
./reconcile.go:212:15: undefined: GroupService
./reconcile.go:216:13: undefined: NewAdminService
./reconcile.go:221:13: undefined: NewGroupService
./reconcile.go:350:19: cannot assign error to err in multiple assignment
```

MadhavJivrajani:
Oh wow, the compile time errors are odd... Will take a closer look soon, thanks for the reviews @spiffxp!

spiffxp:
I suspect we need to change the make target to build and run the binary

groups/service.go
MadhavJivrajani: > I think this is to handle someone moving from manager to member, and not dropping them.

If I understand correctly, this is along the lines of:

If a member m exists in our desired list of members, do nothing.
However, if this member m does not exist in our desired list of
members but is in the role of a MEMBER (non OWNER/MANAGER), still
do nothing since we care only about OWNER/MANAGER roles here.

MadhavJivrajani:
@spiffxp I've addressed the review comments for the most part - the reviews that are still unresolved are to be made part of a follow-up PR.

Could you PTAL? :)

MadhavJivrajani:
I'll be adding the fake client and test coverage in subsequent PRs

groups/reconcile.go
```diff
@@ -181,69 +178,172 @@ func main() {
...
+		log.Fatal(err)
+	}
+}
+
+// Reconciler syncs the actual state of the world with the configuration.
+// It does so by making use of AdminService and GroupService which are mockable
+// interfaces.
+type Reconciler struct {
+	AdminSvc AdminService
+	GroupSvc GroupService
```
spiffxp, line 206: Can be for followup: I'm not sure these necessarily need to be public

groups/service.go
```diff
@@ -0,0 +1,434 @@
...
+
+// AdminService provides functionality to perform high level
+// tasks using a AdminServiceClient.
+type AdminService interface {
+	AddOrUpdateGroupMembers(group GoogleGroup, role string, members []string) error
+	CreateOrUpdateGroupIfNescessary(group GoogleGroup) error
+	DeleteGroupsIfNecessary() error
+	RemoveOwnerOrManagersFromGroup(group GoogleGroup, members []string) error
+	RemoveMembersFromGroup(group GoogleGroup, members []string) error
+	GetClient() AdminServiceClient
```
spiffxp, line 47: For followup: can we remove this? Not clear to me why we would want something other than the admin service (or the thing building it) to have access to the client

groups/service.go
```diff
@@ -0,0 +1,434 @@
...
+}
+
+// GroupService provides functionality to perform high level
+// tasks using a GroupServiceClient.
+type GroupService interface {
+	UpdateGroupSettings(group GoogleGroup) error
+	GetClient() GroupServiceClient
```
spiffxp, line 54: For followup: can we remove this? Not clear to me why we would want something other than the group service (or the thing building it) to have access to the client

groups/reconcile.go
```diff
@@ -181,69 +178,172 @@ func main() {
...
+	}
+
+	var groupsConfig GroupsConfig
+	for _, g := range g.Groups {
+		group := GoogleGroup{
+			EmailId:     g.Email,
+			Name:        g.Name,
+			Description: g.Description,
+		}
+		g2, err := gsClient.Get(g.Email)
```
spiffxp, line 294: e.g. this could probably be a `GroupService` function so we don't need the client

groups/reconcile.go
```diff
@@ -181,69 +178,172 @@ func main() {
...
+		group.Settings["WhoCanViewGroup"] = g2.WhoCanViewGroup
+		group.Settings["WhoCanDiscoverGroup"] = g2.WhoCanDiscoverGroup
+		group.Settings["WhoCanInvite"] = g2.WhoCanInvite
+		group.Settings["WhoCanAdd"] = g2.WhoCanAdd
+		group.Settings["WhoCanApproveMembers"] = g2.WhoCanApproveMembers
+		group.Settings["WhoCanModifyMembers"] = g2.WhoCanModifyMembers
+		group.Settings["WhoCanModerateMembers"] = g2.WhoCanModerateMembers
+		group.Settings["MembersCanPostAsTheGroup"] = g2.MembersCanPostAsTheGroup
+
+		l, err := asClient.ListMembers(g.Email)
```
spiffxp, line 311: ditto here, could be an `AdminService` function

spiffxp (approved):
`make run` is happy now

I've left the comments that can be done as followups unresolved

/approve
/lgtm

groups/reconcile.go
```diff
@@ -181,69 +184,176 @@ func main() {
...
+
+	gs, err := NewGroupService(ctx, clientOption)
+	if err != nil {
+		return nil, err
+	}
+
+	return &reconciler{adminService: as, groupService: gs}, nil
+}
+
+func (r *reconciler) reconcileGroups(groups []GoogleGroup) error {
```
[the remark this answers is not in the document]
spiffxp: Opened followup issue: https://github.com/kubernetes/k8s.io/issues/2772

MadhavJivrajani:
Thank you!

Issue #2625, opened 2021-08-27
Migrate away from gs://kops-ci


ameukam:
Part of umbrella issue to migrate the kubernetes project away from use of GCP project google-containers: #1469

- [x] Create a new GCS bucket writable by k8s-infra-prow-build GKE cluster (`kops-ci` to `k8s-infra-kops-ci-results`)
- [x] Give kOps maintainers admin access to this bucket
- [x] Add a canary jobs (duplicate of the ones writing to gs://kops-ci) that pushes to the new bucket
	- ensure it's building and pushing appropriately
- [ ] update jobs that pull from the new bucket
- [x] get rid of the jobs that runs on the "default" cluster	

/sig cluster-lifecycle
/area jobs
/help wanted

/assign @justinsb @johngmyers 
For Kops

/assign @spiffxp 
For wg-k8s-infra

spiffxp:
/unassign
I'm not actively working this, though I am happily reviewing @ameukam's PRs

spiffxp:
/milestone clear
Clearing from milestone because migrating all of the kops jobs over is likely to affect our spend in a non-significant way

upodroid:
This migration is complete.

Are we missing something @ameukam ?

ameukam:
I see references of the bucket: https://cs.k8s.io/?q=https%3A%2F%2Fstorage.googleapis.com%2Fkops-ci&i=nope&files=&excludeFiles=&repos=. 🤔

justinsb:
I think we're really close, we were able to set up the 1.30 jobs with no kops-ci at all.  I think there are still a few 1.28 / 1.29 jobs that point to kops-ci, but we can now repoint them following the example of 1.30 (we were waiting on releases on 1.28 / 1.29, I think they're now ~done)

ameukam:
/close

What remaining is part of some legacy stuff and will probably go away.

PR #2718, opened 2021-09-13
gcp/bash: Add GCP secrets for kops ssh key

ameukam:
Related:
  - https://github.com/kubernetes/k8s.io/issues/2625.

EDIT: The kops e2e tests needs a SSH key (passwordless is allowed) to be able to access to instances running on AWS. Add GCP secret for SSH key
The value of the secret will be accessible by `k8s-infra-kops-maintainers@kubernetes.io`.

Signed-off-by: Arnaud Meukam <ameukam@gmail.com>

PR #2728, opened 2021-09-15
kubernetes-public: Allow write access to prow-build-trusted SA

ameukam:
Ref: https://github.com/kubernetes/k8s.io/issues/2625

Allow prow-build-trusted SA to write to k8s-infra-kops-ci-results
bucket.
This is required by prowjobs pushing version markers for kops.

### DIFF
--- a/infra/gcp/terraform/kubernetes-public/prowjob-buckets.tf
+++ b/infra/gcp/terraform/kubernetes-public/prowjob-buckets.tf
@@ -174,7 +174,7 @@ data "google_iam_policy" "kops_ci_bucket_bindings" {
   binding {
     role = "roles/storage.objectAdmin"
     members = [
-      "serviceAccount:prow-build@k8s-infra-prow-build.iam.gserviceaccount.com",
+      "serviceAccount:prow-build-trusted@k8s-infra-prow-build-trusted.iam.gserviceaccount.com",
     ]
   }
   // Ensure bucket is world readable


ameukam:
cc @dims @spiffxp

spiffxp (approved):
/approve
/lgtm
/hold
I didn't think the jobs ran in test-infra-trusted? So sorta unclear why we're moving to trusted here. But I trust you know what you're doing

ameukam:
> /approve
> /lgtm
> /hold
> I didn't think the jobs ran in test-infra-trusted? So sorta unclear why we're moving to trusted here. But I trust you know what you're doing

The plan is to add the ssh key added in #2718 to prow-build-trusted cluster using (kubernetes-external-secrets) and move all the jobs concerned. Currently : https://github.com/kubernetes/test-infra/blob/e1f515bd889c0435db458772a317dc994a740a79/config/prow/config.yaml#L690-L707. 

Can we add the new key to prow-build ?

spiffxp:
>Can we add the new key to prow-build ?

We do this today for all of the k8s-infra-e2e projects, see:

- instructions to manually add the key as a secret: https://github.com/kubernetes/k8s.io/blob/main/infra/gcp/terraform/k8s-infra-prow-build/prow-build/README.md#provisioning
- code that sets up projects with the pubkey
https://github.com/kubernetes/k8s.io/blob/d4ba40ad7c11b6a4fbafb57a4463cf521121e60f/infra/gcp/bash/prow/ensure-e2e-projects.sh#L104-L137

I'd like to move the key to an external secret, and then it's not too hard to imagine something that periodically rotates the keys.

The other alternative is to use something like oslogin, but I'm not sure that'd work for kops jobs that run on aws, and might require more of a rewrite to kube-up.sh than anyone has the stomach for.

ameukam:
Considered not needed kubernetes-externel-secrets deployed on prow-build cluster: 

https://github.com/kubernetes/k8s.io/pull/2827

/close

PR #2730, opened 2021-09-15
kube-cross: tagging kube-cross to use the same iamge for v1.21.0-go1.16.8-buster.0

cpanato:
tagging the image that was built for `v1.22.0-go1.16.8-buster.0` to be `v1.21.0-go1.16.8-buster.0` to have a image that follows the tag pattern for each branch

In the image for 1.21 and 1.22 there are no changes so we are good to reuse that same for now.

context: https://github.com/kubernetes/kubernetes/pull/104906#issuecomment-920093745


cpanato:
this is correct to do?

justaugustus:
> this is correct to do?

Yep, thanks @cpanato!
/lgtm
/approve

PR #2766, opened 2021-09-21
groups: add sig-k8s-infra-leads mailing list

ameukam:
Related:
  - Ref: https://github.com/kubernetes/community/issues/6036
  - Followup of : https://github.com/kubernetes/k8s.io/pull/2761

Signed-off-by: Arnaud Meukam <ameukam@gmail.com>

PR #2770, opened 2021-09-21
groups: mv wg-k8s-infra sig-k8s-infra

spiffxp:
Related:
- Part of: https://github.com/kubernetes/community/issues/6036
- Followup to: https://github.com/kubernetes/k8s.io/pull/2766

This moves the location of sig-k8s-infra's groups configs to match our name, then two group changes:
- drop wg-k8s-infra-leads now that zoom / docs permissions have been updated
- rename k8s-infra-team-private to sig-k8s-infra-private


groups/sig-k8s-infra/groups.yaml
```diff
@@ -23,11 +23,11 @@ groups:
       - thockin@google.com
       - linusa@google.com
 
-  # TODO(spiffxp): rename this to wg-k8s-infra-private at some point
-  - email-id: k8s-infra-team-private@kubernetes.io
-    name: k8s-infra-team-private
+  - email-id: sig-k8s-infra-private@kubernetes.io
```
ameukam, line 26: Some cleanup needed ?

ameukam:
/lgtm
/hold
Remove when ready.

groups/sig-k8s-infra/groups.yaml
spiffxp: SG, I'll look at pruning members from -private in a separate PR

PR #2774, opened 2021-09-21
Elekto deployment

jberkus:
Draft deployment for Elekto on k8s infra for @ameukam to finish.
Do not merge in current state.

Unresolved questions:
- need to create secret with credentials
- need to add DB port info to the configmap
- could stand to create "latest" branch in Elekto autobuild so we're not using a SHA
- please check ingress setup here, not sure it's correct (and not tested on GCE)
- Elekto uses some emphemeral cache storage.  Do we need to create an emphemeral volume for that?

Notes:
- yes, only one replica because of GH rate limits and DB locking

/sig contributor-experience
/committee steering
/hold

### DIFF
--- a/apps/elekto/configmap.yaml
+++ b/apps/elekto/configmap.yaml
@@ -1,0 +1,21 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: elekto-config
+  namespace: elekto
+data:
+  APP_NAME: k8s.elections
+  APP_ENV: production
+  APP_KEY:
+  APP_DEBUG: "False"
+  APP_URL: "http://localhost"
+  APP_PORT: 8080
+  APP_HOST: localhost
+  APP_CONNECT: http
+  DB_CONNECTION: postgresql
+  META_REPO: "https://github.com/kubernetes/community.git"
+  ELECTION_DIR: "events/elections"
+  META_DEPLOYMENT: local
+  META_PATH: /tmp/meta
+  META_BRANCH: master
+  GITHUB_REDIRECT: "/oauth/github/callback"
--- a/apps/elekto/deployment.yaml
+++ b/apps/elekto/deployment.yaml
@@ -1,0 +1,42 @@
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: elekto
+  namespace: elekto
+  labels:
+    app: elekto
+spec:
+  replicas: 1
+  selector:
+    matchLabels:
+      app: elekto
+  strategy:
+    type: Recreate
+  template:
+    metadata:
+      labels:
+        app: elekto
+    spec:
+      containers:
+        - image: ghcr.io/elekto-io/elekto:sha-8d58a8e
+          imagePullPolicy: Always
+          name: elekto
+          ports:
+            - containerPort: 8080
+              protocol: TCP
+          resources:
+            limits:
+              memory: 1500M
+          securityContext:
+            allowPrivilegeEscalation: false
+            privileged: false
+            runAsNonRoot: true
+          envFrom:
+            - configMapRef:
+                name: elekto-config
+            - secretRef:
+                name: elekto-secret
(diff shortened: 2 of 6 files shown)


apps/elekto/configmap.yaml
```diff
@@ -0,0 +1,23 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: elekto-config
```
ameukam, line 4:
```suggestion
metadata:
  name: elekto-config
  namespace: elekto
```

apps/elekto/configmap.yaml
```diff
@@ -0,0 +1,23 @@
...
+data:
+  APP_NAME: k8s.elections
+  APP_ENV: production
+  APP_KEY: 
+  APP_DEBUG: "False"
+  APP_URL: "http://localhost"
+  APP_PORT: 8080
+  APP_HOST: localhost
+  APP_CONNECT: http
+  DB_CONNECTION: mysql {or postgresql}
```
ameukam, line 14:
```suggestion
  DB_CONNECTION: postgresql
```

apps/elekto/configmap.yaml
```diff
@@ -0,0 +1,23 @@
...
+  DB_HOST: {CloudSQL address}
+  DB_PORT: {CloudSQL port}
+  DB_DATABASE: {CloudSQL database}
```
ameukam, line 17: There are dynamic and will be added to the secret.

jberkus: ok, works, I'll take them out of the configmap

apps/elekto/configmap.yaml
jberkus: ooops, does namespace need to be added to all the other files as well?

ameukam: Yes for all the files deploying Kubernetes resources. Otherwise the resources will be deployed to the default namespace.

apps/elekto/deployment.yaml
```diff
@@ -0,0 +1,38 @@
...
+  selector:
+    matchLabels:
+      app: elekto
+  strategy:
+    type: Recreate
+  template:
+    metadata:
+      labels:
+        app: elekto
+    spec:
```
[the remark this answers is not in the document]
jberkus: how does UID mapping for non-root containers work in GCE?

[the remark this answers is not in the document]
jberkus: I tested this and the cache seems to work fine, as long as I put it in /tmp.  Just need to figure out how to set the UID.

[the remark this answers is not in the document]
ameukam: A securityContext at the container level should be fine:

```yaml
    securityContext:
      allowPrivilegeEscalation: false
      privileged: false
      runAsNonRoot: true
```

spiffxp:
Related:
- Part of: https://github.com/kubernetes/k8s.io/issues/2575

(so it's easier for me to find this again)

jberkus:
Secrets are populated, so removing hold.

/remove-hold

spiffxp (approved):
/approve
/lgtm
Let's see if we've got everything hooked up properly...

ameukam:
Configmap seems incorrect (cc @jberkus): 

```shell
ameukam@cloudshell:~/k8s.io/apps/elekto (kubernetes-public)$ kubectl apply -f configmap.yaml 
Error from server (BadRequest): error when creating "configmap.yaml": ConfigMap in version "v1" cannot be handled as a ConfigMap: v1.ConfigMap.Data: ReadString: expects " or n, but found 8, error found in #10 byte of ...|PP_PORT":8080,"APP_U|..., bigger context ...|KEY":"N/A","APP_NAME":"k8s.elections","APP_PORT":8080,"APP_URL":"http://localhost","DB_CONNECTION":"|...
```

PR #2775, opened 2021-09-22
Aggregate errors rather than exiting reconcile loop per error

### LINKED ISSUE #2772: groups: attempt to reconcile all groups/members instead of failing on first error
Noticed while reviewing https://github.com/kubernetes/k8s.io/pull/2713

ref: https://github.com/kubernetes/k8s.io/pull/2713#discussion_r710252179

Right now the reconciler is relatively brittle and an invalid group/member can halt all groups reconciliation.  A better pattern to follow is to gather errors during reconciliation, and then fail at the end with the full list of errors.  I believe we do this with other github management tools like peribolos and label_sync

MadhavJivrajani:
Following up on https://github.com/kubernetes/k8s.io/pull/2713

Instead of returning a slice of errors, the aggregation utils defined by apimachienry utils is used mainly in accordance with how peribolos does it

As a slightly related fix - change format specifiers of error types in `fmt.Errorf()` from `%v` to `%w`

Fixes #2772 

/priority important-soon
/assign @spiffxp

### DIFF
--- a/groups/reconcile.go
+++ b/groups/reconcile.go
@@ -227,37 +227,31 @@ func (r *Reconciler) ReconcileGroups(groups []GoogleGroup) error {
 	for _, g := range groups {
 		if g.EmailId == "" {
 			errs = append(errs, fmt.Errorf("group has no email-id: %#v", g))
-			continue
 		}
 
 		err := r.AdminSvc.CreateOrUpdateGroupIfNescessary(g)
 		if err != nil {
 			errs = append(errs, err)
-			continue
 		}
 
 		err = r.GroupSvc.UpdateGroupSettings(g)
 		if err != nil {
 			errs = append(errs, err)
-			continue
 		}
 
 		err = r.AdminSvc.AddOrUpdateGroupMembers(g, OwnerRole, g.Owners)
 		if err != nil {
 			errs = append(errs, err)
-			continue
 		}
 
 		err = r.AdminSvc.AddOrUpdateGroupMembers(g, ManagerRole, g.Managers)
 		if err != nil {
 			errs = append(errs, err)
-			continue
 		}
 
 		err = r.AdminSvc.AddOrUpdateGroupMembers(g, MemberRole, g.Members)
 		if err != nil {
 			errs = append(errs, err)
-			continue
 		}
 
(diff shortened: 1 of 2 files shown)


groups/reconcile.go
```diff
@@ -221,53 +222,66 @@ func NewReconciler(ctx context.Context, clientOption option.ClientOption) (*Reco
...
-			return fmt.Errorf("group has no email-id: %#v", g)
+			errs = append(errs, fmt.Errorf("group has no email-id: %#v", g))
+			continue
 		}
 
 		err := r.AdminSvc.CreateOrUpdateGroupIfNescessary(g)
 		if err != nil {
-			return err
+			errs = append(errs, err)
+			continue
```
spiffxp, line 236: I would prefer to drop the continues from here on. One bad member should not stop attempts to add the rest, etc

spiffxp (changes requested):

groups/reconcile.go
MadhavJivrajani: ahh yes, thanks for catching that

groups/service.go
```diff
@@ -164,12 +169,12 @@ func (as *adminService) CreateOrUpdateGroupIfNescessary(group GoogleGroup) error
...
 				g4, err := as.client.InsertGroup(&g)
 				if err != nil {
-					return fmt.Errorf("unable to add new group %q: %v", group.EmailId, err)
+					return fmt.Errorf("unable to add new group %q: %w", group.EmailId, err)
 				}
 				log.Printf("> Successfully created group %s\n", g4.Email)
 			}
 		} else {
-			return fmt.Errorf("unable to fetch group %q: %#v", group.EmailId, err.Error())
+			return fmt.Errorf("unable to fetch group %q: %q", group.EmailId, err)
```
spiffxp, line 177:
```suggestion
			return fmt.Errorf("unable to fetch group %q: %w", group.EmailId, err)
```

spiffxp (changes requested):
One tiny fix otherwise lgtm

MadhavJivrajani:
Thanks @spiffxp 
I've made the changes

spiffxp (approved):
/approve
/lgtm

Issue #2755, opened 2021-09-20
Clearing retention policy for s390x's  GCS bucket


guirish:
@ameukam  We are trying to execute conformance tests on prow setup with kubetest2 using pod-utilities, and we are getting below mentioned error specific to bucket's retention policy. Can we get this property cleared for s390x GCS bucket -> `k8s-conform-s390x-k8s` . We tried to clear the property at our end but we do not have required rights.

error: 
```{"component":"initupload","error":"failed to upload to blob storage: failed to upload to blob storage: encountered errors during upload: [writer close error: googleapi: Error 403: Object 'k8s-conform-s390x-k8s/logs/demo/latest-build.txt' is subject to bucket's retention policy and cannot be deleted, overwritten or archived until 2031-09-20T12:18:15.93713-07:00, forbidden]","file":"prow/cmd/initupload/main.go:45","func":"main.main","level":"fatal","msg":"Failed to initialize job","severity":"fatal","time":"2021-09-20T07:30:10Z"}```

CC: @dims

guirish:
@ameukam  Any updates on this?
Gentle reminder.

ameukam:
> @ameukam Any updates on this?
> Gentle reminder.

I'm not really familiar with kubetest2 so I don't know why it needs to remove `latest-build.txt` instead of update it. I can drop the file concerned but I'm not sure if it's the right approach to solve this. 
@spiffxp  Any ideas ?

spiffxp:
We don't hold the project to retention policies for its own conformance results, I'm fine dropping them for federated conformance results

/assign

spiffxp:
Opened https://github.com/kubernetes/k8s.io/pull/2779

PR #2779, opened 2021-09-22
infra: remove bucket retention policy from k8s-conform buckets

spiffxp:
Related:
- Fixes: https://github.com/kubernetes/k8s.io/issues/2755

We don't hold the project to retention policies for its own conformance results, so I think we should drop them for federated conformance results

This allows folks to write to these buckets using CI systems like jenkins that need to update files in place

### DIFF
--- a/infra/gcp/bash/lib_gcs.sh
+++ b/infra/gcp/bash/lib_gcs.sh
@@ -154,6 +154,18 @@ function ensure_gcs_bucket_retention() {
     gsutil retention set "${retention}" "${bucket}"
 }
 
+# Ensure the bucket retention policy is not set
+# $1: The GCS bucket (e.g. gs://bucket-name)
+function ensure_removed_gcs_bucket_retention() {
+    if [ $# -lt 1 ] || [ -z "$1" ]; then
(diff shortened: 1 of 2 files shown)


spiffxp:
/hold
I'll drop this and deploy once reviewed

infra/gcp/bash/lib_gcs.sh
```diff
@@ -26,7 +26,7 @@
 # $2: The bucket (e.g. gs://bucket-name)
 function _ensure_gcs_bucket() {
     if [ $# -lt 2 ] || [ -z "$1" ] || [ -z "$2" ]; then
-        echo "_ensure_gcs_bucket(project, bucket) requires 2 arguments" >&2
+        echo "${FUNCNAME[0]}(project, bucket) requires 2 arguments" >&2
```
spiffxp, line 29: DRY'ed these up while I was in this file

rajibmitra (approved):
lgtm

spiffxp:
Ran "./infra/gcp/bash/ensure-conformance-storage.sh`

PR #2791, opened 2021-09-23
give kustomize admins key management permission

natasha41575:
We are trying to run a cloud build job in the k8s-staging-kustomize project. This PR is intended to:

1. Give the members listed here:

https://github.com/kubernetes/k8s.io/blob/7f720723206888b4bfb01036d5028f9852940b8f/groups/sig-cli/groups.yaml#L22-L34

permission to create and manage a keyring by giving it the role `Cloud KMS Admin`.

2. Give the cloud build project associated with k8s-staging-kustomize the roles `Cloud KMS CryptoKey Decrypter` and `Secret Manager Secret Accessor` so that it can read and decrypt the secret stored in our cloudbuild.yaml file. 

It would be helpful if the reviewer could confirm that this PR gives the correct service accounts and people the roles that I intend (since I am new here). 

I also noticed that there is a file here https://github.com/kubernetes/k8s.io/blob/main/audit/projects/k8s-staging-kustomize/iam.json which describes the IAM roles for each service account. Do I need to update that file as well? If so, is there an automated script I can run to do so?

Thank you in advance.

/cc @monopole @KnVerey @spiffxp

### DIFF
--- a/infra/gcp/bash/ensure-staging-storage.sh
+++ b/infra/gcp/bash/ensure-staging-storage.sh
@@ -401,6 +401,22 @@ function staging_special_case__k8s_staging_cluster_api_gcp() {
     ensure_staging_gcb_builder_service_account "cluster-api-gcp" "k8s-infra-prow-build-trusted"
 }
 
+# In order to build the release artifacts, the group needs to be
+# able to create and manage a keyring to encrypt a secret token
+# that will be accessed and decrypted by a cloud build job.
+function staging_special_case__k8s_staging_kustomize() {
+    readonly STAGING_PROJECT="k8s-staging-kustomize"
+
+    local sa_email_group="k8s-infra-staging-kustomize@kubernetes.io"
+    local principal_group="serviceAccount:${sa_email_group}"
+    ensure_project_role_binding "${STAGING_PROJECT}" "${principal_group}" "roles/cloudkms.admin"
+
+    local sa_email_cloudbuild="660796270509@cloudbuild.gserviceaccount.com"
+    local principal_cloudbuild="serviceAccount:${sa_email_cloudbuild}"
+    ensure_project_role_binding "${STAGING_PROJECT}" "${principal_cloudbuild}" "roles/cloudkms.cryptoKeyDecrypter"
+    ensure_project_role_binding "${STAGING_PROJECT}" "${principal_cloudbuild}" "roles/secretmanager.secretAccessor"
+}
+
 #
 # main
 #


spiffxp:
>I also noticed that there is a file here https://github.com/kubernetes/k8s.io/blob/main/audit/projects/k8s-staging-kustomize/iam.json which describes the IAM roles for each service account. Do I need to update that file as well? If so, is there an automated script I can run to do so?

This is currently retroactive dumping of the current state of our infra.  We don't yet use this as source of truth or reconcile it against souce of truth (which is these scripts) short of human review.  So you're doing the right thing by not touching it.

infra/gcp/bash/ensure-staging-storage.sh
```diff
@@ -401,6 +401,22 @@ function staging_special_case__k8s_staging_cluster_api_gcp() {
...
+# able to create and manage a keyring to encrypt a secret token
+# that will be accessed and decrypted by a cloud build job.
+function staging_special_case__k8s_staging_kustomize() {
+    readonly STAGING_PROJECT="k8s-staging-kustomize"
+
+    local sa_email_group="k8s-infra-staging-kustomize@kubernetes.io"
+    local principal_group="serviceAccount:${sa_email_group}"
+    ensure_project_role_binding "${STAGING_PROJECT}" "${principal_group}" "roles/cloudkms.admin"
+
+    local sa_email_cloudbuild="660796270509@cloudbuild.gserviceaccount.com"
```
spiffxp, line 414: Is this who the builds run as?  I see the following from `gcloud projects get-iam-policy k8s-staging-kustomize`

```yaml
bindings
- members:
  - serviceAccount:660796270509@cloudbuild.gserviceaccount.com
  - serviceAccount:gcb-builder@k8s-infra-prow-build-trusted.iam.gserviceaccount.com
  role: roles/cloudbuild.builds.builder
# ...
- members:
  - serviceAccount:service-660796270509@gcp-sa-cloudbuild.iam.gserviceaccount.com
  role: roles/cloudbuild.serviceAgent
# ...
```

IIRC triggering a build as a user is not the same as guaranteeing who that build runs as.

Can probably find this out from logs of recent builds

spiffxp: Answering my own question: yes it is.  And annoyingly I could only find this on console, `gcloud builds describe` doesn't seem to dump this info?

spiffxp (approved):
/approve
/lgtm

infra/gcp/bash/ensure-staging-storage.sh
```diff
@@ -401,6 +401,22 @@ function staging_special_case__k8s_staging_cluster_api_gcp() {
...
 }
 
+# In order to build the release artifacts, the group needs to be
+# able to create and manage a keyring to encrypt a secret token
+# that will be accessed and decrypted by a cloud build job.
+function staging_special_case__k8s_staging_kustomize() {
+    readonly STAGING_PROJECT="k8s-staging-kustomize"
+
+    local sa_email_group="k8s-infra-staging-kustomize@kubernetes.io"
+    local principal_group="serviceAccount:${sa_email_group}"
```
spiffxp, line 411: ahhhhh I just caught this after it merged, I'll open a followup
```suggestion
    local principal_group="group:${sa_email_group}"
```

spiffxp:
Changes deployed, https://github.com/kubernetes/k8s.io/pull/2792 is what I deployed from

spiffxp:
FWIW, https://github.com/kubernetes/k8s.io/pull/2793 is the audit PR that reflects the changes being deployed

PR #2794, opened 2021-09-23
bash/main: provision secrets for elekto

spiffxp:
Related:
- Part of: https://github.com/kubernetes/k8s.io/issues/2575

This creates secrets to host the values as named in https://github.com/kubernetes/k8s.io/issues/2575#issuecomment-923487372


spiffxp:
/cc @jberkus @mrbobbytables 

/cc @ameukam 
FYI

spiffxp:
Missed some apparently, adding more in https://github.com/kubernetes/k8s.io/pull/2797

PR #2795, opened 2021-09-23
Revert "Remove extra label from k8s-infra-prow"

ameukam:
Revert
https://github.com/kubernetes/k8s.io/commit/dd1509e8d8f46cbfadf78e7233f16624bf83d8c4.

This prevent k8s-infra-ci-robot to merge PRs and conflict with
k8s-ci-robot.

Signed-off-by: Arnaud Meukam <ameukam@gmail.com>

PR #2800, opened 2021-09-23
Add comment for k8s-infra-prow tide labels

eddiezane:
This PR is to ensure that #2795 does not happen.

/assign @ameukam

### DIFF
--- a/apps/prow/config.yaml
+++ b/apps/prow/config.yaml
@@ -58,6 +58,8 @@ tide:
     labels:
     - lgtm
     - approved
+    # This label is used to prevent k8s-infra-ci-robot from double merging
+    # PRs because k8s-ci-robot is still the active Tide account.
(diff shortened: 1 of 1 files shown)


ameukam:
/hold
k8s-infra-ci-robot took over again.

spiffxp (approved):
/approve
/lgtm
Sorry we didn't have something like this here in the first place. Good catch

spiffxp:
>k8s-infra-ci-robot took over again.

So the way to get its tide to stop trying to look at this repo is to remove the tide query entirely.  The current status is that both tides report to the same context, so last one to update wins.  It's not very friendly to new contributors to this repo.  It does let us know that k8s-infra's tide is running and doing stuff though.  So I could go either way.

spiffxp:
/close
I'm going to close this given #2803 merged

PR #2813, opened 2021-09-24
bash/staging: prune unexpected services

spiffxp:
Related:
- Part of: https://github.com/kubernetes/k8s.io/issues/1675

Specifically

- add a mechanism for staging projects to specify special case services
  that they expect to have enabled
- use that mechanism for cluster-api-gcp and sig-storage
- set obnoxious env var that causes ensure_only_services to disable any
  service not explicitly requested as enabled
- cleanup: order special case functions by project name

Last step before I call https://github.com/kubernetes/k8s.io/issues/1675
done will be to run `ensure-staging-storage.sh` for all staging projects
just to ensure we've pruned all unexpected services.

### DIFF
--- a/infra/gcp/bash/ensure-staging-storage.sh
+++ b/infra/gcp/bash/ensure-staging-storage.sh
@@ -136,10 +136,17 @@ function ensure_staging_project() {
 
     # Enable services for staging projects and their direct dependencies; prune anything else
     color 6 "Ensuring necessary enabled services staging project: ${project}"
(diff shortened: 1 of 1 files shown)


spiffxp:
I've run `ensure-staging-storage.sh cluster-api-gcp` and `ensure-staging-storage.sh kubetest2` to ensure this works for both projects that have a stock set of services, and projects that have additional services enabled as a special case.

PR #2816, opened 2021-09-24
add readiness probe at /health

jberkus:
Custom readiness probe since / needs to have status 302

Signed-off-by: Josh Berkus <josh@agliodbs.com>

/sig contributor-experience
/committee steering
attn: @ameukam @spiffxp

### DIFF
--- a/apps/elekto/deployment.yaml
+++ b/apps/elekto/deployment.yaml
@@ -18,12 +18,21 @@ spec:
         app: elekto
     spec:
       containers:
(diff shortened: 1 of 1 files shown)


jberkus:
@ameukam is this all that's needed, or do we need to add some config under ingress.yaml as well?

jberkus:
addresses comment in #2575

ameukam:
> @ameukam is this all that's needed, or do we need to add some config under ingress.yaml as well?

Nothing to add to the ingress. I suspect you will need to recreate(delete+apply) the ingress object.

PR #2843, opened 2021-09-28
terraform: add roles to workload-identity-service-account

spiffxp:
First part of an attempt to for_each the sundry workload-identity-service-account modules created in k8s-infra-prow and k8s-infra-prow-build

Add a `project_roles` attribute that will optionally bind the service account to those roles in the project hosting the service account

PR #2844, opened 2021-09-28
terraform: invoke workload-identity-service-account module via for_each

spiffxp:
Related:
- Followup to: https://github.com/kubernetes/k8s.io/pull/2843

This DRYs things up a bit.

Now when we want to add/edit a service account that is usable via workload identity, instead of looking at this:
```terraform
module "prow_build_cluster_sa" {
  source            = "../modules/workload-identity-service-account"
  project_id        = local.project_id
  name              = local.cluster_sa_name
  description       = "default service account for pods in ${local.cluster_name}"
  cluster_namespace = local.pod_namespace
}

module "boskos_janitor_sa" {
  source            = "../modules/workload-identity-service-account"
  project_id        = local.project_id
  name              = local.boskos_janitor_sa_name
  description       = "used by boskos-janitor in ${local.cluster_name}"
  cluster_namespace = local.pod_namespace
}

module "kubernetes_external_secrets_sa" {
  source            = "../modules/workload-identity-service-account"
  project_id        = local.project_id
  name              = local.kubernetes_external_secrets_sa_name
  description       = "sync K8s secrets from GSM in this and other projects"
  cluster_namespace = "kubernetes-external-secrets"
  project_roles     = ["roles/secretmanager.secretAccessor"]
}
```

We look at this:
```terraform
workload_identity_service_accounts = [
  {
    name        = "prow-build",
    description = "default service account for pods in ${local.cluster_name}",
  },
  {
    name        = "boskos-janitor",
    description = "uesd by boskos-janitor in ${local.cluster_name}",
  },
  {
    name          = "kubernetes-external-secrets",
    description   = "sync K8s secrets from GSM in this and other projects",
    project_roles = ["roles/secretmanager.secretAccessor"]
  }
]
```

It's the sort of data that could come from a YAML/JSON file, though I'm not inclined to go that far since I can't see an obvious place to put such a thing.

This requires terraform state surgery to deploy without destroying/recreating some resources. I've included shell scripts `tf-state-mv.sh` in each module to do the needful, which can be reverted out once deployed

### DIFF
--- a/infra/gcp/terraform/k8s-infra-prow-build-trusted/main.tf
+++ b/infra/gcp/terraform/k8s-infra-prow-build-trusted/main.tf
@@ -29,13 +29,39 @@ locals {
   bigquery_location = "US"                 // The bigquery specific location where the dataset should be created
   pod_namespace     = "test-pods"          // MUST match whatever prow is configured to use when it schedules to this cluster
 
+  workload_identity_service_accounts = [
+    {
+      name        = "prow-build-trusted"
+      description = "default service account for pods in ${local.cluster_name}"
+    },
+    {
+      name        = "gcb-builder"
+      description = "trigger GCB builds in all k8s-staging projects"
+    },
+    {
+      name          = "prow-deployer"
+      description   = "deploys k8s resources to k8s clusters"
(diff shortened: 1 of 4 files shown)


spiffxp:
/hold
To deploy when around

PR #2845, opened 2021-09-29
terraform: provision prow-related secrets via terraform

spiffxp:
Related:
- Part of: https://github.com/kubernetes/k8s.io/issues/2220
- Part of: https://github.com/kubernetes/k8s.io/issues/1731

This moves management of k8s-infra-prow-build-trusted's secrets to terraform

Then follows a similar pattern to setup secrets for k8s-infra-prow-build that will be populated and hooked up to ExternalSecret CRDs as followup

This was written with separate locals blocks to avoid merge conflicts with outstanding terraform PRs for prow

Issue #2220, opened 2021-06-15
Migrate all Opaque kubernetes Secrets on k8s-infra clusters to ExternalSecrets


spiffxp:
- Clusters that need kubernetes-external-secrets installed:
  - [x] k8s-infra-prow-build/prow-build: https://github.com/kubernetes/k8s.io/pull/2827, https://github.com/kubernetes/k8s.io/pull/2830

- Secrets that need to be migrated over for k8s-infra-prow-build/prow-build:
```$ k --context=prow-build get secrets -A | grep Opaque | awk '{print "- [ ] "$1"/"$2}';```
  - [x] test-pods/service-account: https://github.com/kubernetes/k8s.io/pull/2845, https://github.com/kubernetes/k8s.io/pull/2858
  - [x] test-pods/ssh-key-secret: https://github.com/kubernetes/k8s.io/pull/2845, https://github.com/kubernetes/k8s.io/pull/2858

- Secrets that need to be migrated over for k8s-infra-prow-build-trusted/prow-build-trusted:
```$ k --context=prow-build-trusted get secrets -A | grep Opaque | awk '{print "- [ ] "$1"/"$2}'```
  - [x] test-pods/cncf-ci-github-token (https://github.com/kubernetes/k8s.io/pull/2219)
  - [x] test-pods/k8s-cip-test-prod-service-account: https://github.com/kubernetes/k8s.io/pull/2859
  - [x] test-pods/k8s-gcr-audit-test-prod-service-account: https://github.com/kubernetes/k8s.io/pull/2859
  - [x] test-pods/k8s-infra-ci-robot-github-token: https://github.com/kubernetes/k8s.io/pull/2859
  - [x] test-pods/k8s-triage-robot-github-token: https://github.com/kubernetes/k8s.io/pull/2859
  - [x] test-pods/service-account: https://github.com/kubernetes/k8s.io/pull/2859
  - [x] test-pods/slack-tempelis-auth: https://github.com/kubernetes/k8s.io/pull/2859
  - [x] test-pods/snyk-token* (https://github.com/kubernetes/k8s.io/pull/2222)

- Secrets that need to be migrated over for kubernetes-public/aaa
``` $ k --context=aaa get secrets -A | grep Opaque | awk '{print "- [ ] "$1"/"$2}' ```
  - [x] ~cert-manager/letencrypt-prod-account-key~ (generated by cert-manager)
[...]

Issue #1731, opened 2021-02-26
audit followup: ensure secretmanager secrets and bindings are managed by scripts


spiffxp:
Caught via https://github.com/kubernetes/k8s.io/pull/1718#discussion_r582443184

I've done the following pattern a few times now (most recently https://github.com/kubernetes/k8s.io/pull/1696#issuecomment-784618406):
- someone needs to pass a secret to me, wonders how to transfer it
- I setup a secretmanager secret in the relevant project
  - intended to be k8s secrets in a cluster: the project hosting the cluster
    - kubernetes-public for apps running on aaa
    - k8s-infra-prow-build-trusted for prow jobs or components runnong on prow-build-trusted
    - etc.
- Setup appropriate permissions on the secret (need to figure out a consistent pattern here)
  - Org admins should implicitly get access via their role/owner on the organization
  - Give some sort of "oncall" team ownership of the secret and its versions for break-glass cases
  - Give whomever is handing the secret to us write access to the secret version
- Setup labels on the secret
  - `app: foo` if its for an app in the `foo` dir running on aaa
  - `sig: foo` if the app is owned by sig-foo
  - ??? maybe the namespace the secret is destined for?
- Request that for things destined to be k8s secrets, people store an actual secret manifest yaml in the secret

Put it all together and this allows for relatively simple / safe deployment: https://github.com/kubernetes/k8s.io/tree/main/slack-infra#how-to-deploy

Problems with the above:
- it needs to be documented instead of me just happening to kinda/sorta do something consistent
- the secret creation should be scripted

spiffxp:
/kind cleanup
/kind documentation
/priority important-soon
/area cluster-mgmt
/area access

Labels and areas for relevant sigs:

/sig testing
/area prow
I'd like to set this up as the standard for managing secrets for prow.k8s.io (with possible exception of kubeconfigs that speak to google-internal clusters)

/sig release
/area release-eng
triage-party, publishing-bot, and possibly other secrets

/sig contributor-experience
slack-infra, groups management

/sig scalability
perf-dash

/sig node
node-perf-dash

spiffxp:
Another piece of followup: I setup a custom `secretLister` role (ref:  https://github.com/kubernetes/k8s.io/pull/1726) which _should_ allow someone to use the GCP console to manage secrets.  It should work such that adding it to a group means they'll be able to list/see secrets, but only those they directly have access to.  Need to trial this with someone (maybe @jeefy)

[...]

PR #2847, opened 2021-09-29
terraform: update secret iam to depend on secrets

spiffxp:
Related:
- Part of: https://github.com/kubernetes/k8s.io/issues/2220
- Part of: https://github.com/kubernetes/k8s.io/issues/1731
- Followup to: https://github.com/kubernetes/k8s.io/pull/2845

This is a subtle ordering bug that appears to only surface when new
secrets are created. The workaround was to just keep re-running
`terraform apply` until it worked

The first time I ran `terraform apply` for k8s-infra-prow-build, there
were secrets to be created, and then iam bindings to be applied on those
secrets. But terraform didn't know about that ordering, because the
iam bindings depended on the same locals that the secrets did, so
terraform tried to create them all in parallel.

This resulted in terraform erroring out on first run, because it was
trying to create iam bindings on secrets that didn't yet exist. When I
reran `terraform apply` the secrets had been created, so the iam
bindings succeeded.

One way to let terraform know about an ordering like this is to have one
terraform resource use the value of another. Another way would be an
explicit `depends_on`, but that's described as a last resort, and we're
not out of options here.

### DIFF
--- a/infra/gcp/terraform/k8s-infra-prow-build-trusted/main.tf
+++ b/infra/gcp/terraform/k8s-infra-prow-build-trusted/main.tf
@@ -173,8 +173,8 @@ resource "google_secret_manager_secret" "build_cluster_secrets" {
 
 resource "google_secret_manager_secret_iam_binding" "build_cluster_secret_admins" {
   for_each  = local.build_cluster_secrets
-  project   = local.project_id
-  secret_id = each.key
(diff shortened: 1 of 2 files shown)


spiffxp:
/hold
to deploy when around

cpanato (approved):
/lgtm

PR #2851, opened 2021-09-29
Add volume populator alpha images

bswartz:


k8s.gcr.io/images/k8s-staging-sig-storage/generate.sh
```diff
@@ -40,6 +41,7 @@ readonly images=(
     snapshot-controller
     snapshot-validation-webhook
     # TODO: validation-webhook?
+    volume-data-source-validator
```
xing-yang, line 44: Can you remove line 43?  That's not needed any more.

PR #2859, opened 2021-09-30
tf/k8s-infra-prow-build-trusted: add secrets

spiffxp:
Related:
- Part of: https://github.com/kubernetes/k8s.io/issues/2220

Specifically GSM secrets that correspond to secrets that already exist
in the k8s-cluster


spiffxp:
/cc @dims 
I have run `terraform apply`

PR #2858, opened 2021-09-30
tf/k8s-infra-prow-build: add ExternalSecrets

spiffxp:
Related:
- Part of: https://github.com/kubernetes/k8s.io/issues/2220
- Followup to: https://github.com/kubernetes/k8s.io/pull/2845

Turns out the k8s secret has two fields: ssh-private and ssh-public, so we'll instead make a GSM secret per field

I populated the GSM secrets, and have added ExternalSecret CRDs to sync the values back to the Kubernetes secrets

PR #2863, opened 2021-10-01
bash/prow: add new ssh key to e2e projects

spiffxp:
Related:
- Part of: https://github.com/kubernetes/kubernetes/issues/105381
- Followup to: https://github.com/kubernetes/k8s.io/pull/2858

While chasing down SSH failures to instances, I took a copy of the ssh key that was used by jobs in k8s-prow-builds and added it to all e2e projects.

This allowed us to rule out `load pubkey "/workspace/.ssh/google_compute_engine": invalid format` as the cause of SSH errors for the pull-kubernetes-node-e2e-containerd


spiffxp:
/cc @dims @BenTheElder 
This was already deployed yesterday, I forgot to check this is in while doing so

Issue #2890, opened 2021-10-07
groups: jobs are continuously failing


spiffxp:
What's failing:
- https://testgrid.k8s.io/sig-k8s-infra-groups#post-k8sio-groups - continuously failing
- https://testgrid.k8s.io/sig-k8s-infra-groups#ci-k8sio-groups - was continuously failing, now flapping

Sample failure from https://prow.k8s.io/view/gs/kubernetes-jenkins/logs/ci-k8sio-groups/1446007265566920704
```
2021/10/07 06:59:57  ======================= Updates =======================
2021/10/07 07:01:13 Removing noah.ispas@googlemail.com from "k8s-infra-staging-ingress-nginx@kubernetes.io" as a MEMBER
2021/10/07 07:01:32 Removing kaslinfields@gmail.com from "release-comms@kubernetes.io" as a MEMBER
2021/10/07 07:01:37 Added Cicih@google.com to "release-team@kubernetes.io" as a MEMBER
2021/10/07 07:01:38 Added DamanArora@cmail.carleton.ca to "release-team@kubernetes.io" as a MEMBER
2021/10/07 07:01:40 Removing cicih@google.com from "release-team@kubernetes.io" as a MEMBER
2021/10/07 07:01:40 Removing Damanarora@cmail.carleton.ca from "release-team@kubernetes.io" as a MEMBER
2021/10/07 07:01:41 Removing kaslinfields@gmail.com from "release-team@kubernetes.io" as a MEMBER
2021/10/07 07:01:44 Added DamanArora@cmail.carleton.ca to "release-team-shadows@kubernetes.io" as a MEMBER
2021/10/07 07:01:45 Removing Damanarora@cmail.carleton.ca from "release-team-shadows@kubernetes.io" as a MEMBER
2021/10/07 07:01:46 Removing kaslinfields@gmail.com from "release-team-shadows@kubernetes.io" as a MEMBER
[...]

PR #2892, opened 2021-10-07
groups: attempt to handle equivalent e-mail addresses

spiffxp:
Related:
- Part of: https://github.com/kubernetes/k8s.io/issues/2890

Attempt to deal with email addresses that are basically equivalent by 
comparing email address with case insensitivity, and stripping dots from 
the local part (before the '@').

These rules make sense for gmail addresses, which comprise the majority of
email addresses used in our groups. It's difficult to know whether they
apply universally to all e-mail addresses.

https://www.rfc-editor.org/rfc/rfc5321#section-2.3.11 says that the 
semantics for the local part of an e-mail address are determined by the 
host. So the only way to know for sure if two email addresses are 
equivalent is to send e-mail to both of them, and inspect the resulting 
mailbox.

It sure seems like there must be some part of the Directory API that knows
how to handle this equivalence, so I wanted to rely on that instead of
trying to build our own hackish local best guess at email equivalence.  But
when I tried using either the GetMember or HasMember API calls with the
semantically equivalent email addresses, I got back a 404 not found, and
400 invalid member key respectively.

So, we'll go with this for now until we receive reports of someone not 
receiving email due to case-sensitivty or dot-sensitivity issues.

Other things included in here:

- To aid in debugging I hacked up my group config to focus just on the
sig-release groups, which led me to add support for relative paths in the
config file.

- I had initially hoped refreshing libraries might get us some kind of fix
for free, so `go.mod` and `go.sum` have a few updates (see that commit
for details)


spiffxp:
If you look at the linked issue there are two other domains failing due to case insensitivity

MadhavJivrajani:
Ah, missed that. Thanks for clarifying :)
LGTM!

spiffxp:
ping @nikhita PTAL

spiffxp:
Rebased to avoid merge conflict

spiffxp:
Missed merge conflict in boilerplate, fixed

groups/service.go
```diff
@@ -503,3 +520,26 @@ func deepCopySettings(a, b interface{}) {
 	byt, _ := json.Marshal(a)
 	json.Unmarshal(byt, b)
 }
+
+// EmailAddressEquals checks equivalence between two e-mail addresses according
+// to the following rules:
+// - email addresses are case-insensitive (e.g. FOO@bar.com == foo@bar.com)
+// - local parts are dot-inensitive (e.g. foo@bar.com == f.o.o@bar.com)
+func EmailAddressEquals(a, b string) bool {
```
nikhita, line 528: Should we be considering the "anything after `+` is ignored" case here?

nikhita (approved):
Apologies for the delay. One question but we can add that as a follow-up if needed and this lgtm so:

/lgtm

PR #2906, opened 2021-10-08
hack: fix verify-boilerplate

spiffxp:
Related:
- Addresses issue discovered here: https://github.com/kubernetes/k8s.io/pull/2710#issuecomment-939091383

Two fixes:
- the previous default of `vendor` would exclude `foo-vendor-bar/baz` from
boilerplate checks, now the default is `vendor/` (and code updated to
support the trailing `/`)
- nested dirs weren't actually getting skipped, they can now be
specified relative to repo root, e.g. `/foo/bar` will still walk `bar`

The first fix addresses the linked PR not failing verify-boilerplate despite
introducing many new files that should have boilerplate added.

The second fix speeds up verify-boilerplate by avoiding a walk of
`audit/projects` which has numerous machine-generated files.

I plan on upstreaming these fixes but for now, use a local copy of `verify_boilerplate.py`

### DIFF
--- a/hack/bin/verify_boilerplate.py
+++ b/hack/bin/verify_boilerplate.py
@@ -45,7 +45,7 @@ def get_args():
             'node_modules',
             '_output',
             'third_party',
-            'vendor',
+            'vendor/',
             'verify/boilerplate/test',
         ],
         action='append',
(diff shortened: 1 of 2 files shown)


spiffxp:
/cc @dims @ameukam 
FYI @BobyMCbobs @hh

PR #2904, opened 2021-10-08
Sandbox for CAPG folks testing

dims:
ACL group for sandbox for CAPG mentees (LFX)

context:
- https://github.com/kubernetes/funding/issues/18
- https://kubernetes.slack.com/archives/CCK68P2Q2/p1633697468176100

Signed-off-by: Davanum Srinivas <davanum@gmail.com>

PR #2908, opened 2021-10-08
infra: add k8s-infra-sandbox-capg project

spiffxp:
Related:
- Part of: https://github.com/kubernetes/funding/issues/18
- Followup to: https://github.com/kubernetes/k8s.io/pull/2904

### DIFF
--- a/infra/gcp/terraform/k8s-infra-sandbox-capg/main.tf
+++ b/infra/gcp/terraform/k8s-infra-sandbox-capg/main.tf
@@ -39,14 +39,24 @@ resource "google_project_service" "project" {
   project = google_project.project.id
 
   for_each = toset([
+    "bigquery.googleapis.com",
+    "bigquerystorage.googleapis.com",
+    "compute.googleapis.com",
     "container.googleapis.com",
+    "containerregistry.googleapis.com",
+    "iam.googleapis.com",
+    "iamcredentials.googleapis.com",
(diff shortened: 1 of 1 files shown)


spiffxp:
Created the project with `terraform apply`

Set a budget on the project via click-ops:

Screen Shot 2021-10-08 at 2 37 13 PM

`terraform apply` doesn't seem to mind, so I've dropped the WIP label

I'd like to try and keep this terraform module scoped such that the people using the project could run terraform to affect changes on it.  I'm assuming so long as nobody touches the billing account info on the project, terraform won't try to use a permission we're not going to grant to these folks.  e.g. one of the [required permissions to update a project's billing info][1] (`billing.resourceAssociations.create`) is missing from this list:

```yaml
$ gcloud iam roles describe roles/owner | grep '[Bb]illing'
- billing.resourceCosts.get
- cloudasset.assets.exportCloudbillingBillingAccounts
- firebase.billingPlans.get
- firebase.billingPlans.update
- resourcemanager.projects.createBillingAssignment
- resourcemanager.projects.deleteBillingAssignment
```

Managing the budget via terraform (or some other client library, looks like gcloud doesn't have it) can be done as a follow up (which I am noodling on now, but unlikely to finish by end of day)  

We'll first got notified when 20% of the budget has been spent

[1]: https://cloud.google.com/billing/v1/how-tos/access-control#tbl_perm

PR #3048, opened 2021-11-08
ksm: promote v2.2.4 images

dgrisonnet:
Promoting v2.2.4 images of kube-state-metrics based on https://console.cloud.google.com/gcr/images/k8s-staging-kube-state-metrics?project=k8s-staging-kube-state-metrics.


dgrisonnet:
I also took the opportunity to reorder the image versions since it was a bit messy.

/cc @mrueg @fpetkovski

mrueg:
/lgtm

@dgrisonnet not sure if cip-mm will change the order back again (I believe it's alphabetically by the hash instead of tag)

dgrisonnet:
I don't think so, it doesn't seem to be reordered for other projects:
- https://github.com/kubernetes/k8s.io/blob/main/k8s.gcr.io/images/k8s-staging-prometheus-adapter/images.yaml
- https://github.com/kubernetes/k8s.io/blob/main/k8s.gcr.io/images/k8s-staging-metrics-server/images.yaml

PR #3049, opened 2021-11-08
groups: move manager groups to members

spiffxp:
Related:
- Part of: https://github.com/kubernetes/k8s.io/issues/2890

Though I can't find anything in the docs saying that groups aren't
allowed to be added as managers, and I see contributors@kubernetes.io
listed as Owners for leads@kubernetes.io, our groups CI job is
continuously failing attempting to add these groups as managers.

We really should strive to avoid manager permissions for most of these
groups anyway, since they're mostly useful for moderation and spam
deletion. These particular groups are more about ACLs or mailing lists,
the membership of which we should have no issue managing via gitops.


spiffxp:
I'll track the manual stuff in the tracking issue

PR #3082, opened 2021-11-17
Promote ip-masq-agent v2.6.1 images for CVE fixes.

MrHohn:
These are essentially a re-build of the v2.6.0 images with a newer base image.

/assign @varunmar


MrHohn:
Giving @prameshj a try :) Thanks!
/assign @prameshj

PR #3107, opened 2021-11-22
Add ii streaming humans to k8s stream-team

hh:
Stephen, Caleb, and myself will be assisting with the Kubernetes Community Celebration Dec 13thish


hh:
Replaces #3106 as the commit had an '@' in it.
/assign @mrbobbytables

PR #3117, opened 2021-11-24
releng: Temporary RM access for palnabarun

Verolop:
Nabarun (palnabarun) is a Release Manager Associate with SIG Release being granted
temporary elevated access to cut the v1.23.0.0 release. Access will be
revoked after the 1.23.0.0 release is cut.

Signed-off-by: Verónica López (verolop) <gveronicalg@gmail.com>

### DIFF
--- a/groups/sig-release/groups.yaml
+++ b/groups/sig-release/groups.yaml
@@ -42,6 +42,7 @@ groups:
       - gveronicalg@gmail.com
       - k8s@auggie.dev
       - mudrinic.mare@gmail.com
+      - pal.nabarun95@gmail.com
       - saschagrunert@gmail.com
 
   - email-id: k8s-infra-release-viewers@kubernetes.io


xmudrii (approved):

saschagrunert (approved):

palnabarun:
@Verolop -- I don't think this would be required now for the RC cut. Shall we close this for now until the final release cut?

saschagrunert:
Let’s keep it open as a placeholder for the release. It’s already approved so all we have to do is releasing the hold when the time is right. 😊

puerco:
> Let’s keep it open as a placeholder for the release. It’s already approved so all we have to do is releasing the hold when the time is right. 
Maybe this can be a good enough practice to add to the handbook? It really proved handy at the right time.

Verolop:
+1, it has been very helpful!

PR #3121, opened 2021-11-24
Promote git-sync v3.3.5

thockin:
This is a minor release.  It includes the following changes and also picks up the latest base image, which addresses some vulnerabilities.

* Flag `--sync-hook-command` (`$GIT_SYNC_HOOK_COMMAND`)  is deprecated (but still works).  Use `--exechook-command` (`$GIT_SYNC_EXECHOOK_COMMAND`) instead (functionally identical).  This adds flags `--exechook-timeout` (`$GIT_SYNC_EXECHOOK_TIMEOUT`) and `--exechook-backoff` (`$GIT_SYNC_EXECHOOK_BACKOFF`) to parallel webhooks.
* Fix a bug that can mis-attribute callers in log-lines.
* Create git worktrees using the specific SHA rather than branch name.  This allows to change branches with persistent volumes.  Otherwise should be functionally identical.
* Add flag `password-file` (`$GIT_SYNC_PASSWORD_FILE`), which reads the password from a file and this is considered as safer than reading from env or flag directly.
  - `--password` and `--password-file` can't be specified at the same time.
  - If `--username` is specified, then one of `--password` or `--password-file` must be specified.


thockin:
Uggh, messed the tag, wait

thockin:
Fixed.  PTAL

PR #3130, opened 2021-11-30
k8s.io: Redirect common Go package URLs to pkg.go.dev

imjasonh:
Today, a valid Go import path like `k8s.io/api/core/v1` redirects to https://kubernetes.io/api/core/v1 when viewed in a browser, which is an unhelpful 404. 😢 

With this change, I'd like to propose that that import path should redirect to the much more helpful Go package documentation at https://pkg.go.dev/k8s.io/api/core/v1

This change currently only applies to:
- `k8s.io/api/...`
- `k8s.io/apimachinery/...`
- `k8s.io/client-go/...`
- `k8s.io/kubernetes/...`

If this is helpful we can look into applying this across more repos. The Go import redirects triggered by the presence of `?go-get=1` should be unaffected.

### DIFF
--- a/apps/k8s-io/configmap-nginx.yaml
+++ b/apps/k8s-io/configmap-nginx.yaml
@@ -55,8 +55,8 @@ data:
           }
 
           # Redirect go import paths to Go package documentation.
-          if ($repo ~ "api|apimachinery|client-go|kubernetes") {
-            return 301 https://pkg.go.dev/k8s.io/$repo/$subpath;
+          if ($repo ~ "api|apimachinery|client-go") {
+            return 302 https://pkg.go.dev/k8s.io/$repo/$subpath;
           }
 
           # Default to redirecting to the "real" site.
--- a/apps/k8s-io/configmap-nginx.yaml
+++ b/apps/k8s-io/configmap-nginx.yaml
@@ -54,6 +54,11 @@ data:
             ';
           }
 
+          # Redirect go import paths to Go package documentation.
+          if ($repo ~ "api|apimachinery|client-go|kubernetes") {
+            return 301 https://pkg.go.dev/k8s.io/$repo/$subpath;
+          }
+
           # Default to redirecting to the "real" site.
           return 301 https://kubernetes.io$request_uri;
         }
--- a/apps/k8s-io/test.py
+++ b/apps/k8s-io/test.py
@@ -153,7 +153,7 @@ class RedirTest(HTTPTestCase):
         self.assert_scheme_redirect(
                 'http://k8s.io/kubernetes',
                 'https://k8s.io/kubernetes', 301)
-        for repo in ('api', 'apimachinery', 'client-go', 'kubernetes'):
+        for repo in ('api', 'apimachinery', 'client-go'):
           self.assert_scheme_redirect(
                   'https://k8s.io/%s' % repo,
                   'https://pkg.go.dev/k8s.io/%s' % repo, 301)
--- a/apps/k8s-io/test.py
+++ b/apps/k8s-io/test.py
@@ -149,6 +149,19 @@ class RedirTest(HTTPTestCase):
                     'http://' + url + path,
                     'https://' + url + path, 301)
 
+    def test_godoc_redirect(self):
+        self.assert_scheme_redirect(
+                'http://k8s.io/kubernetes',
+                'https://k8s.io/kubernetes', 301)
+        for repo in ('api', 'apimachinery', 'client-go', 'kubernetes'):
+          self.assert_scheme_redirect(
+                  'https://k8s.io/%s' % repo,
+                  'https://pkg.go.dev/k8s.io/%s' % repo, 301)
+        # Sub-paths also work.
+        self.assert_scheme_redirect(
+                'https://k8s.io/api/core/v1',
+                'https://pkg.go.dev/k8s.io/api/core/v1', 301)
+
     def test_go_get(self):
         self.assert_scheme_redirect(
                 'http://k8s.io/kubernetes?go-get=1',


apps/k8s-io/configmap-nginx.yaml
```diff
@@ -54,6 +54,11 @@ data:
             ';
           }
 
+          # Redirect go import paths to Go package documentation.
+          if ($repo ~ "api|apimachinery|client-go|kubernetes") {
+            return 301 https://pkg.go.dev/k8s.io/$repo/$subpath;
```
ameukam, line 59: I'm still not familiar with this config but why not :
```suggestion
            return 301 https://git.k8s.io/$repo/$subpath;
```
?
I think you can get the same result.

ameukam:
/assign @spiffxp 
cc @thockin

apps/k8s-io/configmap-nginx.yaml
imjasonh: This would end up redirecting to eg github.com/kubernetes/client-go, and I'm trying to get this to redirect to Go package docs at [pkg.go.dev/k8s.io/client-go](https://pkg.go.dev/k8s.io/client-go)

cblecker:
/ok-to-test
/hold

@imjasonh Thanks for opening this! I could see this being helpful. However I'd want to chat with the sig-docs team that is responsible for the website first if they have any concerns.

@kubernetes/sig-docs-en-owners Hello! This change is to potentially move redirect some URLs that correspond to git packages to the pkg.go.dev.. these URLs would currently go to the website, but as there is unlikely to be content that directly matches the go package URL. Do you have any thoughts?

@liggitt @thockin @nikhita Also wanted to ping the three of you on this.. with the way we publish these packages, are the pkg.go.dev docs the right place to potentially redirect to? I know I've seen some funkiness before with these docs if you have irregular go module tagging practice.

liggitt:
> @liggitt @thockin @nikhita Also wanted to ping the three of you on this.. with the way we publish these packages, are the pkg.go.dev docs the right place to potentially redirect to? I know I've seen some funkiness before with these docs if you have irregular go module tagging practice.

no concerns from my side - https://pkg.go.dev/k8s.io/client-go was the weirdest one and that got cleaned up by https://github.com/kubernetes/kubernetes/issues/98267 and https://github.com/golang/go/issues/43265

imjasonh:
> @imjasonh Thanks for opening this! I could see this being helpful. However I'd want to chat with the sig-docs team that is responsible for the website first if they have any concerns.

No problem! Definitely makes sense to check with anybody and everybody who might see where this can unexpected go wrong.

> @kubernetes/sig-docs-en-owners Hello! This change is to potentially move redirect some URLs that correspond to git packages to the pkg.go.dev.. these URLs would currently go to the website, but as there is unlikely to be content that directly matches the go package URL. Do you have any thoughts?

I actually found that https://k8s.io/kubernetes redirects to https://kubernetes.io/docs/home/ today, so we should probably not change this in case something depends on it.

Ideally I'd like to put in redirects for all the packages in https://github.com/kubernetes/kubernetes/tree/master/staging/src/k8s.io since these tend to be Go packages with useful godoc. But `api`, `apimachinery` and `client-go` are the ones I most often find myself wanting, and I didn't want to jump straight to generating the list (somehow...), so I'm fine starting with those three and seeing how it lands before proceeding with more.

I'd also be interested in feedback about how to make this sort of change easier to test before merging and rolling it out. The `test.py` seems to only test current prod URLs (unless I'm missing something), which makes it a bit scary to propose changes.

sftim:
If we make this change, are we setting a convention that could set us up for a future collision with the website and its short URLs?

apps/k8s-io/configmap-nginx.yaml
```diff
@@ -54,6 +54,11 @@ data:
             ';
           }
 
+          # Redirect go import paths to Go package documentation.
+          if ($repo ~ "api|apimachinery|client-go|kubernetes") {
```
sftim, line 58: I think I'd implement this kind of redirect by editing https://github.com/kubernetes/website/blob/main/static/_redirects

Would that work? It's more visible and less likely to conflict with SIG Docs changes / plans.

apps/k8s-io/configmap-nginx.yaml
sftim, line 59: I'd use a 302 redirect. 301 implies either “this is the new home for the content that used to be here, and we're sure of that” or “having consulted a soothsayer, we know we will never redirect anywhere else” (and I think the second meaning is ambitious).

imjasonh:
> If we make this change, are we setting a convention that could set us up for a future collision with the website and its short URLs?

Good question.

As an example, `k8s.io/community` currently redirects to (valid, useful) https://kubernetes.io/community, even though it's also a valid Go module with godoc: https://pkg.go.dev/k8s.io/community -- I definitely think we should prefer the existing redirect instead of redirecting to potentially-useful-but-probably-less-so godoc pages.

This seems to me like more signal that we should maintain a hand-edited list instead of trying too hard to automate it. If we do automate it, we should have a list of exceptions for things like `community` and any others we find or invent in the future.

apps/k8s-io/configmap-nginx.yaml
imjasonh: Excellent call, done.

apps/k8s-io/configmap-nginx.yaml
imjasonh: Ooh, nice. I didn't know that was also an option, this definitely seems like an easier place to put these, with preview deploys and tests and everything. I'll look into it, and possibly close this PR. 👍

cblecker: Yeah, I think this might be the better overall path as then we are saying: "unless we know for sure they are trying to do a `go get`, then it should go to the website, and then sig-docs can do things there. Especially if this is a manual list.

imjasonh:
Closing this in favor of https://github.com/kubernetes/website/pull/30692

Thanks for you feedback everyone! 🙏

PR #3178, opened 2021-12-13
Revoke Release Manager access for jimangel

jimangel:
James Angel is a Release Manager Associate who was granted temporary elevated access to cut the v1.24.0-alpha.a release.

The release is out, so we are dropping the elevated privileges.

/assign @justaugustus @Verolop @palnabarun

Closes kubernetes/sig-release#1777


justaugustus (approved):
Thanks @jimangel!

PR #3199, opened 2021-12-16
groups: add brandon to steering-emeritus

philips:
Signed-off-by: Brandon Philips <brandon@ifup.org>


justaugustus:
Thanks Brandon!
ref: https://github.com/kubernetes/steering/issues/189

PR #3221, opened 2021-12-23
Add boskos pool for k8s-infra-e2e-scale-5k-project

marseel:
We would like to add boskos pool for 5k project to leverage janitor - currently some of our test runs can leak resources.

/hold
Let's merge if after holiday season.


ameukam:
/lgtm
/approve
/hold

PR #3222, opened 2021-12-24
promote nginx-errors image to v1.0.0

anand-abgis:


gkarthiks (approved):

longwuyuan:
I would close this so fast that it never makes the open list again.

PR #3229, opened 2022-01-03
update dev group name

parispittman:
promotes visibility on members groups list


cblecker:
actually, looking at the failing test, this won't work. the name is standardized.

/lgtm cancel
/close

PR #3230, opened 2022-01-03
Paginate the ListMembers call

cblecker:
This adds pagination to the ListMembers call, as the dev@ group now has too many members to fit on one page.

/hold
/cc @dims @spiffxp @nikhita

### DIFF
--- a/groups/client.go
+++ b/groups/client.go
@@ -28,7 +28,7 @@ type AdminServiceClient interface {
 	GetGroup(groupKey string) (*admin.Group, error)
 	GetMember(groupKey, memberKey string) (*admin.Member, error)
 	ListGroups() (*admin.Groups, error)
-	ListMembers(groupKey string) (*admin.Members, error)
+	ListMembers(groupKey string) ([]*admin.Member, error)
 	InsertGroup(group *admin.Group) (*admin.Group, error)
 	InsertMember(groupKey string, member *admin.Member) (*admin.Member, error)
 	UpdateGroup(groupKey string, group *admin.Group) (*admin.Group, error)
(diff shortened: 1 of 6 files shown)


cblecker:
I should also note that I've tested this locally and it seems to be working

groups/client.go
```diff
@@ -62,8 +62,27 @@ func (asc *adminServiceClient) ListGroups() (*admin.Groups, error) {
...
+	var members []*admin.Member
+	var resp *admin.Members
+	var err error
+
+	call := asc.service.Members.List(groupKey)
+	for {
+		resp, err = call.Do()
+		if err != nil {
+			return nil, err
+		}
```
MadhavJivrajani, line 75:
```suggestion
		if err != nil {
			return members, err
		}
```
Maybe this? In case an error occurs on a page that is not the first one and some number of members have been processed successfully before the error occurred.
(behaviour is same as `return nil, err` since `members` will have `nil` value during the first page)

MadhavJivrajani (commented):
Thank you! ❤️

groups/client.go
cblecker: As this is a list call, we don't really want or care about a partial result, I don't think. Is there a reason we'd want a partial list, rather than relying on the error to just bail the whole call?

MadhavJivrajani: Hmm, that makes sense. I was mainly thinking about calls like `AddOrUpdateGroupMembers` where it does a `ListMembers`, I guess partial reconciliation doesn't really make a lot of sense, makes sense to bail the call in case of an error :)

nikhita: IMO using a partial result will lead to undesired and non-deterministic results.

@cblecker Adding lgtm to unblock the reconcile script. Will let you remove the hold when you are ready.

/lgtm

PR #3245, opened 2022-01-10
add dummy A record for an invalid registry

aojea:
Some e2e test use an invalid registry, currently it is being hardcoded to an external domain `invalid.com` that is handled by an external company.

In order to avoid that e2e users create traffic to an external domain, we can create a dummy record and use it for the e2e tests, so the e2e traffic remains controlled by the project.

Ref: https://github.com/kubernetes/kubernetes/pull/104522#discussion_r781008308

### DIFF
--- a/dns/zone-configs/k8s.io._0_base.yaml
+++ b/dns/zone-configs/k8s.io._0_base.yaml
@@ -51,6 +51,10 @@ docs:
 _acme-challenge.docs:
   type: A
   value: 0.0.0.0
+# Dummy A record used for e2e tests to reach an invalid registry:
+invalid.registry:
+  type: A
+  value: 0.0.0.0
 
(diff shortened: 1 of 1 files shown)


ameukam:
I don't see an issue to redirect this traffic to Internet.

/lgtm
/approve
/hold
For others to review.

aojea:
> I don't see an issue to redirect this traffic to Internet.

the problem is not to redirect traffic to Internet, is because this is a domain that someone can take and use to gather intelligence from these requests, or add a real registry and load malicious images in a cluster, .... is better to be paranoid with these things

thockin (commented):
/lgtm

aojea:
@thockin @BenTheElder Arnaud is not approver here, it needs your blessing :smile:

aojea:
/hold cancel

:crossed_fingers:

PR #3268, opened 2022-01-17
releng: promote kube-cross/go-runner/releng-ci Go1.17.6 and 1.16.13 images

palnabarun:
### What type of PR is this?

Used `cip-mm` to create the changes.

/kind feature
/area dependency

### What this PR does / why we need it:

Tracking issue: [kubernetes/release#2349](https://github.com/kubernetes/release/issues/2379)

  - kube-cross: v1.21.0-go1.16.13-buster.0 / v1.22.0-go1.16.13-buster.0 / v1.23.0-go1.17.6-bullseye.0 / v1.24.0-go1.17.6-bullseye.0
  - go-runner: v2.3.1-go1.16.13-buster.0 / v2.3.1-go1.17.6-bullseye.0
  - releng-ci: 1.16.13-buster-0 / 1.17.6-bullseye-0

/assign @justaugustus @saschagrunert @puerco @xmudrii @Verolop
cc @kubernetes/release-engineering

Signed-off-by: Nabarun Pal <pal.nabarun95@gmail.com>

PR #3271, opened 2022-01-18
remove images that we dont promote

cpanato:
some images that we usually don't promote were added to this PR https://github.com/kubernetes/k8s.io/pull/3268
Removing those, however not 100% sure if that is ok.

/kind cleanup

/assign @justaugustus @saschagrunert @palnabarun @puerco 


saschagrunert:
Is the tool capable of removing already promoted images?

cpanato:
> Is the tool capable of removing already promoted images?

i think if that is promoted it will not remove it

saschagrunert:
> > Is the tool capable of removing already promoted images?
> 
> i think if that is promoted it will not remove it

Then we probably should keep it for integrity reasons.

cpanato:
@saschagrunert thanks, makes sense

/close

PR #3281, opened 2022-01-19
audit: Add Benjamin Coles as GCP auditor

ameukam:
Benjamin will provide orientation and guidance about cost optimisation
for the current infrastructure.


ameukam:
cc @parispittman @dims

PR #3288, opened 2022-01-21
add juanfescobar build admins, drop amwat@google.com

BenTheElder:
https://github.com/kubernetes/org/pull/3201, but this repo

cc @juanfescobar @justaugustus

### DIFF
--- a/groups/sig-k8s-infra/groups.yaml
+++ b/groups/sig-k8s-infra/groups.yaml
@@ -14,7 +14,6 @@ groups:
       ReconcileMembers: "true"
     members:
       - ameukam@gmail.com
-      - amwat@google.com
       - bentheelder@google.com
       - cblecker@gmail.com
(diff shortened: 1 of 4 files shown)


groups/sig-k8s-infra/groups.yaml
```diff
@@ -14,7 +14,6 @@ groups:
       ReconcileMembers: "true"
     members:
       - ameukam@gmail.com
-      - amwat@google.com
       - bentheelder@google.com
```
BenTheElder, line 17: we don't really have an emeritus concept here, so instead just noting here for posterity that if amit wishes to continue being in any of these other groups he should be welcome to add himself, I'm only removing these because amwat@google.com is a dangling pointer 😢

cblecker (approved):
/lgtm
/approve

PR #3302, opened 2022-01-25
Promote Vertical Pod Autoscaler 0.10.0

jbartosik:
I've built & tested images


jbartosik:
@krzysied @mwielgus Please take a look

PR #3336, opened 2022-02-01
[WIP] dockerize dns check

BenTheElder:
we could also do this without docker, but, this is a smaller refactor and has tighter guarantees that we're running the same stack we use to push instead of depending on system python libs / install

### DIFF
--- a/dns/Makefile
+++ b/dns/Makefile
@@ -18,6 +18,8 @@ PROD_ZONES := k8s.io. kubernetes.io. x-k8s.io. canary.k8s-e2e.com.
 ZONE_CONFIGS := zone-configs
 OCTODNS_CONFIG := octodns-config.yaml
 DOCKER_IMAGE ?= k8s.gcr.io/infra-tools/octodns:v20200616-67ce585
+# for the shell scripts
+export DNS_IMAGE=$(DOCKER_IMAGE)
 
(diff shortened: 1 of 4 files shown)


dns/check-zone.sh
```diff
@@ -14,11 +14,16 @@
...
-# This runs as you.  It assumes you have built an image named ${USER}/octodns.
+# This runs as you.  You must have built an image and set ${DNS_IMAGE}.
+#
 # It also requires working gcloud credentials
 #
+# This is best run with `make check`
+#
 # $1: fqdn including final dot... ex "canary.k8s.io."
 
+set -o nounset -o pipefail
```
BenTheElder, line 25: TODO: this script is not errexit safe.

dns/Makefile
```diff
@@ -45,14 +47,8 @@ build:
...
-	docker run -ti \
-		--user $(shell id -u) \
-		--volume ~/.config/gcloud:/.config/gcloud:ro \
-		--volume $(CURDIR):/octodns \
-		--workdir /octodns \
-		$(DOCKER_IMAGE) \
-		./push.sh \
-			$(ARGS)
+	./run-dns-container.sh \
+		./push.sh $(ARGS)
```
BenTheElder, line 51: TODO: untested

PR #3345, opened 2022-02-03
Promote node-driver-registrar and livenessprobe staging images to release

mauriciopoppe:
Generated with step 14 of https://github.com/kubernetes-csi/csi-release-tools/blob/master/SIDECAR_RELEASE_PROCESS.md#release-process

```
./generate.sh > images.yaml
```

/cc @msau42 
/assign @msau42

PR #3351, opened 2022-02-04
remove image hash of iscsiplugin

humblec:
The canary image hash (v0.1.0-canary)  was mentioned as release tag,  considering canary get updated over time, we shouldnt be overwriting it.  This remove the same.
 
Note For reviewer :

A quick followup PR will place the release tag in place. Also both hashes are pointing to the same code/binary.

### DIFF
--- a/k8s.gcr.io/images/k8s-staging-sig-storage/images.yaml
+++ b/k8s.gcr.io/images/k8s-staging-sig-storage/images.yaml
@@ -100,7 +100,6 @@
     "sha256:6029c252dae6178c99b580de72d7776158edbc81be0de15cedc4152a3acfed18": [ "v1.7.3" ]
 - name: iscsiplugin
   dmap:
-    "sha256:31117074c183242b223340f2fc0f886243e97c6cb1a83d98cfb00a69d1746697": [ "v0.1.0" ]
 - name: livenessprobe
   dmap:
     "sha256:f8cec70adc74897ddde5da4f1da0209a497370eaf657566e2b36bc5f0f3ccbd7": [ "v1.1.0" ]


mauriciopoppe:
I'm not sure if removing the line means unpublishing the image, what I think the script is doing is taking all the images listed on the file and trying to copy them from staging to prod

If v0.1.0 is added again with a new tag we might encounter the same problem because v0.1.0 already exists in prod

humblec:
> I'm not sure if removing the line means unpublishing the image, what I think the script is doing is taking all the images listed on the file and trying to copy them from staging to prod
> 
> If v0.1.0 is added again with a new tag we might encounter the same problem because v0.1.0 already exists in prod

What I am noticing or understanding here is, if we remove the image hash as in this PR and if we promote or update the other hash as v0.1.0 , it looks to be working. 

Otherwise, we can just hack the script to forget about this image tag and hash in generate.sh, so it remains as we have today, the new hashes and images would be available for upcomign releases any way.

mauriciopoppe:
> What I am noticing or understanding here is, if we remove the image hash as in this PR and if we promote or update the other hash as v0.1.0 , it looks to be working.

I'm not sure if we can promote v0.1.0 again, we can check that we can't push v0.1.0 again by merging this PR and then running the generate.sh script again which will restore the line in another PR, I think that we'd see an error in the 2nd PR in pull-k8sio-cip similar to this one that I got in https://github.com/kubernetes/k8s.io/pull/3345

`level=fatal msg="runcip run: promoting images: [edge &{{gcr.io/k8s-staging-sig-storage true} {iscsiplugin v0.1.0} sha256:8ff2016fd563f0365ccf6ad4ab09f09a1941cbede5c62c5692e95589c0ed41de {asia.gcr.io/k8s-artifacts-prod/sig-storage k8s-infra-gcr-promoter@k8s-artifacts-prod.iam.gserviceaccount.com false} {iscsiplugin v0.1.0}}: tag 'v0.1.0' in dest points to sha256:31117074c183242b223340f2fc0f886243e97c6cb1a83d98cfb00a69d1746697, not sha256:8ff2016fd563f0365ccf6ad4ab09f09a1941cbede5c62c5692e95589c0ed41de (as per the manifest), but tag moves are not supported; skipping edge &{{gcr.io/k8s-staging-sig-storage true} {iscsiplugin v0.1.0} sha256:8ff2016fd563f0365ccf6ad4ab09f09a1941cbede5c62c5692e95589c0ed41de {eu.gcr.io/k8s-artifacts-prod/sig-storage k8s-infra-gcr-promoter@k8s-artifacts-prod.iam.gserviceaccount.com false} {iscsiplugin v0.1.0}}: tag 'v0.1.0' in dest points to sha256:31117074c183242b223340f2fc0f886243e97c6cb1a83d98cfb00a69d1746697, not sha256:8ff2016fd563f0365ccf6ad4ab09f09a1941cbede5c62c5692e95589c0ed41de (as per the manifest), but tag moves are not supported; skipping edge &{{gcr.io/k8s-staging-sig-storage true} {iscsiplugin v0.1.0} sha256:8ff2016fd563f0365ccf6ad4ab09f09a1941cbede5c62c5692e95589c0ed41de {us.gcr.io/k8s-artifacts-prod/sig-storage k8s-infra-gcr-promoter@k8s-artifacts-prod.iam.gserviceaccount.com false} {iscsiplugin v0.1.0}}: tag 'v0.1.0' in dest points to sha256:31117074c183242b223340f2fc0f886243e97c6cb1a83d98cfb00a69d1746697, not sha256:8ff2016fd563f0365ccf6ad4ab09f09a1941cbede5c62c5692e95589c0ed41de (as per the manifest), but tag moves are not supported; skipping]"`

> Otherwise, we can just hack the script to forget about this image release tag for iscsiplugin in generate.sh, so it remains as we have today, the new hashes and images would be available for upcoming releases.

I think it'd be great if we could just move the tag to the v0.1.0 hash that we already have, I'm not sure about the command that we would need to do that though

humblec:
>>I'm not sure if we can promote v0.1.0 again, we can check that we can't push v0.1.0 again by merging this PR and then running the generate.sh script again which will restore the line in another PR, I think that we'd see an error in the 2nd PR in pull-k8sio-cip similar to this one that I got in https://github.com/kubernetes/k8s.io/pull/3345
>>
iic , it happens only both tags are updated in the same PR.. and seperate PR would work I believe. 

>I think it'd be great if we could just move the tag to the v0.1.0 hash that we already have, I'm not sure about the command that we would need to do that though
>
Sure, that works too.. I will wait for @msau42's suggestion..

msau42:
I will see if I can modify the v0.1.0 tag in staging to the hash that was promoted. 

@humblec can you explain the steps that caused the v0.1.0? I would like to see what we can do to prevent this in the future.

msau42:
I manually changed the tag in the staging repo to match what was promoted.

/close

PR #3371, opened 2022-02-08
Fix releng script location

saschagrunert:
The script location does not seem to exist, so I'm fixing it by using
the next obvious one.

cc @ameukam


ameukam:
Thanks!

/lgtm
/approve

PR #3376, opened 2022-02-09
Update docs for public-log-asn-matcher

BobyMCbobs:
- preparation docs
- updates the image to use, to a production one
- clean up docs


hh:
Thanks @BobyMCbobs !

PR #3391, opened 2022-02-14
storage: update hashes of iscsiplugin and nfs-provisioner container images

humblec:
This commit advance/update the image hashes of iscsiplugin and nfs
provisioner

### DIFF
--- a/k8s.gcr.io/images/k8s-staging-sig-storage/images.yaml
+++ b/k8s.gcr.io/images/k8s-staging-sig-storage/images.yaml
@@ -101,7 +101,7 @@
     "sha256:6029c252dae6178c99b580de72d7776158edbc81be0de15cedc4152a3acfed18": [ "v1.7.3" ]
 - name: iscsiplugin
   dmap:
-    "sha256:31117074c183242b223340f2fc0f886243e97c6cb1a83d98cfb00a69d1746697": [ "v0.1.0" ]
+    "sha256:31117074c183242b223340f2fc0f886243e97c6cb1a83d98cfb00a69d1746697": [ "0.1-canary;v0.1.0" ]
 - name: livenessprobe
   dmap:
     "sha256:f8cec70adc74897ddde5da4f1da0209a497370eaf657566e2b36bc5f0f3ccbd7": [ "v1.1.0" ]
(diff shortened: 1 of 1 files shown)


k8s.gcr.io/images/k8s-staging-sig-storage/images.yaml
```diff
@@ -101,7 +101,7 @@
     "sha256:6029c252dae6178c99b580de72d7776158edbc81be0de15cedc4152a3acfed18": [ "v1.7.3" ]
 - name: iscsiplugin
   dmap:
-    "sha256:31117074c183242b223340f2fc0f886243e97c6cb1a83d98cfb00a69d1746697": [ "v0.1.0" ]
+    "sha256:31117074c183242b223340f2fc0f886243e97c6cb1a83d98cfb00a69d1746697": [ "0.1-canary;v0.1.0" ]
```
pohly, line 104: Did this come out of the `generate.sh` script? Are you sure that this syntax is valid?

We usually don't promote canary images, so isn't just `v0.1.0` enough?

humblec: @pohly it came out of `generate.sh` ( may be a confirmation from your end helps too)  .. so I believe this syntax is valid too..  due to a mistake in the hash recorded here for iscsiplugin , we ( with the help of @msau42 ) adjusted the tag in the backend repo  , thats why both of these in place..

pohly: https://github.com/kubernetes/k8s.io/blob/289fd8c5189e3cbff76ce0d5bf4fc035692da306/k8s.gcr.io/images/k8s-staging-sig-storage/generate.sh#L56

just passes through what it gets from gcloud, which seems to be `0.1-canary;v0.1.0` in this case now. I have no idea whether this is valid in `images.yaml`.

msau42: hmmmm I'm not sure if the promotion tool will work with this. Also we should not promote canary tags. Probably we need to fix generate.sh

PR #3402, opened 2022-02-15
Request kep SAN on TLS cert

enj:
Signed-off-by: Monis Khan <mok@vmware.com>

### DIFF
--- a/apps/k8s-io/certificate-prod.yaml
+++ b/apps/k8s-io/certificate-prod.yaml
@@ -40,6 +40,8 @@ spec:
   - issue.kubernetes.io
   - issues.k8s.io
   - issues.kubernetes.io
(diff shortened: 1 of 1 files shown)


apps/k8s-io/certificate-canary.yaml
```diff
@@ -47,6 +47,8 @@ spec:
   - issue.kubernetes.io
   - issues.k8s.io
   - issues.kubernetes.io
+  - kep.k8s.io
+  - kep.kubernetes.io
```
ameukam, line 51: @enj This is a self-sign certificate we plan to deprecate.
```suggestion
```

enj: Okay, I have reverted the changes to this file.

Issue #3356, opened 2022-02-04
Refactor GSuite Groups Reconciler Application


upodroid:
Hello

I'm a contributor with Knative and we have implemented the GSuite Groups Reconciler at https://github.com/knative/community/pull/905. However, this package has hardcoded references to `kubernetes.io`, critical validation is done `go test` instead of a cli sub command which forced us to clone the code and replace k8s specific references instead of `go install k8s.io/k8s.io/groups@v0.0.0-20220124221514-b894a331a999`.

I'm proposing:
- `groups validate` command - equivalent to `make test`
- `groups plan` command - equivalent to `make run`
- `groups apply` command - equivalent to `make run -- --confirm`
- Make the domain configurable through config.yaml

Thank you

/cc @dims @dprotaso

dims:
@upodroid sounds like the right thing to do. Hoping these are minimal changes since we have a bunch of things we'll need to fix and this is already "production" code for us.

PR #3407, opened 2022-02-16
Refactor the GSuite Reconciler

upodroid:
Fixes: https://github.com/kubernetes/k8s.io/issues/3356

There are a few prowjobs that need to be updated.

/cc @dims @evankanderson


groups/validate.go
```diff
@@ -17,29 +17,43 @@ limitations under the License.
...
 
+	if err := config.Load(*configPath, false); err != nil {
+		fmt.Printf("Could not load main config: %v\n", err)
+		os.Exit(1)
+	}
+
+	// Check if restrictions.yaml is loadable
 	if *restrictionsPath != "" && !filepath.IsAbs(*restrictionsPath) {
 		fmt.Printf("restrictions-path \"%s\" must be an absolute path\n", *restrictionsPath)
 		os.Exit(1)
```
evankanderson, line 59: This feels like it should be extracted from `Reconcile` in `main.go` so that the same loading code can be used here and in the `Reconcile` method. In fact, it seems like the `validate` command should be:

```golang
		RunE: func(cmd *cobra.Command, args []string) error {
			configFile, _ := cmd.Flags().GetString("config")
			config, restrictions, groups, err := ReadConfigFile(configFile)
			if err != nil {
				return err;
			}
			return Validate(config, restrictions, groups)
		},
```

evankanderson: (and `plan` and `apply` would add a call to `Reconcile(print_only, config, groups)` to the end of the `RunE` method above.)

groups/validate.go
```diff
@@ -73,27 +91,39 @@ func TestMain(m *testing.M) {
...
-	os.Exit(m.Run())
+
+	// Run all the Verification Logic
+	VerifyMergedGroupsConfig(config)
+	VerifyStagingEmailLength(config)
+	VerifyDescriptionLength(config)
+	VerifyGroupConventions(config)
+	VerifyK8sInfraGroupConventions(config)
+	VerifyK8sInfraRBACGroupConventions(config)
+	VerifySecurityResponseCommitteeGroups(config)
```
evankanderson, line 106: It feels like it might make sense to use the `config.SkipKubernetesIOTests` bool here to make it clear which verifications are skipped (rather than doing it inside the methods).

groups/validate.go
```diff
@@ -73,27 +91,39 @@ func TestMain(m *testing.M) {
...
 	var containsMergedConfig bool
 	found := sets.String{}
 	dups := sets.String{}
 
-	for _, g := range cfg.Groups {
+	for _, g := range groupsConfig.Groups {
 		name := g.Name
-		if name == "community" {
+
+		if name == "security" {
```
evankanderson, line 126: Why did this change? Was this a data change?

groups/validate.go
```diff
@@ -104,196 +134,232 @@ func TestMergedGroupsConfig(t *testing.T) {
...
-	for _, g := range cfg.Groups {
+	for _, g := range groupsConfig.Groups {
 		if strings.HasPrefix(g.EmailId, "k8s-infra-staging-") {
-			projectName := strings.TrimSuffix(strings.TrimPrefix(g.EmailId, "k8s-infra-staging-"), "@kubernetes.io")
+			projectName := strings.TrimSuffix(strings.TrimPrefix(g.EmailId, "k8s-infra-staging-"), "@"+primaryDomain)
 
 			len := utf8.RuneCountInString(projectName)
 			if len > 18 {
-				errs = append(errs, fmt.Errorf("Number of characters in project name \"%s\" should not exceed 18; is: %d", projectName, len))
+				errs = append(errs, fmt.Errorf("number of characters in project name \"%s\" should not exceed 18; is: %d", projectName, len))
```
evankanderson, line 161: It seems like there's an additional assumption here about the `k8s-staging-` prefix being 12 letters.

(See the comment at the beginning of this method.) I'm fine with leaving this as a "TODO" for now, but it seems like the string `k8s-infra-staging-` should be configurable, and the `18` value should be something like `36 - len(bot-prefix)`.

groups/validate.go
```diff
@@ -104,196 +134,232 @@ func TestMergedGroupsConfig(t *testing.T) {
...
 			}
 		}
 	}
 
 	if len(errs) > 0 {
 		for _, err := range errs {
-			t.Error(err)
+			fmt.Print(err)
+			os.Exit(1)
```
evankanderson, line 169: I think you either want to return a list of errors, or move the `os.Exit` out of the inner part of the loop into the `if len(errs) > 0` part of the loop.

groups/validate.go
```diff
@@ -104,196 +134,232 @@ func TestMergedGroupsConfig(t *testing.T) {
...
+				fmt.Errorf("number of characters in description \"%s\" for group name \"%s\" "+
 					"should not exceed 300; is: %d", description, g.Name, len))
 		}
 	}
 
 	if len(errs) > 0 {
 		for _, err := range errs {
-			t.Error(err)
+			fmt.Print(err)
+			os.Exit(1)
```
evankanderson, line 196: Ditto here.

groups/validate.go
```diff
@@ -104,196 +134,232 @@ func TestMergedGroupsConfig(t *testing.T) {
...
 }
 
-// TestDescriptionLength tests that the number of characters in the
+// VerifyDescriptionLength tests that the number of characters in the
 // google groups description does not exceed 300.
 //
 // This validation is needed because gcloud allows apps:description
 // with length no greater than 300
-func TestDescriptionLength(t *testing.T) {
+func VerifyDescriptionLength(config Config) {
```
evankanderson, line 179: `config` doesn't seem to be used here, but `groupsConfig.Groups` is, and I'd love to see that passed as a parameter.

groups/validate.go
```diff
@@ -104,196 +134,232 @@ func TestMergedGroupsConfig(t *testing.T) {
...
 
-// TestStagingEmailLength tests that the number of characters in the
+// VerifyStagingEmailLength tests that the number of characters in the
 // project name in emails used for staging repos does not exceed 18.
 //
 // This validation is needed because gcloud allows PROJECT_IDs of length
 // between 6 and 30. So after discounting the "k8s-staging" prefix,
 // we are left with 18 chars for the project name.
-func TestStagingEmailLength(t *testing.T) {
+func VerifyStagingEmailLength(config Config) {
```
evankanderson, line 152: Do you want to pass `groupsConfig.Groups` as a parameter here, too, so that it's not referencing the shared global. (Mostly for testing convenience and to make this a pure function.)

groups/validate.go
```diff
@@ -104,196 +134,232 @@ func TestMergedGroupsConfig(t *testing.T) {
...
+	for _, g := range groupsConfig.Groups {
 		if strings.HasPrefix(g.EmailId, "k8s-infra-rbac") {
 			rbacEmails[g.EmailId] = false
 			// this is necessary for group-based rbac to work
 			whoCanViewMembership, ok := g.Settings["WhoCanViewMembership"]
 			if !ok || whoCanViewMembership != "ALL_MEMBERS_CAN_VIEW" {
-				t.Errorf("group '%s': must have settings.WhoCanViewMembership = ALL_MEMBERS_CAN_VIEW", g.Name)
+				log.Printf("group '%s': must have settings.WhoCanViewMembership = ALL_MEMBERS_CAN_VIEW", g.Name)
+				errs++
 			}
```
evankanderson, line 258: I would love to have a pattern of:

```golang
errs := make([]error)
// in a later loop
	append(errs, fmt.Errorf("group %q: must have settings.WhoCanViewMembership = ALL_MEMBERS_CAN_VIEW", g.Name))

if len(errs) > 0 {
  // do something reasonable
}

File: groups/validate.go [hunk 10]
```diff
@@ -104,196 +134,232 @@ func TestMergedGroupsConfig(t *testing.T) {
...
 		}
 	}
 	foundGKEGroup := false
-	for _, g := range cfg.Groups {
-		if g.EmailId == "gke-security-groups@kubernetes.io" {
+	for _, g := range groupsConfig.Groups {
+		if g.EmailId == "gke-security-groups@"+primaryDomain {
```
evankanderson, line 263: I'm not sure swapping this for `primaryDomain` helps much in terms of portability. I'd leave this as k8s-specific and just opt people out of this test.

groups/validate.go
```diff
@@ -104,196 +134,232 @@ func TestMergedGroupsConfig(t *testing.T) {
...
 	}
 	if !foundGKEGroup {
-		t.Errorf("group '%s' is missing", "gke-security-groups@kubernetes.io")
+		log.Printf("group '%s' is missing", "gke-security-groups@"+primaryDomain)
+		errs++
 	}
 	for email, found := range rbacEmails {
 		if !found {
-			t.Errorf("group '%s': must be a member of gke-security-groups@kubernetes.io", email)
+			log.Printf("group '%s': must be a member of gke-security-groups@"+primaryDomain, email)
```
evankanderson, line 287: A 👎  on combining `log.Printf` and string `+` -- but I'd switch these back to `@kubernetes.io` anyway.

groups/validate.go
```diff
@@ -104,196 +134,232 @@ func TestMergedGroupsConfig(t *testing.T) {
...
 			if g.EmailId == pscGroup {
 				if !reflect.DeepEqual(owners, g.Owners) {
-					t.Errorf("group '%s': owners must match owners from security@kubernetes.io, expected: %v, actual: %v", pscGroup, owners, g.Owners)
+					log.Printf("group '%s': owners must match owners from security@kubernetes.io, expected: %v, actual: %v", pscGroup, owners, g.Owners)
+					errs++
 				}
 				break
 			}
 		}
 	}
```
evankanderson, line 326: It may not matter, but it would be more efficient to put the `pscGroups` into a set or map and scan `groupsConfig.Groups` only once (or 1.5 times, since we scan for `security@kubernetes.io` earlier.

Separately, it feels like having `groupsConfig.Groups` be a `map[string]GroupConfig` would be a big win for _some_ of these loops, especially if you could combine the "verification function" for several different verifications using something like the Visitor pattern. This is definitely a TODO item, not a "do it in this PR" item.

groups/validate.go
```diff
@@ -348,37 +417,47 @@ func TestHardcodedGroupsForParanoia(t *testing.T) {
 // and history of threads and also use web interface to operate the group)
 // More info:
 // 	https://developers.google.com/admin-sdk/groups-settings/v1/reference/groups#allowWebPosting
-func TestGroupsWhichShouldSupportHistory(t *testing.T) {
+func VerifyGroupsWhichShouldSupportHistory(config Config) {
```
evankanderson, line 420: I'd love to make this more general to enforce that groups matching a certain pattern must have web history / web posting, but again **not in this PR**.

evankanderson (commented):
A few cleanup comments -- a couple are musings on the code that should at most result in a TODO rather than any actual code changes.

I'm also suggesting you walk back a few of the code changes, because the validation seems Kubernetes-specific enough that it's not worth trying to make it generic right now (and we can always figure out how to do that in the future if someone ends up wanting something similar).

groups/main.go
```diff
@@ -0,0 +1,127 @@
...
+	}
+
+	apply := &cobra.Command{
+		Use:   "apply",
+		Short: "Reconciles the supplied configuration to Google Groups",
+		Run: func(cmd *cobra.Command, args []string) {
+			configFile, _ := cmd.Flags().GetString("config")
+			Reconcile(true, configFile)
+		},
+	}
```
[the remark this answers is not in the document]
upodroid: Can do. Fixed

groups/main.go
```diff
@@ -0,0 +1,127 @@
...
+		admin.AdminDirectoryGroupMemberScope,
+		groupssettings.AppsGroupsSettingsScope)
+	if err != nil {
+		log.Fatalf("Unable to authenticate using key in secret-version %s, %v", config.SecretVersion, err)
+	}
+	credential.Subject = config.BotID
+
+	ctx := context.Background()
+	client := credential.Client(ctx)
+	clientOption := option.WithHTTPClient(client)
```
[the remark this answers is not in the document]
upodroid: Fixed

groups/validate.go
upodroid: security is common for both projects right now.

groups/validate.go
upodroid: This is a useful for feature for anyone doing [Google Groups RBAC for GKE](https://cloud.google.com/kubernetes-engine/docs/how-to/google-groups-rbac). We'll be using this in Knative. In GSuite, there is a primary domain and secondary domains and key features are tied to the primary domain(item 0 in the list)

https://support.google.com/a/answer/7009324?hl=en

upodroid:
Forgot about this, will address the remaining feedback from Evan by tomorrow.

puerco:
CI failure fixed in https://github.com/kubernetes/k8s.io/pull/3495
/test pull-k8sio-verify

groups/reconcile.go
```diff
@@ -493,3 +417,12 @@ func accessSecretVersion(secretVersion string) ([]byte, error) {
...
 	return result.Payload.Data, nil
 }
+
+func PrintConfig(config Config) {
+	log.Printf("config: BotID:            %v", config.BotID)
+	log.Printf("config: SecretVersion:    %v", config.SecretVersion)
+	log.Printf("config: GroupsPath:       %v", config.GroupsPath)
+	log.Printf("config: RestrictionsPath: %v", config.RestrictionsPath)
+	log.Printf("config: ConfirmChanges:   %v", config.ConfirmChanges)
+	log.Printf("config: Domains:   %v", config.Domains)
```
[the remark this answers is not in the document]
upodroid: Fixed

groups/validate.go
upodroid: Fixed

groups/validate.go
upodroid: Fixed

groups/validate.go
upodroid: Fixed

groups/validate.go
upodroid: Added TODO note

groups/validate.go
upodroid: Added TODO note

groups/validate.go
upodroid: Passed as parameters to all relevant functions.

upodroid:
@puerco The CI is still broken. The yamllint error is for k8s.gcr.io/* files. Did rebase from main now.

dims:
let's close this for now. it's not getting eyes and it needs a rebase anyway as many things have changed.

PR #3410, opened 2022-02-16
A group may not be the owner or manager of another group

dims:
We have 2 instances where we fail on $TITLE.

Signed-off-by: Davanum Srinivas <davanum@gmail.com>


endocrimes (approved):
/lgtm
/approve

for wg-batch

Issue #3431, opened 2022-02-21
Sippy is down


wojtek-t:
Sippy is crashlooping. Looking into status, it seems like a broken image - the failures look like this
```
State:          Waiting
      Reason:       CrashLoopBackOff
    Last State:     Terminated
      Reason:       StartError
      Message:      failed to create containerd task: OCI runtime create failed: container_linux.go:380: starting container process caused: exec: "/go/src/sippy/scripts/fetchdata-kube.sh": stat /go/src/sippy/scripts/fetchdata-kube.sh: no such file or directory: unknown
      Exit Code:    128
```

@deads2k

wojtek-t:
I think we should built a stable image and start using it, instead of the current:
https://github.com/kubernetes/k8s.io/blob/main/apps/sippy/deployment.yaml#L21

wojtek-t:
This only partially fixed the problem:
- sippy is running down, but is not fetching the data

The fetching container is failing with:
```
 Last State:     Terminated
      Reason:       StartError
      Message:      failed to create containerd task: OCI runtime create failed: container_linux.go:380: starting container process caused: exec: "/bin/fetchdata-kube.sh": stat /bin/fetchdata-kube.sh: no such file or directory: unknown
      Exit Code:    128
      Started:      Thu, 01 Jan 1970 00:00:00 +0000
      Finished:     Fri, 25 Feb 2022 07:20:50 +0000
    Ready:          False
```

[Which seems a bit strange to me.]

PR #3449, opened 2022-02-24
update sippy to match new docker image location

deads2k:
fixes https://github.com/kubernetes/k8s.io/issues/3431

at some point these got moved in the sippy Dockerfile: https://github.com/openshift/sippy/blob/master/Dockerfile#L10-L13

/assign @wojtek-t


wojtek-t:
/lgtm
/approve

Thanks!

PR #3469, opened 2022-03-03
Promote patch releases for v0.21 and v0.22

damemi:
CVE pointed out to us in this PR also affects Go 1.16, which is on these versions (https://github.com/kubernetes-sigs/descheduler/pull/744)

ref: https://github.com/kubernetes-sigs/descheduler/pull/745 and https://github.com/kubernetes-sigs/descheduler/pull/746


damemi (commented):
/cc @ingvagabund 
ptal, thanks!

PR #3521, opened 2022-03-16
Update public log asn matcher (2022-03-16)

BobyMCbobs:
- add (bash) debug mode
- update buckets.txt
- remove BQ_OUTPUT and redirection
- fix evaluation of bash array


BobyMCbobs:
> /approve /lgtm

Woohoo! Thank you @dims :tada:

PR #3604, opened 2022-04-07
groups/*_test.go: use %v instead of %w

neolit123:
$ go version
go version go1.18 linux/amd64

t.Errorf() and other output functions
for tests do not support %w wrapping.

Reported error:
"..Errorf does not support error-wrapping directive %w"

Use %v instead of %w.

https://stackoverflow.com/questions/70505459/testing-common-errorf-does-not-support-error-wrapping-directive-w
https://github.com/golang/go/issues/35472

for some reason these errors are not visible in the current CI logs.
possibly the go version there is still < 1.18 :thinking:

### DIFF
--- a/groups/reconcile_test.go
+++ b/groups/reconcile_test.go
@@ -703,7 +703,7 @@ func TestReconcileGroups(t *testing.T) {
 			expectedState = c.desiredState
 		}
 		if err = s.isReconciled(expectedState); err != nil {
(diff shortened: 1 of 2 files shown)


MadhavJivrajani (approved):
Thanks!

MadhavJivrajani:
/assign @dims @ameukam 
for approval

Issue #2182, opened 2021-03-09
Where to host cs.k8s.io


ameukam:
https://cs.k8s.io is running on a baremetal server [provided](https://github.com/cncf/cluster/issues/67)  by Equinix Metal(ex Packet) under CNCF budget and operated until now by @dims. 

The question was asked about whether or not we should host CodeSearch on [aaa](https://github.com/kubernetes/k8s.io/blob/main/running-in-community-clusters.md#where-are-the-clusters) cluster. 

Ref: https://kubernetes.slack.com/archives/CCK68P2Q2/p1615204807111900?thread_ts=1615189697.108500&cid=CCK68P2Q2

Issue open to track the discussions and the consensus about this.

jimdaga:
I took a stab at onboarding codesearch; @spiffxp could I get your input? I want to make sure I didn't miss anything. 
I want to stage all the infra, and get it deployed via prow first. Then we can follow up with another PR to cut-over DNS when we are ready.

https://github.com/kubernetes/k8s.io/pull/2513
https://github.com/kubernetes/test-infra/pull/23201

I could also work on adding the docker build logic after, but I haven't worked in that repo yet so I'll have to do some digging. 

cc @dims

justaugustus:
What about using https://sourcegraph.com/kubernetes to minimize the maintenance burden here?
This is something I suggested to @dims in the past, but didn't have the bandwidth to do at the time.

dims:
choices are:
1. leave things where they are 
2. move to k8s wg infra
3. redirect to cs.k8s.io to sourcegraph

- i have been taking care of 1 already for a while with minimal downtime, so i am ok with continuing to do so
- if someone wants to do 2, i am happy to work with help, show how things are setup and we can shut down the equinix vm
- i personally don't like option 3, i love the the hound UX, if the consensus is we should go with 3, that is fine with me. I am happy to run a personal instance on a custom domain for myself (community is welcome to use)

if i missed any other options, please feel free to chime in.

jimdaga:
FYI: If choice #2 is picked, my two PRs are pretty much ready to stage codesearch in the `aaa` cluster. There are a few small things that need to happen after the merge, but that's documented in my PRs. 
- https://github.com/kubernetes/k8s.io/pull/2513
- https://github.com/kubernetes/test-infra/pull/23201

dims:
thanks @jimdaga 

[...]

PR #3679, opened 2022-04-27
Add restart script to codesearch for daily restart

jimdaga:
Resolves: https://github.com/kubernetes/k8s.io/issues/2182 

This change adds a `restart.sh` script that will be run daily by the `periodic-k8sio-deploy-app-codesearch` prow job.
This will ensure we restart codesearch daily and update search indexes to include any new git changes. 
Since we are using rolling upgrades with health checks, restarts will (should) NOT cause downtime.

### DIFF
--- a/apps/codesearch/restart.sh
+++ b/apps/codesearch/restart.sh
@@ -47,4 +47,4 @@ fi
 
 # deploy kubernetes resources
 pushd "${SCRIPT_ROOT}" >/dev/null
-kubectl --context="${context}" --namespace="${namespace}" rollout restart deployment ${app}
+kubectl --context="${context}" --namespace="${namespace}" rollout restart deployment "${app}"
--- a/apps/codesearch/restart.sh
+++ b/apps/codesearch/restart.sh
@@ -1,0 +1,50 @@
+#!/usr/bin/env bash
+
+# Copyright 2021 The Kubernetes Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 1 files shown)


apps/codesearch/restart.sh
```diff
@@ -0,0 +1,50 @@
...
+
+# ensure we have a context to talk to the target cluster
+if ! kubectl config get-contexts "${context}" >/dev/null 2>&1; then
+  gcloud container clusters get-credentials "${cluster_name}" --project="${cluster_project}" --region="${cluster_region}"
+  context="${gke_context}"
+fi
+
+# deploy kubernetes resources
+pushd "${SCRIPT_ROOT}" >/dev/null
+kubectl --context="${context}" --namespace="${namespace}" rollout restart deployment "${app}"
```
ameukam, line 50: the deployment is restarted but we don't know if it's successful. (we should also do that for other deployments)

```suggestion
kubectl --context="${context}" --namespace="${namespace}" rollout restart deployment "${app}"
kubectl --context="${context}" --namespace="${namespace}" rollout status deployment "${app} --watch --timeout=2m"
```

jimdaga: @ameukam do you have a way to run this command manually and get a benchmark of how long it takes? 
```
time kubectl -n codesearch rollout status deployment codesearch --watch --timeout=999m
```
I agree adding these flags make sense, but because there is so much data the init pod takes a long time and we will need a rather large timeout window.

ameukam: You should be able to do it using Cloud Shell https://shell.cloud.google.com/. I ran the command: 

```
ameukam@cloudshell:~ (kubernetes-public)$ time kubectl -n codesearch rollout status deployment codesearch --watch --timeout=999m
Waiting for deployment "codesearch" rollout to finish: 1 out of 2 new replicas have been updated...
error: deployment "codesearch" exceeded its progress deadline

real	9m58.608s
user	0m0.155s
sys	0m0.074s
```

PR #3683, opened 2022-04-27
[DO NOT MERGE] Image promotion for cluster-api v0.99.99

CecileRobertMichon:
Image promotion for cluster-api v0.99.99
This is an automated PR generated from `kpromo`
```
kpromo pr --fork CecileRobertMichon --project cluster-api --reviewers "@CecileRobertMichon @enxebre @fabriziopandini @sbueringer @vincepri" --tag v0.99.99 --image cluster-api-controller --image kubeadm-control-plane-controller --image kubeadm-bootstrap-controller
```

/hold
cc: @CecileRobertMichon @enxebre @fabriziopandini @sbueringer @vincepri

PR #3687, opened 2022-04-28
Cluster API: reorder image tags lexicographically

CecileRobertMichon:
This PR changes the order of the images in the images.yaml manifest to match the order used by kpromo pr (lexicographical order instead of semver to support both semver and non-semver tags) so that future automated PRs opened by kpromo pr don't carry an additional diff (see also https://github.com/kubernetes/k8s.io/pull/3683). This does not add or remove any images.


CecileRobertMichon (commented):
/assign @sbueringer @fabriziopandini

PR #3698, opened 2022-05-03
Adding sig-windows-leads@kuernetes.io group

marosset:
Signed-off-by: Mark Rossetti <marosset@microsoft.com>

Adding sig-windows-leads group.

Group reference already exists in restrictions.yaml https://github.com/kubernetes/k8s.io/blob/634769ad9f9bdf7fa0f4a831812e3d07f5588f3f/groups/restrictions.yaml#L189


jsturtevant:
lgtm

PR #3730, opened 2022-05-12
Promote VPA 0.11.0 images

jbartosik:
I got hashes with gcloud, for example:

```
gcloud container images describe gcr.io/k8s-staging-autoscaling/vpa-admission-controller:0.11.0
```

I tested images by creating a cluster, setting up VPA using those images
and running `full-vpa` e2e tests.


jbartosik:
Let's wait with this a while - [VPA e2e tests are failing](https://github.com/kubernetes/autoscaler/issues/4886). I don't think this is a problem with VPA but I want to make sure.

/hold

jbartosik:
/hold cancel

I fixed VPA E2E tests. Tests were failing because they run with K8s from HEAD and for 1.25 some APIs were removed.

So I think there are no problems with images I built:
- It looks like VPA passed some e2e runs with this code
- I identified problems, fixed them and they shouldn't appear until 1.25
- Images passed my manual tests

jbartosik:
@mwielgus Please take a look

mwielgus (approved):
/lgtm
/approve

PR #3795, opened 2022-06-01
Fix CAA records for anything that uses docs wildcard

cblecker:
netlifyglobalcdn.com has a CAA record, and CAAs follow our CNAMEs to that record. That record is making us unable to use Let's Encrypt. We don't need an extra issuer domain, as netlify doesn't issue the certs directly.

This should hopefully fix all the cert issuance issues, as netlify.app doesn't have a CAA record.


cblecker:
/hold

Holding so I can be around to watch this merge

PR #3820, opened 2022-06-09
promote attacher, volume populator sidecars to the latest release

humblec:
Reference#
https://github.com/kubernetes-csi/external-attacher/releases/tag/v3.5.0
https://github.com/kubernetes-csi/lib-volume-populator/releases/tag/v1.1.0


humblec:
Cc @jsafrane @bswartz

pohly (approved):
/lgtm
/approve

bswartz:
This PR is fine, but for some reason it didn't pick up volume-data-source-validator v1.1.0. I wonder if the script has a bug that caused it to miss that image.

humblec:
> This PR is fine, but for some reason it didn't pick up volume-data-source-validator v1.1.0. I wonder if the script has a bug that caused it to miss that image.

Its actually gettting picked up, however while discarding the chunks of some of the hash updates ( nfsplugin, smbplugin..etc) script brings, it went unnoticed..

pohly:
Humble Devassy Chirammal ***@***.***> writes:

>> This PR is fine, but for some reason it didn't pick up volume-data-source-validator v1.1.0. I wonder if the script has a bug that caused it to miss that image.
>
> Its actually gettting picked up, however while discarding the chunks
> of some of the hash updates ( nfsplugin, smbplugin..etc) script brings

Why are there updates that we don't want to promote? That looks like a
manual step that someday someone will submit accidentally.

pohly:
Looks like my emails just made it through to GitHub with a considerable delay - I wrote this in June...

I think @humblec since then has already cleaned up the script.

PR #3834, opened 2022-06-13
Revert tag move of kueue v0.1

alculquicondor:
It's not supported by the promotion bot.

https://prow.k8s.io/view/gs/kubernetes-jenkins/logs/post-k8sio-image-promo/1536425226437398528

We should keep the old tag as some people might already be using it.

kubernetes-sigs/kueue#272


ArangoGutierrez (approved):
/lgtm

Issue #3836, opened 2022-06-14
Need a new hack/ verify script that validates project name length


dims:
From https://cloud.google.com/resource-manager/docs/creating-managing-projects we have the following restrictions:
```
Project ID requirements:

Must be 6 to 30 characters in length.
Can only contain lowercase letters, numbers, and hyphens.
Must start with a letter.
Cannot end with a hyphen.
Cannot be in use or previously used; this includes deleted projects.
Cannot contain restricted strings, such as google and ssl.
```

For example the following project name is invalid `k8s-staging-cloud-provider-openstack`:
https://github.com/kubernetes/k8s.io/blob/343320108a89136fd4cad92b1dda6765d251dffe/infra/gcp/infra.yaml#L274

So we need a shellscript that uses `yq` to enumerate the projects in the `/infra/gcp/infra.yaml` file and ensures that each project is between `6 to 30 characters in length`.

ameukam:
@Priyankasaggu11929 are we consider this as done ?

Priyankasaggu11929:
yes @ameukam :)

ameukam:
@Priyankasaggu11929 Thanks!

/sig k8s-infra
/milestone v1.25
/close

PR #3887, opened 2022-06-22
add verify-gcp-projects-id.sh hack script

Priyankasaggu11929:
**What this PR do:**

The PR adds a hack script that uses `yq` to enumerate the projects in the `/infra/gcp/infra.yaml` file and ensures that each project is between 6 to 30 characters in length.

The hack script validates the project-id for following criteria:

- Must be 6 to 30 characters in length.
- Can only contain lowercase letters, numbers, and hyphens.
- Must start with a letter.
- Cannot end with a hyphen.

The hack script is missing validations for:

- Cannot be in use or previously used; this includes deleted projects.
- Cannot contain restricted strings, such as google and ssl.

---

**What issue the PR fixes:**

Part of issue: https://github.com/kubernetes/k8s.io/issues/3836

### DIFF
--- a/hack/verify-gcp-projects-id.sh
+++ b/hack/verify-gcp-projects-id.sh
@@ -1,0 +1,55 @@
+#!/usr/bin/env bash
+
+# Copyright 2022 The Kubernetes Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# Usage: `verify-gcp-project-name.sh`
+
+set -o errexit
+set -o nounset
+set -o pipefail
+
+SCRIPT_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+REPO_ROOT="$(cd "${SCRIPT_DIR}/.." && pwd )"
+TMPDIR="${TMPDIR:-${REPO_ROOT}/tmp}"
+
+GCP_INFRA_YAML="${REPO_ROOT}/infra/gcp/infra.yaml"
+LIST_OF_GCP_PROJECTS="${TMPDIR}/list_of_gcp_projects.txt"
+VERIFIED_PROJECTS="${TMPDIR}/verified_project_list.txt"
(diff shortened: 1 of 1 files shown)


Priyankasaggu11929:
cc: @dims

hack/verify-gcp-projects-id.sh
```diff
@@ -0,0 +1,47 @@
+#!/usr/bin/env bash
+
+# Copyright 2016 The Kubernetes Authors.
```
ameukam, line 3:
```suggestion
# Copyright 2022 The Kubernetes Authors.
```

Priyankasaggu11929: Thanks for pointing, Arnaud. 🙂

Priyankasaggu11929:
I'll fix the failing script in my morning. Looks like my local `yq` & the container `yq` are different implementations - go & python

Priyankasaggu11929:
@dims, The script failed with actual error this time. The project id `k8s-staging-cloud-provider-openstack` is invalid

Screenshot 2022-06-23 at 5 38 29 PM

---

**Need input on the following:**
 
The script takes care of all the GCP Project ID naming criterias but the following:

> Cannot be in use or previously used; this includes deleted projects.

Since I can't hit any GCP APIs to grab those "deleted projects" information, to implement the check for "deleted projects", is there any list we maintain for deleted projects? Also is this something we want the script to check?

> Cannot contain restricted strings, such as google and ssl.

Regarding the project-id having restricted strings, I was not able to find out a specific list anywhere from GCP regarding this.

dims:
thanks for working on this @Priyankasaggu11929 !! the script LGTM. I need to fix the broken project related stuff before we can merge this.

/assign

Priyankasaggu11929:
/hold

I think the test failed :|

hack/verify-gcp-projects-id.sh
```diff
@@ -0,0 +1,55 @@
...
+
+ for index in $(seq 0 $loop_index_length);
+ do
+	 project=$(yq ".infra | keys | .[${index}]" "${GCP_INFRA_YAML}")
+	 yq ".infra.${project}.projects | keys | .[]" "${GCP_INFRA_YAML}" >> "${LIST_OF_GCP_PROJECTS}"
+	 yq ".infra.${project}.projects | keys | .[] | match(\"^[a-z][a-z0-9-]{5,28}[^-]$\").string" "${GCP_INFRA_YAML}" >> "${VERIFIED_PROJECTS}"
+done
+
+
+if eval "$(diff "${LIST_OF_GCP_PROJECTS}" "${VERIFIED_PROJECTS}")"
```
dims, line 46: do we need eval here? (please see scripts under hack/ in k/k, there are several that do `diff` without eval)

Priyankasaggu11929: removed eval from the diff

Issue #3840, opened 2022-06-15
DNS REQUEST: cloud-provider-azure.sigs.k8s.io


MartinForReal:
### Type of DNS update:
Create
### Domain being modified:
k8s.io
### New DNS Record:

```yaml
cloud-provider-azure.sigs:
- type: CNAME
  value: kubernetes-sigs-cloud-provide-azure.netlify.app
```

### Reason for update:

The cloud proviider azure subproject has requested a netlify site to host its documentation via https://github.com/kubernetes/org/issues/3268?notification_referrer_id=NT_kwDOAE91trIzMTI0MjY5NzcyOjUyMDc0Nzg#issuecomment-1155121919

PR #3889, opened 2022-06-22
DNS REQUEST: cloud-provider-azure.sigs.k8s.io

upodroid:
Fixes: #3840

### DIFF
--- a/dns/zone-configs/k8s.io._0_base.yaml
+++ b/dns/zone-configs/k8s.io._0_base.yaml
@@ -412,3 +412,7 @@ kubectl.sigs:
 gateway-api.sigs:
   type: CNAME
   value: kubernetes-sigs-gateway-api.netlify.app.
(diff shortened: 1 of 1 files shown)


dns/zone-configs/k8s.io._0_base.yaml
```diff
@@ -412,3 +412,7 @@ kubectl.sigs:
 gateway-api.sigs:
   type: CNAME
   value: kubernetes-sigs-gateway-api.netlify.app.
+# https://github.com/kubernetes-sigs/cloud-provider-azure docs (@MartinForReal)
+cloud-provider-azure.sigs:
+  type: CNAME
+  value: kubernetes-sigs-cloud-provide-azure.netlify.app.
```
dims, line 418: `provide` should be `provider`?

dims: ack ok

PR #3904, opened 2022-06-27
Rework `oci-proxy` Cloud Run Terraform Code

upodroid:
Part of: https://github.com/kubernetes-sigs/oci-proxy/issues/77

I rewrote the terraform code to allow per region modification of each cloud run service now. Right now, you can specify env variables for each specific region but other values can be changed later too.

It still needs more improvements but i'll need privileges to execute those changes. For example, you should be using project-factory module.

/cc @ameukam @BenTheElder


ameukam:
We need to close this PR and move the code to github.com/kubernetes-sigs/oci-proxy.

BenTheElder:
xref: https://github.com/kubernetes-sigs/oci-proxy/issues/76

I would prefer that staging match production closely actually, not matching is causing issues (see above issue) so wherever you want to store production config, we should also do staging as only light overrides on e.g. image version in the same place.

I could see it being easier to store both in oci-proxy repo, either works for me.

upodroid:
/close

The work is continuing at https://github.com/kubernetes-sigs/oci-proxy/pull/95

PR #3911, opened 2022-06-28
Add myself to oci-proxy-admins group

upodroid:
Need this for #3904 sandbox instance of oci-proxy.

/cc @ameukam


ameukam:
/hold

I want to keep the humans capable to operate on registry.k8s.io only SIG chairs and TLs. 

I would prefer we focus on automation of the config changes.

PR #3919, opened 2022-06-30
Add registry.k8s.io AWS account structure docs

BobyMCbobs:
Adds a diagram with docs on accounts and AWS accounts


BobyMCbobs:
Related: https://github.com/cncf-infra/aws-infra/issues/4

Issue #3935, opened 2022-07-06
Add list support to registry.k8s.io S3 buckets


BobyMCbobs:
From the July 6th/7th k8s-infra meeting.

It might be useful to support listing of the blobs for these buckets.

/assign

ameukam:
The main intent is to replicate the existing behavior from the GCS buckets. e.g. 

```bash
❯ gsutil ls gs://us.artifacts.k8s-artifacts-prod.appspot.com
gs://us.artifacts.k8s-artifacts-prod.appspot.com/containers/
❯ gsutil ls gs://us.artifacts.k8s-artifacts-prod.appspot.com/containers
gs://us.artifacts.k8s-artifacts-prod.appspot.com/containers/images/
```

BobyMCbobs:
Depends on https://github.com/cncf-infra/aws-infra/pull/10

PR #3936, opened 2022-07-06
Update list perms for buckets

BobyMCbobs:
also change to use s3admin role for management

Fixes: https://github.com/kubernetes/k8s.io/issues/3935


BobyMCbobs:
Thank you, @ameukam!

PR #3955, opened 2022-07-12
WIP: Add s3 bucket syncing

BobyMCbobs:
S3 supports batch jobs for object replication.
`us-east-2` was temporarily picked as a good enough region and as a "main" bucket.
A replication configuration must be placed in the same region as the source bucket and named according to the source and destination buckets (for identifiability).

Currently the _aws_s3_bucket_replication_configuration_ fails to create due to the provider referenced in it's object.

Extra notes:
- should there be a main bucket?
  - all buckets could sync between each other

Related: https://github.com/cncf-infra/aws-infra/pull/11

### DIFF
--- a/infra/aws/terraform/registry.k8s.io/main.tf
+++ b/infra/aws/terraform/registry.k8s.io/main.tf
@@ -29,101 +29,124 @@ variable "prefix" {
   }
 }
 
+data "aws_s3_bucket" "registry-k8s-io-main" {
+  provider = aws.us-east-2
+  bucket   = "prod-registry-k8s-io-us-east-2"
+}
+
 module "us-west-1" {
   source = "./s3"
 
   providers = {
-    aws = aws.us-west-1
+    aws           = aws.us-west-1
+    aws.us-east-2 = aws.us-east-2
   }
 
-  region = "us-west-1"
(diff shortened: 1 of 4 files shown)


BobyMCbobs:
While the slow process of manually syncing is happening, I've been looking the automated syncing.
If AWS is able to do the heavy lifting then that's a win for speed and progress of this task.

BobyMCbobs:
The current error this is getting is a _s3/PutBucketReplication_, although the IAM policy that the role is using has _s3:*_

BobyMCbobs:
> The current error this is getting is a _s3/PutBucketReplication_, although the IAM policy that the role is using has _s3:*_

This issue is to be raised with AWS support. Currently the account where the buckets are running in needs some sort of support upgrade in order to do that.

BobyMCbobs:
Superseded by https://github.com/kubernetes/k8s.io/pull/4118.

In this PR, I was attempting to sync between all of the buckets so that any bucket can become the main or current bucket.
I'd like to revisit it since it's functionality means for a way to have a random _leader_ with consistency.
It didn't make anymore progress since the Terraform interweaving made it hard to pass the providers through.

cc @Riaankl @hh

PR #3988, opened 2022-07-18
Add registry-k8s-io sync scripts

BobyMCbobs:
These are the scripts that I have built and optimized for manual syncing k8s.gcr.io GCS bucket and the registry.k8s.io S3 buckets

- `initial-copy-to-s3.sh` copies from GCS to S3 (_us-east-2_)
- `initial-copy-between-s3.sh` copies from _us-east-2_ to the other buckets and uses tmate to run in parallel

The scripts are intended to be run by any IAM user with ability to assume the _registry.k8s.io_s3writer_ role in the _registry.k8s.io_ account, such as the _registry.k8s.io-ci_ user in the _k8s-infra-accounts_ account.

### DIFF
--- a/registry.k8s.io/hack/initial-copy-between-s3.sh
+++ b/registry.k8s.io/hack/initial-copy-between-s3.sh
@@ -30,12 +30,14 @@ REGIONS=(
 
 SOURCE=s3:prod-registry-k8s-io-us-east-2
 
+CALLER_ID="$(aws sts get-caller-identity --output json | jq -r .UserId)"
+
 for REGION in "${REGIONS[@]}"; do
     DESTINATION="s3dest:prod-registry-k8s-io-${REGION:-}"
     unset AWS_ACCESS_KEY_ID AWS_SECRET_ACCESS_KEY AWS_SESSION_TOKEN
     JSON=$(aws sts assume-role \
         --role-arn "arn:aws:iam::513428760722:role/registry.k8s.io_s3writer"  \
-        --role-session-name registry.k8s.io_s3writer \
+        --role-session-name "${CALLER_ID:-}-registry.k8s.io_s3writer" \
         --duration-seconds 43200 \
         --output json || exit 1)
 
--- a/registry.k8s.io/hack/initial-copy-between-s3.sh
+++ b/registry.k8s.io/hack/initial-copy-between-s3.sh
@@ -41,10 +41,10 @@ for REGION in "${REGIONS[@]}"; do
         --duration-seconds 43200 \
         --output json || exit 1)
 
-    export \
-        AWS_ACCESS_KEY_ID=$(echo "${JSON}" | jq --raw-output ".Credentials[\"AccessKeyId\"]") \
-        AWS_SECRET_ACCESS_KEY=$(echo "${JSON}" | jq --raw-output ".Credentials[\"SecretAccessKey\"]") \
-        AWS_SESSION_TOKEN=$(echo "${JSON}" | jq --raw-output ".Credentials[\"SessionToken\"]")
+    AWS_ACCESS_KEY_ID=$(echo "${JSON}" | jq --raw-output ".Credentials[\"AccessKeyId\"]")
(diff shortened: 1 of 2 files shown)


registry.k8s.io/hack/initial-copy-between-s3.sh
```diff
@@ -0,0 +1,75 @@
...
+    JSON=$(aws sts assume-role \
+        --role-arn "arn:aws:iam::513428760722:role/registry.k8s.io_s3writer"  \
+        --role-session-name "${CALLER_ID:-}-registry.k8s.io_s3writer" \
+        --duration-seconds 43200 \
+        --output json || exit 1)
+
+    AWS_ACCESS_KEY_ID=$(echo "${JSON}" | jq --raw-output ".Credentials[\"AccessKeyId\"]")
+    AWS_SECRET_ACCESS_KEY=$(echo "${JSON}" | jq --raw-output ".Credentials[\"SecretAccessKey\"]")
+    AWS_SESSION_TOKEN=$(echo "${JSON}" | jq --raw-output ".Credentials[\"SessionToken\"]")
+    export AWS_ACCESS_KEY_ID AWS_SECRET_ACCESS_KEY AWS_SESSION_TOKEN
```
jaypipes, line 47: looks familiar! ;)

registry.k8s.io/hack/initial-copy-between-s3.sh
```diff
@@ -0,0 +1,75 @@
...
+[s3dest]
+type = s3
+provider = AWS
+access_key_id = $AWS_ACCESS_KEY_ID
+secret_access_key = $AWS_SECRET_ACCESS_KEY
+session_token = $AWS_SESSION_TOKEN
+region = ${REGION}
+EOF
+    echo "Running sync between '${SOURCE:-}' and '${DESTINATION:-}'"
+    tmate -F -v -S "$TMATE_SOCKET" new-window -d -c "$PWD" -n sync-to-"${REGION:-}" "rclone sync --config \"${RCLONE_CONFIG:-}\" -P \"${SOURCE:-}\" \"${DESTINATION:-}\""
```
jaypipes, line 74: Interesting use of `tmux` ... nothing wrong with this per-se, but might want to mention in a comment at the top of this file that `tmate` is required to run the script.

registry.k8s.io/hack/initial-copy-between-s3.sh
```diff
@@ -0,0 +1,75 @@
...
+    us-west-1
+    us-west-2
+)
+
+SOURCE=s3:prod-registry-k8s-io-us-east-2
+
+CALLER_ID="$(aws sts get-caller-identity --output json | jq -r .UserId)"
+
+for REGION in "${REGIONS[@]}"; do
+    DESTINATION="s3dest:prod-registry-k8s-io-${REGION:-}"
```
jaypipes, line 36: here and throughout...

little nit: you know that `$REGION` will be defined already so no reason to use the defaulting syntax of `${REGION:-}`. You can just use `$REGION`

registry.k8s.io/hack/initial-copy-between-s3.sh
BobyMCbobs: It's a nifty few lines! Works well :smile:

registry.k8s.io/hack/initial-copy-between-s3.sh
BobyMCbobs: It's definitely all about manual local syncing. I've added a note in 7b1e676e1

registry.k8s.io/hack/initial-copy-between-s3.sh
BobyMCbobs: I write it like `${REGION:-}` instead of `$REGION` since sometimes `shellcheck` likes to have a fallback value, which is allowed to be empty. Something around variable declaration maybe?

jaypipes (approved):
ok dokey.

ameukam:
/approve
/hold

BobyMCbobs:
Looking for an lgtm once again

/unhold

cc @ameukam @jaypipes

dims:
/approve
/hold

BobyMCbobs:
Thank you @ameukam!

PR #4007, opened 2022-07-25
Update public-log-asn-matcher

BobyMCbobs:
Changes include
- add k8s.gcr.io tags table


BobyMCbobs:
Not sure we need the k8s_gcr_io_tags table at this point.

PR #4015, opened 2022-07-26
Create a new sig-scheduling-leads group

Huang-Wei:
There was an old sig-scheduling-leads mailing list named "kubernetes-sig-scheduling-leads@googlegroups.com", which was stale and none of the current scheduling leads had access.

This PR creates a new leads group for sig-scheduling for better communication and sharing private info (like zoom account) among leads.

cc @ahg-g @alculquicondor


alculquicondor (approved):

alculquicondor:
It needs to also be added here: https://github.com/kubernetes/k8s.io/blob/7a292c9613be1864c08f26e4ecfa24d4f5769b3e/groups/restrictions.yaml#L159

PR #4018, opened 2022-07-27
promote baseimage built by pr 8870

longwuyuan:
- This PR promotes the base-image built because of https://github.com/kubernetes/ingress-nginx/pull/8870

/assign @tao12345666333 @rikatz @strongjz


tao12345666333:
Thanks! Did this image build just because we fixed the CI, or did we actually make some changes?

longwuyuan:
This image was built only because of CI change.
But we should use this image because although this is a rebuild based on alpine v3.16.1, it will have patched binaries that are more recent than the one build before this (that was also v3.16.1).

Also, this is cleaner execution as per build history because then we will be consistent with the registry's latest image. Alternative is to put the one previous build's sha here so NGINX_BASE file will have the sha not from the latest images in the registry.

Waiting for this eagerly so that we can get up to date on the important other images like e2e-test-runner with ginkgo2 and cfssl etc.

longwuyuan:
And only after after this we will get the patch of busybox + ssl_client CVE. Then whenever possible , the cherrypick to the branch 1.3 can happen to cut a controller v1.3.1 release with CVE fixes (among other changes). I am eager to see results of e2e with ginkgo2 changes

longwuyuan:
And we also need a Golang bump as per dims mail

tao12345666333 (approved):
make sense

/lgtm
/approve

PR #4026, opened 2022-07-27
Promote cluster-proportional-autoscaler 1.8.6 images

MrHohn:
We are cutting a new release for cpa and need to release these images.

The main change fixing the multi-arch support.

Ref https://github.com/kubernetes-sigs/cluster-proportional-autoscaler/issues/123.
cc @tukwila @tstapler


thockin (commented):
Do we really need to keep doing the per-arch names?  I assume the no-arch version is a manifest-list?

/approve

MrHohn:
> Do we really need to keep doing the per-arch names? I assume the no-arch version is a manifest-list?

Thanks Tim! Yep that one is a manifest-list so we likely won't need to update the per-arch entry anymore.

MrHohn:
Okay updated to add only the manifest-list now :)

ialidzhikov:
@thockin @bowei can we proceed with this PR? We are looking forward to consume 1.8.6.

PR #4040, opened 2022-08-01
promote newimages with alpine-v3.16.1

longwuyuan:
- This PR promotes images built by https://github.com/kubernetes/ingress-nginx/pull/8892
- This is important because there are patches to CVEs that ship with these images
- After these images are promoted, the code in https://github.com/kubernetes/ingress-nginx has to be changes to use all these new images (sha)
- Specifically this is promoting the e2e-test-runner that for the first time ships ginkgo2 so that is a major change

/triage-acceped
/assign @tao12345666333 @strongjz


tao12345666333 (approved):
/lgtm
/approve

PR #4045, opened 2022-08-02
infra/gcp: add a secret for AWS credentials

ameukam:
Add a GCP secret that will be used to store credentials used for
synchronisation between GCS buckets and AWS S3 buckets. the current
secret has fake values.


ameukam:
cc @dims @BenTheElder 

/hold

PR #4054, opened 2022-08-05
Add SRC associates as members of private lists

tallclair:
Fixes https://github.com/kubernetes/committee-security-response/issues/157
Fixes https://github.com/kubernetes/committee-security-response/issues/156
Fixes https://github.com/kubernetes/committee-security-response/pull/152

### DIFF
--- a/groups/committee-security-response/groups.yaml
+++ b/groups/committee-security-response/groups.yaml
@@ -76,6 +76,12 @@ groups:
       - mhausler@amazon.com
       - tabitha.c.sable@gmail.com
       - timallclair@gmail.com
+    members:
+      # Associate SRC members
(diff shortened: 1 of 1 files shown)


enj:
LGTM (purposefully avoiding `/lgtm` to give others a chance to review)

tallclair:
we've already discussed "offline", so I'm not sure we need an explicit lgtm from everyone?

tabbysable:
Agreed, but they look good here / serve as a nice paper trail, so I suggest you unhold it once it feels right. (Maybe now?)

PushkarJ:
Thank you for taking care of this @tallclair and all the rest of you for the welcome!!

PR #4055, opened 2022-08-05
oci-proxy: disable CDN for the sandbox environment

ameukam:
The CDN service is not needed at the moment to serve docker client
requests.
Also:
- ignore changes done on the Cloud Run service.
- Bumping the Google provider to a recent version


ameukam:
cc @dims

PR #4080, opened 2022-08-15
Update community@ Mailing List with KCS NA Reg Team members

palnabarun:
Also, alphabetically sorts the community@ list membership.

Note: @ameukam is already a member of the list.


mrbobbytables (commented):
Thanks! 
/lgtm

and also oof theres some folk that should be removed from this list - in the reg guide, might want to include removal ass an offboarding item. I can clean those up in a follow up PR.

PR #4086, opened 2022-08-16
Grant myself read access to k8s GCP infra

upodroid:
Adding myself to a group that had read access.

/cc @dims @ameukam


spiffxp (approved):
/approve
/hold cancel

PR #4065, opened 2022-08-10
Promote 1.25-rc.0 images to experimental for testing Artifact Registry

upodroid:
Part of #3976 

Promoting the images from #4063 to the experimental subproject. The idea is to test promoting images to AR. 

This PR will promote 28 images to 26 registries.

Before we do this:
- #3956 needs to be merged
- Infra from #3968 needs to be deployed.
```
 REDACTED  MCW0CDP3YY  ~  ERROR  $  gcrane ls europe-west2-docker.pkg.dev/k8s-artifacts-prod/images/
2022/08/10 16:39:35 No matching credentials were found for "europe-west2-docker.pkg.dev"
Error: GET https://europe-west2-docker.pkg.dev/v2/token?scope=repository%3Ak8s-artifacts-prod%2Fimages%2F%3Apull&service=: DENIED: Artifact Registry API has not been used in project 388270116193 before or it is disabled. Enable it by visiting https://console.developers.google.com/apis/api/artifactregistry.googleapis.com/overview?project=388270116193 then retry. If you enabled this API recently, wait a few minutes for the action to propagate to our systems and retry.
```
- Merging #3956 should give me permissions to push images to the `gcr.io/k8s-staging/experimental`, i'll cp the images with gcrane from `gcr.io/k8s-staging-kubernetes` to experimental
- @puerco said something about bugs in the image-promoter that need to be fixed. https://kubernetes.slack.com/archives/CJH2GBF7Y/p1660077867016269?thread_ts=1660057694.134999&cid=CJH2GBF7Y

/cc @kubernetes/release-engineering @dims

PR #4090, opened 2022-08-16
k8s.gcr.io: Remove Artifact registry repositories

ameukam:
Follow-up of:
  - https://github.com/kubernetes/k8s.io/pull/4065

Remove Artifact Registry repositories as destination registries for the
image promoter. An issue with AR IAM block the image promotion
process.


ameukam:
cc @puerco

ameukam:
We should look at a different approach about to test AR promotion process.

puerco:
Should we te also remove the images from the manifest? If we run it l like this they are going to be promoted to GCR and not AR:
https://github.com/kubernetes/k8s.io/blob/6ca9ca57841206b3b7df976c84356e3d62e8b74f/k8s.gcr.io/images/k8s-staging-experimental/images.yaml#L1-L56

ameukam:
> Should we te also remove the images from the manifest? If we run it l like this they are going to be promoted to GCR and not AR:
> 
> https://github.com/kubernetes/k8s.io/blob/6ca9ca57841206b3b7df976c84356e3d62e8b74f/k8s.gcr.io/images/k8s-staging-experimental/images.yaml#L1-L56

@puerco Done.

puerco:
Awesome, let's kick it :rocket: 
/lgtm

PR #4103, opened 2022-08-18
promoting baseimage with alpine-v3.16.2

longwuyuan:
- Because of zlib CVE being critical, built new base image on alpine v3.16.2
- This PR promotes that above mentioned baseimage

/triage accepted
/assign @tao12345666333 @strongjz


thockin:
Was it always on Alpine?  We used to not want anyone to use Alpine, and I
don't recall seeing that guidance changed...

longwuyuan:
Hi @thockin , unsure of the origin of not wanting to be on alpine. If it was intended for core components of k8s, it may be likely (i am unaware). 

This PR though is about the ingress-nginx-controller's baseimage and that has been on alpine always yes. WIll look forward to providing any other info that would be needed.
@rikatz cc

longwuyuan:
This PR related to https://github.com/kubernetes/ingress-nginx/issues/8933  and https://github.com/kubernetes/ingress-nginx/pull/8934

rikatz:
@thockin it was always alpine. 

We have a plan to migrate to debian or even distroless, but due to the amount of dependencies and compilations it is not an easy movement right now. 

After splitting control plane and dataplane, the cp will be distroless for sure and probably dp and tester images will be distroless as well, but will take some more time

strongjz:
Did we scan this base before merge to make sure it got the patch?

longwuyuan:
@strongjz see below
image

PR #4112, opened 2022-08-19
Remove Slackin, as it is being bypassed

jberkus:
Remove the Slackin app from our cloud, since it's broken and is now being completely bypassed.  No need to waste the CPU cycles.

attn: @ameukam @dims


BenTheElder:
manually deleted these from the cluster as the auto-deploy just `kubectl apply`, also dropped from ingress in #4159

PR #4115, opened 2022-08-20
groups: parallelize reconciler

spiffxp:
This adds a `--workers` flag (default: 5) to the groups reconciler. Each group is reconciled in parallel before deleting groups (if necessary).

This took runtime down from from ~5m to ~45s in my environment


ameukam:
/lgtm
/hold

MadhavJivrajani (approved):
/lgtm

Issue #3918, opened 2022-06-29
Configure Terraform presubmits and postsubmit jobs


upodroid:
We have a number of projects managed by terraform at https://github.com/kubernetes/k8s.io/tree/main/infra/gcp/terraform.

However, these currently projects require manual deployment by sig-k8s-infra leads and others which is blocking rapid iteration of the GCP infra.

We need to configure some automation to deploy these changes safely.

Google Cloud Changes:
1. Create a new project that holds a privileged service account. Something like https://github.com/knative/test-infra/tree/main/infra/gcp#bootstrapping-terraform---one-time-setup
2. Grant this ServiceAccount some roles on the organization. https://github.com/knative/test-infra/blob/main/infra/gcp/iam.tf
3. Create a k8s service account
4. Complete the Workload Identity Configuration

Prow Changes:
1. We will need a postsubmit job that runs when changes in infra/gcp/* are merged
2. We will also need a presubmit that runs terraform plan only and prints that output to the PR. This will be running in the trusted cluster and that isn't allowed by default. Need to work something out.

Lets talk about it at the next sig-k8s-infra meeting.

/cc @ameukam

ameukam:
Why this issue is scoped to GCP only ? we have the same problem with AWS. I would be interesting to introduce the same pattern to the existing cloud providers ?

upodroid:
AWS can be enabled by completing #3807. Will add this to the body of the issue.

PR #4127, opened 2022-08-23
Enable running terraform on GCB jobs

upodroid:
Fixes: #3918 
Required for https://github.com/kubernetes-sigs/oci-proxy/pull/95

This PR will allow us to run terraform jobs with privileged service accounts. 

I will raise another PR with the prowjobs in k/test-infra today.

```
in k/test-infra

For oci-proxy in particular:
- presubmit job that runs plan on sandbox/dev project when the terraform folder is changed
- postsubmit job that runs apply on sandbox/dev when the terraform folder is changed
- presubmit job that runs plan on prod instance when prod folder is changed
- postsubmit job that runs apply on prod instance when prod folder is changed
- create an autobumper for sandbox project, prod deployment must be bumped by a human manually.
```

/cc @dims @ameukam @BenTheElder

### DIFF
--- a/infra/gcp/bash/ensure-terraform.sh
+++ b/infra/gcp/bash/ensure-terraform.sh
@@ -38,7 +38,7 @@ fi
 PROJECT=$(k8s_infra_project "prow" "k8s-infra-terraform")
 
 color 6 "Ensuring project exists: ${PROJECT}"
-# ensure_project "${PROJECT}"
+ensure_project "${PROJECT}"
 
 PROJECT_SERVICES=(
     iam.googleapis.com
--- a/infra/gcp/bash/ensure-terraform.sh
+++ b/infra/gcp/bash/ensure-terraform.sh
@@ -1,0 +1,84 @@
+#!/usr/bin/env bash
+
+# Copyright 2022 The Kubernetes Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# This script ensures terraform service accounts for prow jobs and GCB are created.
+# DOCS: https://cloud.google.com/build/docs/securing-builds/configure-user-specified-service-accounts#cross-project_set_up
+
+set -o errexit
+set -o nounset
+set -o pipefail
+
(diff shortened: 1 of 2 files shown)


infra/gcp/bash/ensure-terraform.sh
```diff
@@ -0,0 +1,84 @@
...
+    "oci-proxy-dev" \
+    "oci proxy dev Terraform Account"
+
+color 6 "Creating oci-proxy prod Service Account"
+ensure_service_account \
+    "${PROJECT}" \
+    "oci-proxy-prod" \
+    "oci proxy prod Terraform Account"
+
+oci_proxy_dev_account="$(svc_acct_email "${project}" "${STAGING_SIGNER_SVCACCT}")"
```
ameukam, line 65: what's the purpose of `STAGING_SIGNER_SVCACCT` ?

infra/gcp/bash/ensure-terraform.sh
```diff
@@ -0,0 +1,84 @@
...
+color 6 "Ensuring GCB service agent of k8s-staging-infra-tools can impersonate ${oci_proxy_dev_account} "
+ensure_serviceaccount_role_binding \
+    "${oci_proxy_dev_account}" \
+    "serviceAccount:service-1017132094926@gcp-sa-cloudbuild.iam.gserviceaccount.com" \
+    "roles/iam.serviceAccountTokenCreator"
+
+oci_proxy_prod_account="$(svc_acct_email "${project}" "${STAGING_SIGNER_SVCACCT}")"
+color 6 "Ensuring GCB service agent of k8s-staging-infra-tools can impersonate ${oci_proxy_prod_account} "
```
ameukam, line 73: We want to deploy using a promoted image of `archeio`. why do we need to use the staging project ?

infra/gcp/bash/ensure-terraform.sh
upodroid: that is a typo should be oci-proxy-dev and oci-proxy-prod

infra/gcp/bash/ensure-terraform.sh
upodroid: The GCB job that will run terraform will be in the `k8s-staging-infra-tools` project. Since it runs in that project, it needs to impersonate the designated service accounts in the `k8s-infra-terraform` project.

ameukam: Why not simply use a postsubmit to do that ? we could simply have a service account associated to do that.

upodroid: That is the plan, but GCB is running a different account.

https://github.com/kubernetes-sigs/oci-proxy/blob/main/cloudbuild.yaml is running as `1017132094926@cloudbuild.gserviceaccount.com` which isn't what we want as we need 2 different service accounts used for separate jobs. Unlike most jobs, we need to run terraform plan in presubmits which is a privileged operation, hence the need for multiple service accounts.

We could grant Owner to `1017132094926@cloudbuild.gserviceaccount.com` on both projects but I don't like that idea.

Issue #3411, opened 2022-02-16
[Umbrella issue] Migrate from k8s.gcr.io to registry.k8s.io


ameukam:
Part of:
  - https://github.com/kubernetes/k8s.io/issues/1834

Design doc: https://docs.google.com/document/d/1yNQ7DaDE5LbDJf9ku82YtlKZK0tcg5Wpk9L72-x2S2k. 

This issue focus on the implementation of `oci-proxy` as a proxy/redirect endpoint for `k8s.gcr.io`:

# Sandbox environment 
- [x] Setup a sandox environment for development: `registry-sandbox.k8s.io`.
     - https://github.com/kubernetes/k8s.io/issues/3317
     - https://github.com/kubernetes-sigs/oci-proxy/issues/16
- [x] Migrate test workloads to registry-sandbox.k8s.io:
    - https://github.com/kubernetes/test-infra/pull/25466
    - https://github.com/kubernetes/test-infra/pull/25458
    - https://github.com/kubernetes/test-infra/pull/25608
    - https://github.com/kubernetes/test-infra/pull/25607

# Production environment:

(taken from https://github.com/kubernetes/k8s.io/issues/1834#issuecomment-949916308)
- [ ] Continuous deployment
- [X] Monitoring: Cloud Run provides native metrics based on GCP Monitoring: https://cloud.google.com/run/docs/monitoring
- [x] Logging: built in archeio and using Cloud Logging
- [x] PII Management: Using Linux Foundation privacy policy: https://regisry.k8s.io/privacy
- [ ] Recovery
- [x] Scalability: Cloud Run auto-scaling
- [ ] Incident Management

/milestone v1.24
/priority important-soon

cc @BenTheElder @dims @eddiezane

matthyx:
@ameukam can we ditch already the endpoint config here?
https://github.com/BenTheElder/kind/blob/main/images/base/files/etc/containerd/config.toml#L46-L47

ameukam:
> @ameukam can we ditch already the endpoint config here? [BenTheElder/kind@`main`/images/base/files/etc/containerd/config.toml#L46-L47](https://github.com/BenTheElder/kind/blob/main/images/base/files/etc/containerd/config.toml?rgh-link-date=2022-10-03T12%3A55%3A49Z#L46-L47)

@matthyx We are still in the early days `registry.k8s.io` in production. I would not advise to remove this endpoint at the moment until we are sure we stop use `k8s.gcr.io`.

[...]

PR #4145, opened 2022-08-25
oci-proxy: set alerts for sandbox and prod.

ameukam:
Add a terraform module to ensure alerts are created for:
- SSL certificate expiration: an alert is sent if the certificate for
  given domain is not renewed less than 15 days before the expiration
  date.
- uptime check: given a hostname, ensure the host is alive on port 443
  for the root path (/). A check is done every minute.

Ensure alerts are created using the terraform module.


ameukam:
cc @dims
Already Deployed. I can't fully tested using Terraform. The tests can only be done with the UI.

ameukam:
Ref: https://github.com/kubernetes/k8s.io/issues/3411

PR #4105, opened 2022-08-18
Switch slack.k8s.io to the community inviter link

dims:
related to https://github.com/kubernetes/k8s.io/issues/4099

Signed-off-by: Davanum Srinivas <davanum@gmail.com>

PR #4156, opened 2022-08-30
move slack tools to their own subdomain so they don't conflict with the inviter redirect

BenTheElder:
Should be easier to manage going forward, see:
- https://github.com/kubernetes/community/issues/6841
- https://github.com/kubernetes/k8s.io/pull/4105

I spoke to katharine about reconfiguring the apps to point to this seperate domain, we can do that once this is deployed.

TODO (follow-up):
- actually delete deployed resources from https://github.com/kubernetes/k8s.io/pull/4112
- remove slackin from ingress config
- consider not tagging all apps slackin2 since that's not really accurate 🤷‍♂️

### DIFF
--- a/apps/slack-infra/resources/certificate.yaml
+++ b/apps/slack-infra/resources/certificate.yaml
@@ -4,5 +4,5 @@ metadata:
   name: slack-infra
 spec:
   domains:
-  - slack.k8s.io
-  - slack.kubernetes.io
+  - slack-infra.k8s.io
+  - slack-infra.kubernetes.io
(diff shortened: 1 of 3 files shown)


ameukam:
the certificate controller might not be able to create the certificate after the PR is merged. Safe approach is to deploy the DNS before the certificate.

/lgtm
/hold

Remove the hold when ready.

BenTheElder:
I'll deploy DNS then let this merge, working on that now

BenTheElder:
taking a bit, had to refresh myself on this first, dns push is running but pretty slow

BenTheElder:
DNS is live
/hold cancel

PR #4170, opened 2022-09-01
Add s3-sync role for syncing registry.k8s.io objects

BobyMCbobs:
Adds a service account to the trusted cluster, for running s3 syncing between GCS and S3 for registry.k8s.io objects.
Related: 
- https://github.com/cncf-infra/aws-infra/pull/15
- https://github.com/kubernetes/k8s.io/issues/4094

PR #4185, opened 2022-09-06
Add S3 sync GCP service accounts

upodroid:
Part of #4170

Can someone assign the Storage Object Creator role on the prow bucket for this account?

### DIFF
--- a/infra/gcp/terraform/k8s-infra-prow-build-trusted/serviceaccounts.tf
+++ b/infra/gcp/terraform/k8s-infra-prow-build-trusted/serviceaccounts.tf
@@ -57,6 +57,9 @@ locals {
       project_roles     = ["roles/secretmanager.secretAccessor"]
       cluster_namespace = "kubernetes-external-secrets"
     }
+    s3-sync = {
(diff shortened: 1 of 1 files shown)


listx:
> Part of #4170
> 
> Can someone assign the Storage Object Creator role on the prow bucket for this account?

I don't have enough context to understand what needs to be done here, not to mention whether I even have permissions to do so.

ameukam:
> Part of #4170
> 
> Can someone assign the Storage Object Creator role on the prow bucket for this account?

which prow bucket ? AFAIK we don't need specific permissions to push the logs for all the prowjobs. what's particular about this service account ?

/uncc @listx @cjwagner

upodroid:
Actually, ignore my comment and we don't need this PR.

https://prow.k8s.io/prowjob?prowjob=12723dcb-2e9e-11ed-8857-d61b7b5f68bf I assumed k8s' prow was using Workload Identity for writing to `gs://kubernetes-jenkins` but it seems we are still using service account keys (`spec.decoration_configs.gcs_credentials_secret` is set)

/close

PR #4196, opened 2022-09-08
k8s-infra: bump components

ameukam:
Add aws-cli to k8s-infra image
Bump other components

Signed-off-by: Arnaud Meukam <ameukam@gmail.com>

PR #4198, opened 2022-09-08
Update registry.k8s.io s3 bucket sync scripts

BobyMCbobs:
Changes include:
- add `regions` subcommand to registry.k8s.io/hack/initial-copy-between-s3.sh
- use `env_auth` with rclone configs, if `/var/run/secrets/aws-iam-token/serviceaccount/token` is found in the environment

Conditionally allowing `env_auth` means that the scripts will continue to work as normal outside the ProwJobs for now.

Related: https://github.com/kubernetes/test-infra/pull/27415, https://github.com/kubernetes/k8s.io/pull/4196

### DIFF
--- a/registry.k8s.io/hack/initial-copy-between-s3.sh
+++ b/registry.k8s.io/hack/initial-copy-between-s3.sh
@@ -22,20 +22,26 @@
 #   sync all:       ./initial-copy-between-s3.sh
 #   specify region: ./initial-copy-between-s3.sh <REGION>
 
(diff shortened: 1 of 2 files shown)


BobyMCbobs:
cc @upodroid @ameukam

ameukam:
/approve
/hold 

Leave the lgtm to @upodroid
In case other changes needs to be introduced.

upodroid (approved):
/lgtm

PR #4203, opened 2022-09-12
Cleanup 1.22 release team shadows from sig-testing groups.yaml

reylejano:
The current release cycle is 1.26, this PR removes 1.22 release team shadows from sig-testing/groups.yaml


BenTheElder (approved):
/lgtm
/approve

Issue #3992, opened 2022-07-20
registry.k8s.io S3 bucket auditing


BobyMCbobs:
Currently there's GCP auditing of events and resource usage in https://github.com/kubernetes/k8s.io/tree/main/audit.
I'd would be good to have auditing of the registry.k8s.io S3 buckets, writing to a repo when new objects are uploaded.

The auditing could be as simple as
```
REGIONS=(
    ap-northeast-1
    ap-south-1
    ap-southeast-1

    eu-central-1
    eu-west-1

    us-east-1
    us-east-2
    us-west-1
    us-west-2
)

for REGION in "${REGIONS[@]}"; do
  aws s3api list-objects --bucket "prod-registry-k8s-io-$REGION" --no-sign-request --output json > "bucket-$REGION.json"
done
```
in a CI job, making a PR into the k8s.io repo.

related: https://github.com/kubernetes/k8s.io/issues/1834, https://github.com/kubernetes/k8s.io/issues/3623

BobyMCbobs:
I'm playing around with this idea over here

https://github.com/ii/registry-k8s-io-s3-object-audit

BobyMCbobs:
Any thoughts around including
https://github.com/ii/registry-k8s-io-s3-object-audit
into the kubernetes/k8s.io repo?
Might be pretty easy to turn it into a ProwJob.

cc @ameukam @upodroid

upodroid:
+1 to wrapping that github action as a prowjob and committing the diff to k/k8s.io

BobyMCbobs:
Related: https://github.com/kubernetes/k8s.io/pull/4223

spiffxp:
With the full caveat that I have been away while most of the discussion and work on registry.k8s.io has happened, I don't understand why this is necessary.

Our audit scripts today do not list the contents of GCS buckets, GCR repos, or GAR repos.  I'm not exactly clear why we feel compelled to do so for S3.

asim-reza:
There needs to be another way than just to push the audit into the main repo if this is to be done .. because the audit files will be something like 90% of the repo

justinsb:
So I think what we're trying to audit is that the mirrors accurately reflect the canonical location.  So we want to know if somehow the mirrors have (1) extra files or (2) corrupted files.  We also want to know if the mirrors are missing files, but this could be a natural transitory state during initial publishing of new images.

[...]

PR #4223, opened 2022-09-16
Add registry.k8s.io generate audit script

BobyMCbobs:
Adds a script to generate a blob list of the objects in the registry.k8s.io S3 buckets.

Later, it might be good to group it by aws/s3 when there are more providers and there is more auditing.

This will be consumed by a ProwJob.

### DIFF
--- a/registry.k8s.io/hack/generate-audit.sh
+++ b/registry.k8s.io/hack/generate-audit.sh
@@ -1,0 +1,48 @@
+#!/bin/bash
+
+# Copyright 2022 The Kubernetes Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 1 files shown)


BobyMCbobs:
Fixes https://github.com/kubernetes/k8s.io/issues/3992

asim-reza (commented):
Why are the generated files so big? and why are they needed in the repo?

dims:
@asim-reza these are "audit" which means, we track who did what and when and save it for historical record. say 6 months from now, we wanted to check on something that happened today, we will be able to pull that up. this is the way we assure the community as all actions are open for inspection. also helps us catch if anyone made a mistake.

spiffxp (changes requested):
```
$ find . -type f | grep -v .git | xargs wc -l | tail -n1
  316192 total
``` 
I think I agree with @asim-reza.  A naive line count in this repo is ~300K, while this PR adds ~11M... this doesn't feel right.

With the full caveat that I have been away while most of the discussion and work on registry.k8s.io has happened, I don't understand why this approach is necessary.

Our audit scripts today do not list the contents of GCS buckets, GCR repos, or GAR repos.  I'm not exactly clear why we feel compelled to do so for S3.

dims:
@spiffxp i am with you as well, i hadn't peeked at the files ... yep, it's too big.

spiffxp:
/close
Will take further discussion over to https://github.com/kubernetes/k8s.io/issues/3992

PR #4228, opened 2022-09-16
promoting images with go v1.19.1

longwuyuan:
- Built new images because of https://github.com/kubernetes/ingress-nginx/pull/9057
- Lot of related info is here as well https://github.com/kubernetes/ingress-nginx/issues/8932
- This PR promotes the new images that now contain go v1.19.1
- This PR makes a significant promotion in the test-runner image because that sha has to be updated in the ingress-nginx project's main, for the ingres-nginx-controller's binary to get compiled with go v1.19.1. Reason being the build system uses a docker environ to compile the binary and the docker-environ is built using the test-runner image being promoted here

/tirage accepted

/assign @tao12345666333 @strongjz @rikatz

### DIFF
--- a/apps/prow/cluster/crier_deployment.yaml
+++ b/apps/prow/cluster/crier_deployment.yaml
@@ -33,7 +33,7 @@ spec:
       terminationGracePeriodSeconds: 30
       containers:
       - name: crier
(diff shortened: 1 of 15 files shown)


longwuyuan:
I messed up in rebasing so will close this and create new one

PR #4240, opened 2022-09-19
Update NP-api dns name

astoycos:
Update DNS name here since the original
one was too long for netlify.

see https://github.com/kubernetes/org/issues/3643 for more information

fixes https://github.com/kubernetes/org/issues/3643


rikatz:
/lgtm
Thanks

PR #4265, opened 2022-09-26
Fixed badly formatted link

userbradley:
There is a file under `infra/gcp` that had a badly formatted link


userbradley:
Signed the Document just now. All good to go

PR #4318, opened 2022-10-06
Add support for binary artifact promotion of the cloud-provider-aws project

olemarkus:
The AWS CCM project comes with a kubelet credentials provider binary we'd like to use the binary promotion for.
I believe this PR contains the necessary files to create the staging bucket and add support for binary promotion.


ameukam:
I think a manifest file is missing in the `manifests` path. can you try to add a `files.yaml` with the line `#No files yet` ?

olemarkus:
Doesn't look like it got more happy this time either. I guess it requires an actual file entry, but that is quite a hen/egg issue.

dims:
probably need to remove the empty `artifacts/manifests/k8s-staging-cloud-provider-aws/files.yaml` ?
```
level=fatal msg="run `kpromo run files`: error validating merged manifest: at least one file must be specified"
```

olemarkus:
@dims didn't seem to help, unfortunately.

PR #4331, opened 2022-10-10
groups: update sig-testing chairs

spiffxp:
Related
- Part of: https://github.com/kubernetes/community/issues/6913


spiffxp:
/cc @BenTheElder
/hold
I noticed TL's aren't here, I think this was to reduce access to potentially sensitive credentials.  Do we want to revisit that or keep as-is?

BenTheElder (commented):
/lgtm
/approve

PR #4337, opened 2022-10-11
switch registry.k8s.io to serve from regional AR instances

BenTheElder:
xref staging: https://github.com/kubernetes/registry.k8s.io/blob/d625292f6285c574365d276017bcc0bfcf4998e7/hack/make-rules/deploy.sh#L58

/hold
need to double check that each of these exists.

PR #4338, opened 2022-10-11
update registry.k8s.io to current information

BenTheElder:
- Delete outdated information ... NOTE: this is in git, so it's not _gone_ we just don't have misleading / outdated documents _in the current source state_, they're still reference-able in the git history.
  - We're no longer stress testing this way
  - We are not using the asns for now, and if we were they should probably be in the source repo
  - We are not using any of the "explorations", current approach can be found in registry.k8s.io, these can be referenced in git history (as with prior iterations of the rest of the source code)
- Add current information
  - registry.k8s.io is live, not under construction
  - design docs can be found in the main source repo
  - notes about GCP infra
  - notes about syncing not being manual

The information for what's actually going on with registry.k8s.io is scattered about and often outdated, consolidating between:
- this repo for deployment configuration
- github.com/kubernetes/registry.k8s.io for design, ops, sources ...

Next Up: replacing the wiki page with docs in registry.k8s.io https://github.com/kubernetes/k8s.io/wiki/New-Registry-url-for-Kubernetes-(registry.k8s.io)


ameukam:
cc @spiffxp 

/lgtm
/approve
/hold

We should maybe have all those infos as archives in git.k8s.io/registry.k8s.io ? (No strong opionion, just suggesting)

BenTheElder:
> We should maybe have all those infos as archives in git.k8s.io/registry.k8s.io ? (No strong opionion, just suggesting)

Eh, git is inherently an archive.
We don't specially archive past versions of the registry.k8s.io sources either, git handles this for us cleanly.

hh:
@BenTheElder @ameukam @spiffxp : We are still using the ASN data to calculate the spend:

https://github.com/kubernetes/k8s.io/blob/main/images/public-log-asn-matcher/pg-init.d/lib-04-load-into-a-bigquery-dataset.sh

I'd like some time to talk about this before a merge.

BenTheElder:
IIRC @ameukam asked to stop running that to save on BQ costs.
As-is, this is not part of registry.k8s.io, so if we restore these files, we should organize them better.

We will not be serving from GCR/GCS shortly anyhow: https://github.com/kubernetes/k8s.io/pull/4337

BenTheElder:
As part of mitigating the multi-regional costs and preparing for the future of image host, we are moving to regional AR, which is not transparently backed by GCS, and will not be matched against this data. We are prioritizing this transition to production given the multi-regional billing cost changes are live as of this month.

Last week we rolled it to staging. This week to production. The current movement has been discussed async in #sig-k8s-infra, and long planned in sig-k8s-infra both in the slack and in the meetings.

We have a good idea where traffic has been coming from, based on past runs.

PR #4339, opened 2022-10-11
explain the relationship between k8s.gcr.io folder and registry.k8s.io

BenTheElder:
see also: https://kubernetes.slack.com/archives/CCK68P2Q2/p1665503606889629?thread_ts=1665503352.287139&cid=CCK68P2Q2


BenTheElder:
cc @dims @ameukam

PR #4298, opened 2022-10-03
Fix typo in `etcdadm` promoter manifests

upodroid:
Fixing a typo from #4291

/cc @ameukam 

I haven't seen any other typos.

PR #4366, opened 2022-10-17
Fix typo in Azure Cluster API Repository

upodroid:
Cleanup script is available in https://github.com/kubernetes/k8s.io/pull/4298

We probably need to write a simple go app that validate the manifests as we will be adding many AR repos in the coming months and years.

/cc @puerco @BenTheElder


BenTheElder (approved):
/lgtm
/approve

BenTheElder:
we should cleanup the old names cc @ameukam

PR #4371, opened 2022-10-18
Promote kOps 1.24.4 binary artifacts

hakman:
/cc @olemarkus


artifacts/manifests/k8s-staging-kops/1.24.4.yaml
```diff
@@ -0,0 +1,70 @@
+{}
```
olemarkus, line 1: Should this one be removed?

olemarkus (approved):

PR #4381, opened 2022-10-20
fine-tune registry.k8s.io

BenTheElder:
based on metrics against some less-trivial real live traffic

We peaked at ~500qps and ~50% CPU in us-east4 with today's additional k8s.gcr.io traffic.

Slightly reducing container_concurrency (we don't actually want to hit 100% CPU ...) and tuning down request timeout.

We're still configured to scale up to 8000 qps/region which should be more than plenty.


BenTheElder:
applying this
will merge shortly

BenTheElder:
> Apply complete! Resources: 0 added, 20 changed, 0 destroyed.

/hold cancel

PR #4382, opened 2022-10-20
Add mailing list for CEL related discussions

jpbetz:
We already have a mailing list (https://groups.google.com/a/google.com/g/wg-sig-api-machinery-cel-dev-external) but it is not external or community managed, so we'd like to move to this list.

cc @kubernetes/sig-api-machinery-leads

### DIFF
--- a/groups/restrictions.yaml
+++ b/groups/restrictions.yaml
@@ -21,6 +21,7 @@ restrictions:
     allowedGroups:
       - "^k8s-infra-staging-etcd@kubernetes.io$"
       - "^k8s-infra-staging-storage-migrator@kubernetes.io$"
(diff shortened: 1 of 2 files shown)


ameukam:
@jpbetz please ensure this group is listed in https://github.com/kubernetes/k8s.io/blob/main/groups/restrictions.yaml#L21-L23

jpbetz:
> Joe Betz please ensure this group is listed in https://github.com/kubernetes/k8s.io/blob/main/groups/restrictions.yaml#L21-L23

Added. Thanks!

nikhita (approved):
/lgtm
/approve

PR #4419, opened 2022-11-02
Deploy AR mirrors to remaining US AR regions

upodroid:
The image promoter has been optimised and we should be good with adding additional registries.

I'm also going to do the same for the S3 buckets.

/cc @ameukam @BenTheElder


ameukam:
I don't think we should add more AR/Cloud Run regions. AR costed us more than $10k for Oct and in some regions we barely have 5 RQS.

ameukam:
I think we can convert the entire US multi-region. @upodroid can you only add `us-west3` and `us-west4`?

upodroid:
done

ameukam:
This is done.

PR #4431, opened 2022-11-08
Promote Addon Resizer 1.8.16

kgolab:
Checksums from gcr.io/k8s-staging-autoscaling (obviously).


k8s.gcr.io/images/k8s-staging-autoscaling/images.yaml
```diff
@@ -225,6 +226,7 @@
     "sha256:85c57faa812598eac877f97ef5a3eef7f62897c5aed62d774b36937ca31eb285": ["1.8.13"]
     "sha256:43f129b81d28f0fdd54de6d8e7eacd5728030782e03db16087fc241ad747d3d6": ["1.8.14"]
     "sha256:21417cb15452a59f601ad851a2fed4acefde9c438aeb06f6db0b738296800368": ["1.8.15"]
+    "sha256:0a23d8ee36d7795eba1e21c05c11f7b892528bd3e1ffee7529329eeb8f6877db": ["1.8.16"]
```
jbartosik, line 229: This one seems incorrect:

```
$ gcloud container images describe gcr.io/k8s-staging-autoscaling/addon-resizer-amd64:1.8.16
WARNING: Successfully resolved tag to sha256, but it is recommended to use sha256 directly.
image_summary:
  digest: sha256:e9abc9f1fb796246d4048dbf289faea20a8321c70d42fad4230f2cec4f6817de
```

jbartosik (changes requested):
I stopped after first mismatch I found. Please check other images too

k8s.gcr.io/images/k8s-staging-autoscaling/images.yaml
kgolab: Ouch, thanks for catching it!

Looking at the change it seems like copy from gcr output didn't catch and I pasted the previous value (x86). I will script this step to avoid similar issues in the future.

Checked all other entries and they look fine to me.

jbartosik (approved):
/lgtm
/approve

Issue #4402, opened 2022-10-31
shut down k8s.gcr.io bigquery analysis


BenTheElder:
We have all the existing historical data, **this is still costing ~$7k/mo or $6,920.33 / $314,396.53 * 100 = 2.2%** of our spend. (see: `k8s-infra-ii-sandbox`, `BigQuery` in the billing report at #7).

Also, the historical data is more valuable, because it's 100% traffic there vs now some traffic is at registry.k8s.io

We should immediately disable these jobs.

~~IIRC we already requested this earlier this year multiple times including at the SIG meeting, but it still appears to be spending $$$, so let's turn these off now?~~

EDIT: per https://github.com/kubernetes/k8s.io/issues/4402#issuecomment-1297814983 this is storage costs not processing? even still, this is a quite noticable charge, let's grab high level info and then purge this cost.

<img width="1227" alt="Screen Shot 2022-10-31 at 4 15 37 PM" src="https://user-images.githubusercontent.com/917931/199128058-b42e251d-0110-4374-b50a-631db6e585e2.png">

https://datastudio.google.com/c/u/0/reporting/14UWSuqD5ef9E4LnsCD9uJWTPv8MHOA3e/page/8fWn

ameukam:
The cost is related to active storage not processing or querying but I agree with the idea to reduce the resources cost.

I'll clean up the project in the incoming weeks.

@hh @Riaankl this data is transferable to another GCP (owned by ii or CNCF). Please take a look at other options.

BenTheElder:
Can we grab the high level results if we haven't already and then remove this dataset? This is # 7 on our spend costing about as much as analyzing CI logs for error clusters.

ameukam:
Cleanup was done. One dataset is remaining.

```bash
ameukam@cloudshell:~ (k8s-infra-ii-sandbox)$ gcloud alpha bq datasets list --project k8s-infra-ii-sandbox
ID: k8s-infra-ii-sandbox:etl_script_generated_set
LOCATION: US
```

We should see cost related to BQ go down over the next days.

Riaankl:
Yes, I just deleted all the historical datasets, keeping only the latest set that feed the data studio report.
If new data is needed we can generate it.

BenTheElder:
thank you!!

hh:
This is done, one report data set remains.
/close

BenTheElder:
This is now costing very little, at $6.64 out of $10,100.89 yesterday.
Thanks all!

PR #4448, opened 2022-11-10
Revert deleted ASN data

BobyMCbobs:
The data here is needed for running the [public-log-asn-matcher](https://github.com/kubernetes/k8s.io/tree/main/images/public-log-asn-matcher), it was removed sometime ago and will allow it to work again.

The report generated out of the data from _public-log-asn-matcher_ was requested by @BenTheElder.

cc @Riaankl


Riaankl:
/assign @BenTheElder 
We need to merge these file back in as we need them to run the data query.

BenTheElder:
1) quibble but, this report is for k8s.gcr.io, not registry.k8s.io?
2) Where and why are we still actively processing this? Are we unable to produce a static snapshot of k8s.gcr.io and keep the stats on hand without continuing to process this? (xref #4402, we're really cost constrained right now)

BenTheElder:
I'm actually hoping to see the historical reports, the results from k8s.gcr.io's current data hopefully become increasingly irrelevant as we get people to switch, but it's good to understand where we came from and which other providers to target for local copies.

hh:
Agreed. We need this back to generate a static report. We will run the job manually, but it references the ASN files restored in this PR. The historical data has been purged. We gotta regenerate it to view the report again.

ameukam:
Data query and storage should not be done on the `kubernetes.io` org for the moment.

Since `registry.k8s.io` use Artifact Registry, we lose access to the audit logs for provide relevant information. The data collected will be partial. 
I would suggest to do this on a different GCP org.

BobyMCbobs:
> 1. quibble but, this report is for k8s.gcr.io, not registry.k8s.io?

@BenTheElder, the report uses $BUCKET_usage data from buckets found in kubernetes org projects.
If there are buckets which the usage data is being written to there, it will pull from that that.

>     2. Where and why are we still actively processing this? 

This is a manually run job. By default, it produces a BigQuery dataset in _k8s-infra-ii-sandbox_ as _etl_script_generated_set_$DATE_.

> Are we unable to produce a static snapshot of k8s.gcr.io and keep the stats on hand without continuing to process this? (xref [shut down k8s.gcr.io bigquery analysis #4402](https://github.com/kubernetes/k8s.io/issues/4402), we're really cost constrained right now)

How do you mean?
The process generates a database containing the required information for the report. It would be good to keep at least the latest around for that reason.

hh:
> I would suggest to do this on a different GCP org.

Are you asking the CNCF provide another GCP org to process this dataset and resulting queries to avoid touching the remaining credits for this year?

I don't think generating a single data set, or the report will have a huge impact on cost.

ameukam:
> > I would suggest to do this on a different GCP org.
> 
> Are you asking the CNCF provide another GCP org to process this dataset and resulting queries to avoid touching the remaining credits for this year?

Pause the generation of this data until we have a better control of the GCP budget is what I'm asking. We can later revisit how we optimize that BQ storage. 

> I don't think generating a single data set, or the report will have a huge impact on cost.

PR #4460, opened 2022-11-11
infra/aws: create S3 buckets for artifacts.k8s.io

justinsb:
Copying registry.k8s.io configuration, but removed most of the mirror
buckets for now.


justinsb:
/hold for discussion

cc @ameukam I think we can create some AWS buckets fairly easily, and then we don't need a ton of code to map to github names.

justinsb:
https://github.com/cncf-infra/aws-infra/pull/21 should be how we can create the needed AWS IAM roles etc

ameukam:
> /hold for discussion
> 
> cc @ameukam I think we can create some AWS buckets fairly easily, and then we don't need a ton of code to map to github names.

A more simple approach is to create one S3 bucket shielded by a AWS Cloudfront distribution. Less maintenance and less complexity in the overall. Unlike `registry.k8s.io` we have room to make things very simple.

justinsb:
> A more simple approach is to create one S3 bucket shielded by a AWS Cloudfront distribution. Less maintenance and less complexity in the overall. Unlike registry.k8s.io we have room to make things very simple.

Yes, and I did cut down the number of buckets.  I think the question is whether we want to maintain two approaches here (particularly as these are both HTTP serving of largish static blobs).  Or is the plan to move the S3 buckets to CloudFront also?

ameukam:
> > A more simple approach is to create one S3 bucket shielded by a AWS Cloudfront distribution. Less maintenance and less complexity in the overall. Unlike registry.k8s.io we have room to make things very simple.
> 
> Yes, and I did cut down the number of buckets. I think the question is whether we want to maintain two approaches here - one for registry.k8s.io, one for artifacts.k8s.io (particularly as these are both HTTP serving of largish static blobs). Or is the plan to move the S3 buckets for registry.k8s.io to CloudFront also?

I think mid-term we should maintain both approaches. That give us the opportunity to observe impact on the budgets we are getting in 2023.

justinsb:
> I think mid-term we should maintain both approaches. 

OK, well if we create the S3 buckets then I can figure out how to put CloudFront in front of them.  Doesn't seem too bad: https://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/DownloadDistS3AndCustomOrigins.html#using-s3-as-origin

One thing that wasn't clear (to me) was whether CloudFront discounts traffic going to AWS.  So yes, I agree that we should compare S3 vs CloudFront for our traffic.

ameukam:
@justinsb Also note that we still don't receive new credits from AWS (possibly early 2023), so going in prod may take a while (or not). In the meantime we have access to an account where we can experiment S3+Cloudfront.

infra/aws/terraform/artifacts.k8s.io/main.tf
```diff
@@ -0,0 +1,63 @@
...
+module "us-east-2" {
+  source = "./s3"
+
+  providers = {
+    aws = aws.us-east-2
+  }
+
+  prefix = var.prefix
+
+  s3_replication_iam_role_arn = "arn:aws:iam::513428760722:role/artifacts.k8s.io_s3writer"
```
ameukam, line 52: @justinsb I don't have access to this account. The account is currently under AWS org. I would suggest to create only one bucket for the staging environment in the AWS account we used for `registry.k8s.io` until we have better clarity on AWS credits donation.

infra/aws/terraform/artifacts.k8s.io/providers.tf
```diff
@@ -0,0 +1,62 @@
...
+*/
+
+terraform {
+  backend "s3" {
+    bucket = "artifacts-k8s-io-tfstate"
+    key    = "terraform.tfstate"
+    region = "us-east-2"
+  }
+
+  required_version = ">= 1.0.0"
```
ameukam, line 24:
```suggestion
  required_version = ">= 1.2.0"
```

infra/aws/terraform/artifacts.k8s.io/main.tf
```diff
@@ -0,0 +1,63 @@
...
+    error_message = "The string must end with a hyphen or be empty."
+  }
+}
+
+
+module "us-west-2" {
+  source = "./s3"
+
+  providers = {
+    aws = aws.us-west-2
```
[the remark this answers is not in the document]
justinsb: Done - us-west-2 -> eu-west-1

infra/aws/terraform/artifacts.k8s.io/main.tf
justinsb: This should be the AWS account we use for registry.k8s.io - it's a copy paste.  That said, I agree that the TF configuration is super complicated and so I'm creating a simplified configuration.

justinsb:
I created a PR specifically for the dev configuration in https://github.com/kubernetes/k8s.io/pull/4560 ... I think the stage/prod are useful when we get there, but the dev configuration doesn't need to test the exact prod configuration I don't think (at least not at first, when we have no real dependencies here)

ameukam:
we have a new AWS org. we probably need to rethink how we organize resources and workloads through OUs (folders in GCP terms). 

Let's merge this for the moment.

/lgtm
/approve

infra/aws/terraform/artifacts.k8s.io/README.md
```diff
@@ -0,0 +1,44 @@
+# infra/aws/terraform/artifacts.k8s.io
+
+Terraform to provision AWS infra for supporting sandbox-artifacts.k8s.io (and in future the prod equivalent artifacts.k8s.io)
+
+## Goals
+
+- create publicly-readable regional S3 buckets for serving Kubernetes container images
```
sftim, line 7: Are these container images, or something else?

infra/aws/terraform/artifacts.k8s.io/main.tf
```diff
@@ -0,0 +1,63 @@
...
+  prefix = var.prefix
+
+  s3_replication_iam_role_arn = "arn:aws:iam::513428760722:role/artifacts.k8s.io_s3writer"
+
+  s3_replication_rules = [
+    {
+      id                               = "artifacts-k8s-io-us-east-2-to-artifacts-k8s-io-us-west-2"
+      status                           = "Enabled"
+      priority                         = 2
+      destination_bucket_arn           = "arn:aws:s3:::${var.prefix}artifacts-k8s-io-us-west-2"
```
sftim, line 59: This could use a tidy up.
For example, interpolate in the name of the source and target regions.

infra/aws/terraform/artifacts.k8s.io/providers.tf
```diff
@@ -0,0 +1,63 @@
...
+*/
+
+terraform {
+  backend "s3" {
+    bucket = "artifacts-k8s-io-tfstate"
+    key    = "terraform.tfstate"
+    region = "us-east-2"
+  }
+
+  required_version = ">= 1.2.0"
```
sftim, line 24:
```suggestion
  required_version = "~> 1.2.0"
```
Safer (in case of a TF 2.x coming out).

infra/aws/terraform/artifacts.k8s.io/s3/variables.tf
```diff
@@ -0,0 +1,32 @@
...
+
+variable "prefix" {
+  type    = string
+  default = ""
+}
+
+variable "s3_replication_iam_role_arn" {
+  description = "IAM role assumed by S3 service for replication"
+  type        = string
+  default     = ""
```
sftim, line 25: It doesn't make sense to default to `""`. Either default to `null` and make replication condition, or set this to `nullable = false`, _don't_ set a default, and require the caller to set it.

sftim (commented):
Thanks. Some thoughts. I know a bunch of this code is copied; if we make improvements, we might want to fix the place we copied from too.

infra/aws/terraform/artifacts.k8s.io/s3/variables.tf
justinsb: Thanks - it's triggered by s3_replication_rules, so it's not required.  I'll default it to null

infra/aws/terraform/artifacts.k8s.io/providers.tf
justinsb: Done - thanks!

infra/aws/terraform/artifacts.k8s.io/README.md
justinsb: Thanks - fixed!

infra/aws/terraform/artifacts.k8s.io/main.tf
justinsb: I could use some help here.  So I'm focusing on test-artifacts.k8s.io at the moment.  I put the replication in #4714.  You can see that I was able to get some interpolation, but I still need to repeat the provider blocks.  So it's better, but still suboptimal.  If you know any tricks, happy to review PRs / test them out!

infra/aws/terraform/artifacts.k8s.io/main.tf
```diff
@@ -0,0 +1,70 @@
...
+  default     = "test-"
+  description = "The prefix for all resources"
+
+  validation {
+    condition     = can(regex(".*-$|^$", var.prefix))
+    error_message = "The string must end with a hyphen or be empty."
+  }
+}
+
+module "us-east-2" {
```
sftim, line 32: I'd do:
```suggestion
module "bucket_us-east-2" {
```

etc

infra/aws/terraform/artifacts.k8s.io/providers.tf
```diff
@@ -0,0 +1,60 @@
...
+  }
+
+  required_version = "~> 1.2.0"
+
+  required_providers {
+    aws = {
+      source  = "hashicorp/aws"
+      version = "~> 4.0"
+    }
+  }
```
sftim, line 31: Do we want to lock the remote state? See https://developer.hashicorp.com/terraform/language/settings/backends/s3 for details.

infra/aws/terraform/artifacts.k8s.io/s3/bucket.tf
```diff
@@ -0,0 +1,129 @@
...
+
+resource "aws_s3_bucket" "artifacts-k8s-io" {
+  provider = aws
+  bucket = "${var.prefix}artifacts-k8s-io-${data.aws_region.current.name}"
+}
+
+resource "aws_s3_bucket_acl" "artifacts-k8s-io" {
+  provider = aws
+  bucket = aws_s3_bucket.artifacts-k8s-io.bucket
+  # This clears the ACL list, so we can apply object_ownership = "BucketOwnerEnforced"
```
sftim, line 25: I'm not sure that this does have that effect.

infra/aws/terraform/artifacts.k8s.io/s3/bucket.tf
```diff
@@ -0,0 +1,129 @@
...
+        "Principal" : "*"
+      },
+      {
+        "Action" : "s3:GetObject",
+        "Effect" : "Allow",
+        "Resource" : "${aws_s3_bucket.artifacts-k8s-io.arn}/*",
+        "Principal" : "*"
+      },
+      {
+        "Sid" : "AllowSSLRequestsOnly",
```
sftim, line 51: (nit)
```suggestion
        "Sid" : "RequireTLSForObjectAccess",
```

infra/aws/terraform/artifacts.k8s.io/s3/providers.tf
```diff
@@ -0,0 +1,23 @@
...
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+terraform {
+  required_providers {
+    aws = {
+      source  = "hashicorp/aws"
+    }
```
sftim, line 21:
```suggestion
    aws = {
      source  = "hashicorp/aws"
      version = "~> 4.0"
    }
```
I'd pin to a major version here too.

infra/aws/terraform/artifacts.k8s.io/s3/variables.tf
```diff
@@ -0,0 +1,32 @@
...
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+variable "prefix" {
+  type    = string
+  default = ""
```
sftim, line 19:
```suggestion
  type     = string
  default  = ""
  nullable = false
```
?

infra/aws/terraform/artifacts.k8s.io/s3/variables.tf
```diff
@@ -0,0 +1,32 @@
...
+variable "s3_replication_iam_role_arn" {
+  description = "IAM role assumed by S3 service for replication"
+  type        = string
+  default     = null
+}
+
+variable "s3_replication_rules" {
+  description = "List of maps for S3 replication rules"
+  type        = list(map(string))
+  default     = []
```
sftim, line 31: Could make this non-`nullable` too.

infra/aws/terraform/test-artifacts.k8s.io/s3/providers.tf
```diff
@@ -0,0 +1,23 @@
...
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+terraform {
+  required_providers {
+    aws = {
+      source  = "hashicorp/aws"
+    }
```
sftim, line 21:
```suggestion
    aws = {
      source  = "hashicorp/aws"
      version = "~> 4.0"
    }
```
I'd pin to a major version here too.

sftim (commented):
A few thoughts. Hope this feedback helps.

infra/aws/terraform/artifacts.k8s.io/s3/bucket.tf
justinsb: Well it seemed to work when I accidentally applied an ACL the first time :-)

I hit the InvalidBucketAclWithObjectOwnership described [here](https://docs.aws.amazon.com/AmazonS3/latest/userguide/object-ownership-error-responses.html)

And took the hint from [here](https://github.com/hashicorp/terraform-provider-aws/issues/26164#issuecomment-1260995693)

If there's a better way, that would be great!

infra/aws/terraform/artifacts.k8s.io/s3/variables.tf
justinsb: Good idea - will do!

infra/aws/terraform/artifacts.k8s.io/providers.tf
justinsb: That could be a good idea, but we aren't doing it elsewhere, and it introduces a dependency on DynamoDB, so I suggest we add it separately.  TBH I'm not even sure how we apply this terraform for production, we might well have some protection against concurrent execution at a different layer.

justinsb:
Thanks for the pointers @sftim 

For using `for_each`, I'd love to, but I don't think it's possible in terraform because of https://github.com/hashicorp/terraform/issues/24476

That said, I'll happily review a PR if you want to send one to dedup this - you definitely know terraform much better than I do :-)

justinsb:
/area artifacts

@ameukam are you OK if we merge this?  It reflects our configuration as applied...

justinsb:
This configuration is currently applied to AWS.  I propose we merge it as is to avoid any drift.  I'm happy to review any PRs for further changes / fix any issues etc.

cc @ameukam

Issue #4528, opened 2022-12-07
Shield dl.k8s.io with Fastly


ameukam:
Ensure we serve binaries through Fastly

/area infra
/milestone v1.27
/priority important-longterm

ameukam:
TODO:
- Add staging bucket to use with Fastly
- add a periodic e2e test with `KUBERNETES_CI_RELEASE_URL=cdn.dl.k8s.io`
- add regional periodic e2e tests to validate Fastly shielding.

ameukam:
We can consider this done. THere are a few optimisations to do and they already tracked

/close

PR #4555, opened 2022-12-16
fastly: Add configuration for cdn.dl.k8s.io

ameukam:
Related:

Add Fastly configuration for cdn.dl.k8s.io

### DIFF
--- a/infra/fastly/terraform/dl.k8s.io/provider.tf
+++ b/infra/fastly/terraform/dl.k8s.io/provider.tf
@@ -1,0 +1,33 @@
+/*
+Copyright 2023 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+/*
+This file defines:
+- Required provider versions
+- Storage backend details
+*/
+
+terraform {
+  required_providers {
+    fastly = {
+      source  = "fastly/fastly"
+      version = "~> 4.3.0"
+    }
+  }
+}
+
+# Configure the Fastly Provider
+provider "fastly" {}
--- a/infra/fastly/terraform/dl.k8s.io/services.tf
+++ b/infra/fastly/terraform/dl.k8s.io/services.tf
@@ -1,0 +1,117 @@
+/*
+Copyright 2023 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+locals {
+  domain_name = "cdn.dl.k8s.io"
+}
+
+resource "fastly_service_vcl" "files" {
+  name     = local.domain_name
+  activate = true
+
+  domain {
+    name = local.domain_name
+  }
+
+  backend {
+    name             = "GCS"
+    auto_loadbalance = false
+
+    healthcheck = "GCS Health"
+
+    address           = "storage.googleapis.com"
+    port              = 443
+    use_ssl           = true
+    ssl_cert_hostname = "storage.googleapis.com"
+    ssl_sni_hostname  = "storage.googleapis.com"
+
+    override_host = "${var.bucket}.storage.googleapis.com"
+
+    connect_timeout       = 5000  # milliseconds
+    first_byte_timeout    = 60000 # milliseconds
+    between_bytes_timeout = 15000 # milliseconds
+    error_threshold       = 5
+  }
+
+  healthcheck {
+    name = "GCS Health"
+
+    host           = "${var.bucket}.storage.googleapis.com"
+    method         = "GET"
+    path           = "/"
+    check_interval = 3000
+    timeout        = 2000
+    threshold      = 2
+    initial        = 2
+    window         = 4
+  }
+
+  product_enablement {
+    origin_inspector = true
+  }
+
+  snippet {
+    content  = <<-EOT
+      if (req.url.path ~ "^/release/") {
+        set req.enable_segmented_caching = true;
+      }
+    EOT
+    name     = "Enable segment caching for large files"
+    priority = 60
+    type     = "recv"
+  }
+
+  # Allow CORS GET requests.
+  header {
+    destination = "http.access-control-allow-origin"
+    type        = "response"
+    action      = "set"
+    name        = "CORS Allow"
+    source      = "\"*\""
+  }
+
+  # Do not cache 'not found' & authenticated requests:
+  condition {
+    type      = "CACHE"
+    name      = "is-not-found"
+    statement = "beresp.status == 404"
+  }
+
+  cache_setting {
+    name            = "pass-not-found"
+    cache_condition = "is-not-found"
+    action          = "pass"
+    ttl             = 120
+    stale_ttl       = 120
+  }
+
+  vcl {
+    name    = "Main"
+    content = file("${path.module}/vcl/binaries.vcl")
+    main    = true
+  }
+
+  force_destroy = true
+}
+
+resource "fastly_tls_subscription" "files" {
+  domains               = [for domain in fastly_service_vcl.files.domain : domain.name]
+  certificate_authority = "lets-encrypt"
+}
+
+output "files_managed_dns_challenge" {
+  value = fastly_tls_subscription.files.managed_dns_challenges
+}
--- a/infra/fastly/terraform/dl.k8s.io/vcl/binaries.vcl
+++ b/infra/fastly/terraform/dl.k8s.io/vcl/binaries.vcl
@@ -1,0 +1,116 @@
+sub vcl_recv {
+  # configure purges to require api authentication:
+  # https://docs.fastly.com/en/guides/authenticating-api-purge-requests
+  #
+  if (req.method == "FASTLYPURGE") {
+      set req.http.Fastly-Purge-Requires-Auth = "1";
+  }
+
+  # Prevent edge from caching stale content served from shield
+  # https://developer.fastly.com/learning/concepts/stale/#shielding-considerations
+  if (fastly.ff.visits_this_service != 0) {
+    set req.max_stale_while_revalidate = 0s;
+  }
+
+
+#FASTLY recv
+
+  # don't bother doing a cache lookup for a request type that isn't cacheable
+  if (req.method !~ "(GET|HEAD|FASTLYPURGE)") {
+    return(pass);
+  }
+
+  return(lookup);
+}
+
+sub vcl_fetch {
+
+  /* handle 5XX (or any other unwanted status code) */
+  if (beresp.status >= 500 && beresp.status < 600) {
+    /* deliver stale if the object is available */
+    if (stale.exists) {
+      return(deliver_stale);
+    }
+
+    if (req.restarts < 1 && (req.request == "GET" || req.request == "HEAD")) {
+      restart;
+    }
+
+    set beresp.cacheable = false;
+    set beresp.ttl = 0s;
+    /* else go to vcl_error to deliver a synthetic */
+  error beresp.status;
+  }
+
+  if (beresp.http.Surrogate-Control !~ "(stale-while-revalidate|stale-if-error)") {
+    set beresp.stale_if_error = 31536000s; // 1 year
+    set beresp.stale_while_revalidate = 60s; // 1 minute
+  }
+
+#FASTLY fetch
+  if (req.restarts > 0) {
+    set beresp.http.Fastly-Restarts = req.restarts;
+  }
+
+  if (beresp.http.Set-Cookie) {
+    set req.http.Fastly-Cachetype = "SETCOOKIE";
+    return(pass);
+  }
+
+if (beresp.http.Cache-Control ~ "private") {
+    set req.http.Fastly-Cachetype = "PRIVATE";
+    return(pass);
+  }
+
+  if (beresp.http.Expires || beresp.http.Surrogate-Control ~ "max-age" || beresp.http.Cache-Control ~ "(s-maxage|max-age)") {
+    # keep the ttl here
+  } else {
+    # apply the default ttl
+    set beresp.ttl = 3600s;
+  }
+
+  return(deliver);
+}
+
+sub vcl_deliver {
+#FASTLY deliver
+
+  if (req.http.X-Monitoring == "true") {
+    set resp.http.X-Monitoring = req.http.X-Monitoring;
+  }
+
+  return(deliver);
+}
+
+sub vcl_hit {
+#FASTLY hit
+
+    # If the object we have isn't cacheable, then just serve it directly
+    # without going through any of the caching mechanisms.
+    if (!obj.cacheable) {
+        return(pass);
+    }
+
+    return(deliver);
+}
+
+sub vcl_deliver {
+    if (resp.status >= 500 && resp.status < 600) {
+        if (stale.exists) {
+            restart;
+        }
+    }
+}
+
+sub vcl_miss {
+#FASTLY miss
+  return(fetch);
+}
+
+sub vcl_error {
+#FASTLY error
+}
+
+sub vcl_pass {
+#FASTLY pass
+}
--- a/infra/fastly/terraform/dl.k8s.io/variables.tf
+++ b/infra/fastly/terraform/dl.k8s.io/variables.tf
@@ -1,0 +1,26 @@
+/*
+Copyright 2023 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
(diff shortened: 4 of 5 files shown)


ameukam:
Got:

```bash
$curl -I http://cdn.dl.k8s.io/release/v1.21.2/kubernetes-test-linux-amd64.tar.gz
HTTP/1.1 503 Response object too large
Connection: keep-alive
Content-Length: 466
Server: Varnish
Retry-After: 0
Content-Type: text/html; charset=utf-8
Accept-Ranges: bytes
Date: Fri, 16 Dec 2022 16:34:04 GMT
X-Varnish: 1152098499
Via: 1.1 varnish
```

ameukam:
cc @chrispoole643

infra/fastly/terraform/dl.k8s.io/services.tf
```diff
@@ -0,0 +1,89 @@
...
+  activate = true
+
+  domain {
+    name = "cdn.dl.k8s.io"
+  }
+
+  backend {
+    name             = "GCS"
+    auto_loadbalance = false
+    shield           = "bfi-wa-us"
```
aojea, line 12: > [shield](https://registry.terraform.io/providers/fastly/fastly/latest/docs/resources/service_vcl#shield) (String) The POP of the shield designated to reduce inbound load

is not kind of contradictory for using a cdn?

aojea:
so , the goal is to serve the release binaries from fastly cdn

>       if (req.url.path ~ "^/release/") {

>     host           = "${var.bucket}.storage.googleapis.com"

>     override_host = "${var.bucket}.storage.googleapis.com"


is not easy to get the details from the config, at least until you know terraform and fastly, can we add comments on the code explaining the configuration details?

infra/fastly/terraform/dl.k8s.io/services.tf
```diff
@@ -0,0 +1,89 @@
+resource "fastly_service_vcl" "files" {
+  name     = "cdn.dl.k8s.io"
+  activate = true
+
+  domain {
+    name = "cdn.dl.k8s.io"
```
justinsb, line 6: I suggest `fastly.dl.k8s.io` - it sets up a dynamic where we ack the contributions of the CDNs, and thereby incentivize more CDNs to contribute bandwidth.

dl.k8s.io can then use archeio one day, and likely a simple DNS redirect in the short term

infra/fastly/terraform/dl.k8s.io/services.tf
```diff
@@ -0,0 +1,89 @@
...
+    EOT
+    name     = "Enable segment caching for large files"
+    priority = 60
+    type     = "recv"
+  }
+
+  # Allow CORS GET requests.
+  header {
+    destination = "http.access-control-allow-origin"
+    type        = "cache"
```
chrispoole643, line 75: I'd suggest setting `type` to `response`, instead of `cache`. This way the CORS header will be set right at the point Fastly serves the object, instead of caching the header. Unless you have a specific use case for this?

(E.g., at time of any particular security incident, you would then be able to change this `source` value quickly and easily, instead of having to purge the cache and update the header values.)

infra/fastly/terraform/dl.k8s.io/vcl/binaries.vcl
```diff
@@ -0,0 +1,144 @@
...
+    return(pass);
+  }
+
+  return(lookup);
+}
+
+sub vcl_fetch {
+
+  if (beresp.status >= 200 && beresp.status < 300) {
+    set beresp.cacheable = true;
```
chrispoole643, line 29: This [isn't needed](https://developer.fastly.com/reference/vcl/variables/backend-response/beresp-cacheable/) as you have `req.enable_segmented_caching` set to true, but no harm in being explicit.

infra/fastly/terraform/dl.k8s.io/vcl/binaries.vcl
```diff
@@ -0,0 +1,144 @@
...
+    /* deliver stale if the object is available */
+    if (stale.exists) {
+      return(deliver_stale);
+    }
+
+    if (req.restarts < 1 && (req.request == "GET" || req.request == "HEAD")) {
+      restart;
+    }
+
+    set beresp.cacheable = true;
```
chrispoole643, line 43: Not sure this should be true? Do you want to cache 5xx responses?

infra/fastly/terraform/dl.k8s.io/vcl/binaries.vcl
```diff
@@ -0,0 +1,144 @@
...
+    set beresp.ttl = 1s;
```
chrispoole643, line 44: Same with this also.

infra/fastly/terraform/dl.k8s.io/vcl/binaries.vcl
```diff
@@ -0,0 +1,144 @@
...
+  error beresp.status;
+  }
+
+  if (beresp.http.Surrogate-Control !~ "(stale-while-revalidate|stale-if-error)") {
+    set beresp.stale_if_error = 31536000s; // 1 year
+    set beresp.stale_while_revalidate = 60s; // 1 minute
+  }
+
+#FASTLY fetch
+  if ((beresp.status == 500 || beresp.status == 503) && req.restarts < 1 && (req.request == "GET" || req.request == "HEAD")) {
```
chrispoole643, line 55: I think this `if` block is redundant.

infra/fastly/terraform/dl.k8s.io/vcl/binaries.vcl
```diff
@@ -0,0 +1,144 @@
...
+  }
+
+  if (beresp.http.Cache-Control ~ "private") {
+    set req.http.Fastly-Cachetype = "PRIVATE";
+    return(pass);
+  }
+
+  # If we've gotten an error after the restarts we'll deliver the response
+  # with a very short cache time.
+  if (http_status_matches(beresp.status, "500,502,503")) {
```
chrispoole643, line 75: I don't think this `if` block is needed either; it'll follow the synthetic error path above.

infra/fastly/terraform/dl.k8s.io/vcl/binaries.vcl
```diff
@@ -0,0 +1,144 @@
...
+  return(deliver);
+}
+
+sub vcl_deliver {
+#FASTLY deliver
+
+  # fastly internal state information:
+  # https://docs.fastly.com/en/guides/useful-variables-to-log
+  #
+  set resp.http.Fastly-State = fastly_info.state;
```
chrispoole643, line 98: I'd suggest not setting this header, unless you're doing this now for debug of course. But better to remove this I'd suggest, and set a `Fastly-Debug` header on your requests.

infra/fastly/terraform/dl.k8s.io/vcl/binaries.vcl
```diff
@@ -0,0 +1,144 @@
...
+
+sub vcl_miss {
+#FASTLY miss
+  return(fetch);
+}
+
+sub vcl_error {
+#FASTLY error
+
+    if (obj.status >= 500 && obj.status < 600) {
```
chrispoole643, line 135: Seems like given the above, where this error flow is triggered, this will have already been handled. Better here to return a custom error page.

chrispoole643 (commented):
Broadly looks good, with some comments 🙂

infra/fastly/terraform/dl.k8s.io/services.tf
ameukam: Yeah. we use this ensure objects are pulled through Fastly network.

infra/fastly/terraform/dl.k8s.io/services.tf
```diff
@@ -0,0 +1,89 @@
...
+    timeout        = 2000
+    threshold      = 2
+    initial        = 2
+    window         = 4
+  }
+
+  condition {
+    name      = "NeverReq"
+    type      = "REQUEST"
+    statement = "req.http.Fastly-Client-IP == \"127.0.0.1\" && req.http.Fastly-Client-IP != \"127.0.0.1\""
```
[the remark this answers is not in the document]
ameukam: I did this for tests purpose. I'll remove it.

infra/fastly/terraform/dl.k8s.io/vcl/binaries.vcl
ameukam: No. we don't want to cache 5xx responses. will switch to `false`.

infra/fastly/terraform/dl.k8s.io/vcl/binaries.vcl
ameukam: Removed.

infra/fastly/terraform/dl.k8s.io/vcl/binaries.vcl
ameukam: Removed

infra/fastly/terraform/dl.k8s.io/vcl/binaries.vcl
ameukam: Removed

infra/fastly/terraform/dl.k8s.io/vcl/binaries.vcl
ameukam: I'll drop this for the moment. I'll do define a custom error page as follow-up

infra/fastly/terraform/dl.k8s.io/vcl/binaries.vcl
ameukam: Set to 0.

infra/fastly/terraform/dl.k8s.io/services.tf
```diff
@@ -0,0 +1,89 @@
...
+  }
+
+  condition {
```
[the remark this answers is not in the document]
ameukam: We probably don't need all those conditions. Do we need to put a condition for `404` responses from the origin ?

infra/fastly/terraform/dl.k8s.io/services.tf
ameukam: Put the type to `response`.

infra/fastly/terraform/dl.k8s.io/services.tf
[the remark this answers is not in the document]
chrispoole643: So, HTTP 404 is one of the response codes that are [cached by default](https://docs.fastly.com/en/guides/caching-best-practices#http-status-codes-cached-by-default).

You might want to reduce the cache TTL for a 404 response, from 3600 seconds to 120 seconds for example, as [this guide](https://docs.fastly.com/en/guides/overriding-caching-defaults-based-on-a-backend-response) talks about. This is optional but might well be a good idea for you.

infra/fastly/terraform/dl.k8s.io/provider.tf
```diff
@@ -0,0 +1,35 @@
...
+This file defines:
+- Required provider versions
+- Storage backend details
+*/
+
+terraform {
+  required_providers {
+    fastly = {
+      source  = "fastly/fastly"
+      version = ">= 3.0.4"
```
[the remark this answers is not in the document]
ameukam: Done.

infra/fastly/terraform/dl.k8s.io/services.tf
BenTheElder: +1, the rest looks pretty fine but let's call out fastly here. thanks @chrispoole643 🙏

sftim (commented):
Thanks.

Very informal LGTM, I haven't used this provider before.
Some nits.

thockin (commented):
Should we have a turnup party - we all get on a Zoom, bring our own beer, and turn up fastly?

ameukam:
> Should we have a turnup party - we all get on a Zoom, bring our own beer, and turn up fastly?

Definitely. beer, wine, champagne...gonna be awesome.

infra/fastly/terraform/dl.k8s.io/services.tf
```diff
@@ -0,0 +1,114 @@
...
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+resource "fastly_service_vcl" "files" {
+  name     = "cdn.dl.k8s.io"
```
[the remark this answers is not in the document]
BenTheElder: +1, and I still think this should be like `fastly.dl.k8s.io`

infra/fastly/terraform/dl.k8s.io/services.tf
ameukam: I think it's fine to use this domain for the moment. there is no direct benefict to use a different subdomain.
we can revisit that later.

infra/fastly/terraform/dl.k8s.io/provider.tf
```diff
@@ -0,0 +1,35 @@
...
+This file defines:
+- Required provider versions
+- Storage backend details
+*/
+
+terraform {
+  required_providers {
+    fastly = {
+      source  = "fastly/fastly"
+      version = "~> 3.1.0, < 4.0.0"
```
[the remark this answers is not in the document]
ameukam: Applied and provider bumped.

infra/fastly/terraform/dl.k8s.io/services.tf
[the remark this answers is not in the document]
ameukam: Defined a local variable.

infra/fastly/terraform/dl.k8s.io/services.tf
```diff
@@ -0,0 +1,114 @@
...
+  activate = true
+
+  domain {
+    name = "cdn.dl.k8s.io"
+  }
+
+  backend {
+    name             = "GCS"
+    auto_loadbalance = false
+    shield           = "bfi-wa-us"
```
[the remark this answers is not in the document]
ameukam: For the moment, we don't need to use Fastly shield. Removed.

infra/fastly/terraform/dl.k8s.io/services.tf
```diff
@@ -0,0 +1,114 @@
...
+    port              = 443
+    use_ssl           = true
+    ssl_cert_hostname = "storage.googleapis.com"
+    ssl_sni_hostname  = "storage.googleapis.com"
+
+    override_host = "${var.bucket}.storage.googleapis.com"
+
+    connect_timeout       = 5000
+    first_byte_timeout    = 60000
+    between_bytes_timeout = 15000
```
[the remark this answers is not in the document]
ameukam: Applied

infra/fastly/terraform/dl.k8s.io/provider.tf
```diff
@@ -0,0 +1,35 @@
...
+    }
+  }
+}
+
+# Configure the Fastly Provider
+provider "fastly" {
+  api_key = var.fastly_api_key
+}
```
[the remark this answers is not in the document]
ameukam: variable dropped.

infra/fastly/terraform/dl.k8s.io/services.tf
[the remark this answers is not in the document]
ameukam: I'll handle this as a follow-up PR.

infra/fastly/terraform/dl.k8s.io/services.tf
```diff
@@ -0,0 +1,117 @@
...
+  }
+
+  healthcheck {
+    name = "GCS Health"
+
+    host           = "${var.bucket}.storage.googleapis.com"
+    method         = "GET"
+    path           = "/"
+    check_interval = 3000
+    timeout        = 2000
```
sftim, line 56:
```suggestion
    check_interval = 3000 # milliseconds
    timeout        = 2000 # milliseconds
```

infra/fastly/terraform/dl.k8s.io/services.tf
```diff
@@ -0,0 +1,117 @@
...
+    window         = 4
+  }
+
+  product_enablement {
+    origin_inspector = true
+  }
+
+  snippet {
+    content  = <<-EOT
+      if (req.url.path ~ "^/release/") {
```
sftim, line 68: Could move `/release` into a local or an input. Very much optional.

infra/fastly/terraform/dl.k8s.io/services.tf
```diff
@@ -0,0 +1,117 @@
...
+    name      = "is-not-found"
+    statement = "beresp.status == 404"
+  }
+
+  cache_setting {
+    name            = "pass-not-found"
+    cache_condition = "is-not-found"
+    action          = "pass"
+    ttl             = 120
+    stale_ttl       = 120
```
sftim, line 98: I assume:
```suggestion
    ttl             = 120 # seconds
    stale_ttl       = 120 # seconds
```

sftim (commented):
Just nits.

[How] do we test this?

infra/fastly/terraform/dl.k8s.io/services.tf
ameukam: From a code perspective, it make sense but this path is hardecoded in too many places to make variable. I prefer we make it explicit for the moment.

BenTheElder (commented):
/lgtm
/approve
let's merge and iterate

ameukam:
> Just nits.
> 
> [How] do we test this?

You'll need a local dns configuration according to https://docs.fastly.com/en/guides/working-with-domains#testing-your-domain

ameukam:
Fastly service is created after #5327:

```bash
Plan: 1 to add, 0 to change, 1 to destroy.
fastly_service_vcl.files: Destroying... [id=FkseKokv0pOWtVwjJcXd63]
fastly_service_vcl.files: Destruction complete after 1s
fastly_service_vcl.files: Creating...
fastly_service_vcl.files: Creation complete after 6s [id=myOKWFV3A3TGNBOXkU5kk2]
```

ameukam:
Got a 403 while trying to create the TLS certificate:

```bash
fastly_tls_subscription.files: Creating...
╷
│ Error: 403 - Forbidden:
│ 
│   with fastly_tls_subscription.files,
│   on services.tf line 106, in resource "fastly_tls_subscription" "files":
│  106: resource "fastly_tls_subscription" "files" {
│ 
```

I checked with token permissions

image

ameukam:
Opened a ticket (`609514`) to get IPv6 enabled and  possibly a TLS configuration supporting IP dualstack.

cc @chrispoole643

PR #4580, opened 2022-12-28
Bump network proxy to 0.0.35

jkh52:
(There are no 0.0.34 images because https://github.com/kubernetes/test-infra/pull/28328 was missing at the commit time of commits.)


jkh52:
/assign @andrewsykim 

(are you around? lots of approvers are OOO)

PR #4599, opened 2023-01-06
chore: group changes for capg

richardcase:
This change updates the staging group for CAPG to match the current maintainers of CAPG.

/cc @cpanato @dims 

Signed-off-by: Richard Case <richard.case@suse.com>


cpanato (approved):
/lgtm
/approve

PR #4606, opened 2023-01-09
Remove mikedanese@google.com from k8s-infra-artifact-security

enj:
It is unclear to me why anyone outside of SRC is on this list.


enj:
cc @mikedanese @cjcullen @tallclair

enj:
@kubernetes/security-response-committee

PR #4607, opened 2023-01-10
Update csi-driver-host-path image

xing-yang:
Update images for csi-driver-host-path and external-health-monitor controller.


pohly (approved):
/lgtm

PR #4610, opened 2023-01-10
update SIG release groups for 1.27 leads

salaxander:
Initial PR to update the SIG release groups for 1.27 leads. There will be an additional PR to add shadows once the selection is completed, but in the meantime we need to get the leads the needed access.


cpanato (approved):
/lgtm

Issue #4626, opened 2023-01-13
Tracking 73 Kubernetes AWS Account Migrations from CNCF


hh:
There are 73 total Kubernetes accounts in the CNCF AWS Account. We will need to track these migrations.

arn:aws:servicequotas:::organizations/L-29A0C5DF is a request to raise the default max accounts from 10 to 200.

The process for moving accounts between organisations is outlined here: https://aws.amazon.com/premiumsupport/knowledge-center/organizations-move-accounts/

# Kubernetes OU (3)

-   [x] sig-infra-aws-admins+accounts@kubernetes.io (previously: k8s-infra-accounts@kubernetes.io)
-   [x] sig-release-leads@kubernetes.io (password shared with @ cpanato)
-   [x] k8s-infra-aws-root-account@kubernetes.io (password known by @ameukam and @dims)
- [registry.k8s.io OU (1)](#org817d4e3)
  - [aws-ebs-csi-driver OU (1)](#orgb7326f2)
  - [kops OU (4)](#orgc496c06)
  - [sig-k8s-infra OU (12)](#org25da404)
  - [capa OU (52)](#orgbead97b)

<a id="org817d4e3"></a>

# registry.k8s.io OU (1)

-   [x] k8s-infra-aws-admins+registry-k8s-io@kubernetes.io (previously k8s-infra-aws-registry-k8s-io-admin@kubernetes.io)

<a id="orgb7326f2"></a>

# aws-ebs-csi-driver OU (1)

-   [x] aws-ebs-csi-driver@lists.cncf.io

<a id="orgc496c06"></a>

# kops OU (4)

-   [x] k8s-infra-aws-admins+e2e-kops-2020-03-15-c@kubernetes.io
-   [x] k8s-infra-aws-admins+e2e-kops-2020-03-15-d@kubernetes.io
-   [x] k8s-infra-aws-admins+e2e-kops-2020-03-15-a@kubernetes.io
-   [x] k8s-infra-aws-admins+e2e-kops-2020-03-15-b@kubernetes.io

<a id="org25da404"></a>

# sig-k8s-infra OU (12)

-   [x] k8s-infra-aws-admins+image2020111905@kubernetes.io
-   [x] k8s-infra-aws-admins+image2020111904@kubernetes.io
-   [x] k8s-infra-aws-admins+image2020111906@kubernetes.io
-   [x] k8s-infra-aws-admins+image2020111907@kubernetes.io
[...]

PR #4646, opened 2023-01-18
Adding cncf-aws-admins to sig-release-leads

hh:
This list is associated with an AWS account that is part of a the migration to a new Kubernetes AWS Org in https://github.com/kubernetes/k8s.io/issues/4626 This is to enable a password reset for that account and ensure someone within #sig-k8s-infra has password access to all Kubernetes AWS Org member accounts.

### DIFF
--- a/groups/sig-release/groups.yaml
+++ b/groups/sig-release/groups.yaml
@@ -294,6 +294,7 @@ groups:
       - jeremy.r.rickard@gmail.com
       - k8s@auggie.dev
       - saschagrunert@gmail.com
+      - cncf-aws-admins@lists.cncf.io
 
   - email-id: release-team@kubernetes.io
     name: release-team


justaugustus (changes requested):
@hh -- This would grant access to cncf-aws-admins@lists.cncf.io to receive private email sent to @kubernetes/sig-release-leads, which is a non-starter.

I would suggest a separation of concerns here, with a new Google Group created explicitly for this purpose (that includes senior Release Managers / SIG Release leads and SIG K8s Infra leads).

/hold

hh:
@justaugustus @saschagrunert I'll created a tracking ticket as your right about possibly creating a new group. 

https://github.com/kubernetes/k8s.io/issues/4659m

I'll send an email to the list, but generally we need to have someone with password access to this AWS account assist in the migration and possibly updating the associated email to a new group once created.

```
aws organizations list-accounts --query 'sort_by(Accounts,&Email)[?contains(Email,`sig-release-leads@kubernetes.io`)==`true`]'        

[
    {
        "Id": "433650573627",
        "Arn": "arn:aws:organizations::768319786644:account/o-6gmjmvfg2i/433650573627",
        "Email": "sig-release-leads@kubernetes.io",
        "Name": "sig-release-leads",
        "Status": "ACTIVE",
        "JoinedMethod": "CREATED",
        "JoinedTimestamp": "2022-02-18T15:24:09.548000+13:00"
    }
]
```

hh:
Email sent.
/close

PR #4650, opened 2023-01-18
Add jeefy as org auditor

ameukam:
Jeefy will be the new point of contact from CNCF.


ameukam:
cc @dims @jeefy

PR #4679, opened 2023-01-25
AWS: delegate core services to the audit account

ameukam:
Ensure core services of the organization are delegated to the audit account.

### DIFF
--- a/infra/aws/terraform/management-account/accounts.tf
+++ b/infra/aws/terraform/management-account/accounts.tf
@@ -27,7 +27,7 @@ module "security_incident_response" {
   source = "../modules/org-account"
 
   account_name = "k8s-infra-security-incident-response"
-  email = "k8s-infra-aws-admins+incident-response@kubernetes.io"
+  email = "k8s-infra-aws-admins+security-incident-response@kubernetes.io"
   parent_id = aws_organizations_organizational_unit.security.id
 }
 
--- a/infra/aws/terraform/management-account/delegated-admins.tf
+++ b/infra/aws/terraform/management-account/delegated-admins.tf
@@ -1,0 +1,52 @@
+/*
+Copyright 2023 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+# https://docs.aws.amazon.com/organizations/latest/userguide/orgs_integrate_services_list.html
+
+resource "aws_organizations_delegated_administrator" "config_multiaccount" {
+  account_id        = module.security_audit.account_id
+  service_principal = "config-multiaccountsetup.amazonaws.com"
+}
+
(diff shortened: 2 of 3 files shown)


ameukam:
cc @sftim

infra/aws/terraform/management-account/accounts.tf
```diff
@@ -27,7 +27,7 @@ module "security_incident_response" {
   source = "../modules/org-account"
 
   account_name = "k8s-infra-security-incident-response"
-  email = "k8s-infra-aws-admins+incident-response@kubernetes.io"
+  email = "k8s-infra-aws-admins+security-incident-response@kubernetes.io"
```
sftim, line 30: Is this going to cause drift? Will we / you fix it manually?

infra/aws/terraform/management-account/delegated-admins.tf
```diff
@@ -0,0 +1,52 @@
...
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+# https://docs.aws.amazon.com/organizations/latest/userguide/orgs_integrate_services_list.html
+
+resource "aws_organizations_delegated_administrator" "config_multiaccount" {
+  account_id        = module.security_audit.account_id
```
sftim, line 20: How about:
```suggestion
  account_id        = module.security_audit.account.id
```
and then don't change the `outputs.tf`

?

infra/aws/terraform/management-account/accounts.tf
ameukam: Yes. We need to wait 90 days to use `k8s-infra-aws-admins+incident-response@kubernetes.io` to create a new account. Since we are bootstrapping the org, create another account should be ok.

infra/aws/terraform/management-account/delegated-admins.tf
ameukam: We can't technically do that with HCL since `module.security_audit` is an object and there is no `account` ojbect in the module. the `id` is private to the module so it is necessary to modify the module to expose it.

sftim: Insufficient coffee. Sure, that sounds right.

infra/aws/terraform/management-account/delegated-admins.tf
```diff
@@ -0,0 +1,52 @@
...
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+# https://docs.aws.amazon.com/organizations/latest/userguide/orgs_integrate_services_list.html
+
+resource "aws_organizations_delegated_administrator" "config_multiaccount" {
```
sftim, line 19: You could use a `for_each` here if you want. Might read better.

sftim (commented):
/hold

/lgtm

@ameukam, unhold this when you want it to go live.

infra/aws/terraform/management-account/delegated-admins.tf
ameukam: 🤔 I'm hesitant about using `for_each` to handle delegated administration. I'm gonna a `TODO`as a follow-up of this PR and we can revisit this.

sftim: No worries.

I wouldn't use it without having support for [`moved`](https://developer.hashicorp.com/terraform/language/modules/develop/refactoring#moved-block-syntax).

PR #4688, opened 2023-01-27
AWS: create cost and usage reports

ameukam:
- create encrypted S3 buckets for AWS cost and usage reports
- create AWS CUR reports with and without integrations to AWS AThena.

### DIFF
--- a/infra/aws/terraform/management-account/s3.tf
+++ b/infra/aws/terraform/management-account/s3.tf
@@ -1,0 +1,187 @@
+/*
+Copyright 2023 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+# Cost and Usage
+
+module "cur_reports_s3_bucket" {
+  providers = {
+    aws = aws.shared-services
+  }
+
+  source = "terraform-aws-modules/s3-bucket/aws"
+
+  bucket = "k8s-infra-cur-reports-bucket"
+
+  # S3 bucket-level Public Access Block configuration
+  block_public_acls       = true
+  block_public_policy     = true
+  ignore_public_acls      = true
+  restrict_public_buckets = true
+
+  attach_deny_insecure_transport_policy = true
+  attach_policy                         = true
+  policy                                = data.aws_iam_policy_document.cur_reports_s3_bucket.json
+
+  # Note: Object Lock configuration can be enabled only on new buckets
+  # https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/s3_bucket_object_lock_configuration
+  object_lock_enabled = true
+
+  versioning = {
+    enabled    = true
+    mfa_delete = false
+  }
+}
+
+module "cur_reports_integrations_s3_bucket" {
+  providers = {
+    aws = aws.shared-services
+  }
+
+  source = "terraform-aws-modules/s3-bucket/aws"
+
+  bucket = "cur_reports_integrations_s3_bucket"
+
+  # S3 bucket-level Public Access Block configuration
+  block_public_acls       = true
+  block_public_policy     = true
+  ignore_public_acls      = true
+  restrict_public_buckets = true
+
+  attach_deny_insecure_transport_policy = true
+  attach_policy                         = true
+  policy                                = data.aws_iam_policy_document.cur_reports_integrations_s3_bucket.json
+
+  # Note: Object Lock configuration can be enabled only on new buckets
+  # https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/s3_bucket_object_lock_configuration
+  object_lock_enabled = true
+
+  versioning = {
+    enabled    = true
+    mfa_delete = false
+  }
+}
+
+data "aws_iam_policy_document" "cur_reports_s3_bucket" {
+  version = "2008-10-17"
+
+  statement {
+    sid    = "AWSBillingDeliveryAclCheck"
+    effect = "Allow"
+    principals {
+      type = "Service"
+      identifiers = [
+        "billingreports.amazonaws.com"
+      ]
+    }
+    actions = [
+      "s3:GetBucketPolicy",
+      "s3:GetBucketAcl"
+    ]
+    resources = [module.cur_reports_s3_bucket.s3_bucket_arn]
+
+    condition {
+      test     = "StringEquals"
+      variable = "aws:SourceArn"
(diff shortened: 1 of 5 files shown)


ameukam:
cc @sftim

infra/aws/terraform/management-account/s3.tf
```diff
@@ -0,0 +1,155 @@
...
+        sse_algorithm     = "aws:kms"
+      }
+    }
+  }
+
+  versioning = {
+    enabled    = true
+    mfa_delete = false
+  }
+}
```
sftim, line 58: This bucket should live in an AWS account in the infrastructure OU.
KMS encryption is nice to have but IMO not essential.

The only exception would be if CUR requires that the bucket is in the management account.

This is because we should minimize the amount of infrastructure that we define in the management account.

sftim: Also, if we can make the bucket be `BucketOwnerEnforced`, let's do that (it turns off ACLs).

AWS' documentation isn't clear, but I have a feeling that CUR doesn't yet support `BucketOwnerEnforced` buckets. If that support is already in place, we should take advantage.

infra/aws/terraform/management-account/s3.tf
```diff
@@ -0,0 +1,155 @@
...
+  statement {
+    sid    = "AWSBillingDeliveryWrite"
+    effect    = "Allow"
+    actions   = ["s3:PutObject"]
+    resources = ["${module.cur_reports_s3_bucket.s3_bucket_arn}/*"]
+
+    principals {
+      type        = "Service"
+      identifiers = ["billingreports.amazonaws.com"]
+    }
```
sftim, line 121:
```suggestion
    }
    
    condition {
      test     = "StringEquals"
      variable = "aws:SourceArn"
      values   = ["arn:aws:cur:us-east-1:${data.aws_caller_identity.current.account_id}:definition/*"]
      # I think this region is ALWAYS us-east-1 but I'm not certain
    }

    condition {
      test     = "StringEquals"
      variable = "aws:SourceAccount"
      values   = [data.aws_caller_identity.current.account_id]
    }
```
along with a `data "aws_caller_identity" "current" {}` somewhere else

infra/aws/terraform/management-account/s3.tf
```diff
@@ -0,0 +1,155 @@
...
+  statement {
+    sid    = "AWSBillingDeliveryWrite"
+    effect    = "Allow"
+    actions   = ["s3:PutObject"]
+    resources = ["${module.cur_reports_integrations_s3_bucket.s3_bucket_arn}/*"]
+
+    principals {
+      type        = "Service"
+      identifiers = ["billingreports.amazonaws.com"]
+    }
```
sftim, line 153:
```suggestion
    }
    
    condition {
      test     = "StringEquals"
      variable = "aws:SourceArn"
      values   = ["arn:aws:cur:us-east-1:${data.aws_caller_identity.current.account_id}:definition/*"]
      # I think this region is ALWAYS us-east-1 but I'm not certain
    }

    condition {
      test     = "StringEquals"
      variable = "aws:SourceAccount"
      values   = [data.aws_caller_identity.current.account_id]
    }
```
along with a `data "aws_caller_identity" "current" {}` somewhere else

infra/aws/terraform/management-account/s3.tf
```diff
@@ -0,0 +1,155 @@
...
+    principals {
+      type = "Service"
+      identifiers = [
+        "billingreports.amazonaws.com"
+      ]
+    }
+    actions = [
+      "s3:GetBucketPolicy",
+      "s3:GetBucketAcl"
+    ]
```
sftim, line 108:
```suggestion
    ]
    
    condition {
      test     = "StringEquals"
      variable = "aws:SourceArn"
      values   = ["arn:aws:cur:us-east-1:${data.aws_caller_identity.current.account_id}:definition/*"]
      # I think this region is ALWAYS us-east-1 but I'm not certain
    }

    condition {
      test     = "StringEquals"
      variable = "aws:SourceAccount"
      values   = [data.aws_caller_identity.current.account_id]
    }
```
along with a `data "aws_caller_identity" "current" {}` somewhere else

sftim (commented):
- We can prevent a confused deputy attack with a source ARN check
- I would define the S3 bucket using a provider for a shared-services account and then configure writes to it from the management account
- if that's difficult to set up with KMS, switch to server-side AES256 encryption

If we might ever want the CUR bucket to support [Object Lock](https://docs.aws.amazon.com/AmazonS3/latest/userguide/object-lock.html), then we should enable Object Lock at bucket creation time.

ameukam:
>     * I would define the S3 bucket using a provider for a shared-services account and then configure writes to it from the management account

Let's create the shared-services account https://github.com/kubernetes/k8s.io/pull/4690 and I'll revamp the PR.

infra/aws/terraform/management-account/s3.tf
ameukam: > To ensure the bucket owner can access these reports, AWS further applies a BucketOwnerFullControl ACL to the reports when delivering them. By default, Amazon S3 objects such as these reports can only be read by the user or service principal who wrote them. To provide you or the bucket owner with permission to read the reports AWS must apply the BucketOwnerFullControl ACL. The ACL grants the bucket owner Permission.FullControl for these reports.

From: https://docs.aws.amazon.com/cur/latest/userguide/cur-s3.html

we should set the ACL to `log-delivery-write` so we don't apply ACLs at the object level.

infra/aws/terraform/management-account/s3.tf
ameukam: Conditions added.

infra/aws/terraform/management-account/s3.tf
ameukam: Conditions added.

infra/aws/terraform/management-account/s3.tf
ameukam: Conditions added.

ameukam:
@sftim Added a object lock rule. PTAL

infra/aws/terraform/management-account/s3.tf
```diff
@@ -0,0 +1,203 @@
...
+  # https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/s3_bucket_object_lock_configuration
+  object_lock_enabled = true
+  object_lock_configuration = {
+    rule = {
+      default_retention = {
+        mode = "COMPLIANCE"
+        years = 3
+      }
+    }
+  }
```
sftim, line 85: See https://github.com/kubernetes/k8s.io/pull/4688/files#r1091009106

infra/aws/terraform/management-account/s3.tf
```diff
@@ -0,0 +1,203 @@
...
```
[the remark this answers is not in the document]
ameukam: `s3_bucket_object_lock_configuration` is part of the module I'm using to create the bucket: https://github.com/terraform-aws-modules/terraform-aws-s3-bucket/blob/6fa011f63ff55992f40751bf24fdf35520b7b61d/main.tf#L324-L338

> This is a “[for but not with](https://en.wikipedia.org/wiki/For_but_not_with)” scenario in case we want to turn on object lock one day in the future - that way, we won't need to use an object lock token.

So we can just turn on the object locking mechanism with `object_lock_enabled = true` and remove the rule.

PR #4693, opened 2023-01-28
Create conduct-emeritus email list

cblecker:
Creating list for CoC emeritus folks. I've only added the current members as owners + @tpepper as an example. I'll let the CoCC folks add the rest of the addresses.


hlipsig (approved):
LGTM

PR #4695, opened 2023-01-28
add distribution lists for github bots

ameukam:
Add distribution lists for github accounts currently owned by Google teams. Those MLs will allow credentials rotation and ownership transfer to the community.

### DIFF
--- a/groups/sig-testing/groups.yaml
+++ b/groups/sig-testing/groups.yaml
@@ -169,6 +169,32 @@ groups:
   # required for membership in these groups.
   #
 
+  - email-id: k8s-ci-robot@kubernetes.io
+    name: k8s-ci-robot
+    description: |-
+      Github admins for github account k8s-ci-robot
+    settings:
+      ReconcileMembers: "true"
+      WhoCanPostMessage: "ANYONE_CAN_POST"
+      WhoCanViewMembership: "ALL_MEMBERS_CAN_VIEW"
+    members:
+    - sig-k8s-infra-leads@kubernetes.io
+    - github@kubernetes.io
+    - sig-testing-leads@kubernetes.io
+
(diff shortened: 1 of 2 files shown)


ameukam:
cc @michelle192837 @BenTheElder

michelle192837 (approved):

groups/sig-testing/groups.yaml
```diff
@@ -169,6 +169,32 @@ groups:
   # required for membership in these groups.
   #
 
+  - email-id: k8s-ci-robot@kubernetes.io
```
BenTheElder, line 172: just to confirm, this is https://groups.google.com/a/kubernetes.io/g/k8s-ci-robot/ which is *NOT* public read?

ameukam: `WhoCanPostMessage: "ANYONE_CAN_POST"` is only needed to ensure we can receive e-mails.

BenTheElder: Yeah, but I don't see the config for "who can read this". 
The live link appears that it is restricted, wanted to confirm.

groups/sig-testing/groups.yaml
```diff
@@ -169,6 +169,32 @@ groups:
...
+
+  - email-id: k8s-github-robot@kubernetes.io
+    name: k8s-github-robot
+    description: |-
+      Github admins for github account k8s-github-robot
+    settings:
+      ReconcileMembers: "true"
+      WhoCanPostMessage: "ANYONE_CAN_POST"
+      WhoCanViewMembership: "ALL_MEMBERS_CAN_VIEW"
+    members:
```
BenTheElder, line 193: can we promote someone to OWNER of this?
right now nobody can e.g. moderate, as we only have members

Issue #4708, opened 2023-02-01
DNS Update Request


justinsb:
### Type of DNS update:
Create

### Domain being modified:
`k8s.io`

### New DNS Record:

```yaml
# Sandbox artifact redirector service.
artifacts-sandbox:
  - type: A
    value: 34.110.218.166
  - type: AAAA
    value: "2600:1901:0:3b54::"
```

### Reason for update:

Creating artifacts-sandbox.k8s.io pointing to load balancer IP addresses in k8s-infra-porche-sandbox

justinsb:
I had a go at implementing in #4712 if that's helpful.

justinsb:
#4712 has merged (thanks!), so closing!

/close

PR #4712, opened 2023-02-02
DNS: add A and AAAA records for artifacts-sandbox.k8s.io

justinsb:
In case it helps with #4708 :-)


dims:
@justinsb Why Not!! let's give it a shot :)

/approve
/lgtm

PR #4713, opened 2023-02-02
Update terraform config for artifacts-sandbox.k8s.io

justinsb:
This is the terraform configuration as currently applied to artifacts-sandbox.k8s.io.

Some small tweaks were needed, nothing particularly significant.


ameukam:
/approve
/hold

Remove the hold when ready.

PR #4714, opened 2023-02-02
Add replication to test-artifacts.k8s.io terraform

justinsb:
Builds on #4460 , adding replication to two regions, just to give a more realistic test environment and so we can reuse this configuration in prod with more regions.

- infra/aws: create S3 buckets for artifacts.k8s.io
- test-artifacts.k8s.io terraform: Add replication configuration


justinsb:
/hold

I don't think we want to use AWS replication here, let's see how it goes directly copying to each of our mirrors.

infra/aws/terraform/artifacts.k8s.io/main.tf
```diff
@@ -0,0 +1,63 @@
+/*
+Copyright 2022 The Kubernetes Authors.
```
TerryHowe, line 2:
```suggestion
Copyright The Kubernetes Authors.
```
Year is old. Not sure if year from copyright notices are allowed here. It saves a lot of hassle.

justinsb:
We are not currently using AWS replication and it seems to be working well (at least when I set up IAM correctly :-) ), so going to close this; we can always reopen if we change our mind.

/close

PR #4715, opened 2023-02-02
Mirror artifacts to S3

justinsb:
This is the configuration I used to populate the S3 sandbox mirrors.

Comments welcome!

Edit: We thought about whether we should use S3 replication vs copying to each bucket separately, this PR does not use S3 replication; we can revisit in future if needed.  ~In particular, I'm not sure if we should use S3 replication vs just copying to each individually.~

Cons of S3 replication:
* S3 replication is a little tricky to set up (https://github.com/kubernetes/k8s.io/pull/4714)
* Having one bucket as the source makes that bucket a (likely) single point of failure (a change is replicated everywhere, if the bucket is down we can't publish to the replicas).
* We likely want to audit the replica buckets also, so we probably need _a_ configuration for them also.

Pros of S3 replication:
* Less bandwidth used during publish - we only need to write to S3 once.  The cost seems insiginficant compared to download costs, but it will be slower to publish to each replica.  But transfers are incremental, so after the initial copy it should be pretty fast.

### DIFF
--- a/artifacts/mirroring-staging/filestores/k8s-staging-cri-tools/filepromoter-manifest.yaml
+++ b/artifacts/mirroring-staging/filestores/k8s-staging-cri-tools/filepromoter-manifest.yaml
@@ -1,0 +1,7 @@
+filestores:
+- base: gs://k8s-artifacts-prod/binaries/cri-tools/
+  src: true
+- base: s3://test-artifacts-k8s-io-us-east-2/binaries/cri-tools/
+- base: s3://test-artifacts-k8s-io-us-west-2/binaries/cri-tools/
+- base: s3://test-artifacts-k8s-io-eu-west-2/binaries/cri-tools/
+- base: s3://test-artifacts-k8s-io-ap-southeast-1/binaries/cri-tools/
--- a/artifacts/mirroring-staging/README.md
+++ b/artifacts/mirroring-staging/README.md
@@ -1,0 +1,7 @@
+# Kubernetes Artifacts - Mirroring
+
+TODO
+
+```
+kpromo run files --manifests artifacts/mirroring/
+```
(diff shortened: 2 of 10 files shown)


justinsb:
Based on some discussions with @BenTheElder I'm going to pursue the idea of _not_ using AWS replication here.  It saves us very little money (if any), and it really complicates things by creating a dependency on a service over which we have little control.  Moreover, it doesn't save us a lot of work, in that we'd likely have to create the audit jobs anyway.

justinsb:
First prow job to try using these is https://github.com/kubernetes/test-infra/pull/28676

artifacts/mirroring-staging/filestores/k8s-staging-cri-tools/filepromoter-manifest.yaml
```diff
@@ -0,0 +1,7 @@
+filestores:
+- base: gs://k8s-artifacts-prod/binaries/cri-tools/
+  src: true
+- base: s3://test-artifacts-k8s-io-us-east-2/binaries/cri-tools/
```
BenTheElder, line 4: do we have auth setup such that this will work?

justinsb: So this can work without auth, it's all publicly accessible (currently).  I am not sure whether it will try to get an auth token and fail or whether it will simply not try to get one (and then hopefully succeed).

I'm also hoping we'll find out now that you're merged the test-infra job - thank you!

justinsb:
/test pull-k8sio-file-promo-with-mirroring

Maybe this will trigger our new test (maybe!)

justinsb:
It did trigger it, but looks like we need a newer version of kpromo (which is to be expected, TBH, we only added S3 support recently):

```
level=fatal msg="run `kpromo run files`: error validating merged manifest: filestore has unsupported scheme in base \"s3://prod-artifacts-k8s-io-us-east-2/binaries/cri-tools/\""
```

ameukam:

justinsb:
/test pull-k8sio-file-promo-with-mirroring

Updated image :crossed_fingers:

xmudrii (approved):
/lgtm
/approve
/hold
if @jeremyrickard wants to take a look

justinsb:
/test pull-k8sio-file-promo-with-mirroring

Still running with older job definition, hopefully I just pushed the button too soon....

PR #4716, opened 2023-02-02
artifacts-sandbox terraform: Add cloudfront distribution

justinsb:
We can add a cloudfront distribution, it's a good option for when we don't have a bucket local to the user.  TBD whether we should send the user to cloudfront when we do have a cheaper/faster bucket.

### DIFF
--- a/infra/aws/terraform/test-artifacts.k8s.io/cloudfront.tf
+++ b/infra/aws/terraform/test-artifacts.k8s.io/cloudfront.tf
@@ -1,0 +1,76 @@
+/*
+Copyright 2023 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+resource "aws_cloudfront_distribution" "artifacts" {
+  enabled             = true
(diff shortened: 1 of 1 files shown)


justinsb:
cc @ameukam and @upodroid  - the third commit adds cloudfront. You're the experts here - I did the minimum configuration - suggested changes greatly appreciated!

ameukam:
@justinsb This is awesome! 

Few things we can do as follow-ups:
- Use the [price class ](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/cloudfront_distribution#price_class) to cover all the Cloud regions. See [CloudFront Price Class](https://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/PriceClass.html)
- Enable [Origin Shield](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/cloudfront_distribution#origin_shield). See [docs](https://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/origin-shield.html) for CloudFront Origin Shield.
- For production enable the access logs (you'll need a S3 bucket)

We probably need to write a TF module we can reuse for production and other workloads. 

We also define the policy we want to establish for object caching.

infra/aws/terraform/test-artifacts.k8s.io/cloudfront.tf
```diff
@@ -0,0 +1,60 @@
...
+
+  origin {
+    domain_name = "test-artifacts-k8s-io-eu-west-2.s3.eu-west-2.amazonaws.com"
+    origin_id   = "failoverS3"
+  }
+
+  default_cache_behavior {
+    allowed_methods  = ["GET", "HEAD", "OPTIONS"]
+    cached_methods   = ["GET", "HEAD"]
+    viewer_protocol_policy = "allow-all"
```
ameukam, line 34:
```suggestion
    viewer_protocol_policy = "redirect-to-https"
```

justinsb:
Added missing license header, probably lost LGTM

PR #4721, opened 2023-02-03
AWS: create more accounts

ameukam:
- add different OUs to establish a hierarchy between production and non-production accounts
- create a network account for Network services (Cloudfront, Shield, etc...)
- create a account as env prod for artifacts.k8s.io


ameukam:
cc @dims

PR #4737, opened 2023-02-08
artifacts.k8s.io: Create prod configuration for CloudRun

justinsb:
Copying the existing configuration with a few tweaks to remove sandbox etc.

### DIFF
--- a/infra/gcp/terraform/k8s-infra-porche-prod/network.tf
+++ b/infra/gcp/terraform/k8s-infra-porche-prod/network.tf
@@ -1,0 +1,103 @@
+/*
+Copyright 2022 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+resource "google_compute_global_address" "default_ipv4" {
+  project      = google_project.project.project_id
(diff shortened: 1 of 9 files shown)


infra/gcp/terraform/k8s-infra-porche-prod/network.tf
```diff
@@ -0,0 +1,103 @@
...
+  create_address      = false
+  create_ipv6_address = false
+  enable_ipv6         = true
+  https_redirect      = true
+  #TODO(ameukam): current the TF resource google_compute_global_address don't have
+  #the value of IP in his attribute. But it's accessible with the data source:
+  #https://registry.terraform.io/providers/hashicorp/google/latest/docs/data-sources/compute_global_address
+  address      = data.google_compute_global_address.default_ipv4.address
+  ipv6_address = data.google_compute_global_address.default_ipv6.address
+  managed_ssl_certificate_domains = [
```
justinsb, line 97: So when we repoint DNS I think there will be about 15 minutes where we won't have a valid cert etc.  Not sure if there's any way around this...

ameukam: We need to define a maintenance window of 1 hour to do the switch (happy to do the switch on a week-end).

ameukam:
/lgtm
/hold

Remove the hold when ready.

justinsb:
Well this won't really "do" anything until we move the DNS so I think we can apply this whenever we want to.  I don't know if we need to block cutover on having AWS artifact copying be fully automatic - we don't _have_ to, but I've been delaying completing the artifact copying until we have at least one prow job running kpromo with these mirrors.

ameukam:
Project created, services enabled and Groups empowered as `Owner`:

```bash
$terraform apply -auto-approve -target 'google_project.project' -target 'google_project_service.project' -target 'google_project_iam_member.k8s_infra_porche_admins'
...

Releasing state lock. This may take a few moments...

Apply complete! Resources: 11 added, 0 changed, 0 destroyed.
```

PR #4739, opened 2023-02-08
first pass mapping cloud run regions to S3 regions

BenTheElder:
xref: https://github.com/kubernetes/registry.k8s.io/pull/147

this environment variable doesn't do anything yet, it will after we update to https://github.com/kubernetes/registry.k8s.io/pull/147

### DIFF
--- a/infra/gcp/terraform/k8s-infra-oci-proxy-prod/terraform.tfvars
+++ b/infra/gcp/terraform/k8s-infra-oci-proxy-prod/terraform.tfvars
@@ -19,7 +19,14 @@ project_id = "k8s-infra-oci-proxy-prod"
 digest     = "sha256:fc811af33c78e9765ee95d6994a6e0dac4c382ae63a960872404b7190110c769"
 cloud_run_config = {
   asia-east1 = {
+    // TODO: switch DEFAULT_AWS_BASE_URL to cloudfront or else refine the region mapping
+    // GCP asia-east1 is Changhua County, Taiwan
     environment_variables = [
+      {
+        name = "DEFAULT_AWS_BASE_URL",
+        // AWS ap-southeast-1 is Singapore
+        value = "https://prod-registry-k8s-io-ap-southeast-1.s3.dualstack.ap-southeast-1.amazonaws.com",
+      },
       {
         name  = "UPSTREAM_REGISTRY_ENDPOINT",
         value = "https://asia-east1-docker.pkg.dev"
@@ -30,8 +37,14 @@ cloud_run_config = {
       }
     ]
   }
+  // GCP asia-northeast1 is Tokyo, Japan
   asia-northeast1 = {
     environment_variables = [
+      {
+        name = "DEFAULT_AWS_BASE_URL",
+        // AWS ap-northeast-1 is Tokyo
+        value = "https://prod-registry-k8s-io-ap-northeast-1.s3.dualstack.ap-northeast-1.amazonaws.com",
+      },
       {
         name  = "UPSTREAM_REGISTRY_ENDPOINT",
         value = "https://asia-northeast1-docker.pkg.dev"
@@ -42,8 +55,14 @@ cloud_run_config = {
       }
     ]
   }
+  // GCP asia-northeast2 is Osaka, Japan
   asia-northeast2 = {
     environment_variables = [
+      {
+        name = "DEFAULT_AWS_BASE_URL",
+        // AWS ap-northeast-1 is Tokyo
+        value = "https://prod-registry-k8s-io-ap-northeast-1.s3.dualstack.ap-northeast-1.amazonaws.com",
+      },
       {
         name  = "UPSTREAM_REGISTRY_ENDPOINT",
         value = "https://asia-northeast2-docker.pkg.dev"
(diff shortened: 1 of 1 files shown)


infra/gcp/terraform/k8s-infra-oci-proxy-prod/terraform.tfvars
```diff
@@ -162,8 +235,14 @@ cloud_run_config = {
       }
     ]
   }
+  // GCP southamerica-west1 is Santiago, Chile
   southamerica-west1 = {
     environment_variables = [
+      {
+        name = "DEFAULT_AWS_BASE_URL",
+        // AWS us-east-1 is Virginia, USA
```
justinsb, line 243: This one is pretty surprising and maybe comment worthy

infra/gcp/terraform/k8s-infra-oci-proxy-prod/terraform.tfvars
```diff
@@ -246,8 +361,14 @@ cloud_run_config = {
       }
     ]
   }
+  // GCP us-west2 is California, USA
   us-west2 = {
     environment_variables = [
+      {
+        name = "DEFAULT_AWS_BASE_URL",
+        // AWS us-west-1 is California, USA
+        value = "https://prod-registry-k8s-io-us-west-1.s3.dualstack.us-west-2.amazonaws.com",
```
justinsb, line 370: Typo: us-west-1 and us-west-2 in same URL, should match I think

justinsb:
LGTM though I think you need to fix that second URL

Do you want to do this in sandbox first / as well?

infra/gcp/terraform/k8s-infra-oci-proxy-prod/terraform.tfvars
BenTheElder: Fair, we don't have a bucket in south america currently https://github.com/kubernetes/registry.k8s.io/blob/8d1d604a30942e89b7e085534e4d39f2819440e4/cmd/archeio/app/buckets.go#L38-L67

I don't think any of the other regions we have are closer

BenTheElder: Regions we have a bucket in were based on traffic data xref https://github.com/kubernetes/registry.k8s.io/issues/39

infra/gcp/terraform/k8s-infra-oci-proxy-prod/terraform.tfvars
BenTheElder: fixed!

infra/gcp/terraform/k8s-infra-oci-proxy-prod/terraform.tfvars
BenTheElder: Added a self-referential comment to this discussion with context links

dims: @BenTheElder there's one in sao paolo `sa-east-1`

dims: do we mean we don't have a bucket there and choose to serve from us-east-1?

BenTheElder:
> LGTM though I think you need to fix that second URL

which URL? I think I fixed everything?

> Do you want to do this in sandbox first / as well?

good idea, though we auto deploy to the sandbox using some bash in the registry repo currently, we don't use the terraform currently

BenTheElder:
I think letting the sandbox default to one region is OK, we only have some test CI aimed at it. No end users should be using the sandbox endpoint and only a very limited subset of CI. We have a default-default.

ameukam:
/approve
/hold

Remove the hold when ready.

BenTheElder:
terraform applyed

BenTheElder:
https://prow.k8s.io/view/gs/kubernetes-jenkins/logs/post-oci-proxy-push-images/1623444771215249408 rerun post tag (manually triggered, not sure why tag push didn't trigger...)

gcr.io/k8s-staging-infra-tools/archeio:v20230208-v0.1.0@sha256:43904e32c58803bed1dbf6457e1b5abdd1409c976bb951f5bb32cf3b37772b2f

infra/gcp/terraform/k8s-infra-oci-proxy-prod/terraform.tfvars
BenTheElder: Yes, to clarify there is an AWS region in South America but kubernetes does not have a bucket there, while we did wind up having cloud run / AR there.

Most places we have reasonably close proximity between AR/Cloud Run and S3, but not here.

infra/gcp/terraform/k8s-infra-oci-proxy-prod/terraform.tfvars
dims: thanks Ben

infra/gcp/terraform/k8s-infra-oci-proxy-prod/terraform.tfvars
BenTheElder: (Also we only have a few QPS here, and previously k8s.gcr.io only had EU,US,Asia multi-regional storage, so while this feels off it's probably not the most pressing)

BenTheElder: this is now https://github.com/kubernetes/registry.k8s.io/issues/172

PR #4742, opened 2023-02-08
update archeio deployment to v0.1.0

BenTheElder:
deploying this will have to wait on https://github.com/kubernetes/k8s.io/pull/4741 to land, but that will be soon, and merging it will not auto-deploy it currently.


BenTheElder:
promo will take a bit, there's a fast postsubmit but it has to wait for the slow periodic because they're in the same prow concurrency restriction

BenTheElder:
watching https://prow.k8s.io/?repo=kubernetes%2Fk8s.io

might make more sense to deploy from staging, knowing that either way we're pinning a digest.

BenTheElder:
image is live, terraform apply-ing

Issue #4763, opened 2023-02-14
artifacts: set up prow authentication to S3


justinsb:
We need to set up some IAM roles / federation so that prow can authenticate to S3, so that it can copy artifacts to the mirror buckets.

/area artifacts

justinsb:
We have a proposed PR here: https://github.com/cncf-infra/aws-infra/pull/21, which is a copy of the configuration for registry.k8s.io (the configurations should be pretty similar)

It's not clear whether we should be doing the configuration in this repo or in cncf-infra/aws-infra - anyone know?  I'm happy to submit a version of the above PR to this repo if that's easier.

justinsb:
I also tried https://github.com/kubernetes/k8s.io/pull/4770, not sure whether that is the right place to do things now?

justinsb:
/assign

I think #4770 is a good way to go

justinsb:
The periodic job is running and prow is using identity federation to authenticate to AWS/S3

/close

ameukam:
Part of:
  - https://github.com/kubernetes/k8s.io/issues/4401

PR #4770, opened 2023-02-14
artifacts.k8s.io: IAM policy for prow promotion

justinsb:
Add IAM policy to artifacts.k8s.io, to recognize the trusted prow
cluster as a federated identity provider, and to give the k8s service
account permission to write to buckets.


justinsb:
Issue is #4763

PR #4779, opened 2023-02-15
artifacts.k8s.io DNS: create DNS record for certificate challenge

justinsb:
If we can provision a certificate here, we should be able to cut over
with no downtime.

### DIFF
--- a/dns/zone-configs/k8s.io._0_base.yaml
+++ b/dns/zone-configs/k8s.io._0_base.yaml
@@ -111,6 +111,12 @@ artifacts:
   type: A
   values:
   - 35.244.192.77
+
+# DNS challenge for issuing (transition) TLS certificate
+_acme-challenge.artifacts:
+  type: CNAME
+  value: c72842dd-afa6-4ce7-b722-546baf89701f.5.authorize.certificatemanager.goog.
+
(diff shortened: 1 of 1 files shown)


justinsb:
This is my attempt to create a Google-managed certificate so that we can switch to artifacts.k8s.io without downtime (hopefully).

Values can be verified by running this command:

```
> gcloud certificate-manager dns-authorizations describe artifacts-k8s-io --project k8s-infra-porche-prod
createTime: '2023-02-15T22:50:47.627435159Z'
dnsResourceRecord:
  data: c72842dd-afa6-4ce7-b722-546baf89701f.5.authorize.certificatemanager.goog.
  name: _acme-challenge.artifacts.k8s.io.
  type: CNAME
domain: artifacts.k8s.io
name: projects/k8s-infra-porche-prod/locations/global/dnsAuthorizations/artifacts-k8s-io
updateTime: '2023-02-15T22:50:48.797276846Z'
```

Because this is a managed certificate, we should not be able to get the key after provisioning (a good thing).

ameukam:
cc @dims @cblecker

BenTheElder (commented):
/lgtm
/approve

PR #4781, opened 2023-02-16
artifacts.k8s.io: add managed certificates

justinsb:
The module doesn't seem to be ready for managed certificates yet, but
it seems much clearer when it is explicitly inlined anyway.

### DIFF
--- a/infra/gcp/terraform/k8s-infra-porche-prod/certificate.tf
+++ b/infra/gcp/terraform/k8s-infra-porche-prod/certificate.tf
@@ -1,0 +1,51 @@
+/*
+Copyright 2023 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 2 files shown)


justinsb:
This escalated quickly!  Using the DNS challenge with managed certificates (so I can't see the key) requires me to use the new google_certificate_manager_certificate, but that doesn't work well with our serverless_negs module (it complains that we specify both ssl_certificates and certificate_map on google_compute_target_https_proxy).  Also trying to use the version of the module that supports managed certificates kicks off an upgrade storm which I think we don't want right now.

I figure the raw objects are actually more easy to understand than the module, albeit slightly more tedious.  But the line count is not ridiculous - 90 lines instead of 50 lines - to me that 's worth it.

justinsb:
And we can now do things like:

```
curl -L -v --resolve artifacts.k8s.io:443:34.110.216.12 https://artifacts.k8s.io/binaries/kops/1.26.0-beta.2/darwin/amd64/kops.sha256
```

and

```
curl -L -v --resolve artifacts.k8s.io:80:34.110.216.12 http://artifacts.k8s.io/binaries/kops/1.26.0-beta.2/darwin/amd64/kops.sha256
```

The IPv6 address is `2600:1901:0:a641::` if anyone is feeling fancy.

BenTheElder (commented):
/lgtm
/approve

ameukam:
I think we can merge this.
/approve

PR #4782, opened 2023-02-16
artifacts: Fix typo in FALLBACK_LOCATION env var

justinsb:
Is overridden in staging, but in prod we probably want the correct
value!


ameukam:
/lgtm
/approve
/hold

Remove the hold when ready.

PR #4793, opened 2023-02-16
AWS: Enable services

ameukam:
Enable SSO and Storage Lens at the org level.


ameukam:
cc @sftim

sftim:
OK, but we should document why we want these enabled.

PR #4794, opened 2023-02-17
artifacts.k8s.io: Switch to new infrastructure (porche)

justinsb:
Switch to the new redirector service for artifacts.k8s.io.

### DIFF
--- a/dns/zone-configs/k8s.io._0_base.yaml
+++ b/dns/zone-configs/k8s.io._0_base.yaml
@@ -105,12 +105,12 @@ releases:
   value: redirect.k8s.io.
 
 # artifacts.k8s.io is for the (under development) binary artifact serving.
-# The IP points to a GCLB in front of a GCS bucket,
-# in the k8s-artifacts-prod project.
+# The IP points to a GCLB in front of CloudRun in the k8s-infra-porche-prod project.
 artifacts:
   type: A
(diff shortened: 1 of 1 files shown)


justinsb:
/hold

Plan is to do this tomorrow morning

ameukam:
/lgtm
/approve

A follow-up will be to add a AAAA record for the global IPv6 address.

dns/zone-configs/k8s.io._0_base.yaml
```diff
@@ -110,7 +110,7 @@ releases:
 artifacts:
   type: A
   values:
-  - 35.244.192.77
+  - 34.110.216.12
```
ameukam, line 113: Do we need to define a short TTL to ensure DNS update is quickly propagated ?

justinsb: That's a good idea, if we can.  I put this through in case you didn't have DNS permissions, but it looks like you do.

On IPv6, I wasn't sure whether we wanted to change "two things at once", in case IPv6 broke some people... happy to do it at the same time though.

ameukam: For the TTL, you can add a `ttl` field according to https://github.com/octodns/octodns/blob/v0.9.9/docs/dynamic_records.md.

For IPv6 we can do it later.

PR #4795, opened 2023-02-17
Rollback artifacts.k8s.io

justinsb:
In case of emergency, this is the PR to rollback DNS for artifacts.k8s.io


BenTheElder:
/close
been a month now :-) 🎉

PR #4798, opened 2023-02-17
artifacts.k8s.io: stand up mirrors in 2 more regions

justinsb:
These regions are seeing some traffic:

* us-east-1
* eu-central-1


ameukam:
/lgtm
/approve
/hold

Remove the hold when ready.

justinsb:
Thanks, I think it should be good to go.  Still trying to deal with the backlog of kOps images, I think our current promoter does not like compressed images...  I might just remove them for the moment.

/hold cancel

PR #4801, opened 2023-02-17
Revert "artifacts: add backfill for old kOps binaries"

justinsb:
This reverts commit 6f7b4defd90790955b62ddcd5b9e375afbb02104.

These binaries were compressed (which we should investigate
supporting!), and our current promoter does not support compressed
binaries.


justinsb:
This should fix https://prow.k8s.io/?job=ci-k8sio-file-promo-mirrors again.  We don't need these old artifacts to be mirrored anyway, it was just an attempt to describe the world as it is.  We should probably support compression though!

PR #4802, opened 2023-02-17
Grant justinsb access to the logs behind k8s.gcr.io

justinsb:
I would like to investigate whether there is a slice of traffic we can
redirect from k8s.gcr.io.  Only need access for about a week.

I will keep PII confidential and I will not retain raw information
past Feb 28.


justinsb:
I'm not sure what the process is for getting access here, but I would like to see if there is a slice of the k8s.gcr.io traffic that we could safely redirect.

dims:
/approve
/lgtm

please file a revert which we can land later (we can keep it on `/hold`). thanks!

justinsb:
Thank you - pre-filed #4803

PR #4804, opened 2023-02-18
artifacts: create all the buckets, replicate to 3 more regions

justinsb:
We should precreate all the buckets.

Also add the next 3 top regions by observed traffic.


justinsb:
Thanks - starting the replication now!

/hold cancel

PR #4814, opened 2023-02-21
Adjust OIDC bindings for registry.k8s.io AWS account

upodroid:
/cc @ameukam 

Can you deploy this change please?

I copied over the iam_role config from https://github.com/cncf-infra/aws-infra/blob/main/terraform/iam/main.tf

I can't see the AWS infra so I can't tell what it looks like.

Line 67 is what I really care about.

### DIFF
--- a/infra/aws/terraform/registry.k8s.io/iam.tf
+++ b/infra/aws/terraform/registry.k8s.io/iam.tf
@@ -1,0 +1,110 @@
+/*
+Copyright 2023 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+# Recognize federated identities from the prow trusted cluster
+resource "aws_iam_openid_connect_provider" "k8s-infra-trusted-cluster" {
+  url             = "https://container.googleapis.com/v1/projects/k8s-infra-prow-build-trusted/locations/us-central1/clusters/prow-build-trusted"
+  client_id_list  = ["sts.amazonaws.com"]
+  thumbprint_list = ["08745487e891c19e3078c1f2a07e452950ef36f6"]
+}
+
+# s3writer iam role for artifacts management
+# We allow the kubernetes service account to assume this role
+resource "aws_iam_role" "registry-k8s-io-s3writer" {
+  name = "registry.k8s.io_s3writer"
+  assume_role_policy = jsonencode({
+    Version = "2012-10-17"
+    Statement = [
+      {
+        Action = "sts:AssumeRole"
+        Effect = "Allow"
+        Principal = {
+          AWS = "arn:aws:iam::768319786644:root"
+        }
+      },
+      {
+        Action = "sts:AssumeRole"
+        Effect = "Allow"
+        Principal = {
+          Service = "s3.amazonaws.com"
+        }
+      },
+      {
+        Action = "sts:AssumeRole",
+        Effect = "Allow",
+        Principal = {
+          Service = "batchoperations.s3.amazonaws.com"
+        },
+      },
+      {
+        Action = "sts:AssumeRole"
+        Effect = "Allow"
+        Principal = {
+          AWS = "arn:aws:iam::585803375430:user/registry.k8s.io-ci"
+        }
+      },
+      {
+        "Effect" : "Allow",
+        "Principal" : {
+          "Federated" : aws_iam_openid_connect_provider.k8s-infra-trusted-cluster.arn
+        },
+        "Action" : "sts:AssumeRoleWithWebIdentity",
+        "Condition" : {
+          "StringEquals" : {
+            "container.googleapis.com/v1/projects/k8s-infra-prow-build-trusted/locations/us-central1/clusters/prow-build-trusted:sub" : "system:serviceaccount:test-pods:k8s-infra-gcr-promoter"
+          }
+        }
+      }
+    ]
+  })
+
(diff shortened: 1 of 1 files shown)


infra/aws/terraform/registry.k8s.io/iam.tf
```diff
@@ -0,0 +1,110 @@
...
+    Version = "2012-10-17"
+    Statement = [
+      {
+        Action = "sts:AssumeRole"
+        Effect = "Allow"
+        Principal = {
+          AWS = "arn:aws:iam::768319786644:root"
+        }
+      },
+      {
```
sftim, line 38: Any risk of a confused deputy attack here? If so, what mitigations are we using?

infra/aws/terraform/registry.k8s.io/iam.tf
```diff
@@ -0,0 +1,110 @@
...
+          Service = "s3.amazonaws.com"
+        }
+      },
+      {
+        Action = "sts:AssumeRole",
+        Effect = "Allow",
+        Principal = {
+          Service = "batchoperations.s3.amazonaws.com"
+        },
+      },
```
sftim, line 51: Any risk of a confused deputy attack here? If so, what mitigations are we using?

infra/aws/terraform/registry.k8s.io/iam.tf
```diff
@@ -0,0 +1,110 @@
...
+resource "aws_iam_role" "registry-k8s-io-s3writer" {
+  name = "registry.k8s.io_s3writer"
+  assume_role_policy = jsonencode({
+    Version = "2012-10-17"
+    Statement = [
+      {
+        Action = "sts:AssumeRole"
+        Effect = "Allow"
+        Principal = {
+          AWS = "arn:aws:iam::768319786644:root"
```
sftim, line 35: Maybe we can look up that account ID, or take it from a `local`?

infra/aws/terraform/registry.k8s.io/iam.tf
```diff
@@ -0,0 +1,110 @@
...
+        Effect = "Allow",
+        Principal = {
+          Service = "batchoperations.s3.amazonaws.com"
+        },
+      },
+      {
+        Action = "sts:AssumeRole"
+        Effect = "Allow"
+        Principal = {
+          AWS = "arn:aws:iam::585803375430:user/registry.k8s.io-ci"
```
sftim, line 56: - I'd make this a list
- Where's this ARN from? Can we either look it up or take it as an input?
- Can we comment about the intent here?

infra/aws/terraform/registry.k8s.io/iam.tf
```diff
@@ -0,0 +1,110 @@
...
+      },
+      {
+        "Effect" : "Allow",
+        "Principal" : {
+          "Federated" : aws_iam_openid_connect_provider.k8s-infra-trusted-cluster.arn
+        },
+        "Action" : "sts:AssumeRoleWithWebIdentity",
+        "Condition" : {
+          "StringEquals" : {
+            "container.googleapis.com/v1/projects/k8s-infra-prow-build-trusted/locations/us-central1/clusters/prow-build-trusted:sub" : "system:serviceaccount:test-pods:k8s-infra-gcr-promoter"
```
sftim, line 67: I'd take the key-value pair here from an input variable or a `local`. We're hard coding various details here.

infra/aws/terraform/registry.k8s.io/iam.tf
```diff
@@ -0,0 +1,110 @@
...
+
+  name = "registry.k8s.io_s3writer_policy"
+  role = aws_iam_role.registry-k8s-io-s3writer.id
+
+  policy = jsonencode({
+    Version = "2012-10-17"
+    Statement = [
+      {
+        Action = [
+          "s3:*Object",
```
sftim, line 94: Watch out for future API calls, such as the (imaginary) `s3:DoExpensiveThingToObject`. It's better to be explicit.

Also, we're allowing `PutObject` and I'm not sure we need to. See https://docs.aws.amazon.com/AmazonS3/latest/userguide/setting-repl-config-perm-overview.html for relevant docs.

infra/aws/terraform/registry.k8s.io/iam.tf
```diff
@@ -0,0 +1,110 @@
...
+        "Condition" : {
+          "StringEquals" : {
+            "container.googleapis.com/v1/projects/k8s-infra-prow-build-trusted/locations/us-central1/clusters/prow-build-trusted:sub" : "system:serviceaccount:test-pods:k8s-infra-gcr-promoter"
+          }
+        }
+      }
+    ]
+  })
+
+  max_session_duration = 43200
```
sftim, line 74: Why this value? A comment will help explain.

infra/aws/terraform/registry.k8s.io/iam.tf
```diff
@@ -0,0 +1,110 @@
...
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+# Recognize federated identities from the prow trusted cluster
+resource "aws_iam_openid_connect_provider" "k8s-infra-trusted-cluster" {
+  url             = "https://container.googleapis.com/v1/projects/k8s-infra-prow-build-trusted/locations/us-central1/clusters/prow-build-trusted"
```
sftim, line 19: Where does this value come from? We repeat a similar value later, so maybe it could come from an input variable?

sftim (commented):
Thanks.

Here's my feedback on the code, based on what I understand right now.

infra/aws/terraform/registry.k8s.io/iam.tf
upodroid: @sftim Most of this is already deployed. My intention is that @ameukam will import the resource and then add one new Action to the role. This mode of access is no longer needed as humans never need to assume this role but it needs to be handled in a separate PR(One that examines the impact of changing the root accounts for this particular).

sftim:
@upodroid please add some details to the PR description so that the context is more clear to other contributors. We might refer back to this a long time from now.

PR #4824, opened 2023-02-22
AWS: Redefine delegated administrators

ameukam:
Redefine the delegate administrators for some services. According the [AWS
SRA](https://docs.aws.amazon.com/prescriptive-guidance/latest/security-reference-architecture/security-tooling.html) we should have:
- GuardDuty, CloudTrail, Config and Security Hub in the security eng account
- Identity Center is the infra-shared services account.
- Storage lens S3 and Access Analyzer in the Audit account.

### DIFF
--- a/infra/aws/terraform/management-account/delegated-admins.tf
+++ b/infra/aws/terraform/management-account/delegated-admins.tf
@@ -16,28 +16,42 @@ limitations under the License.
 
 # https://docs.aws.amazon.com/organizations/latest/userguide/orgs_integrate_services_list.html
 
-resource "aws_organizations_delegated_administrator" "config_multiaccount" {
-  account_id        = module.security_audit.account_id
-  service_principal = "config-multiaccountsetup.amazonaws.com"
-}
-
-resource "aws_organizations_delegated_administrator" "config" {
-  account_id        = module.security_audit.account_id
-  service_principal = "config.amazonaws.com"
-}
-
-resource "aws_organizations_delegated_administrator" "guardduty" {
-  account_id        = module.security_audit.account_id
-  service_principal = "guardduty.amazonaws.com"
-}
+# Audit Account
 
 resource "aws_organizations_delegated_administrator" "access_analyzer" {
   account_id        = module.security_audit.account_id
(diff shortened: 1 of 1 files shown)


sftim (commented):
Yep, makes sense.

/lgtm

infra/aws/terraform/management-account/delegated-admins.tf
```diff
@@ -16,28 +16,42 @@ limitations under the License.
...
+resource "aws_organizations_delegated_administrator" "guardduty" {
+  account_id        = module.security_engineering.account_id
+  service_principal = "guardduty.amazonaws.com"
 }
 
 resource "aws_organizations_delegated_administrator" "securityhub" {
-  account_id        = module.security_audit.account_id
+  account_id        = module.security_engineering.account_id
   service_principal = "securityhub.amazonaws.com"
 }
```
sftim, line 56: When we want to actually use this, it might be easiest to enable it via the AWS web console, once per region, and then have Security Hub take the responsibility for managing hubs within the organization.

We'd only need to access Security Engineering to do this, and the role we assume only need a few specific permissions.

deobieta: _If you are enrolled in AWS Organizations, the organization management account can designate a Security Hub administrator account in each Region._   [securityhub-restrictons-recommendations](https://docs.aws.amazon.com/securityhub/latest/userguide/securityhub-account-restrictions-recommendations.html)

So if we want to set this through Terraform we'll have to create an AWS  provider for each region and repeat "aws_organizations_delegated_administrator"  " securityhub" resource for each provider. Correct? 

I was looking at this other Terraform resource but I think it also has to be enabled for each region.  [aws_securityhub_organization_admin_account](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/securityhub_organization_admin_account#example-usage) 

Do we know what regions we want to use?

ameukam: @sftim Do you mind open an issue about Security Hub ? 

> Do we know which regions we want to use?
I would say all the regions enabled by default.

Issue #4746, opened 2023-02-09
AWS Identity Center Setup


upodroid:
Top-level problem statement: We need an Identity Provider(IdP) to use with AWS SSO to access all the various AWS accounts of the project.

In a typical company/enterprise, you often use either Google, Okta or Azure AD as your main IdP. When someone joins the company, HR/IT add this new employee to the IdP and they get access to various applications. Then employees are added to various groups that grant access to different systems. Also, when they leave the company, they get offboarded and their access gets revoked.

The Kubernetes project's primary IdP is Github as it has all our contributors(similar to employees) and their group memberships. GitHub isn't really designed as an IdP and apps that support GitHub as an IdP have custom code to process OIDC claims and group mappings.

For AWS Identity Center(previously known as AWS SSO), we need to use an IdP that supports SAML and SCIM to have a fully automated user access/creation for AWS. GitHub doesn't support these protocols/standards so we need a middleware to bridge the gap. We can explore Keycloak or Azure AD to plug this gap.

My personal preference would be to acquire an Azure AD tenant with many P1 licenses and write a small program(terraform works just fine) that creates AAD users and groups based on information in GitHub and we would then configure AWS SSO to use Azure AD as the IdP and create the permissions accordingly.

Keycloak can work but I need to test this. It doesn't support SCIM out of the box but someone wrote an external plugin that can do it.

Reading Material
- https://docs.aws.amazon.com/singlesignon/latest/userguide/provision-automatically.html#auto-provisioning-considerations This is about SCIM and autoprovisioning

Out of scope issues:
- Most guides online for enabling SSO for AWS often describe SAML for a **SINGLE** account. This is not scalable and will not be in scope.
- Federating GSuite is out of scope. Most of do not have @kubernetes.io Google accounts and the groups won't be useful for SAML.

I would like to hear more about any other IdPs that can be integrated directly with AWS Identity Center. This is the list. https://docs.aws.amazon.com/singlesignon/latest/userguide/supported-idps.html

/cc @jeefy @ameukam 

Next steps:
- I'm going to put together a Keycloak demo.
- The Azure AD approach is something I already do for prod at work. More than happy to talk about it.

/priority important-soon
/area infra/aws

ameukam:
We should evaluate also https://www.auth0.com as an authentication hub. They have a non-profit pricing: https://auth0.com/nonprofits

ameukam:
cc @mrbobbytables @cblecker @nikhita @palnabarun 
[...]

PR #4831, opened 2023-02-23
Initial Azure AD PoC

upodroid:
Fixes: #4746 

Accessing our AWS accounts and potential Azure subscriptions require access to a proper Identity Provider with support for SAML & SCIM to eliminate manual user provisioning. Azure AD meets that requirement rather well and I had access to some dev tenants to develop this.

The idea is rather simple, you would commit a line in to the csv file which creates an Azure AD user for you. You can use this user to login to AWS and access any accounts that you have permissions for.

We can use terraform to configure which groups should access which accounts via AWS APIs.

If we want to go forward with this, we'll need to create an Azure AD tenant for kubernetes.io and do one-off manual config of the Enterprise application and tenant.

The terraform code for the groups is a bit buggy, I need some help with that.

/cc @ameukam @jeefy

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,9 @@
 
 # Ignore vscode config files
 .vscode
+
+
+# Terraform
+.terraform
+terraform.tfstate
+terraform.tfstate.backup
--- a/infra/azure/azure-ad/users.tf
+++ b/infra/azure/azure-ad/users.tf
@@ -1,0 +1,49 @@
+locals {
+  domain_name = "k8s.borg.dev"
+  users       = csvdecode(file("./users.csv"))
+}
+
+resource "random_string" "random" {
+  length           = 16
+  special          = true
+}
+# Create users
+resource "azuread_user" "users" {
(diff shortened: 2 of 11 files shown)


.gitignore
```diff
@@ -4,3 +4,9 @@
 
 # Ignore vscode config files
 .vscode
+
+
+# Terraform
+.terraform
+terraform.tfstate
+terraform.tfstate.backup
```
sftim, line 12: I recommend separating this change into its own commit.

infra/azure/azure-ad/users.tf
```diff
@@ -0,0 +1,49 @@
...
+resource "azuread_user" "users" {
+  for_each = { for user in local.users : user.github_username => user }
+
+  user_principal_name = format(
+    "%s@%s",
+    each.value.github_username,
+    local.domain_name
+  )
+
+  password = uuid()
```
sftim, line 20: I recommend putting a comment about why this is OK.

PR #4833, opened 2023-02-24
artifacts.k8s.io: replicate to ap-northeast-1 region

justinsb:
This region seems to be about half of our remaining cache misses.


ameukam:
/lgtm
/approve
/hold

Remove the hold when ready.

PR #4834, opened 2023-02-24
AWS: ignore credits in the global budget.

ameukam:
The credits donated rebalance the AWS organization's monthly spend. Thus the defined threshold is never respected.
Also increase the percentage of the threshold


ameukam:
cc @dims

PR #4843, opened 2023-02-28
Initial Terraform configs for prow-build-cluster on AWS

xmudrii:
This PR adds initial Terraform configs for the `prow-build-cluster` EKS cluster. This cluster will be used for running test pods on AWS.

xref #4686 

The build cluster is provisioned in the following way:

- Using Terraform and appropriate AWS Community modules
  - This setup is recommended by Hahicorp: https://developer.hashicorp.com/terraform/tutorials/kubernetes/eks
- VPC is IPv4/IPv6 dual-stack with CIDR `10.0.0.0/16`
  - There are 3 * 3 subnets (public, private, intra, in all 3 availability zones)
  - Public subnets are connected to IGW
  - Private subnets are connected to a NAT Gateway
  - Intra subnets do not have connectivity to the Internet and external services
- EKS cluster
  - Currently running 1.23 in order to properly support Docker-in-Docker (DinD)
  - Uses EKS-optimized Ubuntu
  - Up to 4 `r5ad.2xlarge` instances (to be bumped up after getting limits increased)
  - AWS VPC CNI, CoreDNS, kube-proxy, EBS CSI driver, and Cluster Autoscaler are already deployed
  - Nodes' volumes are encrypted

This is the initial setup, we still need to deploy a few more components mentioned in #4668, as well as, to figure out how to get access to some GCP resources.

/assign @ameukam @upodroid

PR #4844, opened 2023-02-28
Increase subnets in prow-build-cluster

xmudrii:
As discussed on Slack, we figured out that `/16` VPC with `/22` subnets is insufficient because AWS VPC CNI assigns IP addresses to pods from those subnets. Additionally, some IP addresses are reserved for each instance (e.g. 30 IP addresses are preallocated for one `r5ad.2xlarge` instance).

We'll do the following:

- We'll try to use `10.0.0.0/12` VPC
- We'll use `/16` subnets

Each `/16` subnet has about 64k IP addresses. There is a subnet in each AZ, so we have three subnets in total. That gives us about 192k IP addresses in total across three AZs. That should be more than enough for running 300 nodes that we're planning to run.

We might not need to increase public and intra subnets to `/16`, but let's do that to be on the safe side.

This PR also increases IOPS and throughput to the maximum for `gp3` and enables Nitro Enclaves.

Ref: https://kubernetes.slack.com/archives/CCK68P2Q2/p1677588727123469

/assign @ameukam @pkprzekwas


ameukam:
/hold 

can you address https://github.com/kubernetes/k8s.io/pull/4843#discussion_r1120035962 ?

xmudrii:
@ameukam Sure, let me see.

infra/aws/terraform/prow-build-cluster/vpc.tf
```diff
@@ -21,20 +21,24 @@ limitations under the License.
 # VPC is IPv4/IPv6 Dual-Stack, but our cluster is IPv4 because EKS doesn't
 # support dual-stack yet.
 
+locals {
+  vpc_cidr = "10.0.0.0/8"
```
sftim, line 25: We might want to leave _some_ room. How about `10.128.0.0/9`, or a /10 even?

xmudrii: I think even `/12` should work. WDYT @ameukam @pkprzekwas?

infra/aws/terraform/prow-build-cluster/vpc.tf
```diff
@@ -21,29 +21,33 @@ limitations under the License.
 # VPC is IPv4/IPv6 Dual-Stack, but our cluster is IPv4 because EKS doesn't
 # support dual-stack yet.
 
+locals {
+  vpc_cidr = "10.0.0.0/12"
+}
```
sftim, line 26: This could still be an input variable, right?

xmudrii: It was that way before this, but I thought it might be safer this way. I don't have strong preferences.

ameukam: +1 on having this as a variable.

xmudrii: It's an input variable again.

infra/aws/terraform/prow-build-cluster/vpc.tf
ameukam: Let go with `10.128.0.0/10`.

xmudrii: @ameukam Updated, PTAL.

PR #4852, opened 2023-02-28
Grant s3-sync service account permissions to read buckets

upodroid:
/cc @ameukam 

This is another way to fix https://github.com/kubernetes/k8s.io/pull/4814

Instead of making AWS IAM changes for Kubernetes k8s-infra-gcr-promoter service account, we make GCP IAM changes to s3-sync Kubernetes service account.

This is the last infra change required for https://github.com/kubernetes/enhancements/issues/3720


infra/gcp/terraform/k8s-infra-prow-build-trusted/prow-build-trusted/resources/test-pods/test-pods-serviceaccounts.yaml
```diff
@@ -131,5 +131,8 @@ metadata:
 kind: ServiceAccount
 apiVersion: v1
 metadata:
+  # TODO(upodroid) remove this annotation once the k8s-infra-gcr-promoter k8s SA can assume an AWS IAM Role
```
sftim, line 134: Should we also track an issue?

PR #4859, opened 2023-03-02
prow-build-cluster: switch back to IPv4

xmudrii:
As discussed on the SIG K8s infra meeting held on 2023-03-01:

- We are switching back `prow-build-cluster` to IPv4 because:
  - we want to ensure the best compatibility and avoid any potential issues that can be caused by running workload in an IPv6-only cluster
  - there are issues with Ubuntu EKS optimized AMI and IPv6 and we don't want to spend too much time debugging and fixing that
- We are going to use one `/16` VPC with multiple CIDRs
  - `10.0.0.0/16` is going to used for **public** subnets
  - `10.1.0.0/16` is going to used for **private** subnets
  - `10.2.0.0/16` is going to used for **intra** subnets
- Each subnet is going to be `/18` giving us about 16k IP addresses **per availability zone**
  - This should be more than enough for our needs

Additionally:

- The cluster is upgraded to Kubernetes 1.25. We'll use containerd as a container runtime
  - Ubuntu AMI that we use doesn't seem to support Docker anyways, at least, it doesn't come with it installed
  - 1.23 is going to reach EOL at some point (October 2023) and we'll need to migrate anyways
  - I believe that Docker is not a requirement for DinD and I want to test that
    - GKE cluster is also running `ubuntu_containerd` image

/assign @ameukam @upodroid 
/hold
will merge the PR after reconciling changes

### DIFF
--- a/infra/aws/terraform/prow-build-cluster/eks.tf
+++ b/infra/aws/terraform/prow-build-cluster/eks.tf
@@ -30,18 +30,9 @@ module "eks" {
   # Manage aws-auth ConfigMap.
   manage_aws_auth_configmap = true
 
-  # We use IPv6 because we require a large number of nodes and pods.
-  # With IPv4, we can use only /16 VPC, which is a blocker for running
-  # 100+ nodes.
-  cluster_ip_family = "ipv6"
-
-  # We are using the IRSA created below for permissions
-  # However, we have to deploy with the policy attached FIRST (when creating a fresh cluster)
-  # and then turn this off after the cluster/node group is created. Without this initial policy,
-  # the VPC CNI fails to assign IPs and nodes cannot join the cluster
-  # See https://github.com/aws/containers-roadmap/issues/1666 for more context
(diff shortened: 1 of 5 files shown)


upodroid:
> I believe that Docker is not a requirement for DinD and I want to test that
> * GKE cluster is also running `ubuntu_containerd` image

Knative uses `cos_containerd` for prow-build clusters and we have many DinD jobs that work properly.

xmudrii:
I disabled `iam_role_attach_cni_policy` because it's only needed to create the cluster. @ameukam Please approve again. :)
/hold cancel

upodroid:
I'll submit a PR that allows prow control plane service accounts on GKE to assume AWS roles on this account.

@xmudrii Did you configure this yet?

https://docs.aws.amazon.com/eks/latest/userguide/add-user-role.html

xmudrii:
@upodroid Yes, I think this should be working.

Issue #4686, opened 2023-01-26
AWS: create build clusters with EKS


ameukam:
Now we have credits for 2023, we should investigate on moving some prowjobs to AWS. 

Create EKS build cluster(s) that matches existing [GKE clusters](https://github.com/kubernetes/test-infra/blob/9ca1de830646eb0c0fba6b8c4e6d396f07a88c02/prow/create-build-cluster.sh#L49-L52): 

- CPU (Intel|AMD): 8 (minimum)
- Memory: 52 GB (minimum)
- Local SSD disks: 2 (minimum)
- boot disk: 100 GB (minimum)
- OS: Ubuntu/Debian
- Container runtime: containerd

The EKS build clusters should also be able to sync secrets from [AWS Secret Manager](https://docs.aws.amazon.com/secretsmanager/latest/userguide/intro.html)

(I probably forget a few things. Will update the issue)
 
 /milestone v1.27
 /area infra
/area infra/aws
/priority important-soon

sftim:
Let's detail what addons / features we want (eg: IAM Roles For ServiceAccounts).

That can be a separate issue or even 1 issue per addon.

ameukam:
> Let's detail what addons / features we want (eg: IAM Roles For ServiceAccounts).
> 
> That can be a separate issue or even 1 issue per addon.

on top of my mind:
- private clusters (Public endpoint for the API server with private nodes)
- Node auto-scaling
- Integration with AWS Secret manager
- Integration with ALB
- Network Dual-Stack
- OS Kernel configuration
- AWS Nitro enclaves

sftim:
Maybe some of:
- `metrics-server`
- Prometheus (or as part of a bigger component)
- Node exporter
- Node problem reporter
- persistent storage
- AWS VPC network plugin (as an example)
- `kube-state-metrics`

ameukam:
> Maybe some of:
> 
> * `metrics-server`
> * Prometheus (or as part of a bigger component)
> * Node exporter
> * Node problem reporter
> * persistent storage
> * AWS VPC network plugin (as an example)
> * `kube-state-metrics`

LGTM. we can start with this list and expand the list of addons depending on the issues/needs we will face.

ameukam:
@jeefy would be great to have Kubermatic work on that.

xmudrii:
I'll be taking care of this next week.
/assign

ameukam:
[...]

PR #4874, opened 2023-03-06
prow-build-cluster: deploy AWS Load Balancer Controller

xmudrii:
As stated in #4686, our EKS build cluster should have ability to use ALB. This PR ensures that by deploying and configuring AWS Load Balancer Controller.

This change is already reconciled and tested.

/assign @ameukam

### DIFF
--- a/infra/aws/terraform/prow-build-cluster/kubernetes.tf
+++ b/infra/aws/terraform/prow-build-cluster/kubernetes.tf
@@ -40,6 +40,39 @@ module "metrics_server" {
   ]
 }
 
+# AWS Load Balancer Controller (ALB/NLB integration).
+resource "helm_release" "aws_lb_controller" {
+  name       = "aws-load-balancer-controller"
+  namespace  = "kube-system"
+  repository = "https://aws.github.io/eks-charts"
+  chart      = "aws-load-balancer-controller"
+  version    = "1.4.8"
+
+  set {
+    name  = "clusterName"
+    value = module.eks.cluster_name
+  }
+
+  set {
+    name  = "serviceAccount.create"
+    value = "true"
+  }
+
+  set {
+    name  = "serviceAccount.name"
+    value = "aws-load-balancer-controller"
+  }
+
+  set {
+    name  = "serviceAccount.annotations.eks\\.amazonaws\\.com/role-arn"
+    value = module.aws_load_balancer_controller_irsa.iam_role_arn
+  }
+
+  depends_on = [
+    module.eks
+  ]
+}
+
 # AWS Secrets Manager integration
 resource "helm_release" "secrets_store_csi_driver" {
   name       = "secrets-store-csi-driver"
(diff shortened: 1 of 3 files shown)


infra/aws/terraform/prow-build-cluster/kubernetes.tf
```diff
@@ -40,6 +40,39 @@ module "metrics_server" {
   ]
 }
 
+# AWS Load Balancer Controller (ALB/NLB integration).
+resource "helm_release" "aws_lb_controller" {
+  name       = "aws-load-balancer-controller"
+  namespace  = "kube-system"
```
sftim, line 46: I usually put the ALB load balanced controller into a less privileged namespace.

infra/aws/terraform/prow-build-cluster/kubernetes.tf
```diff
@@ -40,6 +40,39 @@ module "metrics_server" {
...
+  set {
+    name  = "serviceAccount.name"
+    value = "aws-load-balancer-controller"
+  }
+
+  set {
+    name  = "serviceAccount.annotations.eks\\.amazonaws\\.com/role-arn"
+    value = module.aws_load_balancer_controller_irsa.iam_role_arn
+  }
+
```
sftim, line 70: I'd also set:

- `resources{}`
- `priorityClassName`
- `image.repository`
- `podDisruptionBudget.minAvailable`

and explicitly manage via Terraform:

- `image.tag`

sftim (commented):
LGTM

We could tweak the values for the Helm chart.

infra/aws/terraform/prow-build-cluster/kubernetes.tf
xmudrii: Do you have recommended values for:

I'd also set:

- resources{}
- priorityClassName
- podDisruptionBudget.minAvailable (I guess 1 since we have 2 replicas)

> and explicitly manage via Terraform:
>
> `image.tag`

Do we really need to do so? Doesn't we get that implicitly since we bind the chart version? Or they can keep the same chart version but bump the app version?

sftim: For resources, try:
```
resources:
  limits:
    memory: 256Mi
  requests:
    cpu: 200m
    memory: 128Mi
```

sftim: What priority classes do we use in the existing cluster (on GCP)? I'd use one of them if we have some, and skip if we don't.

sftim: The image tag thing isn't essential. Setting it puts drift control even more in Terraform's hands.

TerryHowe: On most projects like this, I'd pin image tags.

PR #4881, opened 2023-03-07
Create nfs-server-provisioner google group and staging repo

kvaps:
Signed-off-by: Andrei Kvapil <kvapss@gmail.com>


infra/gcp/infra.yaml
```diff
@@ -318,6 +318,7 @@ infra:
       k8s-staging-multitenancy:
       k8s-staging-networking:
       k8s-staging-nfd:
+      k8s-staging-nfs-server-provisioner:
```
msau42, line 321: Thanks for digging into this! Could you double check if https://github.com/kubernetes/k8s.io/blob/main/k8s.gcr.io/images/k8s-staging-sig-storage/generate.sh#L38 already covers the repo?

kvaps: Yeah, you're right, thanks for pointing to it

/close

PR #4883, opened 2023-03-07
Add staging project for cloud provider for kind

ameukam:


ameukam:
cc @aojea @BenTheElder

aojea:
the ci complains about the name?

ameukam:
more specifically length of the project id. we need to drop at least 1 character.

groups/sig-testing/groups.yaml
```diff
@@ -282,3 +282,13 @@ groups:
     members:
     - eddiezane@gmail.com
     - k8s-infra-prow-oncall@kubernetes.io
+
+  - email-id: k8s-infra-staging-cloud-provider-kind@kubernetes.io
+    name: k8s-infra-staging-cloud-provider-kind
```
aojea, line 287: and if we drpo the `infra` ??

ameukam: The tooling we are using to create those container repos requires we put `k8s-infra-staging` as prefix.

We can use the existing `k8s-staging-kind` to push the images. Are intending to do promotion of the images built for this project?

aojea: I'm find using the kind bucket if there is no more concerns

> Are intending to do promotion of the images built for this project?

AFAIK the official images are pushed directly to dockerhub

ameukam: @BenTheElder Ok we use `k8s-staging-build` to host images for the cloud provider ?

BenTheElder: we have an existing project for kind, while we wanted two source repos for go module reasons I don't mind if both source repos share the staging project.

PR #4914, opened 2023-03-10
Promote kOps 1.26.0, 1.26.1 and 1.26.2 images

hakman:
/cc @justinsb @olemarkus @zetaab @rifelpet

PR #4915, opened 2023-03-10
Promote kOps 1.26.0, 1.26.1 and 1.26.2 binary artifacts

hakman:
/cc @justinsb @olemarkus @zetaab @rifelpet


justinsb:
As with my comment on #4914 this lgtm, I just don't know if we want to publish 1.26.0 and 1.26.1 if we're not going to recommend them.  For example, do you think we should do a github release for them, vs just jumping to 1.26.2?

/approve
/lgtm

/hold

hakman:
Ah, saw your PRs and approved them.

PR #4916, opened 2023-03-10
Promote kOps 1.26.0, 1.26.1 and 1.26.2 images

hakman:
/cc @justinsb @olemarkus @zetaab @rifelpet


justinsb:
Ah, just saw this after sending the 1.26.2 PR - sorry!.  Do you think we should promote the 1.26.0 and 1.26.1 if we have 1.26.2 ready?  In favor of promoting: more consistent to have each version available.  Argument to not promote: people should presumably not be using 1.26.0 or 1.26.1.

/approve
/lgtm

/hold

hakman:
I wanted to have this available in GCB only, not actually create GitHub releases for them.
If you think it's not useful, I can remove the .0 and .1 from the PRs.

hakman:
Ah, saw your PRs and approved them.

PR #4922, opened 2023-03-10
deploy registry.k8s.io v0.2.0

BenTheElder:
picks up fix for https://github.com/kubernetes/registry.k8s.io/issues/162 and latest IP data

I'm also switching to deploy from staging (we already deploy pinned by digest) to enable us to ship fixes faster to this production service (without blocking on image promotion) and more closely align our staging and production deployments.

Pulls from cloud run are GCP <=> GCP either way, and we were only using a single AR prod region and not pulling circularly through the registry.k8s.io frontend anyhow.


ameukam:
/lgtm
/hold

Remove the hold when ready.

BenTheElder:
`terraform apply`ing

BenTheElder:
This is live. so far so good

PR #4930, opened 2023-03-13
Deploy ESO on EKS

upodroid:
/cc @dims @xmudrii @ameukam @BenTheElder 

This PR deploys ESO on EKS using Workload Identity Federation. Marko will need to initially apply this manifest, but I intend on creating a prowjob that applies it automatically. We should probably consider using ArgoCD to deploy complex apps bundled as helm charts.

I'll fix the secrets later as I need to upgrade them to new format expected by ESO

PR #4943, opened 2023-03-15
Load secrets via ESO for eks build clusters

upodroid:
/cc @xmudrii @ameukam 

I wonder what the secret contents look like. You might have to do something like this if they are base64 encoded and stored as json:

```
apiVersion: external-secrets.io/v1beta1
kind: ExternalSecret
metadata:
  name: gcs-upload
  namespace: test-pods
spec:
  refreshInterval: 10m
  secretStoreRef:
    kind: ClusterSecretStore
    name: knative-tests
  target:
    creationPolicy: Owner
    template:
      type: Opaque
      engineVersion: v2
      data:
        service-account.json: "{{ .test_account | toString | b64dec}}" # Used by gcsupload decorator
  data:
  - secretKey: test_account
    remoteRef:
      key: prow-google-service-accounts
      property: test_account
```

### DIFF
--- a/infra/aws/terraform/prow-build-cluster/resources/test-pods/externalsecrets.yaml
+++ b/infra/aws/terraform/prow-build-cluster/resources/test-pods/externalsecrets.yaml
@@ -4,78 +4,114 @@
 # NOTE !!!
 # THIS CLUSTER USES EXTERNAL SECRETS OPERATOR INSTEAD OF KUBERNETES EXTERNAL SECRETS.
 # The CRDs are different from other prow clusters. Sample guide that explains the diff https://wiki.cac.washington.edu/display/MCI/Transition+from+KES+to+ESO
-# ---
-# apiVersion: kubernetes-client.io/v1
-# kind: ExternalSecret
-# metadata:
-#   name: service-account             # The name of the Kubernetes Secret
-#   namespace: test-pods
-# spec:
-#   backendType: gcpSecretsManager
-#   projectId: k8s-infra-prow-build
-#   data:
-#   - key: prow-build-service-account # The name of the GSM Secret
-#     name: service-account.json      # The key to write to in the Kubernetes Secret
-#     version: latest                 # The version of the GSM Secret
-# ---
-# apiVersion: kubernetes-client.io/v1
-# kind: ExternalSecret
-# metadata:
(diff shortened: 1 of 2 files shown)


upodroid:
@ameukam Did you deploy the infra changes from #4930 ?

infra/aws/terraform/prow-build-cluster/resources/test-pods/externalsecrets.yaml
```diff
@@ -4,78 +4,114 @@
...
-#     name: aws-ssh-private                                                        # The key to write to in the Kubernetes Secret
-#     version: latest                                                              # The version of the GSM Secret
-#     property: aws-ssh-private
-#   - key: gke_k8s-prow-builds_us-central1-f_prow__test-pods__aws-ssh-key-secret   # The name of the GSM Secret
-#     name: aws-ssh-public                                                         # The key to write to in the Kubernetes Secret
-#     version: latest                                                              # The version of the GSM Secret
-#     property: aws-ssh-public
+---
+apiVersion: external-secrets.io/v1beta1
+kind: ClusterSecretStore
```
ameukam, line 9: We only want to sync secrets in a specific namespace. we should scope this to the namespace.

xmudrii:
@upodroid I noticed that sync time for ExternalSecrets is 1 hour. That seems too long, AFAIK, it's 10 seconds for the Prow control plane cluster. Should we reduce that to like 10 minutes at least?

infra/aws/terraform/prow-build-cluster/resources/test-pods/externalsecrets.yaml
```diff
@@ -4,78 +4,114 @@
...
+  data:
+  - secretKey: service-account.json # The key to write to in the Kubernetes Secret
+    remoteRef:
+      key: prow-build-service-account # The name of the GSM Secret
+      version: latest                 # The version of the GSM Secret    
+---
+apiVersion: external-secrets.io/v1beta1
+kind: ExternalSecret
+metadata:
+  name: ssh-key-secret
```
ameukam, line 38: We don't need this SSH key. This is needed for e2e tests on GCP. we may need to create a new key pair on EC2.

upodroid:
Yes

PR #4949, opened 2023-03-15
registry.k8s.io: add more AR regions

ameukam:
Add additional regions:
 - asia-east2 (Hong Kong)
 - europe-west3 (Frankfurt)


ameukam:
cc @dims

PR #4982, opened 2023-03-18
fix bug in generate-role-yaml.sh: return can only be used in sourced …

BenTheElder:
…bash or in a function


BenTheElder:
still trying to figure out what version of yq we support and how best to install it ...

BenTheElder:
script from feb 26th 2021 https://github.com/kubernetes/k8s.io/commit/e4da8547f064b9a20adbc745e13949a0ecdb1e56 

sets yq to 2.12.0 in requirements.txt which never changed since them.

yq is currently written in go, so ... that's odd? probably it was rewritten at some point?

we should probably rework this to not depend on a super old yq in system path. recent yq causes this to drop everything from the output files.

BenTheElder:
anyhow `pip3 install yq==2.12.0` for now (and make sure that's what winds up in your path instead of a current install)

Issue #4981, opened 2023-03-18
Restrict log viewing for k8s-infra-gcp-auditors


BenTheElder:
This role is clearly intended to allow visibility into what infrastructure we have on a non-sensitive, read-only basis.
The documentation suggests it should be relatively open to join.

Currently this permits viewing all logs in all projects. We have projects that are too sensitive for that.

Off the top of my head https://registry.k8s.io production (k8s-infra-oci-proxy-prod) and probably AAA cluster (slack automation) logs should require specific, granular permission.

We should either remove log viewing and related roles that allow seeing any sort of service / cluster / loadbalancer logs from this role and restrict it to viewing what assets we have, or else we should figure out how to enable per-project restricting logging to more specific groups.

Thankfully the _current_ set of members can be relatively trusted here, but this violates the intent of the group.

I think removing log access is reasonable. 
People should join project specific groups for access to logs, and some of these groups should be small and highly restricted due to PII.

https://github.com/kubernetes/k8s.io/blob/3874ef7b852c0d9de991ad7920ed36686217cb66/groups/sig-k8s-infra/groups.yaml#L232-L246

sftim:
> removing log access

Perhaps with a long term ambition to provide canned log access, maybe allow querying Cloud Audit Logs for elevation-of-privilege actions, that kind of thing. Definitely access to historic security finding records (if there are any).

BenTheElder:
I think it makes sense to escalate permissions for more granular log access than organization wide. IE joining groups for specific sensitive projects.

Unfortunately we've made this rather complex and with limited bandwidth I'm not super confident in the PRs to update this yet.

So for now I think we need to actually be very restrictive about membership in this group, since it may have PII access, and not actually openly permit joining.

PR #4984, opened 2023-03-18
regenerate roles

BenTheElder:
This PR is just `infra/gcp/bash/roles/generate-role-yaml.sh`

We probably need to check over the output.

Also not sure how we've actually been applying these.

Doing this before https://github.com/kubernetes/k8s.io/pull/4983 where I'd actually change the source spec, because regenerating with no changes to the inputs creates this large diff already.


BenTheElder:
/hold

probably not the best time to change org roles.

I'd like to see us resolve https://github.com/kubernetes/k8s.io/issues/4981 soon though.

PR #4989, opened 2023-03-20
eks-prow-build-cluster: fix root volume for node groups and setup IAM role with admin access

xmudrii:
This PR brings two improvements to the eks-prow-build-cluster:

- Root volume is now properly defined, so Terraform is going to modify the default root volume instead of creating a new one
- Create an IAM role with the administrator privileges
  - The idea is that IAM users are supposed to assume that role instead of giving permissions to each IAM user
  - That IAM role is also specified in the aws-auth ConfigMap
  - That IAM role is also specified as service user for the KMS key so it can be used with Terraform
  - Additionally, that IAM role is Administrator in AWS, so it can be used to run Terraform
  - The only problem with this IAM role is upon creating the cluster for the first time. In that case, assume role must be commented in `main.tf` for the first Terraform run

/hold for discussing
/assign @ameukam @upodroid

### DIFF
--- a/infra/aws/terraform/prow-build-cluster/eks.tf
+++ b/infra/aws/terraform/prow-build-cluster/eks.tf
@@ -112,8 +112,10 @@ module "eks" {
       enable_monitoring = true
 
       block_device_mappings = {
-        xvda = {
-          device_name = "/dev/xvda"
+        # This must be sda1 in order to match the root volume,
+        # otherwise a new volume is created.
+        sda1 = {
+          device_name = "/dev/sda1"
           ebs = {
             volume_size           = var.node_volume_size
             volume_type           = "gp3"
--- a/infra/aws/terraform/prow-build-cluster/eks.tf
+++ b/infra/aws/terraform/prow-build-cluster/eks.tf
@@ -32,13 +32,41 @@ module "eks" {
 
   # Configure aws-auth
   aws_auth_roles = [
+    # Allow access to the Prow-EKS-Admin IAM role (used by Prow directly).
     {
       "rolearn"  = aws_iam_role.eks_admin.arn
       "username" = "eks-admin"
       "groups" = [
-        "system:masters"
+        "eks-prow-cluster-admin"
       ]
-    }
+    },
+    # Allow access to the Prow-Cluster-Admin IAM role (used with assume role with other IAM accounts).
+    {
+      "rolearn"  = aws_iam_role.iam_cluster_admin.arn
+      "username" = "eks-cluster-admin"
+      "groups" = [
+        "eks-cluster-admin"
+      ]
+    },
+  ]
+  # Allow EKS access to the root account.
+  aws_auth_users = [
+    {
+      "userarn"  = local.root_account_arn
+      "username" = "root"
+      "groups" = [
+        "eks-cluster-admin"
+      ]
+    },
+  ]
+
+  # Allow access to the KMS key used for secrets encryption to the root account.
+  kms_key_administrators = [
+    local.root_account_arn
+  ]
+  # Allow service access to the KMS key to the Prow-Cluster-Admin role.
+  kms_key_service_users = [
+    aws_iam_role.iam_cluster_admin.arn
   ]
 
   # We use IPv4 for the best compatibility with the existing setup.
--- a/infra/aws/terraform/prow-build-cluster/main.tf
+++ b/infra/aws/terraform/prow-build-cluster/main.tf
@@ -20,6 +20,14 @@ limitations under the License.
 
 provider "aws" {
   region = var.cluster_region
+
+  # We have a chicken-egg problem here. This role is not going to exist
+  # when creating the cluster for the first time. In that case, this must
+  # be commented, than uncommented afterwards.
+  assume_role {
+    role_arn     = "arn:aws:iam::468814281478:role/Prow-Cluster-Admin"
+    session_name = "prow-build-cluster-terraform"
+  }
 }
 
 provider "kubernetes" {
@@ -30,7 +38,7 @@ provider "kubernetes" {
   exec {
     api_version = "client.authentication.k8s.io/v1beta1"
     command     = "aws"
-    args        = ["eks", "get-token", "--cluster-name", module.eks.cluster_name]
+    args        = ["eks", "get-token", "--cluster-name", module.eks.cluster_name, "--role-arn", aws_iam_role.iam_cluster_admin.arn]
   }
 }
 
@@ -43,7 +51,7 @@ provider "helm" {
     exec {
       api_version = "client.authentication.k8s.io/v1beta1"
       command     = "aws"
-      args        = ["eks", "get-token", "--cluster-name", module.eks.cluster_name]
+      args        = ["eks", "get-token", "--cluster-name", module.eks.cluster_name, "--role-arn", aws_iam_role.iam_cluster_admin.arn]
     }
   }
 }
@@ -52,6 +60,8 @@ data "aws_caller_identity" "current" {}
 data "aws_availability_zones" "available" {}
 
 locals {
+  root_account_arn = "arn:aws:iam::${data.aws_caller_identity.current.account_id}:root"
+
   tags = {
     Cluster = var.cluster_name
   }
--- a/infra/aws/terraform/prow-build-cluster/terraform.tfvars
+++ b/infra/aws/terraform/prow-build-cluster/terraform.tfvars
@@ -30,9 +30,9 @@ node_instance_types = ["r5ad.4xlarge"]
 node_volume_size    = 100
 
 # TODO(xmudrii): Increase this later.
-node_min_size                   = 15
-node_max_size                   = 20
-node_desired_size               = 15
+node_min_size                   = 5
+node_max_size                   = 5
+node_desired_size               = 5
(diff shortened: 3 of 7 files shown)


infra/aws/terraform/prow-build-cluster/main.tf
```diff
@@ -20,6 +20,13 @@ limitations under the License.
 
 provider "aws" {
   region = var.cluster_region
+
+  # We have a chicken-egg problem here. This role is not going to exist
+  # when creating the cluster for the first time. In that case, this must
+  # be commented, than uncommented afterwards.
+  assume_role {
+    role_arn = "arn:aws:iam::468814281478:role/Prow-Cluster-Admin"
```
pkprzekwas, line 28: Is it possible to use `aws_iam_role.iam_cluster_admin.arn` here?

infra/aws/terraform/prow-build-cluster/eks.tf
```diff
@@ -133,3 +163,29 @@ module "eks" {
...
+  name        = "Prow-Cluster-Admin"
+  description = "IAM role used to delegate access to prow-build-cluster"
+
+  assume_role_policy = jsonencode({
+    Version = "2012-10-17"
+    Statement = [
+      {
+        "Effect" : "Allow",
+        "Principal" : {
+          "AWS" : "468814281478"
```
pkprzekwas, line 178: It would be nice to have `468814281478` in a named variable.

ameukam: we have `root_account_arn` defined below as a variable but I would prefer we use the data resource [caller_identity](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/data-sources/caller_identity).

infra/aws/terraform/prow-build-cluster/terraform.tfvars
```diff
@@ -30,9 +32,9 @@ node_instance_types = ["r5ad.4xlarge"]
 node_volume_size    = 100
 
 # TODO(xmudrii): Increase this later.
-node_min_size                   = 15
-node_max_size                   = 20
-node_desired_size               = 15
+node_min_size                   = 5
+node_max_size                   = 5
+node_desired_size               = 5
```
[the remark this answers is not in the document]
xmudrii: The reason why reducing is to make rolling out easier, as I'm doing that more frequently now (to make the SSDs work). Will be reverted later.

infra/aws/terraform/prow-build-cluster/eks.tf
```diff
@@ -32,13 +32,41 @@ module "eks" {
...
+      ]
+    },
+  ]
+  # Allow EKS access to the root account.
+  aws_auth_users = [
+    {
+      "userarn"  = var.root_account_arn
+      "username" = "root"
+      "groups" = [
+        "system:masters"
```
sftim, line 58: Are we sure about this? I tend to not trust the root principal for any use.
Perhaps comment about why it's OK here.

infra/aws/terraform/prow-build-cluster/eks.tf
```diff
@@ -32,13 +32,41 @@ module "eks" {
...
         "system:masters"
       ]
-    }
+    },
+    # Allow access to the Prow-Cluster-Admin IAM role (used with assume role with other IAM accounts).
+    {
+      "rolearn"  = aws_iam_role.iam_cluster_admin.arn
+      "username" = "eks-cluster-admin"
+      "groups" = [
+        "system:masters"
```
sftim, line 48: Can we provide access less than `system:masters`?

infra/aws/terraform/prow-build-cluster/eks.tf
```diff
@@ -133,3 +163,29 @@ module "eks" {
...
+        "Condition" : {}
+      }
+    ]
+  })
+}
+
+# Give administrator access to the admin IAM role so it can be used with Terraform.
+resource "aws_iam_role_policy_attachment" "iam_policy_cluster_admin" {
+  role       = aws_iam_role.iam_cluster_admin.name
+  policy_arn = "arn:aws:iam::aws:policy/AdministratorAccess"
```
sftim, line 190: This is more access than is needed. Consider a permissions boundary or a more specific policy.

For example, this allows `organizations:LeaveOrganization`. A deployer should not need to be allowed to do so.

infra/aws/terraform/prow-build-cluster/main.tf
```diff
@@ -20,6 +20,13 @@ limitations under the License.
 
 provider "aws" {
   region = var.cluster_region
+
+  # We have a chicken-egg problem here. This role is not going to exist
+  # when creating the cluster for the first time. In that case, this must
+  # be commented, than uncommented afterwards.
+  assume_role {
```
sftim, line 27: Consider setting a session name.

infra/aws/terraform/prow-build-cluster/terraform.tfvars
```diff
@@ -14,6 +14,8 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
+root_account_arn = "arn:aws:iam::468814281478:root"
```
sftim, line 17: Better to look this up.

sftim (changes requested):

infra/aws/terraform/prow-build-cluster/eks.tf
```diff
@@ -32,13 +32,41 @@ module "eks" {
 
   # Configure aws-auth
   aws_auth_roles = [
+    # Allow access to the Prow-EKS-Admin IAM role (used by Prow directly).
     {
       "rolearn"  = aws_iam_role.eks_admin.arn
       "username" = "eks-admin"
       "groups" = [
         "system:masters"
```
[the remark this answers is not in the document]
xmudrii: Prow needs [cluster-admin permissions](https://docs.prow.k8s.io/docs/getting-started-deploy/#create-cluster-role-bindings). I guess we can eventually try to bind it to some SA instead, but that's something not relevant to this PR (i.e. we can do it in a follow up).

infra/aws/terraform/prow-build-cluster/eks.tf
xmudrii: This is to give access to the EKS options via the Web Console. We can later drop this if Web Console access is not needed, but at this stage, it can be useful. I'm not sure we can easily scope this down as I'm not sure what permissions are needed by the Web Console.

infra/aws/terraform/prow-build-cluster/eks.tf
xmudrii: Similar to as Prow, we can eventually bind to some SA that's cluster-admin.

infra/aws/terraform/prow-build-cluster/eks.tf
xmudrii: The problem here is how to determine what are needed permissions. This account should be able to deploy those Terraform configs, and it needs quite a lot of access, so going Administrator level made sense. If anyone has any idea how to properly collect needed permissions, it would be appreciated.

infra/aws/terraform/prow-build-cluster/eks.tf
xmudrii: If we want to go that route, I think we should create IAM user group and give access to it. Please note that this is currently supposed to be used by users manually (e.g. me and @pkprzekwas), eventually by some automation (but for automation I would go via IRSA).

infra/aws/terraform/prow-build-cluster/eks.tf
sftim: How about just granting read for now?

infra/aws/terraform/prow-build-cluster/eks.tf
sftim: Try https://skildops.com/blog/generate-restricted-aws-iam-policy-via-cloudtrail as a starting point.

infra/aws/terraform/prow-build-cluster/eks.tf
xmudrii: I changed this so only @pkprzekwas and I can assume this role.

infra/aws/terraform/prow-build-cluster/eks.tf
xmudrii: This is now using a ClusterRoleBinding/Group with cluster-admin access. We'll probably remove this after the POC phase, so I wouldn't spend too much time trying to scope down permissions.

infra/aws/terraform/prow-build-cluster/terraform.tfvars
TerryHowe:
```suggestion
root_account_arn = "arn:aws:iam::468814281478:user/root"
```

xmudrii: It doesn't work with `user/`, but I found a way to look up the account ID, so this comment should be addressed now.

infra/aws/terraform/prow-build-cluster/eks.tf
dims: +1 to merge and iterate

xmudrii:
Changes are already applied.
/hold cancel

Issue #4913, opened 2023-03-10
`ci-k8sio-file-promo-mirrors` error uploading - Access Denied


hakman:
The artifacts mirroring job is failing after promoting `ecr-credential-provider` for `cloud-provider-aws`:
https://prow.k8s.io/job-history/gs/kubernetes-jenkins/logs/ci-k8sio-file-promo-mirrors
```
********** FINISHED WITH ERRORS **********
error uploading "s3://test-artifacts-k8s-io-us-east-2/binaries/cloud-provider-aws/v1.26.1/linux/amd64/ecr-credential-provider-linux-amd64": AccessDenied: Access Denied
	status code: 403, request id: XNEQFB0VXK8XMFZ3, host id: 4/HTmTAVc2CPZ4YyWHOZsOtZmGJhbkW65oK/RGGqbbxTnMGIoegk0Fy+H59xMzK9bwBC/It2u+8=
```

dims:
looking at https://prow.k8s.io/view/gs/kubernetes-jenkins/logs/ci-k8sio-file-promo-mirrors/1634028450891173888

there's a lot more failing. seeing `promote-to-mirrors-staging-build-log.txt`:
```
********** FINISHED WITH ERRORS **********
error uploading "s3://test-artifacts-k8s-io-us-east-2/binaries/kops/1.25.4/darwin/amd64/kops": AccessDenied: Access Denied
	status code: 403, request id: 3JJD7ZS52G13FD4B, host id: 6kfa/FkAPaGiB/c6k5HxlBIWhEB1pchCSSZqdNtI+pNHRux66r8/qtmseWaNP/CZPtiydSoUZh4=
error uploading "s3://test-artifacts-k8s-io-us-east-2/binaries/kops/1.25.4/darwin/amd64/kops.sha256": AccessDenied: Access Denied
	status code: 403, request id: 14QXS6ZZG89AXW3S, host id: +EPeocBg9XRW5sH5ApnhrdZJH3bCm5fetjzuOgi3jpxC5YufHf25aiCKSm2o5ofiBVC9tih+A2Y=
error uploading "s3://test-artifacts-k8s-io-us-east-2/binaries/kops/1.25.4/darwin/arm64/kops": AccessDenied: Access Denied
[...]

PR #5018, opened 2023-03-23
Add IAM federation to test artifacts

justinsb:
We already have this configured in prod artifacts.


justinsb:
This should fix #4913

justinsb:
I rebased and repushed just to be sure, because it's been a minute...

PR #5035, opened 2023-03-27
Freeze the old `k8s.gcr.io` image registry

upodroid:
Fixes: https://github.com/kubernetes/release/issues/2947
Fixes: https://github.com/kubernetes/enhancements/issues/3720

Merging this PR will freeze the old registry.

/cc @kubernetes/release-engineering @dims @BenTheElder

### DIFF
--- a/registry.k8s.io/manifests/k8s-staging-addon-manager/promoter-manifest.yaml
+++ b/registry.k8s.io/manifests/k8s-staging-addon-manager/promoter-manifest.yaml
@@ -2,12 +2,6 @@
 registries:
 - name: gcr.io/k8s-staging-addon-manager
   src: true
-- name: us.gcr.io/k8s-artifacts-prod/addon-manager
-  service-account: k8s-infra-gcr-promoter@k8s-artifacts-prod.iam.gserviceaccount.com
-- name: eu.gcr.io/k8s-artifacts-prod/addon-manager
-  service-account: k8s-infra-gcr-promoter@k8s-artifacts-prod.iam.gserviceaccount.com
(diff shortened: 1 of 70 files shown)


ameukam:
> 2nd commit will add the new AR registries we created. @ameukam What are the new rules?

Not sure to understand the question but let's add the new AR regions in a different PR and probably after the redirect is done. the regional-equivalent buckets for those new AR repositories are still not created. 

I also suggest we merge this PR on April 3th morning CET.

ameukam:
> Correct. I meant regions instead of rules
See:
- https://github.com/kubernetes/registry.k8s.io/issues/172#issuecomment-1485206732

upodroid:
@puerco Did we manage to fix the missing sigstore layers in the backing registries?

puerco:
@upodroid  We merged some improvements to the image promoter which should allow it to run with the AR rate limit, this should stop the problem from getting bigger.

The missing sigantures are still missing. We haven't run the remediation jobs yet as we agreed to wait until the redirect rollout is done, or at least getting it green lighted when there is enough confidence enough that the redirect is still within its error budget.

TerryHowe:
Ready on this yet?

BenTheElder:
Yes, yesterday we lifted https://github.com/kubernetes/registry.k8s.io/issues/181, however the ROI is still lower than pre-redirect and many folks are away at KubeCon so it might not be the best week for this. I do still think we should do this.

sftim:
We're now post-KubeCon.

upodroid:
I rebased the PR and picked up the latest set of registries. This is ready now

BenTheElder (commented):
/lgtm
/approve

PR #5054, opened 2023-03-28
eks-prow-build-cluster: use instance SSDs for containerd and kubelet, and change instance type

xmudrii:
As discussed on Slack and previous SIG K8s Infra meetings, this PR is bringing two very important changes to the prow-build EKS cluster:

- We're now using instance NVMe SSDs for containerd and kubelet instead of EBS storage
  - NVMe SSDs are much faster than EBS storage. This ensures that our tests are faster and less flaky
  - This is done by using a pre-bootstrap script which picks the first SSD, creates a xfs filesystem, then mounts it to be used by containerd (`/var/lib/containerd`) and kubelet (`/var/lib/kubelet`)
  - The pre-bootstrap script is based on the script that AWS plans to include in their AL2 AMI: https://github.com/awslabs/amazon-eks-ami/pull/1171
  - This approach utilizes only the first SSD. If there are additional SSDs, those remain unused. I didn't want to fiddle with RAID because that would to some degree increase complexity, but we wouldn't gain much
- We're now using Intel-backed instances instead of AMD-backed instances
  - The Kubernetes unit tests are constantly flaking when running them on the AMD-backend instances
  - We don't know the exact root cause, but they might be very CPU intensive. The AWS docs mention the following regarding AMD-backend instances: `R5a and R5ad instances feature AMD EPYC 7000 series processors with an all core turbo clock speed of 2.5 GHz. The AMD-based instances provide additional options for customers that do not fully utilize the compute resources and can benefit from a cost savings of 10%.`
  - Based on that, we concluded that AMD processors have lower clock speed and might not be suitable for us
- We're now using again 20 instances considering that we have more jobs running in the cluster

Changes are already applied to the cluster. Some references: https://kubernetes.slack.com/archives/CCK68P2Q2/p1679920767006369

/assign @dims @ameukam

### DIFF
--- a/hack/verify-executable.sh
+++ b/hack/verify-executable.sh
@@ -31,6 +31,7 @@ readonly exec_regexes=(
 # files that should NOT be executable (exceptions to the regexes above)
 readonly noexec_regexes=(
     .*/lib.*\.sh # lib_foo.sh should be sourced by other scripts, not executed
+    infra/aws/terraform/prow-build-cluster/bootstrap/node_bootstrap.sh # this script is injected into another script, not executed
 )
 
 # convert arrays to extended regexes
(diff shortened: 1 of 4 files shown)


cartermckinnon (approved):

ameukam:
this has already been deployed.

/approve

PR #5057, opened 2023-03-29
make oci-proxy verbosity configurable

BenTheElder:
needed for deploying staging with production config

ref: https://github.com/kubernetes/registry.k8s.io/pull/210 (blocked primarily on this change)

PR #5067, opened 2023-04-03
move oci-proxy configuration to common module

BenTheElder:
So staging and production can match with explicit input deltas.

Future changes to production tf should be auto-applied to staging (working on that) and tested there before manually rolling out to production when ready.

This required upgrading to terraform 1.3.0 to support cross-module `moved`.

This diff is hard to read so I'll attach the `terraform plan` output to help confirm what actually changes in prod:

https://gist.github.com/BenTheElder/cf48cbec11719ddf9648c12b9c09e3e2

My read is that for each of the cloud run services it's setting the memory limit from 512mi=>null because that's a default we don't configure, and adding the `-v=0` to each service (which is the default, we're being explicit and allowing the `0` to be configured). 

`-v=0` is is new from the first commit, superseding #5057.

Note also at the end: "Plan: 0 to add, 20 to change, 0 to destroy." AKA the 20 cloud run services entries are changing because the verbosity flag / memory allocation, the rest are unchanged.

Next step would be porting k8s-infra-oci-proxy (staging) to consume the new common module and then in https://github.com/kubernetes/registry.k8s.io/pull/210 we'll sort out auto-deploying staging.


infra/gcp/terraform/modules/monitoring/uptime-alert/versions.tf
```diff
@@ -15,7 +15,7 @@ limitations under the License.
 */
 
 terraform {
-  required_version = "~> 1.2.0"
+  required_version = ">= 1.2.0"
```
ameukam, line 18: I think it's fine we stick to one major version:
```suggestion
  required_version = "~> 1.3.0"
```

we want to be consistent across the different projects.

infra/gcp/terraform/modules/oci-proxy/variables.tf
```diff
@@ -23,11 +23,9 @@ variable "project_id" {
 variable "digest" {
   type = string
 }
-variable "cloud_run_config" {
-  type = map(object({
-    environment_variables = list(object({
-      value = string
-      name  = string
-    }))
-  }))
```
ameukam, line 32: I was hoping we could reuse this module to deploy the sandbox env. By removing this, testing specific configs in the sandbox env is not straightforward.

infra/gcp/terraform/modules/monitoring/uptime-alert/versions.tf
BenTheElder: I didn't want to force the rest of the repo to upgrade yet.

infra/gcp/terraform/modules/oci-proxy/variables.tf
BenTheElder: It will be used to deploy the sandbox env. The sandbox env leads prod and should not skew on the cloud run config except explicit inputs. For example the verbosity is an explicit input to the cloud run config that varies between them.

You will be able to see the full diff by comparing the two root modules.

BenTheElder: https://github.com/BenTheElder/k8s.io/tree/module-staging / https://github.com/kubernetes/k8s.io/commit/a26bc4508c0995e0dc6eb8015487e6b266924752 cover staging

I split it out because this change is already large and that change is not quite complete (need to vet it carefully and write the tedious `moved` statements for everything).

But in particular staging vs prod:
https://github.com/kubernetes/k8s.io/blob/a26bc4508c0995e0dc6eb8015487e6b266924752/infra/gcp/terraform/k8s-infra-oci-proxy/main.tf

https://github.com/kubernetes/k8s.io/blob/a26bc4508c0995e0dc6eb8015487e6b266924752/infra/gcp/terraform/k8s-infra-oci-proxy-prod/main.tf

PR #5080, opened 2023-04-04
set oci-proxy memory limit explicitly

BenTheElder:
prevents spurious terraform plan differences flapping on null vs 512mi

this change has **zero** `terraform plan` deltas and does not need to be applied.

prior to this change, `terraform plan` would flap on null vs 512mi (default).

TODO: evaluate switching to second generation execution environment


BenTheElder:
/hold cancel
no-op change, makes future config deltas clearer

PR #5086, opened 2023-04-05
Promote v1.8.8 CPA images.

MrHohn:
Note that the version tag now comes with a prefix `v` to be consistent with the versioning of other components (e.g. CPVA).

Ref https://github.com/kubernetes-sigs/cluster-proportional-autoscaler/issues/138.


MrHohn:
I know folks are busy with kubecon these weeks. Adding assignees in case that brings better visibility.
/assign @thockin @bowei

thockin (commented):
Thanks!

/lgtm
/approve

PR #5087, opened 2023-04-05
update registry-sandbox DNS

BenTheElder:
New static IPs following reconciling converged terraform with production.

I think we actually had ephemeral before 😅 

https://console.cloud.google.com/networking/addresses/list?project=k8s-infra-oci-proxy


BenTheElder:
https://prow.k8s.io/view/gs/kubernetes-jenkins/logs/post-k8sio-dns/1643726674124083200

BenTheElder:
should be updated but will take a while to propagate

PR #5065, opened 2023-04-01
infra/aws: Tag Policy and Service Control Policy Terraform module

deobieta:
Ref: [#4684](https://github.com/kubernetes/k8s.io/issues/4684)

Enforcing tag policies and utilizing Service Control Policies for tagging at the time of resource creation has the potential to impact our current AWS resource creation practices. It is worth noting that certain AWS services lack the functionality to tag all resources through the AWS console UI during the creation process. For instance, while EC2 allows for the creation of a Security Group during instance creation, the option to tag the group is absent. Consequently, if we decide to enforce Security Groups tags, this may lead to instances failing to launch and an error message indicating insufficient authorization to create the resource. Similarly, ECR repositories do not offer the option to tag repositories during their creation, making ECR console UI ineffective to create repositories. It is essential to exercise caution when enabling tag policy enforcement and service control policies as their efficacy is highly dependent on the agreed upon AWS resource creation methodology.

Tag policy enforcement:  Only has effect on resources that are created with enforced tags. Non-compliant tagging requests on specified resource types are prevented from completing.

Service Control Policy: Enforce tagging at resource creation and deny tag deletion with IAM policies  (with RequestTag conditional).

### Create S3 resources from AWS console

| Resource | Tag policy Enforced | Service Control Policy | Without tags  | With random tag key and value | With tag key and wrong tag value | With correct key and value |
|---|---|---|---|---|---|---|
| bucket   | Yes | No | Succeded | Succeded | ***Succeded with error notice*** | Succeded |

### Delete/Edit Tags S3 resources

| Resource | Tag policy Enforced | Service Control Policy | Add a new random tag key / value  | Remove the random tag key / value | Remove the enforced tag | Edit enforced tag with wrong value
|---|---|---|---|---|---|---|
| bucket   | Yes | No | Succeded | Succeded | ***Succeded*** | Failed |

I couldn't find a way to enforce tags on bucket creation so the SCP is useless in this case.

### Create EC2 resources from AWS console (using "Launch instance" button) 

| Resource | Enforced | Service Control Policy | Without tags  | With random tag key and value | With tag key and wrong tag value | With correct key and value |
|---|---|---|---|---|---|---|
| instance | Yes | Yes | ***Failed*** | Succeded | ***Failed*** | Succeded |
| volume | Yes | Yes | ***Failed*** | Succeded | ***Failed*** | Succeded |

### Create EC2 resources from AWS console (using "Auto Scaling groups")

[...]

Issue #4684, opened 2023-01-26
AWS: Establish tag policies for the organization


ameukam:
Establish and implement tag policies across the organization.

- Tags at the org level
- Tags at the account level
- Tags at the ressource level

/area infra
/area infra/aws
/priority important-soon
/milestone v1.27

ameukam:
cc @deobieta @sftim

sftim:
What outcomes do we have / want in GCP land? Let's aim for equivalent outcomes on AWS.

ameukam:
We currently don't have a tags strategy on GCP so I think we can start with AWS and replicate/extand to GCP. I don't have specific outcomes of this. The ultimate goal will be to easily track costs across the entire organization (what's are the specific costs of a specific workload, service, OU, etc...)

deobieta:
Hello, I've been reading about tag policies and these are my findings: 

1. Tag policies only describe accepted values for tags and not their presence on resources. Resources with non-compliant tags can be fetched via the AWS Resource Groups console so they can be corrected using the console or tagging API operations of the AWS service, or the Resource Group Tag Editor console.
2.  Tag policies can be enforced on [certain resources](https://docs.aws.amazon.com/organizations/latest/userguide/orgs_manage_policies_supported-resources-enforcement.html). This means that non-compliant tagging requests on specified resource types are prevented from completing; however, enforcement does not affect untagged resources so users with the right IAM permissions could still create untagged resources or delete compliant tags.
3. To ensure that AWS resources are created with the appropriate tags and deny tag deletion we need to use Tag policies in conjunction with Service Control Policies.

Questions.

1. What tag strategy should we use?   Tag policies, Enforced Tag policies, Enforced Tag policies + SCPs.
2. Which resources should we enforce compliance?
3. What tag convention should we use?  For example, Costcenter, costcenter, or CostCenter.
4.  Which tags should we start using for cost management ? For example, Workload, Service, OU, SIG, WG, UG.

@ameukam Is it possible to [create an OU for policy staging](https://github.com/kubernetes/k8s.io/pull/4797) and add at least one AWS account to the OU for Tag policy and inheritance testing.

ameukam:
> What tag strategy should we use? Tag policies, Enforced Tag policies, Enforced Tag policies + SCPs.

I think we want to use Enforced tag policies + SCP so we are sure every resource created is tagged and properly tracked in the cost explorer.

> Which resources should we enforce compliance?

Every resource supporting tags.

[...]

PR #5097, opened 2023-04-06
infra/aws: Infra tests in policy-staging-account for tag policies

deobieta:
Ref: [#4684](https://github.com/kubernetes/k8s.io/issues/4684)

@ameukam 

This pull request includes Terraform code designed to test the efficacy of [AWS Tag Policies and Service Control Policies](https://github.com/kubernetes/k8s.io/pull/5065) which have been attached to our policy staging organizational unit (OU). As per inheritance, any policies enforced at this level should impact all child accounts nested within the OU.

Any tags that are removed or possess an invalid value from the code and are subjected to enforcement will result in an unauthorized error at creation time.

### DIFF
--- a/infra/aws/terraform/policy-staging-account-1/terraform.tf
+++ b/infra/aws/terraform/policy-staging-account-1/terraform.tf
@@ -1,0 +1,30 @@
+/*
+Copyright 2023 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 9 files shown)


infra/aws/terraform/policy-staging-account-1/terraform.tf
```diff
@@ -0,0 +1,30 @@
...
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+terraform {
+  backend "s3" {
+    bucket = "k8s-aws-root-account-terraform-state"
```
ameukam, line 19: why not use a bucket local to the policy staging account ? so we can deploy this yourself ? feel to create the bucket manually if needed. 

we want to test policies application.

deobieta: @ameukam I've changed the Terraform state bucket to use a bucket in the policy staging account as suggested.

PR #5099, opened 2023-04-06
oci-proxy: don't depend on project creation in tf

BenTheElder:
So we can auto apply infra config without over granting permissions

/hold

Requires `terraform state rm module.oci-proxy.google_project.project` from both staging and prod.

We could check in project creation somewhere else but ... it's just the project name / org as the inputs and we definitely don't want to delete the whole project so ...

other than `module.oci-proxy.google_project.project`, this has no plan diffs in prod or staging

### DIFF
--- a/infra/gcp/terraform/modules/oci-proxy/monitoring.tf
+++ b/infra/gcp/terraform/modules/oci-proxy/monitoring.tf
@@ -16,7 +16,7 @@ limitations under the License.
 
 resource "google_monitoring_notification_channel" "emails" {
   display_name = "k8s-infra-alerts@kubernetes.io"
-  project      = google_project.project.project_id
+  project      = var.project_id
   type         = "email"
   labels = {
     email_address = "k8s-infra-alerts@kubernetes.io"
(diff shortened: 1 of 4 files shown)


infra/gcp/terraform/modules/oci-proxy/monitoring.tf
```diff
@@ -16,23 +16,23 @@ limitations under the License.
...
+  project_id         = var.project_id
   source             = "../monitoring/uptime-alert"
   documentation_text = "${var.domain} is down"
   domain             = var.domain
 
   notification_channels = [
     # Manually created. Monitoring channels can't be created with Terraform.
     # See: https://github.com/hashicorp/terraform-provider-google/issues/1134
-    "${google_project.project.id}/notificationChannels/${var.notification_channel_id}",
+    "projects/${var.project_id}/notificationChannels/${var.notification_channel_id}",
```
BenTheElder, line 35: https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/google_project#id

> an identifier for the resource with format `projects/{{project}}`

I've also confirmed this is the value with `terraform state show` + there's no `terraform plan` diff.

BenTheElder: In other references we're just passing the project to terraform resources and either format works

ameukam:
Make sense.

/lgtm
/approve

BenTheElder:
ran `terraform state rm module.oci-proxy.google_project.project`  and confirmed no plan changes for both prod and staging afterwards.
/hold cancel

PR #5112, opened 2023-04-10
Onboard new contribex leads

nikhita:
Ref: https://github.com/kubernetes/community/issues/7246

/assign @mrbobbytables


MadhavJivrajani:
Do we also want to update
https://github.com/kubernetes/k8s.io/blob/c236d672f4447bad423aa8e88d260406ef4dfad2/OWNERS#L5-L7

?

groups/groups.yaml
```diff
@@ -106,8 +108,9 @@ groups:
       - natalivlatko@gmail.com
       - neolit123@gmail.com
       - nospam.wong@gmail.com
-      - pal.nabarun95@gmail.com #CoCC
+      - pal.nabarun95@gmail.com
       - patrick.ohly@intel.com
+      - priyankasaggu11929@gmail.com>
```
mrbobbytables, line 113:
```suggestion
      - priyankasaggu11929@gmail.com
```
trailing `>`

mrbobbytables (commented):
Other thing I just noticed...you're not actually on the leads email list O_o

nikhita:
> Other thing I just noticed...you're not actually on the leads email list O_o

I'm listed as a `manager` for the list :) 

> Do we also want to update k8s.io/OWNERS?

@MadhavJivrajani IIRC contribex leads being included there was due to historic reasons when this repo was part of WG K8s Infra and was technically "owned" by SIG Contribex. I don't think we should add all contribex TLs to root OWNERS by default.

@dims root OWNERS might need a refresh.

mrbobbytables:
> I'm listed as a `manager` for the list :)

ah XD woops

/lgtm

PR #5125, opened 2023-04-13
Promote AWS CCM image 1.27.0

olemarkus:


ConnorJC3:
/lgtm

I don't mind approving this as well, but if possible please have somebody from the AWS CCM chime in here to confirm the image is as expected, as I don't normally contribute to it and thus don't know what verification procedures are normally done for CCM images.

Issue #5127, opened 2023-04-13
AWS: migrate kOps away from CNCF account


ameukam:
All kOps e2e tests currently run under the CNCF management account.

There should be migrated to the AWS Kubernetes organization. 

/assign @ameukam @justinsb 
/area aws
/area infra/aws
/priority important-soon
/milestone v1.28

ameukam:
A public bucket is needed otherwise we will get:

```
error writing s3://k8s-kops-ci-prow-sandbox/discovery/e2e-kops-eks-sandbox-cilium-deb12.tests-kops-aws.k8s.io/.well-known/openid-configuration (with ACL="public-read"): AccessDenied: Access Denied
```

I created a new one with the aws cli using: https://kops.sigs.k8s.io/getting_started/aws/#cluster-oidc-store

```bash
$aws s3api create-bucket \
    --bucket k8s-kops-ci-prow-sandbox \
    --region us-east-2 \
    --object-ownership BucketOwnerPreferred \
    --create-bucket-configuration LocationConstraint=us-east-2
$aws s3api put-public-access-block \
    --bucket k8s-kops-ci-prow-sandbox \
    --public-access-block-configuration BlockPublicAcls=false,IgnorePublicAcls=false,BlockPublicPolicy=false,RestrictPublicBuckets=false
$aws s3api put-bucket-acl \
    --bucket k8s-kops-ci-prow-sandbox \
    --acl public-read
```

rifelpet:
@ameukam with all Kops jobs migrated to the new prow clusters, can we close this out or is there additional work to be done?

ameukam:
@rifelpet migrate the prowjobs was needed to unblock the migration of the prow control plane. Second phase is about:
- migrate away from `s3://k8s-kops-prow` 
- rotate or drop the usage of the AWS static credentials for the E2E tests

BenTheElder:
> rotate or drop the usage of the AWS static credentials for the E2E tests

is that directly related to moving to the kubernetes specific accounts?

ameukam:
> > rotate or drop the usage of the AWS static credentials for the E2E tests
> 
> is that directly related to moving to the kubernetes specific accounts?

if you mean the AWS specific accounts, yes. we need to migrate to only one k8s-owned AWS account (kops-infra-ci). Currently we use the CNCF account.

ameukam:
Currently blocked by:
[...]

PR #5149, opened 2023-04-19
AWS: Add a DNS zone for kops tests

ameukam:
Ref: https://github.com/kubernetes/k8s.io/issues/5127

Add a Route53 public zone used by kops tests.


ameukam:
Talked with @justinsb. It's not required at the moment.

/close

PR #5163, opened 2023-04-25
eks-prow-build-cluster: Replace Terraform workspaces with dedicated backends

xmudrii:
We want to use two different AWS accounts for production and canary setups. The accounts are already created and configured, but we need to move canary cluster to the new account.

Our Terraform scripts are using [the workspaces concept](https://developer.hashicorp.com/terraform/cli/workspaces) right now. This worked well so far, however, workspaces are not suitable (as per the Terraform docs) when:

* there are multiple backends
    > Workspaces let you quickly switch between multiple instances of a single configuration within its single backend.
* multiple accounts are used
    > We recommend using [alternative approaches](https://developer.hashicorp.com/terraform/cli/workspaces#alternatives-to-workspaces) for complex deployments requiring separate credentials and access controls.

That said, we should replace workspaces with something else. This PR replaces workspaces with dedicated backends. We already had this before, but we decided to use workspaces because we had a single account for both clusters at that time. Now that we have different accounts, let's give dedicated backends another try.

/assign @pkprzekwas @ameukam @dims

### DIFF
--- a/infra/aws/terraform/prow-build-cluster/terraform.canary.tfvars
+++ b/infra/aws/terraform/prow-build-cluster/terraform.canary.tfvars
@@ -14,6 +14,7 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
+aws_account_id = "054318140392"
 eks_admins = [
   "pkprzekwas",
   "xmudrii"
--- a/infra/aws/terraform/prow-build-cluster/terraform.canary.tfvars
+++ b/infra/aws/terraform/prow-build-cluster/terraform.canary.tfvars
@@ -21,6 +21,7 @@ eks_admins = [
 
 prow_build_cluster = false
 
+cluster_name               = "prow-build-canary-cluster"
 cluster_version            = "1.25"
 cluster_autoscaler_version = "v1.25.0"
 
--- a/infra/aws/terraform/prow-build-cluster/terraform.canary.tfvars
+++ b/infra/aws/terraform/prow-build-cluster/terraform.canary.tfvars
@@ -21,8 +21,6 @@ eks_admins = [
 
 prow_build_cluster = false
 
-cluster_name               = "prow-build-canary-cluster"
-cluster_region             = "us-east-2"
 cluster_version            = "1.25"
 cluster_autoscaler_version = "v1.25.0"
 
--- a/infra/aws/terraform/prow-build-cluster/terraform.canary.tfvars
+++ b/infra/aws/terraform/prow-build-cluster/terraform.canary.tfvars
@@ -15,7 +15,7 @@ limitations under the License.
 */
 
 eks_admins = [
-  "pprzekwa",
+  "pkprzekwas",
   "xmudrii"
 ]
 
--- a/infra/aws/terraform/prow-build-cluster/terraform.canary.tfvars
+++ b/infra/aws/terraform/prow-build-cluster/terraform.canary.tfvars
@@ -19,18 +19,12 @@ eks_admins = [
   "xmudrii"
 ]
 
-assume_role                 = true
-deploy_kubernetes_resources = true
+prow_build_cluster = false
 
-cluster_name    = "prow-build-canary-cluster"
-cluster_region  = "us-east-2"
-cluster_version = "1.25"
-
-vpc_cidr                  = "10.3.0.0/16"
-vpc_secondary_cidr_blocks = ["10.4.0.0/16", "10.5.0.0/16"]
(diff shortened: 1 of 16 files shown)


infra/aws/terraform/prow-build-cluster/terraform.canary.tfvars
```diff
@@ -19,28 +19,19 @@ eks_admins = [
   "xmudrii"
 ]
 
-assume_role                 = true
-deploy_kubernetes_resources = true
+prow_build_cluster = false
```
pkprzekwas, line 22: Maybe instead of introducing this variable we could check the account ID?

```suggestion
data "aws_caller_identity" "current" {}

locals {
  is_prod = data.aws_caller_identity.current.account_id == "111111111111"
}
```

pkprzekwas:
What about modifying [backend configuration](https://github.com/kubernetes/k8s.io/blob/main/infra/aws/terraform/prow-build-cluster/providers.tf#L18-L21) ? We probably need to create a dedicated S3 bucket for canary's state.

infra/aws/terraform/prow-build-cluster/terraform.canary.tfvars
xmudrii: We could, but I don't think that's important because canary will hopefully become a build cluster too.

xmudrii:
> What about modifying [backend configuration](https://github.com/kubernetes/k8s.io/blob/main/infra/aws/terraform/prow-build-cluster/providers.tf#L18-L21) ? We probably need to create a dedicated S3 bucket for canary's state.

That's a pretty good point, I'll update the PR.

infra/aws/terraform/prow-build-cluster/variables.tf
```diff
@@ -37,6 +37,14 @@ variable "deploy_kubernetes_resources" {
   nullable    = false
 }
 
+# This variable defines if this cluster is used as a Prow build cluster.
+variable "prow_build_cluster" {
```
pkprzekwas, line 41: I think `is_prod` or `is_canary` is easier to grasp for code reader.

infra/aws/terraform/prow-build-cluster/terraform.prod.tfvars
```diff
@@ -19,29 +19,19 @@ eks_admins = [
...
-
-vpc_cidr                  = "10.0.0.0/16"
-vpc_secondary_cidr_blocks = ["10.1.0.0/16", "10.2.0.0/16"]
-vpc_public_subnet         = ["10.0.0.0/18", "10.0.64.0/18", "10.0.128.0/18"]
-vpc_private_subnet        = ["10.1.0.0/18", "10.1.64.0/18", "10.1.128.0/18"]
-vpc_intra_subnet          = ["10.2.0.0/18", "10.2.64.0/18", "10.2.128.0/18"]
+cluster_name               = "prow-build-cluster"
+cluster_region             = "us-east-2"
+cluster_version            = "1.25"
+cluster_autoscaler_version = "v1.25.0"
```
pkprzekwas, line 27: I assume we want to keep those in dedicated files to test version upgrades without interfering with the other cluster. Why region is not in the shared file?

pkprzekwas: We could also move those versions to shared file and add them only to `canary` while testing upgrades. With the current order of `-var-files` that you introduced in Makefile, env specific tfvars have a higher priority than shared one.

infra/aws/terraform/prow-build-cluster/variables.tf
pkprzekwas: As discussed on slack, this is only temporary state. We want to add canary cluster to prow and remove this var soonish.

infra/aws/terraform/prow-build-cluster/terraform.prod.tfvars
xmudrii: Yes. We can keep both region and cluster name in the shared file.

infra/aws/terraform/prow-build-cluster/variables.tf
xmudrii: We discussed this and we'll keep prow_build_cluster as it is because we plan on removing it later.

pkprzekwas:
/hold

as canary started to fail.

xmudrii:
Both canary and prod are reconciled with the latest Terraform configs.
/hold cancel

PR #5174, opened 2023-04-25
rate limit registry

BenTheElder:
This will deploy to registry-sandbox.k8s.io after merge.

It will not deploy to registry.k8s.io until we've confirmed we're happy with the behavior and opt to roll it out to production

I've selected roughly the same rate limit as the backing AR instances, 5000 requests / minute per IP, however this applies to *all* calls coming through registry.k8s.io and helps shield us from excessive traffic even on S3 etc. 

(excluding users that bypass registry.k8s.io and call through to backing storage directly, hopefully most users will be nice and not do that...)

Part of: https://github.com/kubernetes/registry.k8s.io/issues/195

### DIFF
--- a/infra/gcp/terraform/modules/oci-proxy/cloud-armor.tf
+++ b/infra/gcp/terraform/modules/oci-proxy/cloud-armor.tf
@@ -151,7 +151,7 @@ resource "google_compute_security_policy" "cloud-armor" {
     preview = false
   }
 
-  # Reject all traffic that hasn't been whitelisted.
+  # Permit all other traffic
   rule {
     action      = "allow"
     description = "Default rule, higher priority overrides it"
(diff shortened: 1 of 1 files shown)


upodroid (approved):
/lgtm

BenTheElder:
this is the first time running the postsubmit against this repo https://prow.k8s.io/view/gs/kubernetes-jenkins/logs/post-k8s-deploy-registry/1650949377948127232

which should work fine but we'll see, it's roughly the postsubmit from registry.k8s.io but triggered by the terraform configs for the registry

BenTheElder:
`ulimit -n 6000 && curl --parallel --parallel-immediate --parallel-max 1000 "https://registry-sandbox.k8s.io/v2/?[1-6000]"` works as expected, producing 429 responses.

BenTheElder:
Whereas registry.k8s.io does not yet produce any 429 rate limit for the same command, as expected.

We should probably consider rolling this out to production soon, so we can establish a baseline for users.

PR #5175, opened 2023-04-25
upgrade registry production to v0.3.0

BenTheElder:
We don't have any new backend changes today, but we do have some improvements since last time:

- enhanced tests (no change here but important)
- rewritten blob cache to be more generalized and avoid lock contention
- timeout on http client performing blob HEAD checks to avoid potential hang there
- automated staging deployments with terraform (no change here but important)
- go 1.20

https://github.com/kubernetes/registry.k8s.io/compare/v0.2.1...v0.3.0

For the terraform, we have #5174, which is behaving as expected in staging.


BenTheElder:
https://testgrid.k8s.io/sig-k8s-infra-registry#Summary

In Particular:
https://testgrid.k8s.io/sig-k8s-infra-registry#registry-sandbox-e2e-gcp
https://testgrid.k8s.io/sig-k8s-infra-registry#registry-sandbox-e2e-aws
https://testgrid.k8s.io/sig-k8s-infra-registry#kops-grid-gcr-mirror-canary

BenTheElder:
/hold
will punt until tomorrow, I want to be around to watch it and it's getting late here

BenTheElder:
We can't deploy this currently because of https://github.com/kubernetes/registry.k8s.io/issues/234#issuecomment-1526903171, so ....

We could switch off of lb-http module like porche did and choose a different BES name ourselves, or we can wait for a response to b/279998015

BenTheElder:
This isn't pressing and we have the weekend coming, so for now I'm going to leave it.

I've started looking at a plan to pivot to a new non-classic GCLB so we can use new features re: failover anyhow, we'll have to practice that in staging

BenTheElder:
Stuck backends from the incident has been resolved. Will look at reconciling finally.

PR #5176, opened 2023-04-26
promote kube-scheduler-simulator v0.1.0

sanposhiho:
This PR promotes kube-scheduler-simulator v0.1.0 images.
(I created k8s-staging-sched-simulator dir in [the past](https://github.com/kubernetes/k8s.io/pull/2681), but it's somehow deleted now. 🤔  )

PR #5178, opened 2023-04-26
add promoter for kube-scheduler-simulator

sanposhiho:
We need to have promoter-manifest.yaml so that the images are promoted.


sanposhiho:
We already have images.yaml. But do we need any other stuff?

https://github.com/kubernetes/k8s.io/pull/5176

ameukam:
> We already have images.yaml. But do we need any other stuff?
> 
> #5176

Oh. Sorry I confused the path. Should be good. 

/lgtm
/approve

PR #5179, opened 2023-04-26
Normalize nfd

ArangoGutierrez:


ArangoGutierrez:
We have noticed that by not being quoted `""` NFD promote PR's take longer than other projects. This Patch aims to fix that

fmuyassarov (commented):
/lgtm

PR #5191, opened 2023-04-28
remove adtac from the simulator group

sanposhiho:
- remove @adtac from the simulator group

ref: https://github.com/kubernetes-sigs/kube-scheduler-simulator/pull/297


Huang-Wei (commented):
/lgtm
/approve

PR #5213, opened 2023-05-04
Automate managing service quotas for eks-e2e-boskos accounts using Terraform and CloudFormation

xmudrii:
Reference: https://kubernetes.slack.com/archives/CCK68P2Q2/p1683127657110489

Notes:

- I'm aware that there are [Service Quotas requests templates](https://docs.aws.amazon.com/servicequotas/latest/userguide/organization-templates.html), but these require creating new accounts, which is too complicated at this point because of permissions
- I'm aware that CAPA tests can handle requesting service quotas to be increased, however, I was looking for an efficient way to accomplish the same, without having CAPA tests to run on each account

The main goal of this PR is to centralize managing service quotas, as well as, to be able to manage service quotas easily for all boskos accounts at the same time. I've observed what service quotas are required by CAPA tests and those requirements are "documented" in the `service-quotas/modules/capa` Terraform module. This module is applied to all boskos accounts via the `service-quotas/boskos` Terraform module.

However, I encountered some technical problems along the way:

- To run Terraform, we need IAM user or IAM role in each boskos account
- If we go with a dedicated IAM user in each account, we have to use dedicated Terraform backends, and that would require us to have a S3 bucket for Terraform state in each boskos account, which is kinda inefficient

I solved these two obstacles in the following way:

- We have an IAM user (`provisioner`) in the first Boskos account (`eks-e2e-boskos-001`)
  - This IAM user can write to S3 bucket and assume other roles
- We have IAM roles (`Provisioner`) in all Boskos accounts
- The `Provisioner` IAM role is configured in a way that `provisioner` user can assume it
- The `Provisioner` IAM role has a restrictive policy with as least permissions as possible
- Resources that support tagging are tagged with `Boskos=Ignore` to ensure that aws-janitor ignores these resources

That way, we have only one IAM user, only one set of credentials, and only one Terraform state and S3 bucket. This should make apply these configs much easier and more efficient.

However, I ran into another issue. I initially wanted to manage IAM for those accounts with Terraform, but there's a chicken-egg problem. Terraform requires a S3 bucket to store Terraform state somewhere. I would also need an IAM user in all boskos accounts to be able to run those Terraform configs for IAM, and that's what I wanted to avoid.

The way I solved this is with using CloudFormation. CloudFormation makes this easy because we can define IAM roles in JSON (or YAML) and "copy-paste" that JSON to AWS. AWS will take care of creating IAM roles and policies, and we don't need to care about Terraform, state, and IAM users and credentials. This is inline with what @pkprzekwas and I discussed with @sftim on KubeCon as a way to manage IAM for AWS accounts.

TODO:

- Ensure boskos is configured to exclude resources with `Boskos=Ignore` tag (#5233)

/assign @pkprzekwas @dims @ameukam @sftim 
/hold for discussion and reviews


pkprzekwas (commented):
I left some nitpicks, most of them requires no action.

infra/aws/terraform/iam/eks-e2e-boskos/README.md
```diff
@@ -0,0 +1,42 @@
+# IAM for Boskos accounts
+
+This directory contains CloudFormation templates needed to provision IAM resources
```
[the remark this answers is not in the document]
pkprzekwas: As discussed outside GH, it was done to avoid creating dedicated S3 bucket for TF state.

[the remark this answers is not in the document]
xmudrii: Let me clarify this further for the reference. If we wanted to use Terraform to manage IAM user and roles for these accounts, we would need to:

- Create an IAM user and access key for that user in each account
- Create a S3 bucket in each account to be used for storing Terraform state
- Have a dedicated Terraform backend for each account

In any case, tasks 1 and 2 must be done manually or with CloudFormation. That said, I think that completely managing IAM with CloudFormation is easier and more scalable.

infra/aws/terraform/service-quotas/boskos/main.tf
```diff
@@ -0,0 +1,112 @@
...
+  }
+}
+
+module "capa_quotas_001" {
+  source = "../modules/capa"
+
+  providers = {
+    aws = aws.eks-e2e-boskos-001
+  }
+}
```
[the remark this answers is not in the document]
xmudrii: I tried this, but it doesn't work:

- String extrapolation is not supported for providers, so `"aws.eks-e2e-boskos-${count.index}"` fails
- It doesn't seem possible to reference `count.index` without quotes in a way that it's supported in the providers block

That said, I think we should ignore this comment for now.

infra/aws/terraform/service-quotas/boskos/main.tf
```diff
@@ -0,0 +1,112 @@
...
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+terraform {
+  required_version = "~> 1.3.0"
+
+  backend "s3" {
+    bucket = "eks-e2e-boskos-tfstate"
```
pkprzekwas, line 21: Where this bucket is located? Maybe we could create it with CF?

infra/aws/terraform/service-quotas/boskos/main.tf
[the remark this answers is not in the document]
pkprzekwas: It does make sense after giving a second thought

infra/aws/terraform/service-quotas/boskos/providers.tf
```diff
@@ -0,0 +1,119 @@
...
+provider "aws" {
+  region = var.region
+}
+
+provider "aws" {
+  region = var.region
+  alias  = "eks-e2e-boskos-001"
+
+  assume_role {
+    role_arn     = "arn:aws:iam::995654820765:role/Provisioner"
```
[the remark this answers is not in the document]
xmudrii: This is done.

infra/aws/terraform/service-quotas/boskos/main.tf
xmudrii: That's a good idea, but let's leave it for follow up.

infra/aws/terraform/iam/eks-e2e-boskos/README.md
```diff
@@ -0,0 +1,42 @@
...
+that aws-janitor ignores these resources.
+
+### IAM User
+
+This Stack is applied only to the first Boskos account. The Stack is located in
+`cloudformation-user.json`.
+
+Once the Stack is applied, `cloudformation-role.json` needs to be modified to replace
+principal for the `Provisioner` role. Also, credentials needs to be created manually
+once the user is created.
```
sftim, line 37: Why an IAM user, not a role?

infra/aws/terraform/iam/eks-e2e-boskos/cloudformation-role.json
```diff
@@ -0,0 +1,67 @@
...
+      "Properties": {
+        "RoleName": "Provisioner",
+        "AssumeRolePolicyDocument": {
+          "Version": "2012-10-17",
+          "Statement": [
+            {
+              "Action": "sts:AssumeRole",
+              "Effect": "Allow",
+              "Principal": {
+                "AWS": "arn:aws:iam::995654820765:user/provisioner"
```
sftim, line 14: Try something like:
```yaml
                "AWS": "arn:${AWS::Partition}:iam::${SomeNamedVariable}:user/provisioner"
```

infra/aws/terraform/iam/eks-e2e-boskos/cloudformation-role.json
sftim, line 1: If you use YAML, you can add explanatory comments. I recommend this approach.

BTW, this document is already valid YAML (all JSON documents are valid YAML and deserialize to the same thing).

infra/aws/terraform/iam/eks-e2e-boskos/cloudformation-user.json
```diff
@@ -0,0 +1,41 @@
...
+        "Policies": [
+          {
+            "PolicyName": "ProvisionerUserRole",
+            "PolicyDocument": {
+              "Version": "2012-10-17",
+              "Statement": [
+                {
+                  "Action": "sts:AssumeRole",
+                  "Effect": "Allow",
+                  "Resource": "*"
```
sftim, line 17: This seems quite broad. I guess we're relying on role trust policies here. Is that safe?

infra/aws/terraform/iam/eks-e2e-boskos/cloudformation-user.json
```diff
@@ -0,0 +1,41 @@
...
+                },
+                {
+                  "Action": [
+                    "s3:GetObject",
+                    "s3:PutObject",
+                    "s3:PutObjectAcl"
+                  ],
+                  "Effect": "Allow",
+                  "Resource": "*"
```
sftim, line 26: Allow writing objects to _any_ S3 bucket in the same account? IMO this is too broad.

infra/aws/terraform/service-quotas/boskos/main.tf
```diff
@@ -0,0 +1,112 @@
...
+    key    = "service-quotas/terraform.tfstate"
+    region = "us-west-2"
+  }
+
+  required_providers {
+    aws = {
+      source  = "hashicorp/aws"
+      version = ">= 4.47"
```
sftim, line 29:
```suggestion
      version = "~> 4.47"
```
(this will balk at a major version change)

infra/aws/terraform/service-quotas/boskos/main.tf
sftim, line 1: Please add a comment explaining the intent behind the numbered provider aliases.

infra/aws/terraform/service-quotas/boskos/providers.tf
sftim, line 1: Please add a comment explaining the intent behind the numbered provider aliases.

(There may be an alternative, such as multiple invocations and Terraform workspaces)

infra/aws/terraform/service-quotas/boskos/terraform.tfvars
```diff
@@ -0,0 +1,31 @@
...
+    "995654820765",
+    "252228595925",
+    "757254795865",
+    "236737742629",
+    "568640253687",
+    "325389058785",
+    "819381709038",
+    "080553712153",
+    "187041256112",
+    "502839862640",
```
sftim, line 30: Consider adding an inline comment to name what these represent.

infra/aws/terraform/service-quotas/modules/capa/main.tf
```diff
@@ -0,0 +1,82 @@
...
+limitations under the License.
+*/
+
+terraform {
+  required_version = "~> 1.3.0"
+
+  required_providers {
+    aws = {
+      source  = "hashicorp/aws"
+      version = ">= 4.47"
```
sftim, line 23:
```suggestion
      version = "~> 4.47"
```
(this will balk at a major version change)

infra/aws/terraform/service-quotas/modules/capa/main.tf
```diff
@@ -0,0 +1,82 @@
...
+#######################################
+# Service Quotas needed by CAPA tests #
+#######################################
+
+# Amazon EventBridge (CloudWatch Events): Maximum number of rules an account can have per event bus
+resource "aws_servicequotas_service_quota" "events_rules" {
+  quota_code   = "L-244521F2"
+  service_code = "events"
+  value        = 500
+}
```
sftim, line 37: This could possibly use a `for_each`. I would, honestly.

infra/aws/terraform/service-quotas/modules/capa/output.tf
```diff
@@ -0,0 +1,43 @@
...
+  value = aws_servicequotas_service_quota.ec2_max_elastic_ips.adjustable
+}
+
+output "ec2_max_vcpus_adjustable" {
+  value = aws_servicequotas_service_quota.ec2_max_vcpus.adjustable
+}
+
+output "ec2_max_vcpus_g_vt_adjustable" {
+  value = aws_servicequotas_service_quota.ec2_max_vcpus_g_vt.adjustable
+}
```
sftim, line 43: I think I'd have one output, typed as a `list` of `maps`.

sftim (commented):
Thanks.

A quick review, hope it's useful.

infra/aws/terraform/service-quotas/boskos/providers.tf
sftim: If the list of accounts is static, workspaces probably aren't worth it.

infra/aws/terraform/service-quotas/boskos/main.tf
ameukam: At least, can we comment about in which account it's created ?

infra/aws/terraform/iam/eks-e2e-boskos/README.md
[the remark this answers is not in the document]
sftim: I wouldn't put these 3 things into the main Terraform:

1. S3 bucket for Terraform state (do we need one per account, though?)
2. lock table for Terraform state locking
3. IAM role for executing Terraform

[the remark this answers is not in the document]
sftim: Also, we could use StackSets if we do want CloudFormation deployed to more than one account.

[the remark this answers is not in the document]
xmudrii: > S3 bucket for Terraform state (do we need one per account, though?)

No, the way it's working right now is that it's enough to have bucket in only one account (in this case boskos-001). With IAM user in that account, we assume roles in other accounts and do the magic. But state for all those accounts is stored in a single bucket.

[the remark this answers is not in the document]
sftim: > state for all those accounts is stored in a single bucket

I think that's OK. We can use a bucket policy if we need to restrict access to state, and that means we don't need to make more buckets.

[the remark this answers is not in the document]
xmudrii: > Also, we could use StackSets if we do want CloudFormation deployed to more than one account.

@sftim I looked into StackSets and my understanding is that basically it allows applying CloudFormation in more than one account. That requires creating a role in each account or integrating with AWS Organization. I like that idea, but I'd leave it for a potential follow up, maybe it would be better to have a StackSet in the management account and then manage boskos accounts from there. However, we currently don't have access to the management account.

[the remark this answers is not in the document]
sftim: You _can_ delegate access, but it's not fun. Let's rule out StackSets.

infra/aws/terraform/service-quotas/boskos/main.tf
xmudrii: That's a good point @ameukam, will be added.

infra/aws/terraform/iam/eks-e2e-boskos/cloudformation-role.json
xmudrii: `${AWS::Partition}` substitution didn't work, but Account ID is now a parameter.

infra/aws/terraform/iam/eks-e2e-boskos/cloudformation-role.json
xmudrii: Good point, I converted both stacks to YAML.

infra/aws/terraform/iam/eks-e2e-boskos/cloudformation-user.json
xmudrii: I agree with that and I did this instead:

```yaml
              - Action: 'sts:AssumeRole'
                Effect: Allow
                Resource: 'arn:aws:iam::*:role/Provisioner'
```

Is this any better?

infra/aws/terraform/iam/eks-e2e-boskos/cloudformation-user.json
xmudrii: I agree, I limited this only to the TFState S3 bucket.

infra/aws/terraform/service-quotas/boskos/providers.tf
xmudrii: > If the list of accounts is static, workspaces probably aren't worth it.

The list is indeed static. I agree that workspaces and multiple backends aren't worth it, hence the idea to do it this way.

infra/aws/terraform/service-quotas/boskos/main.tf
xmudrii: S3 bucket is now managed via CloudFormation as well.

infra/aws/terraform/iam/eks-e2e-boskos/README.md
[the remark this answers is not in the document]
xmudrii: The reasoning is now documented in README for IAM CloudFormation.

infra/aws/terraform/iam/eks-e2e-boskos/README.md
xmudrii: There were some typos in that README so I fixed those. The intention of using IAM user is that we need some static credentials to be able to run Terraform locally, so IAM user is the only way.

xmudrii:
@sftim @pkprzekwas All but last two comments are addressed. I'll try to address remaining two comments tomorrow, but those are very minor, so the PR is ready for review.

xmudrii:
@sftim @pkprzekwas Remaining two comments are addressed as well.

infra/aws/terraform/service-quotas/boskos/main.tf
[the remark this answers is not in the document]
xmudrii: Official reference: https://github.com/hashicorp/terraform/issues/19932

pkprzekwas (commented):
/lgtm

infra/aws/terraform/service-quotas/boskos/README.md
```diff
@@ -0,0 +1,27 @@
+# Service Quotas for Boskos Accounts
+
+This directory contains Terraform configs to apply service quotas in boskos
+accounts. Currently, these configs are only applying
```
pkprzekwas, line 4: Maybe let's be specific and mention that those are AWS accounts?

xmudrii: It's in `infra/aws` directory, so I thought that's kinda implicit 😄

xmudrii: Fixed

xmudrii:
Applied to all accounts.
/hold cancel

PR #5235, opened 2023-05-08
eks-prow-build-cluster: set permission boundary for roles created via EKS and IRSA modules

pkprzekwas:
Add permission boundaries to roles previously created with EKS and IRSA terraform modules. Definition of permission boundary is located [here](https://github.com/kubernetes/k8s.io/blob/main/infra/aws/terraform/modules/eks-prow-iam/boundary.tf).

/assign @xmudrii 
/cc @sftim


pkprzekwas:
/hold
wait for applying on k8s-infra-prow account

xmudrii (approved):
/lgtm
/approve
/hold to apply to prod

infra/aws/terraform/modules/eks-prow-iam/policies.tf
```diff
@@ -143,6 +143,8 @@ data "aws_iam_policy_document" "eks_apply" {
       "iam:TagPolicy",
       "iam:TagRole",
       "iam:UpdateOpenIDConnectProviderThumbprint",
+      "iam:PutRolePermissionsBoundary",
+      "iam:DeleteRolePermissionsBoundary",
```
xmudrii, line 147: Should `DeleteRole*` go here or in the destroy policy?

infra/aws/terraform/prow-build-cluster/eks.tf
```diff
@@ -44,6 +44,8 @@ module "eks" {
     },
   ]
 
+  iam_role_permissions_boundary = "arn:aws:iam::${local.account_id}:policy/ProvisionerPermissionBoundary"
```
xmudrii, line 47: Is it possible to automatically figure out this ARN with data resource? Doing that will also ensure that this permission boundary exists.

xmudrii (commented):
We should add [`permission_boundary`](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/iam_role#permissions_boundary) to all IAM roles that we create via Terraform. It's going to be mandatory once we start enforcing our management IAM role to create IAM roles with the permission boundary.

sftim:
We might want a break-glass role that's allowed to edit the boundary policy, and guidelines around using it.

infra/aws/terraform/modules/eks-prow-iam/policies.tf
pkprzekwas: Good point

infra/aws/terraform/prow-build-cluster/eks.tf
pkprzekwas: Yeah, that's a great idea

pkprzekwas:
> We might want a break-glass role that's allowed to edit the boundary policy, and guidelines around using it.

E.g. `InfraAdmin` role that could be assumed just by very narrow set of users and would contain `AdministratorAccess` policy?

sftim:
> E.g. `InfraAdmin` role that could be assumed just by very narrow set of users and would contain `AdministratorAccess` policy?

That's quite broad!

The role I'm imagining could have an inline policy that _only_ allows:
- IAM read access
- the ability to modify the IAM policies that we use for permissions boundaries

It depends what we want to achieve. Allowing `*:*` really is a lot of access.

xmudrii:
I'm wondering if we should have such role at all. Instead, I think SIG leads should have access to the root account (shared via 1Password). They can then create an IAM user with needed access (e.g. AdministratorAccess). That way, we reduce potential attack surface (you can't assume role if there's no such role). This is based on assumption that break-glass access should be needed very rare (if at all).

pkprzekwas:
I think access to the root account may not be sufficient in this case. Or maybe you meant root access to the k8s-infra-prow account?

xmudrii:
@pkprzekwas Yes, I meant root access to the k8s-infra-prow account.

xmudrii (approved):
/lgtm
/approve

pkprzekwas:
All changes applied on prow-build-cluster.

xmudrii (approved):

PR #5240, opened 2023-05-08
dns challenge: add required project field

BenTheElder:
This is what we get when we only run in postsubmit. This time I've tested locally with:
`terraform apply -auto-approve -var digest=$(crane digest gcr.io/k8s-staging-infra-tools/archeio:$(crane ls gcr.io/k8s-staging-infra-tools/archeio | sort -V | tail -n1))`


BenTheElder:
success, I'm also applying this to the prod project so we can roll out the DNS entries for both then proceed to the next stage of terraform config.

PR #5242, opened 2023-05-09
registry: create managed certificate using DNS challenge

BenTheElder:
next step in #4779 

adapted from porche

tested in staging with `terraform apply -var digest=$(crane digest gcr.io/k8s-staging-infra-tools/archeio:$(crane ls gcr.io/k8s-staging-infra-tools/archeio | sort -V | tail -n1))`

once certs are provisioned in staging and then prod I'll follow with adding the new EXTERNAL_MANAGED LBs, and then we can cut over to those in DNS and phase out the old LBs.

### DIFF
--- a/infra/gcp/terraform/modules/oci-proxy/network.tf
+++ b/infra/gcp/terraform/modules/oci-proxy/network.tf
@@ -17,6 +17,11 @@ limitations under the License.
 # Resources for new envoy-based loadbalancer / EXTERNAL_MANAGED mode.
 ################################################################################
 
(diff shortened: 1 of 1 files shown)


BenTheElder:
I can see the cert provisioned already in staging, excellent.

/hold cancel

BenTheElder:
https://prow.k8s.io/view/gs/kubernetes-jenkins/logs/post-k8s-deploy-registry/1655805024732712960

\+ applied to prod

PR #5241, opened 2023-05-08
registry: add cert DNS challenges, update domain point of contact

BenTheElder:
```
$ gcloud certificate-manager dns-authorizations --project=k8s-infra-oci-proxy list
NAME                             DOMAIN                   DNS_RECORD                                RECORD_TYPE  DNS_VALUE
registry-sandbox-k8s-io-dnsauth  registry-sandbox.k8s.io  _acme-challenge.registry-sandbox.k8s.io.  CNAME        893400bf-1899-4ca4-8cc3-1599310bb39b.12.authorize.certificatemanager.goog.

$ gcloud certificate-manager dns-authorizations --project=k8s-infra-oci-proxy-prod list
NAME                     DOMAIN           DNS_RECORD                        RECORD_TYPE  DNS_VALUE
registry-k8s-io-dnsauth  registry.k8s.io  _acme-challenge.registry.k8s.io.  CNAME        bc87a114-671a-4ff9-8b32-c93230aa3136.14.authorize.certificatemanager.goog.
```

PR #5249, opened 2023-05-09
add envoy based loadbalancer using cert map

BenTheElder:
Next step in #5241, after this is in staging and prod we can test the LB a bit and then PR rotating DNS to point at the new LBs. 

Then we can come back later and tear down the old ones when we're confident there's no more inbound traffic (from stale DNS client-side).

Tested in staging with `terraform apply -var digest=$(crane digest gcr.io/k8s-staging-infra-tools/archeio:$(crane ls gcr.io/k8s-staging-infra-tools/archeio | sort -V | tail -n1)` and then `curl --verbose --resolve registry-sandbox.k8s.io:443:107.178.255.82 https://registry-sandbox.k8s.io/v2/pause/manifest/3.1` (IP of new https LB)

### DIFF
--- a/infra/gcp/terraform/modules/oci-proxy/network.tf
+++ b/infra/gcp/terraform/modules/oci-proxy/network.tf
@@ -58,6 +58,127 @@ resource "google_certificate_manager_certificate_map_entry" "default" {
   matcher      = "PRIMARY"
 }
 
+# IP Addresses for the loadbalancer
+resource "google_compute_global_address" "ipv4" {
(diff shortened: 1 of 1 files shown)


BenTheElder:
https://prow.k8s.io/view/gs/kubernetes-jenkins/logs/post-k8s-deploy-registry/1656085630561554432 🟢 for staging

\+ applied to prod

Will follow up with rotating staging DNS, then prod later.

PR #5253, opened 2023-05-10
eks-prow-build-cluster: create scripts for TF state buckets

pkprzekwas:
Initially we created S3 buckets for terraform stated manually both for k8s-infra-prow and k8s-infra-prow-canary accounts. This PR introduces terraform scripts for those buckets. Both buckets have been imported so that terraform state is in sync.

/assign @xmudrii

### DIFF
--- a/infra/aws/terraform/s3/k8s-infra-prow-canary/.gitignore
+++ b/infra/aws/terraform/s3/k8s-infra-prow-canary/.gitignore
@@ -1,0 +1,21 @@
+# Local .terraform directories
+**/.terraform/*
+
+# .tfstate files
+*.tfstate
+*.tfstate.*
+*.tfplan
+
+# Crash log files
+crash.log
+
+# Ignore override files as they are usually used to override resources locally and so
+# are not checked in
+override.tf
+override.tf.json
+*_override.tf
+*_override.tf.json
+
+# Ignore CLI configuration files
+.terraformrc
+terraform.rc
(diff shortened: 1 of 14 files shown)


ameukam:
@pkprzekwas @xmudrii do we still need this ?

xmudrii:
@ameukam Perhaps yes, but we need to decide about direction. Do we want to manage Terraform state buckets with Terraform or do we want to use CloudFormation? Recently I used CloudFormation for this use case (see https://github.com/kubernetes/k8s.io/tree/main/infra/aws/cloudformation) and it's a really neat way, so we should definitely consider it.

ameukam:
> @ameukam Perhaps yes, but we need to decide about direction. Do we want to manage Terraform state buckets with Terraform or do we want to use CloudFormation? Recently I used CloudFormation for this use case (see https://github.com/kubernetes/k8s.io/tree/main/infra/aws/cloudformation) and it's a really neat way, so we should definitely consider it.

I think it make sense to use CloudFormation (or something else like aws-cli,etc...) to create S3 buckets dedicated tf storage.

pkprzekwas:
I prefer sticking to Terraform, but I can see that in this case it might be confusing. Should I close this PR @ameukam @xmudrii ?

xmudrii:
@pkprzekwas Let's close it for now and we can revisit this topic in the future.

PR #5329, opened 2023-05-27
promote images with alpine318

longwuyuan:
- New baseimage along with a few others built with alpine v3.18 from https://github.com/kubernetes/ingress-nginx/pull/9997
- This PR promotes those 3 new images to prod registry
- After this merges, we need to manually trigger rebuild of e2e-test-runner image as well. Because auto trigger did not occur for it but it e2e-test-runner also needs to upgrade to alpine v3.18 . Its important because even though e2e-test-runner is used only in CI, the current e2e-test-runner image built on alpine v3.17 reports at least 10 CVEs. And the alpine bump is required for avoiding false positives in CI

cc @tao12345666333 @strongjz @rikatz


tao12345666333 (approved):
/lgtm
/approve

tao12345666333 (approved):
/lgtm
/approve

PR #5333, opened 2023-05-27
promote new e2etestecho image

longwuyuan:
- Promoting newly built e2etestecho image that uses alpine v3.18
- After this merges, the ingressnginx project needs to be upates with this sha + tag

cc @tao12345666333 @strongjz @rikatz


tao12345666333 (approved):
/lgtm
/approve

PR #5345, opened 2023-05-31
Update SIG-CLI chairs and TL

soltysh:
Per the announcement https://groups.google.com/g/kubernetes-sig-cli/c/J_WelkrSwBM/m/brDiEhVRAgAJ this updates the leads in SIG-CLI

/assign @mrbobbytables


mrbobbytables:
One last place to update it in here:
https://github.com/kubernetes/k8s.io/blob/main/groups/groups.yaml
That will add them to the leads mailing list / leads calendar invite etc

Issue #5354, opened 2023-06-04
EKS Prow build cluster: remove cadvisor


xmudrii:
We should consider removing cadvisor from EKS Prow build cluster. Kubelet has cadvisor integrated and we should just scrape kubelet instead. Benefits to that include that metrics served by kubelet have additional labels such as node name, pod IP address, and more, and that can be helpful for various queries and charts.

/assign

PR #5355, opened 2023-06-04
EKS Prow build cluster: add monitoring for Kubelet and ProwJob pods

xmudrii:
This PR adds the following components to the monitoring stack in EKS Prow build cluster:

- ServiceMonitor for Kubelet (taken from [the kube-prometheus repository](https://github.com/prometheus-operator/kube-prometheus/blob/158cb5422ebe95a7cad82efe0999bcff1d764443/manifests/kubernetesControlPlane-serviceMonitorKubelet.yaml))
- Prometheus rules for ProwJob pods (taken from [`loodse/prow-dashboards`](https://github.com/loodse/prow-dashboards))
- Grafana dashboards for ProwJob pods (taken from [`loodse/prow-dashboards`](https://github.com/loodse/prow-dashboards))

Prometheus rules and Grafana dashboards for ProwJob pods are based on metrics that kubelet provides. Metrics provided by cadvisor directly are not compatible with those rules and dashboards, and because of that, ServiceMonitor for cadvisor has been temporarily removed from the production cluster. See #5354 for more details.

@pkprzekwas I recommend merging this PR before deploying Flux on production to ensure we don't lose those dashboards.

Grafana dashboards should support migrating jobs without resource requests and limits because those dashboards can be used to determine how much CPU and memory is used by a ProwJob pod. I'm planning on making presentation about this at the next SIG K8s infra meeting.

As an example, see this dashboard: https://monitoring-eks.prow.k8s.io/d/96Q8oOOZk/builds

/assign @pkprzekwas @wozniakjan 
cc @dims @ameukam @BenTheElder


ameukam:
Micro-nit: `deployment` instead of `deplyment`for the `kube-state-metrics`deployment file.

PR #5371, opened 2023-06-06
Fastly: ensure GCS cookies are dropped

ameukam:
Ensure GCS cookies are dropped in the http response from requests to cdn.dl.k8s.io


ameukam:
cc @dims

PR #5378, opened 2023-06-07
Promote nfs csi driver image v4.4.0

andyzhangx:
/assign @msau42 @xing-yang


larivierec:
hello, this should most probably be published before the helm chart is released.
as the helm chart depends on this plugin.

ref: v4.4.0 released 2 hours ago.
https://github.com/kubernetes-csi/csi-driver-nfs

this plugin has not been promoted

larivierec:
@andyzhangx i'm not sure what you mean by cut the release however, if this `nfsplugin` is always tied to the helm-chart then this project should be parent of the helm chart and should eliminate your problem.

1. this project will have its tag created first
2. this project is deployed and released
3. the helm chart project can use this project and be released properly.

msau42:
I'm not sure if this helps, but what we've tried for other csi projects is to create the tag without "creating a release" and then after promoting the image, do the github release.

PR #5380, opened 2023-06-07
Switch Flux git source to main branch of k8s.io

pkprzekwas:
As we now have everything merged to the `main` branch, we can safely switch over.

/assign @wozniakjan


wozniakjan (approved):
/lgtm

PR #5400, opened 2023-06-09
Update SIG-Testing leads list with all leads.

michelle192837:
Related to https://github.com/kubernetes/community/issues/7346, only added chairs which was an oversight. (Everyone, please double-check these are the correct emails to use!)


groups/sig-testing/groups.yaml
```diff
@@ -19,9 +19,14 @@ groups:
       WhoCanPostMessage: "ANYONE_CAN_POST"
       WhoCanViewGroup: "ALL_MEMBERS_CAN_VIEW"
     owners:
+    - aaleman@redhat.com
```
aojea, line 22: I have doubts @alvaroaleman still has this email

PR #5421, opened 2023-06-14
Promote volume/iscsi images

jsafrane:
These images were not updated for a long time.

* Add 2.4 just for the reference, I am not sure it was ever used.
* Add 2.5, which is currently used in e2e test and [they're failing](https://prow.k8s.io/view/gs/kubernetes-jenkins/logs/ci-kubernetes-e2e-gce-iscsi/1668640602431950848) because of that.
* Add 2.6, which is the latest image.

2.6 will be needed to fix https://github.com/kubernetes/kubernetes/issues/118446


jsafrane:
/assign @dims
for review & approval

PR #5427, opened 2023-06-14
Update TestGrid to use load balancer

michelle192837:
Fixes #5392

PR #5429, opened 2023-06-15
Revert "Update TestGrid to use load balancer"

michelle192837:
Reverts kubernetes/k8s.io#5427

This should have minimal downtime, but it doesn't seem to. Rolling back to verify all the correct steps are taken before trying again.


TerryHowe (approved):
/lgtm

PR #5488, opened 2023-06-28
Change alpine registry to avoid docker.io rate limiting 

pkprzekwas:
/assign @ameukam @dims

### DIFF
--- a/infra/aws/terraform/prow-build-cluster/resources/kube-system/create-loop-devs_daemonset.yaml
+++ b/infra/aws/terraform/prow-build-cluster/resources/kube-system/create-loop-devs_daemonset.yaml
@@ -40,7 +40,7 @@ spec:
                 done
                 sleep 100000000
               done
-          image: alpine:3.6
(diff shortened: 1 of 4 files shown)


aojea:
@pkprzekwas @ameukam @dims 

I see two possible problems with this image, correct me if I'm wrong:
1. depends on `gcr.io`
2. uses `latest` tag

can we use `registry.k8s.io/alpine-with-bash:1.0`instead, or anything from the public registry and with a tag that is not `latest`? I just pick that from https://oci.dag.dev/?repo=registry.k8s.io, but I do not have clear where it is coming from,  we have several images on the public repo that can work for this, `agnhost` per example

ameukam:
@aojea Dockerfile for this image: https://github.com/kubernetes/test-infra/blob/master/images/alpine/Dockerfile

aojea:
yeah, you clarified on the other comment, I think that my suggestion is to pin it

PR #5490, opened 2023-06-28
k8s-infra: bump components for the container image

ameukam:
- switch the base image to bookworm
- bump different components
- ensure python packages can be installed with of [PEP 668](https://peps.python.org/pep-0668/)
- drop old versions of terraform


ameukam:
cc @BenTheElder @dims

BenTheElder (approved):
/lgtm
/approve

PR #5505, opened 2023-07-03
eks-prow-build-cluster: ssh access to eks nodes

pkprzekwas:
This PR:
- creates key pair and configures nodes of EKS cluster to use it
- introduces conditional creation of bastion host (jumphost)
- adds doc describing how to access EKS nodes via SSH

/hold rolling out this change may take a while as it requires replacing nodes
/assign @xmudrii @ameukam @wozniakjan


pkprzekwas:
> Why do we need a predefined ssh key to access a build cluster ?

I was trying to configure Session Manager but failed doing so... Using EC2 serial console option also doesn't work as we don't set password for the user inside AMI and EC2 Instance Connect Endpoint requires having ssh key pair as well. The benefit of using traditional ssh key with bastion, is that someone without AWS account can access the node.

I don't have strong opinion on this solution. Just wanted to implement sth quickly in case AMI upgrade didn't resolve issue with nodes.

pkprzekwas:
/hold cancel

Those changes have been rolled out to prod. I used the time around 4th of July as the cluster was less busy.

ameukam:
since this is already rolled out in prod
/lgtm

infra/aws/terraform/prow-build-cluster/docs/Debug.md
```diff
@@ -0,0 +1,30 @@
+## SSH Access EKS Nodes
+
+To access cluster nodes via ssh, first you need to provision bastion host:
+
+```bash
+TF_ARGS="-var bastion_install=true" make apply
```
xmudrii, line 6: Let's recommend doing this via `.tfvars` so that we don't lose bastion if someone forgets the flag.

infra/aws/terraform/prow-build-cluster/bastion.tf
```diff
@@ -0,0 +1,82 @@
...
+resource "aws_key_pair" "eks_nodes" {
+  key_name_prefix = "${var.cluster_name}-nodes"
+  public_key      = var.public_key
+}
+
+resource "aws_instance" "bastion" {
+  count = var.bastion_install ? 1 : 0
+
+  ami                         = data.aws_ami.ubuntu.id
+  instance_type               = "t3.nano"
```
xmudrii, line 41: We should use `bastion_instance_type` here.

xmudrii (commented):
Belated nits, otherwise LGTM.

infra/aws/terraform/prow-build-cluster/docs/Debug.md
pkprzekwas: I made it this way on purpose to point out the ephemeral nature of bastion host. In case someone removes it, it's easy to provision it again.
Another reason is that we don't have any gitops in place for applying terraform, so at the end of the say someone has to run the script manually.
But I'm also fine having this placed in `.tfvars`.

Issue #5473, opened 2023-06-26
eks-prow-build-cluster instability issues as of 2023-06-26


xmudrii:
We observed stability issues in the EKS Prow build cluster. Nodes tend to randomly stop working and become NotReady or Ready,SchedulingDisabled.

Instability indicators:

- All nodes are between several hours and 4 days old, but cluster-autoscaler didn't trigger scale down since 2023-06-23
- NotReady nodes are reporting events about blocked tasks
- There were two nodes with the same issue

We checked logs, disk usage, and some other metrics, but everything looks fine for affected nodes. We tried restarting an affected node, the node becomes Ready for a few minutes, but then eventually gets deleted.

Screenshot of events for an affected node:

![image](https://github.com/kubernetes/k8s.io/assets/18719127/5f1eba14-9a34-440b-bab3-ff51b1db4508)

/assign @xmudrii @pkprzekwas 
/sig k8s-infra
/priority critical-urgent

ameukam:
Dow  we have samples for kubelet logs?

aojea:
> Screenshot of events for an affected node:

it seems clear the node hung and lost the connection against the control plane, several process are blocked for more than 121 seconds, we need to undeerstand why they are blocked

xmudrii:
Let's keep a track of kubeadm revert PRs:

- https://github.com/kubernetes/kubeadm/pull/2911
- https://github.com/kubernetes/test-infra/pull/30283
- https://github.com/kubernetes/test-infra/pull/30284

tzneal:
When you try this again, can you run this script on the failed node:

xmudrii:
We have migrated to Bottlerocket Linux on October 11 as recommended by AWS as a way to mitigate this issue. Since then, we had total of three nodes that failed, but in a different way that's not related to this issue. This AMI is known to work:

- AMI ID: ami-0064cd291d1583c3c
- AMI name: bottlerocket-aws-k8s-1.28-x86_64-v1.15.1-264e294c

Short context story: AWS originally recommended migrating to Amazon Linux 2 (AL2) because they have more debugging options for AL2 compared to Ubuntu. However, AL2 comes with a very old kernel that doesn't support kind well (it's not possible to run two kind cluster at the same time at the same node). A lot of our tests/jobs are using kind, so that wasn't an option. Bottlerocket comes with a much newer kernel version (6.x), so we decided to give it a try instead.

[...]

PR #5516, opened 2023-07-06
Add AlertManager to EKS Prow build cluster

xmudrii:
This PR adds initial alerting stack to the EKS Prow build cluster. Right now, it:

- has only one alert to notify us if there's a node that's NotReady (to help us triage #5473)
- sends notification to Kubermatic's internal Slack

Additional alerts might be added in the future. I'll also check if it's possible to route alerts to Kubernetes Slack.

This has been already tested in the canary cluster and it works well.

/assign @pkprzekwas 
cc @ameukam


xmudrii:
We also need an external secret with a token for Slack.
/hold

Issue #4841, opened 2023-02-27
Feature: AWS Route 53 Health Checks for internet based dependencies


raykrueger:
Recently an issue surfaced where the Google GPG keys were returning 500 errors in various locations around the world after rotation. This was reported here a few times over the years https://github.com/kubernetes/release/issues/1007 One thing we could do is setup Health Checks in the Amazon Route 35 service on AWS to monitor these internet dependencies and alarm on failure before the community is impacted.

Example output from the health checks can be seen here https://github.com/kubernetes/release/issues/2860
The health checks I am testing with have no alarm notifications and are hosted in my own AWS account. We could build something in the community account to monitor and send notifications via email/slack.

PR #5529, opened 2023-07-07
infra/aws/tf/infrastructure-services: set up healthcheck for google-pacakges-apt

iftachk:
- introduce two new modules: 
  -  sns-via-email: setup sns topic and associated email subscription 
  - external-resource-health-check/https-health-check: setup HTTPS health checks per provided config

Requires a new backend s3 bucket per discussion with @ameukam 

ref: #4841 
/cc @xmudrii
/assign @ameukam

### DIFF
--- a/infra/aws/terraform/infrastructure-services/main.tf
+++ b/infra/aws/terraform/infrastructure-services/main.tf
@@ -1,0 +1,86 @@
+/*
+Copyright 2023 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+terraform {
+  required_version = "~> 1.5.0"
+
+  backend "s3" {
+    bucket = "<NEW_BUCKET_NAME>"
+    key    = "infrastructure-services/terraform.tfstate"
+    region = "us-east-2"
+  }
+
+  required_providers {
+    aws = {
+      source  = "hashicorp/aws"
+      version = ">= 5.7"
+    }
+  }
+}
+
+locals {
+  name = "external-dependency-health-checks"
+  external_dependency_health_checks_targets = {
+    google_cloud_pkgs_apt_gpg = {
+      name = "google-cloud-pkgs-apt-gpg"
+      fqdn = "packages.cloud.google.com"
+      resource_path = "/apt/doc/apt-key.gpg"
+    }
+  }
+  emails = [
+    "k8s-infra-alerts@kubernetes.io"
+  ]
+  sns_topic_delivery_policy = jsonencode({
+    "http" : {
+      "defaultHealthyRetryPolicy" : {
+        "minDelayTarget" : 20,
+        "maxDelayTarget" : 20,
+        "numRetries" : 3,
+        "numMaxDelayRetries" : 0,
+        "numNoDelayRetries" : 0,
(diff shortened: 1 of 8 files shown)


infra/aws/terraform/infrastructure-services/main.tf
```diff
@@ -0,0 +1,61 @@
...
+  backend "s3" {
+    bucket = "<NEW_BUCKET_NAME>"
+    key    = "infrastructure-services/terraform.tfstate"
+    region = "us-east-2"
+  }
+
+  required_providers {
+    aws = {
+      source  = "hashicorp/aws"
+      version = ">= 4.47"
```
ameukam, line 29: Do we want to use the v5 version ?

infra/aws/terraform/modules/sns-via-email/main.tf
```diff
@@ -0,0 +1,28 @@
+resource "aws_sns_topic" "this" {
+  name = var.name
+  delivery_policy = jsonencode({
```
ameukam, line 3: This will enforce the policy for every SNS topic and don't give room for customization. Can use pass the path of a json file to the variable ?

infra/aws/terraform/modules/external-resource-health-check/https-health-check/variables.tf
```diff
@@ -0,0 +1,65 @@
...
+  default = "30"
+}
+
+variable "regions" {
+  type = list(string)
+  description = "AWS regions - us-east-1 | us-west-1 | us-west-2 | eu-west-1 | ap-southeast-1 | ap-southeast-2 | ap-northeast-1 | sa-east-1"
+  default = [
+    "us-east-1",
+    "us-west-1",
+    " us-west-2",
```
ameukam, line 35:
```suggestion
    "us-west-2",
```

xmudrii:
@iftachk @ameukam This looks interesting, but is this something that's actually helpful to us at this point? I have some concerns:
- who is monitoring the given email address?
- is it possible to receive notifications only via email? Can we eventually integrate with Slack?
- how are we going to act on alerts?
- how are we going to pass those notifications to users?

My biggest concern is that we don't have any control over that repository and we don't have a good way to pass notifications down to users, meaning that this is eventually useful only to us as a signal. Given that Google repositories are going to be phased out soon™, I think that we should focus on making a solution that's going to work with OBS. That's something that we should collaborate with OBS team and see what's the best way to accomplish something similar.

Regardless, I think this is a really nice idea, we should just see how can we shape it so it provides value to a broader group (including end users).

iftachk:
@xmudrii 
Thank you for your feedback

- who is monitoring the given email address? 
That is a good question, I think the objective is to increase observability.

- is it possible to receive notifications only via email? Can we eventually integrate with Slack?
I can add another HTTPS endpoint SNS to utilize Slack's incoming webhook
Or we can [notify-slack](https://registry.terraform.io/modules/terraform-aws-modules/notify-slack/aws/latest) TF module or we use this 

- how are we going to act on alerts?
Another good question, the focus in my opinion is to have these alerts delivered to the people with access to remedy the issue, if they do not have any alerts set already (shrug)

- how are we going to pass those notifications to users?
Can you clarify what you mean by users? End users eg consumers / those that created issues of the APT GPG errors

Bottom line, responsibility needs to be established and agreed upon with all relevant stakeholders

xmudrii:
> Another good question, the focus in my opinion is to have these alerts delivered to the people with access to remedy the issue, if they do not have any alerts set already (shrug)

The issue here is that we don't know who have access to the platform. There's no one from our community (not even Google Build Admins) that have access to remedy the issue. I think it's safe to assume that Google employees maintaining that platform have some sort of monitoring and alerting and this wouldn't be much of use to them.

This is going to significantly change with OBS, however. Part of OBS infra is maintained by the OBS team at SUSE, but another part is maintained by us (SIG K8s infra and SIG Release). Proper alerting will be very helpful to know if there's anything going on that needs our attention.

> Can you clarify what you mean by users? End users eg consumers / those that created issues of the APT GPG errors

Yes, consumers that are using those repos.

PR #5532, opened 2023-07-10
Add alert mailing list for Cluster API Provider vSphere

killianmuldoon:
Add a mailing list for test failure alerts from the Cluster API vSphere Infrastructure provider.


killianmuldoon:
Will remove the hold later on today if no-one else wants to be on the list. Folks can add themselves later on if they want the notification.

PR #5539, opened 2023-07-11
eks-prow-build-cluster: kustomize flux manifests

pkprzekwas:
This pull request (PR) marks the beginning of the process of kustomizing manifests within the "resources" directory. Currently, we face limitations in configuring installation-specific settings, such as IRSA annotations on Service Accounts.

To address this, we have adopted the solution from the [official example repository](https://github.com/fluxcd/flux2-kustomize-helm-example), utilizing Kustomize.

However, this implementation introduces some boilerplate and complicates the process of adding new resources.

So far, only "boskos" and "external-secrets" have been migrated to this new structure. Before proceeding with updating the remaining manifests, I would appreciate gathering some feedback.

Here's an outline of the remaining work:
- create kustomizations for the remaining manifests (consider potentially splitting the monitoring kustomization).
- update the documentation to provide instructions on modifying and extending the current setup.
- roll out the changes on the production cluster.

/hold for discussion
/assign @xmudrii 
/cc @ameukam @dims


ameukam:
> this implementation introduces some boilerplatw and complicates the process of adding new resources.

If this implementation make onboarding new resources complicated, why do we want to pursue this approach.

pkprzekwas:
> > this implementation introduces some boilerplate and complicates the process of adding new resources.
> 
> If this implementation make onboarding new resources complicated, why do we want to pursue this approach.

The main motivation for pursuing this approach is to ensure that deployments of Kubernetes resources for both canary and production clusters are as similar as possible. By adopting this implementation, it becomes easier to maintain consistency between the two environments and reduce potential issues caused by differences in configurations.

Additionally, this approach allows for a clear distinction between canary and production environments, making it easier to identify and manage the variances between them. This clarity can streamline troubleshooting and improve overall system understanding.

It's worth noting that currently, the canary cluster is primarily utilized for manual testing purposes and is not connected to Prow. While using Kustomize with overlays follows a well-known and battle-tested pattern, the simplicity of the existing setup should also be considered.

Considering the current use case and the preference for simplicity, it is reasonable to lean towards maintaining the existing setup.

ameukam:
> > > this implementation introduces some boilerplate and complicates the process of adding new resources.
> > 
> > 
> > If this implementation make onboarding new resources complicated, why do we want to pursue this approach.
> 
> The main motivation for pursuing this approach is to ensure that deployments of Kubernetes resources for both canary and production clusters are as similar as possible. By adopting this implementation, it becomes easier to maintain consistency between the two environments and reduce potential issues caused by differences in configurations.
> 
> Additionally, this approach allows for a clear distinction between canary and production environments, making it easier to identify and manage the variances between them. This clarity can streamline troubleshooting and improve overall system understanding.
> 
> It's worth noting that currently, the canary cluster is primarily utilized for manual testing purposes and is not connected to Prow. While using Kustomize with overlays follows a well-known and battle-tested pattern, the simplicity of the existing setup should also be considered.
> 
> Considering the current use case and the preference for simplicity, it is reasonable to lean towards maintaining the existing setup.

SGTM.

/lgtm

pkprzekwas:
Closing this PR as a result of discussion above. We can get back to this pattern in the future.

Issue #5392, opened 2023-06-08
DNS REQUEST: testgrid


michelle192837:
### Type of DNS update:

Update

### Domain being modified:

testgrid.k8s.io and testgrid.kubernetes.io

### Existing DNS Record:

```yaml
testgrid:
  - type: A
    values:
    - 216.239.32.21
    - 216.239.34.21
    - 216.239.36.21
    - 216.239.38.21
  - type: AAAA
    values:
    - 2001:4860:4802:32::15
    - 2001:4860:4802:34::15
    - 2001:4860:4802:36::15
    - 2001:4860:4802:38::15
```

### New DNS Record:

```yaml
testgrid:
  - type: A
    values:
    - 34.120.51.46
```

### Reason for update:

Updating the TestGrid site to use a load balancer instead of the default App Engine access.

BenTheElder:
@michelle192837 if you want to speed these along you can file a PR with the yaml you wrote here and we can just review it / let automation enact it.

michelle192837:
I meant to do so and forgot to get to it 😅  Thanks for the reminder, I'll make one today or tomorrow.

michelle192837:
Re-opening until I verify this works.

PR #5542, opened 2023-07-11
Point testgrid.k8s.io to a load balancer.

michelle192837:
Ref #5392. The load balancer should be set up to switch without downtime now, so theoretically this should work well!

/hold

Will remove hold when I can watch this, in case of rollback.


michelle192837:
Thanks folks!

PR #5554, opened 2023-07-12
Revert "Point testgrid.k8s.io to a load balancer."

michelle192837:
Reverts kubernetes/k8s.io#5542

Unfortunately while the load balancer setup and certificate seem to be provisioned correctly, something must be wrong between them because https://testgrid.k8s.io doesn't appear to resolve.


michelle192837:
/assign @ameukam

Assigning in case you're around since you approved the last one. Thanks in advance! ^^

michelle192837:
I tried to fiddle around with the load balancer and certificate, and somewhere in there it seems to have fixed things! Closing this PR since I don't think it's necessary anymore, apologies for the noise.

PR #5559, opened 2023-07-13
Fix IAM permissions for obs-admin users to allow using rclone

xmudrii:
The way we configured `obs-admin` user caused `rclone lsd` and `rclone copy` operations to fail with permission errors. That's because:

- `s3:ListAllMyBuckets` cannot be scoped down to particular buckets, it must use `*` as a resource
- There's a bug in `rclone` requiring a `s3:CreateBucket` permission in some cases (https://github.com/rclone/rclone/issues/5119)
  - This is no-op in our case because the bucket already exists

/assign @pkprzekwas @ameukam


saschagrunert (approved):

xmudrii:
Applied to prod and the issue is solved.
/hold cancel

PR #5597, opened 2023-07-21
cdn.dl.k8s.io: Reduce the first byte timeout

ameukam:
Related to:

Following guidance from https://developer.fastly.com/learning/concepts/cache-freshness/#best-practices and remove the value set in the terraform, the first byte timeout is reduce to 15000 milliseconds.


TerryHowe (approved):
/lgtm

PR #5598, opened 2023-07-21
cdn.dl.k8s.io: Add a BigQuery dataset for logging

ameukam:
Ensure a BigQuery dataset exists if we want to use Fasty's Real-Time Log streaming. See: https://docs.fastly.com/en/guides/log-streaming-google-bigquery

A minimal configuration is defined to ensure logs are sent to BigQuery


ameukam:
cc @dims

PR #5603, opened 2023-07-24
dl.k8s.io: Switch to Fastly

ameukam:
Related to:


BenTheElder (commented):
/lgtm
/approve

aojea:
This is great, is there any plan on how to measure the impact of this?

BenTheElder:
We (=googlers) have observability on the bandwidth to the old GCS bucket and k8s infra leads have observability to the bandwidth through fastly and cache hit rate etc.

We're monitoring currently.

In the future other community members will have observability for the backing bucket when we replace it with a k8s infra owned release bucket, but first we'll get this working.

BenTheElder:
https://github.com/kubernetes/k8s.io/issues/5726 for follow up data

PR #5606, opened 2023-07-24
cdn.dl.k8s.io: Fix BigQuery for Fastly logging

ameukam:
Follow-up of:
  - https://github.com/kubernetes/k8s.io/pull/5598

Ensure a service account to write log to the BigQuery dataset and impersonated by the Fastly service account.


ameukam:
cc @dims @BenTheElder

PR #5615, opened 2023-07-24
fix the other affected dl.k8s.io test case

BenTheElder:
cdn.dl.k8s.io is also used even for invalid versions in the current config, and this is expected

follow up to #5614

### DIFF
--- a/apps/k8s-io/test.py
+++ b/apps/k8s-io/test.py
@@ -293,7 +293,7 @@ class RedirTest(HTTPTestCase):
             # Not a valid release version (gamma)
             self.assert_temp_redirect(
                 base + '/v1.2.3-gamma.4/kubernetes.tar.gz',
-                'https://storage.googleapis.com/kubernetes-release/v1.2.3-gamma.4/kubernetes.tar.gz')
+                'https://cdn.dl.k8s.io/v1.2.3-gamma.4/kubernetes.tar.gz')
             # A few /ci/ tests
             self.assert_temp_redirect(
                 base + '/ci/v$ver/$path',


apps/k8s-io/test.py
```diff
@@ -293,7 +293,7 @@ def test_dl(self):
             # Not a valid release version (gamma)
             self.assert_temp_redirect(
                 base + '/v1.2.3-gamma.4/kubernetes.tar.gz',
-                'https://storage.googleapis.com/kubernetes-release/v1.2.3-gamma.4/kubernetes.tar.gz')
+                'https://cdn.dl.k8s.io/v1.2.3-gamma.4/kubernetes.tar.gz')
```
xmudrii, line 296: It appears that this change is causing the `post-k8sio-deploy-app-k8s-io` postsubmit to consistently fail (e.g. https://prow.k8s.io/view/gs/kubernetes-jenkins/logs/post-k8sio-deploy-app-k8s-io/1684700242429087744). The nginx rule is only matching alpha, beta, and rc prereleases:

https://github.com/kubernetes/k8s.io/blob/7e04d9ed4436049ccaaaae597b7af4ad27024167/apps/k8s-io/configmap-nginx.yaml#L187-L188

@BenTheElder @dims @ameukam I think that we should revert this change.

xmudrii: Created revert PR #5652, we can discuss there

BenTheElder: we should not ever be redirecting to the GCS bucket, we need to be able to replace the GCS bucket with a community owned bucket shielded by fastly. the bug is in the redirect rules then :-)

PR #5645, opened 2023-07-27
aaa: Drop workloads monitoring

ameukam:
aaa cluster is upgraded to 1.25. "WORKLOADS" option has been removed since 1.24. Enable managed prometheus to a replacement


ameukam:
cc @dims @BenTheElder

PR #5649, opened 2023-07-27
k8s.io: update redirect config for dl.k8s.io

ameukam:
Ensure we can also redirect any request matching `dl.k8s.io/$1` to cdn.dl.k8s.io


BenTheElder (commented):
probably need to update tests? :-)

ameukam:
> probably need to update tests? :-)

Done. only one right ?

BenTheElder (commented):
/lgtm

Issue #5641, opened 2023-07-27
eks-prow-build-cluster: Configure EKS Node Group dedicated for KIND jobs


pkprzekwas:
While investigating [cluster instability issues](https://github.com/kubernetes/k8s.io/issues/5473), we noticed some correlation between failing nodes and types of jobs running on them. We suspect that kind jobs might be the reason for node problems. 

To proof our hypothesis we want to configure a dedicated node group with appropriate labels and taints so that we could schedule there kind jobs.

BenTheElder:
Erm ... can we CC kind maintainers on issues like this in the future? cc @aojea

PR #5651, opened 2023-07-28
eks-prow-build-cluster: Introduce node selector and taint for blue node group

pkprzekwas:
As mentioned in https://github.com/kubernetes/k8s.io/issues/5641 we need this setup to test if the node rotation problem is caused by kind prow jobs. This PR decorates the blue node group with label and taint so that we can use it exclusively for kind prow jobs.

/assign @xmudrii 
/cc @ameukam @dims


xmudrii (approved):
/lgtm
/approve

PR #5656, opened 2023-07-28
Fix affected redirection tests for pkgs.k8s.io

xmudrii:
`pkgs.k8s.io` is using temporary redirection same as old endpoints (`apt.kubernetes.io` and `yum.kubernetes.io`). This requires adjusting redirection tests as done in this PR.

/assign @dims @BenTheElder @cblecker


cblecker (approved):
/lgtm

PR #5669, opened 2023-08-01
Update distributors-announce group settings

enj:
Defaults for config from https://github.com/kubernetes/k8s.io/blob/85e692779d056ec2c890b69c90d7d63db2469731/groups/service.go#L456-L491 (only non-default values are manually set).


joelsmith:
Thanks!
/lgtm

PR #5682, opened 2023-08-04
GKE: Enable GKE CSI driver

ameukam:
Starting with GKE 1.25, GKE CSI driver is enabled by default. Reconciciling the terraform code to reflect the change introduced. See: https://cloud.google.com/kubernetes-engine/docs/release-notes#June_08_2023


ameukam:
cc @dims @BenTheElder

PR #5696, opened 2023-08-08
Adding k8s-infra-kops-scale-tests bucket with public read

pkprzekwas:
Creating dedicated bucket with public read access for kops scale tests (as discussed [here](https://kubernetes.slack.com/archives/CCK68P2Q2/p1691083904395479?thread_ts=1689009672.882279&cid=CCK68P2Q2))

/assign @ameukam @dims 
/cc @prateekgogia

### DIFF
--- a/infra/aws/terraform/s3/k8s-infra-e2e-boskos-scale-001/.gitignore
+++ b/infra/aws/terraform/s3/k8s-infra-e2e-boskos-scale-001/.gitignore
@@ -1,0 +1,21 @@
+# Local .terraform directories
+**/.terraform/*
+
+# .tfstate files
+*.tfstate
+*.tfstate.*
+*.tfplan
+
+# Crash log files
+crash.log
+
+# Ignore override files as they are usually used to override resources locally and so
+# are not checked in
+override.tf
+override.tf.json
+*_override.tf
+*_override.tf.json
+
+# Ignore CLI configuration files
+.terraformrc
+terraform.rc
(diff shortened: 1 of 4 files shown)


infra/aws/terraform/s3/k8s-infra-e2e-boskos-scale-001/providers.tf
```diff
@@ -0,0 +1,28 @@
...
+*/
+
+terraform {
+  required_providers {
+    aws = {
+      source  = "hashicorp/aws"
+      version = "~> 4.0"
+    }
+  }
+}
```
ameukam, line 24: @pkprzekwas   all resources created here should live in a remote backend. Which backend is used ?

pkprzekwas: Ah, fair point. We don't have any bucket for TF state on this account yet. Let me create and use it.

pkprzekwas: @ameukam added configuration for remote tf state.

ameukam:
/approve

Leave to @prateekgogia

infra/aws/terraform/s3/k8s-infra-e2e-boskos-scale-001/.gitignore
```diff
@@ -0,0 +1,21 @@
+# Local .terraform directories
+**/.terraform/*
+
```
ameukam, line 3: @pkprzekwas suggestion: we should probably move this .gitignore file to `../infra/aws/terraform` so all the sub-repos inherit this config.

pkprzekwas: We can think about having some generic ignore at the top level of the terraform directory. But that should be a dedicated PR that would also remove other ignores.

prateekgogia:
LGTM `(Don't have permissions to approve)`

PR #5700, opened 2023-08-08
AWS: add EKS cluster for kOps CI

ameukam:
Related to:
  - https://github.com/kubernetes/k8s.io/issues/5127

Add an EKS cluster that will be used as build cluster for kOps


ameukam:
cc @pkprzekwas @hakman

PR #5763, opened 2023-08-28
sbom.k8s.io: Allow file extensions

saschagrunert:
We should allow to retrieve the `.sig`, `.cert`, `.sha256` or `.sha512` files of the SBOM's via sbom.k8s.io. This mean we now accept any suffix to the nginx rewrite and pass that down to the actual URL.

Refers to https://github.com/kubernetes/release/issues/3222#issuecomment-1695197942

### DIFF
--- a/apps/k8s-io/configmap-nginx.yaml
+++ b/apps/k8s-io/configmap-nginx.yaml
@@ -329,8 +329,8 @@ data:
         server_name sbom.k8s.io sbom.kubernetes.io
         listen 80;
 
-        rewrite ^/(.*)?/release$ https://storage.googleapis.com/kubernetes-release/release/$1/kubernetes-release.spdx redirect;
-        rewrite ^/(.*)?/source$  https://storage.googleapis.com/kubernetes-release/release/$1/kubernetes-source.spdx redirect;
+        rewrite ^/(.*)?/release(\.cert|\.sig|\.sha256|\.sha512)?$ https://storage.googleapis.com/kubernetes-release/release/$1/kubernetes-release.spdx$2 redirect;
+        rewrite ^/(.*)?/source(\.cert|\.sig|\.sha256|\.sha512)?$  https://storage.googleapis.com/kubernetes-release/release/$1/kubernetes-source.spdx$2 redirect;
       }
 
       server {


palnabarun (commented):
/lgtm

apps/k8s-io/configmap-nginx.yaml
```diff
@@ -329,8 +329,8 @@ data:
         server_name sbom.k8s.io sbom.kubernetes.io
         listen 80;
 
-        rewrite ^/(.*)?/release$ https://storage.googleapis.com/kubernetes-release/release/$1/kubernetes-release.spdx redirect;
-        rewrite ^/(.*)?/source$  https://storage.googleapis.com/kubernetes-release/release/$1/kubernetes-source.spdx redirect;
+        rewrite ^/(.*)?/release(.*)?$ https://storage.googleapis.com/kubernetes-release/release/$1/kubernetes-release.spdx$2 redirect;
+        rewrite ^/(.*)?/source(.*)?$  https://storage.googleapis.com/kubernetes-release/release/$1/kubernetes-source.spdx$2 redirect;
```
xmudrii, line 333: Nit: given that we already know possible extensions (`.sig` and `.cert`), should we limit this regex to these two extensions? For example, something like:

```suggestion
        rewrite ^/(.*)?/release(\.cert|\.sig)?$ https://storage.googleapis.com/kubernetes-release/release/$1/kubernetes-release.spdx$2 redirect;
        rewrite ^/(.*)?/source(\.cert|\.sig)?$  https://storage.googleapis.com/kubernetes-release/release/$1/kubernetes-source.spdx$2 redirect;
```

One more benefit from this is that potentially malicious/spam requests are going to be terminated by nginx instead of getting routed to the origin bucket.

xmudrii (approved):
Left a nit, but in general looks okay to me.
/lgtm
/approve
/assign @dims

apps/k8s-io/configmap-nginx.yaml
saschagrunert: Changed as suggested, but added the SHA extensions as well.

xmudrii (approved):
/lgtm
/approve

justinsb:
Thanks for addressing @saschagrunert 

/approve
/lgtm

PR #5781, opened 2023-08-30
Updating to current sig cloud provider chairs

bridgetkromhout:
Update to match https://github.com/kubernetes/community/blob/master/OWNERS_ALIASES#L28-L33.


elmiko:
/lgtm

cc @dims might you be able to help here?

PR #5833, opened 2023-09-12
Add a service account for provider-aws-test-role

dims:
we have one for `node-e2e-tests` role, but not for `provider-aws-test-role` so adding it here


ameukam:
/lgtm

I think this is need to be deployed manually. cc @xmudrii @pkprzekwas

Issue #1458, opened 2020-12-02
Registries for used in k/k CI should be on Kubernetes Community


justaugustus:
ref: @spiffxp at https://github.com/kubernetes/kubernetes/issues/97002#issuecomment-737435131:

> https://kubernetes.slack.com/archives/C09QZ4DQB/p1606896985218000
> 
> The project hosting the GCR repo was swept up by a security audit because it hadn't been properly accounted for. That change has been reverted. Now waiting to see affected jobs go back to green.
> 
> We should create a community-owned equivalent project, I'll open a followup issue for that

`gcr.io/k8s-authenticated-test` is the specific case here, but we should try to cover as many as we find.

/assign
/sig release
/area release-eng artifacts
cc: @kubernetes/release-engineering @kubernetes/ci-signal @thockin 

Tagging since they also expressed interest:
@thejoycekung @jeremyrickard @aojea

spiffxp:
/milestone v1.21
This covers a subset of projects in umbrella issue https://github.com/kubernetes/k8s.io/issues/1469

spiffxp:
Per https://github.com/kubernetes/k8s.io/issues/1522#issuecomment-911980801 we are now down to the test-case-specific registries mentioned in the description.  We need assistance from specific SIGs to move forward on deciding if the test cases are valid, need to be rewritten, or can be dropped.

spiffxp:
@hh proposing a rewrite of tests to standup private registry in cluster

spiffxp:
I'll punt this out of v1.23 once we're in test freeze.  We managed to complete kubernetes-e2e-test-images this cycle, but waiting on a deprecation announcement

BenTheElder:
> gcr.io/k8s-authenticated-test is the specific case here, but we should try to cover as many as we find.

this and the other authenticated test repo are pretty much the worst, and just an overall antipattern

> gke-release: https://github.com/kubernetes/k8s.io/issues/1525

at this point this is basically only used to setup GPUs on GCE, the problem with exiting this one is thorny licensing, we're not depending on it in test cases, release artifacts, etc, just setting up GPUs on kube-up.sh

we could probably figure it out, but I'm not sure it's worth it. we're always going to be dependent on a third-party for the forseeable future on gpu setup, and we only have like one job using this.

the authenticated test tests are not great, and I'm leaning towards we should just delete the tests due to lack of response on sorting out a sustainable alternative (which would require input from the owners of the code under test)

[...]

PR #5854, opened 2023-09-15
Recreate the `k8s-authenticated-test` GCP project as `k8s-staging-authenticated-test` on kubernetes.io GCP Org

upodroid:
Fixes: https://github.com/kubernetes/kubernetes/issues/97026
Part of https://github.com/kubernetes/k8s.io/issues/1458

The following tests in k/k are failing quite frequently on kops clusters and for anyone who doesn't run e2e tests on specific projects.

```
[sig-apps] ReplicaSet should serve a basic image on each replica with a private image
[sig-apps] ReplicationController should serve a basic image on each replica with a private image
```

Sample failure:
https://prow.k8s.io/view/gs/kubernetes-jenkins/logs/e2e-kops-gce-cni-cilium-k8s-ci/1702296867670331392

This particular image `gcr.io/k8s-authenticated-test/agnhost:2.6` is pulled for testing and runs in a google.com GCP org.

We can't get rid of these tests till the in-tree Kubelet auth providers are gone.

I'm proposing the images live in a new location that is owned by the community and the images can be pulled by any google service account.

The new pull location for the image will be `us-central1.docker.pkg.dev/k8s-staging-authenticated-test/images/agnhost:2.6`

/cc @dims @ameukam @bentheelder

### DIFF
--- a/infra/gcp/bash/ensure-main-project.sh
+++ b/infra/gcp/bash/ensure-main-project.sh
@@ -73,6 +73,7 @@ readonly DNS_GROUP="k8s-infra-dns-admins@kubernetes.io"
 # - entry syntax is "bucket_name:owners_group" (: is invalid bucket name char)
 readonly TERRAFORM_STATE_BUCKET_ENTRIES=(
     "${LEGACY_CLUSTER_TERRAFORM_BUCKET}:${CLUSTER_ADMINS_GROUP}"
+    k8s-staging-authenticated-test-tf:"${CLUSTER_ADMINS_GROUP}"
     k8s-infra-tf-aws:k8s-infra-aws-admins@kubernetes.io
     k8s-infra-tf-fastly:k8s-infra-fastly-admins@kubernetes.io
     k8s-infra-tf-gcp:k8s-infra-gcp-org-admins@kubernetes.io
--- a/infra/gcp/infra.yaml
+++ b/infra/gcp/infra.yaml
@@ -345,3 +345,5 @@ infra:
       k8s-staging-win-op-rdnss:
       k8s-staging-gcb-untrusted:
         managed_by: infra/gcp/bash/ensure-staging-gcb-untrusted.sh
+      k8s-staging-authenticated-test:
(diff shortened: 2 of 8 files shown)


ameukam:
I wonder if we want to migrate this registry.  There was a [conversation](https://github.com/kubernetes/kubernetes/issues/97026#issuecomment-738500525) about removing the tests using `gcr/k8s-authenticated-test` . 
Do we want input from sig-node before doing this ? @dims @BenTheElder

upodroid:
most recent convo https://kubernetes.slack.com/archives/C7J9RP96G/p1694796571586009

BenTheElder:
I would prefer to see an alternative as I think trying to host a permanent authenticated image endpoint is a liability (also don't forget when something like this happens now old release tests are broken)

> We can't get rid of these tests till the in-tree Kubelet auth providers are gone.

I don't think that's true, again you can pass generic auth that isn't specific to say, GCP, by way of e.g. a secret in a namespace. That's generic and I don't see that being ripped out of tree.


Suggestions:
- Run a local registry in the cluster (see past discussion though of why this may not work portably due to k8s networking)
- Custom app similar to registry.k8s.io and the tiniest image possible w/ custom auth (lots of extra work / infra for one test ...)
- This PR or something similar (... but acknowledge that it's likely to break in the future)

upodroid:
> Suggestions:
> 
> * Run a local registry in the cluster (see past discussion though of why this may not work portably due to k8s networking)
> * Custom app similar to registry.k8s.io and the tiniest image possible w/ custom auth (lots of extra work / infra for one test ...)
> * This PR or something similar (... but acknowledge that it's likely to break in the future)

Let's go with the 3rd option and revisit this problem in the future. I don't see anyone who is ready to do 1 and 2. Replacing kube-up clusters is a better use of our time than doing a ton of extra work for a test that shouldn't even be there in the first place.

ameukam:
@aojea Any concern about picking the third option ?

PR #5857, opened 2023-09-16
Create the k8s-kops-test SA on k8s-infra-prow-build cluster

upodroid:
Part of https://github.com/kubernetes/test-infra/issues/29722

/cc @ameukam @hakman 

The new test I created in https://github.com/kubernetes/test-infra/pull/30719 is failing to start because the k8s-infra-prow-build cluster is missing the k8s-kops-test service account.

https://prow.k8s.io/?job=*cos-105*&state=error

If you look at https://github.com/kubernetes/test-infra/blob/506a5f2d067db7231a96d3773028d0f7186fa1c5/config/prow/cluster/build/build_serviceaccounts.yaml#L47 it is configured to use the service account that is used for running tests in the boskos pools.


ameukam:
/approve
/hold

@hakman @justinsb do we need to set a new gcs bucket for kops state ?

upodroid:
No, the tester creates a state bucket on each run and deletes it

PR #5861, opened 2023-09-17
prow-build: fix test-pods PodDisruptionBudget resource

ameukam:
PodDisruptionBudget is now GA since GKE 1.25


ameukam:
cc @hakman @upodroid

Issue #5629, opened 2023-07-25
dl.k8s.io: Set long TTLs in Fastly cache


ameukam:
Part of:
  - #4528 

According to https://developer.fastly.com/learning/concepts/cache-freshness/, in order to maximize our hit ratio, we increase our TTLs in the Cache-Control headers and caching everything in Fastly.

Referemce:
  - https://www.fastly.com/blog/maintainable-vcl

cc @haubles @ttriblondon 

/area infra
/kind feature
/milestone v1.28

ameukam:
cc @triblondon

ameukam:
TTL is settled for 30 days. 

/close

PR #5871, opened 2023-09-19
dl.k8s.io: Set the TTL of the Fastly service

ameukam:
Related to:
  - https://github.com/kubernetes/k8s.io/issues/5629

Attempting to set the TTL of the Fastly to 30 days. this cannot be set at the origin. This configuration should ensure the objects are cached at the edge.

Also remove HTTP headers from GCS.


ameukam:

cc @dims

PR #5882, opened 2023-09-22
add mailing list members migration functionality

Priyankasaggu11929:
### Description

This PR attempts to add functionality for migrating mailing list members across two google groups.

- add `MigrateMailingListMembers` function in `groups/service.go` to implement the migration logic

- add `PerformMailingListMigration function` function and new flags ( `--migrate`, `--destinationGroup`,  `--sourceGroup`) in `groups/service.go` to trigger mailing list members migration

- add a new Makefile target –`migrate-members` to execute the migration with specified source and destination groups as:

  ``` 
  make migrate-members -- --sourceGroup "source-group-email@example.com" --destinationGroup "destination-group-email@example.com" --migrate --confirm
  ```
- add test `TestMigrateMailingListMembers` in `groups/service_test.go`

**Note**:

I haven't been able to test it yet due to having no credentials for API calls.
Once it's reviewed and merged, I'll try it in a prow job similar to [existing ones](https://github.com/kubernetes/test-infra/blob/fa128e1d50e064a60f83cf523b3b78663e2170f6/config/jobs/kubernetes/wg-k8s-infra/trusted/wg-k8s-infra-groups.yaml#L2-L26).
  
/assign @mrbobbytables @MadhavJivrajani (for early feedback)
/sig contributor-experience

### DIFF
--- a/groups/reconcile.go
+++ b/groups/reconcile.go
@@ -131,6 +131,9 @@ func main() {
 	confirmChanges := flag.Bool("confirm", false, "false by default means that we do not push anything to google groups")
 	printConfig := flag.Bool("print", false, "print the existing group information")
 	numWorkers := flag.Int("workers", defaultNumWorkers, "number of concurrent workers to use")
+	sourceGroupEmail := flag.String("source-group", "", "Email address of the source google group")
+	destinationGroupEmail := flag.String("destination-group", "", "Email address of the destination google group")
+	migrateMembers := flag.Bool("migrate", false, "Enable migration of members")
 
 	flag.Usage = Usage
 	flag.Parse()
@@ -191,6 +194,20 @@ func main() {
 		log.Fatal(err)
 	}
 
+	if *migrateMembers && *confirmChanges{
+		log.Printf("print: migrate flag set to %v, will perform mailing lists migration", *migrateMembers)
+		if *sourceGroupEmail == "" || *destinationGroupEmail == "" {
+			log.Fatal("Both sourceGroup and destinationGroup must be provided")
+		}
+		err := r.PerformMailingListMigration(*sourceGroupEmail, *destinationGroupEmail)
+		if err != nil {
+			fmt.Printf("Error migrating members: %v\n", err)
+			return
+		}
+		log.Printf("Members migrated successfully from %v to %v!", *sourceGroupEmail, *destinationGroupEmail)
+		return
+	}
+
 	if *printConfig {
 		err = r.printGroupMembersAndSettings()
 		if err != nil {
@@ -227,6 +244,13 @@ func NewReconciler(ctx context.Context, clientOption option.ClientOption, numWor
 	}
 
 	return &Reconciler{adminService: as, groupService: gs, numWorkers: numWorkers}, nil
+}
+
+
+func (r *Reconciler) PerformMailingListMigration(sourceGroupEmail, destinationGroupEmail string) error {
+	sourceGroup := GoogleGroup{EmailId: sourceGroupEmail}
+	destinationGroup := GoogleGroup{EmailId: destinationGroupEmail}
+	return r.adminService.MigrateMailingListMembers(sourceGroup, destinationGroup)
 }
 
 func (r *Reconciler) ReconcileGroups(groups []GoogleGroup) error {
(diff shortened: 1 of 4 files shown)


groups/service.go
```diff
@@ -507,6 +508,46 @@ func (gs *groupService) UpdateGroupSettings(group GoogleGroup) error {
...
+	// iterate through source group members and add/update them in the destination group
+	for _, sourceMember := range sourceGroupMembers {
+		err := as.AddOrUpdateGroupMembers(destinationGroup, sourceMember.Role, []string{sourceMember.Email})
+		if err != nil {
+			logErr := fmt.Errorf("unable to add/update %s to %q as %s: %w", sourceMember.Email, destinationGroup.EmailId, sourceMember.Role, err)
+			log.Printf("%s\n", logErr)
+			errs = append(errs, logErr)
+			continue
+		}
+	}
```
MadhavJivrajani, line 546: Is it possible to replace this loop with a single call like such:
```go
sourceMemberEmailIDs := getEmailIDs(sourceGroupMembers)
err := as.AddOrUpdateGroupMembers(destinationGroup, sourceMember, sourceMemberEmailIDs)
```

groups/reconcile.go
```diff
@@ -131,6 +131,9 @@ func main() {
 	confirmChanges := flag.Bool("confirm", false, "false by default means that we do not push anything to google groups")
 	printConfig := flag.Bool("print", false, "print the existing group information")
 	numWorkers := flag.Int("workers", defaultNumWorkers, "number of concurrent workers to use")
+	sourceGroupEmail := flag.String("sourceGroup", "", "Email address of the source google group")
+	destinationGroupEmail := flag.String("destinationGroup", "", "Email address of the destination google group")
```
MadhavJivrajani, line 135: Non-blocking nit
I'd prefer flag names to be as `source-group` and `destination-group`.

MadhavJivrajani (commented):
LGTM on a high level, just one question. Thanks @Priyankasaggu11929!

MadhavJivrajani:
> I haven't been able to test it yet due to having no credentials for API calls.

@Priyankasaggu11929 there is a fake client for groups: https://github.com/kubernetes/k8s.io/blob/main/groups/fake/fake_client.go

We should also update the methods there so that we can test the overall logic. Agreed that behaviour with actual API calls can't really be tested.

groups/service.go
Priyankasaggu11929: Two things here:

- In the `AddOrUpdateGroupMembers` function, the second argument is taking the member's `Role`.
   ```
   func (as *adminService) AddOrUpdateGroupMembers(group GoogleGroup, role string, members []string) error {...}
   ```

- So, we're using a loop to process one Mailing List member at a time from `sourceGroupMembers` list.
  We pass their `Role` and `Email` to `AddOrUpdateGroupMembers()` to update the destination group without affecting their existing role.

  ```
  err := as.AddOrUpdateGroupMembers(destinationGroup, sourceMember.Role, []string{sourceMember.Email})
  ```

Please let me know if I misunderstood your suggestion.

groups/reconcile.go
Priyankasaggu11929: updated to ^ suggested flags. Thanks! :slightly_smiling_face:

Priyankasaggu11929:
> We should also add tests for this (as per your TODO): https://github.com/kubernetes/k8s.io/pull/5882#issuecomment-1731840770

added tests in the latest commit refresh.

@MadhavJivrajani, PTAL again. Thanks for the suggestions! :slightly_smiling_face:

cblecker:
@Priyankasaggu11929 How do you expect this to be run? Is it something a user (who presumably has credentials with sufficient permissions) runs on their local system?

The thought process I have looking at this:
- There aren't many users with sufficient credentials to do this, outside of automation
- Automation needs to be declarative in a repo in order for us to point a bot to it
- We want owners/managers to always be declared in the yaml in the repo as authoritative
- For some lists (like a leads list), we want all the users to always be declared in the yaml too
- We don't want the members of some e-mail lists (like a sig list that anyone can just click the button to join) to be declared in the yaml in the repo
- But how do we migrate an old list to a new list without declaring the individual members in the repo?
- Also, there is throttling by Google for how many emails you can add to a list in a 24 hour period, so migrating a large list could take many runs over the course of the day. And you'd want to pick up where you left off adding folks.

Priyankasaggu11929:
/remove-lifecycle stale

Ping @cblecker @MadhavJivrajani for another round of review. :slightly_smiling_face:

Priyankasaggu11929:
@palnabarun, @mrbobbytables – could you help (with steering elevated permissions) to perform the mailing list migrations?

I'll then close this PR in favor of that. Thanks!

PR #5933, opened 2023-10-05
eks-prow-build-cluster: Fix instance size for green node group

xmudrii:
The green node group has been using a wrong instance type (Intel instances instead of AMD instances). This PR fixes this by switching green node group to AMD instances.

/assign @pkprzekwas @ameukam @dims


iftachk (approved):

upodroid (approved):
LGTM

PR #5935, opened 2023-10-05
chore: Standardize and consolidate the configurations in the `infra/aws/terraform/` projects

bryantbiggs:
Note: happy to split this out into separate PRs per directory. However, despite the large changeset, the truth will lie in the Terraform plan output which should be a no-op diff (unfortunately I cannot see that to confirm/validate)

- Consolidate the `.gitignore` files to use just one `.gitignore` at the root of the `infra/aws/terraform` directory; these are all nearly identi
- Consolidated the boilerplate requirements for a project workspace into `main.tf`
  - This contains the providers, the remote backend,  and any common locals or data sources. Since this contains the entrypoint required for Terraform, its conventionally called `main.tf` for a project workspace. Common locals and data sources are those that are used throughout the workspace and are not specific to any section of configuration - those that are specific to a particular section are defined in close proximity of their use
  - This reduces the number of small files with 1-3 lines of relevant configuration, and makes the layout consistent 
- Move any variables and outputs to their respective `variables.tf` and `outputs.tf` while ensuring they have descriptions
  - Variables are now consistently ordered with `desription`, `type`, `default`
- Region data sources replaced with local variables in project workspaces since we already have this information - modules retain their data sources since these are dynamic via their reflection use
- Reduce the use of aliased providers when using the default provider is sufficient - we only need an aliased provider when access is required to two or more regions/accounts (we should try to reduce this pattern since it conflicts with regional isolation)
- The required Terraform version for workspaces is set at `~> 1.1` - I didn't see any 1.2+ features in use that warrant a higher required version
- Fixed various lints identified by `tflint` - the `terraform_naming_convention` is disabled since fixing that requires resource naming and the use of move blocks so that would be handled in a separate PR

Note: I did not do much in `prow-build-cluster` other than fix the lint checks since this was already massive and theres a lot in that directory


infra/aws/terraform/.pre-commit-config.yaml
```diff
@@ -17,7 +17,7 @@ repos:
           - '--args=--only=terraform_documented_variables'
           - '--args=--only=terraform_typed_variables'
           - '--args=--only=terraform_module_pinned_source'
-          - '--args=--only=terraform_naming_convention'
+          # - '--args=--only=terraform_naming_convention'
```
bryantbiggs, line 20: disabled for now to get reset of checks passing - this will require renaming resources and that requires a bit more care

infra/aws/terraform/audit-account/main.tf
```diff
@@ -0,0 +1,64 @@
...
+################################################################################
+# Common Locals
+################################################################################
+
+locals {
+  region = "us-east-2"
+
+  audit-account-name  = "k8s-infra-security-audit"
+  audit-account-index = index(data.aws_organizations_organization.current.accounts[*].name, local.audit-account-name)
+  audit-account-id    = data.aws_organizations_organization.current.accounts[local.audit-account-index].id
```
bryantbiggs, line 52: this is unfortunately a cyclical loop - using the data source which needs to make an AWS API call to fetch some data -> that requires an authenticated provider -> the authenticated provider is trying to assume a role with a computed value -> that computed value is being derived from the data source API call -> repeat

 
image

infra/aws/terraform/iam/k8s-infra-prow-canary/terraform.tfvars
```diff
@@ -1,23 +0,0 @@
...
-Unless required by applicable law or agreed to in writing, software
-distributed under the License is distributed on an "AS IS" BASIS,
-WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
-See the License for the specific language governing permissions and
-limitations under the License.
-*/
-
-region = "us-east-2"
-
-eks_infra_admins = [
```
bryantbiggs, line 19: FYI - having access provisioned this way could allow someone to supply a different list of admins than what is shown here. I suspect these changes are manually applied by only a select group of users so less of a concern, but its still something to be aware of.

infra/aws/terraform/kops-infra-ci/eks.tf
```diff
@@ -132,21 +131,20 @@ module "eks" {
       }
 
       tags = merge(var.tags, {
-        "region" = "${data.aws_region.current.name}"
+        "region" = local.region
```
bryantbiggs, line 134: I am curious to learn more about the use of a region tag - is this used by something?

infra/aws/terraform/kops-infra-ci/main.tf
```diff
@@ -14,18 +14,67 @@ See the License for the specific language governing permissions and
...
+  region = "us-east-2"
 
-  # AWS wants the thumbprint of the the top intermediate certificate authority.
-  thumbprint_list = [
-    # GlobalSign root certificate (Google Managed Certficates)
-    "08745487e891c19e3078c1f2a07e452950ef36f6"
-  ]
+  kops-infra-ci-name       = "kops-infra-ci"
+  kops-infra-ci-index      = index(data.aws_organizations_organization.current.accounts[*].name, local.kops-infra-ci-name)
+  kops-infra-ci-account-id = data.aws_organizations_organization.current.accounts[local.kops-infra-ci-index].id
```
bryantbiggs, line 71: this too suffers from the same cyclical loop issue

image

infra/aws/terraform/modules/tag-policy/outputs.tf
```diff
@@ -15,13 +15,21 @@ limitations under the License.
...
 
 output "scp_deny_tag_deletion_id" {
-  value = aws_organizations_policy.deny_tag_deletion.id
+  description = "The unique identifier (ID) of the deny tag policy"
+  value       = aws_organizations_policy.deny_tag_deletion.id
+}
+
+output "name" {
+  description = "TODO - name does not quite match the value"
+  value       = local.selected_services
```
bryantbiggs, line 34: FYI - there was already this output and I don't know what might be using it but it does appear that the name of the output doesn't quite align with the value. It didn't have a description but it would be good to get clarification on this output and update accordingly

infra/aws/terraform/prow-build-cluster/bastion.tf
```diff
@@ -16,7 +16,7 @@ limitations under the License.
 
 data "aws_ami" "ubuntu" {
   most_recent = true
-  owners      = ["099720109477"] // Canonical
+  owners      = ["099720109477"] # Canonical
```
bryantbiggs, line 19: in some places we use data sources (great!) for getting the AMI ID, and others we specify the AMI ID as a static string - any reason for the static string or can we standardize on the data source so the correct AMI is pulled based on the given Kubernetes version?

infra/aws/terraform/artifacts.k8s.io/main.tf
```diff
@@ -14,189 +14,110 @@ See the License for the specific language governing permissions and
...
-    condition     = can(regex(".*-$|^$", var.prefix))
-    error_message = "The string must end with a hyphen or be empty."
-  }
-}
-
-
-module "bucket_ap-northeast-1" {
-  source = "./s3"
+terraform {
+  required_version = "~> 1.1"
```
ameukam, line 18:
```suggestion
  required_version = "~> 1.5.0, <=1.5.4"
```

infra/aws/terraform/infrastructure-services/main.tf
```diff
@@ -15,35 +15,39 @@ limitations under the License.
...
 
 terraform {
-  required_version = "~> 1.5.0"
-
-  backend "s3" {
-    bucket = "k8s-infra-tf-shared-services"
-    key    = "infrastructure-services/terraform.tfstate"
-    region = "us-east-2"
-  }
+  required_version = "~> 1.1"
```
ameukam, line 18: Why do we downgrade the version ?

infra/aws/terraform/kops-infra-ci/eks.tf
ameukam: Not really. some of tests pick regions randomly during their run. we want to be able to track cost per region.

ameukam:
@bryantbiggs I think it's better we split this since there are multiple operators affected by this PR.

infra/aws/terraform/kops-infra-ci/eks.tf
bryantbiggs: costs are tracked via AWS CUR (Cost and Usage Reporting), correct? Region is a very common facet that is carried over in nearly everything, including CUR reports https://docs.aws.amazon.com/cur/latest/userguide/product-columns.html#product-details-R

bryantbiggs:
I figured, its quite large and broad 😅 - let me parcel this out

PR #5942, opened 2023-10-07
chore: Standardize and consolidate the CDN packages and registry `infra/aws/terraform/*` projects

bryantbiggs:
- `cdn.packages.k8s.io`, `registry-sandbox-k8s-io-image-layers`, `registry.k8s.io` from #5935 to reduce the scope of change set

See #5935 for details on the changes


infra/aws/terraform/cdn.packages.k8s.io/main.tf
```diff
@@ -15,22 +15,22 @@ limitations under the License.
...
-  backend "s3" {
-    bucket       = "cdn-packages-k8s-io-tfstate"
-    key          = "terraform.tfstate"
-    region       = "eu-central-1"
-    role_arn     = "arn:aws:iam::309501585971:role/Provisioner"
-    session_name = "cdn-packages-k8s-io-terraform"
-  }
-
-  required_version = "~> 1.5.0"
+  required_version = "~> 1.1"
```
xmudrii, line 18: Why is version downgraded to 1.1?

infra/aws/terraform/registry-sandbox-k8s-io-image-layers/main.tf
```diff
@@ -14,23 +14,32 @@ See the License for the specific language governing permissions and
...
-- Storage backend details
-- alias for the provider per region
-*/
-
 terraform {
-  backend "gcs" {
-    bucket = "k8s-infra-tf-aws"
-    prefix = "oci-proxy"
-  }
+  required_version = "~> 1.1"
```
xmudrii, line 18: I think we should use a newer Terraform version.

infra/aws/terraform/registry.k8s.io/main.tf
```diff
@@ -14,168 +14,112 @@ See the License for the specific language governing permissions and
...
-variable "prefix" {
-  type        = string
-  default     = "test-"
-  description = "The prefix for all resources"
-
-  validation {
-    condition     = can(regex(".*-$|^$", var.prefix))
-    error_message = "The string must end with a hyphen or be empty."
+terraform {
+  required_version = "~> 1.1"
```
xmudrii, line 18: I think we should use a newer Terraform version.

xmudrii (commented):
Overall looks good to me, but I'd like @BenTheElder to verify the registry.k8s.io part. I also have some nits regarding the required Terraform version.

infra/aws/terraform/cdn.packages.k8s.io/main.tf
bryantbiggs: because there aren't any features used that are only available in 1.1+ 

for example - if we started utilizing the new Terraform native test functionality, we would need to change this to 
```hcl
  required_version = "~> 1.6"
```

Its not like a runtime or programming language where there are performance improvements, later versions mean newer features and you typically raise the MSV in response to the features utilized

xmudrii: @bryantbiggs I verified that using `~> 1.1` allows using newer Terraform versions (e.g. 1.5) so this can stay IMO

infra/aws/terraform/registry-sandbox-k8s-io-image-layers/main.tf
xmudrii: I verified that using `~> 1.1` allows using newer Terraform versions (e.g. 1.5) so this can stay IMO

infra/aws/terraform/registry.k8s.io/main.tf
xmudrii: I verified that using `~> 1.1` allows using newer Terraform versions (e.g. 1.5) so this can stay IMO

PR #5955, opened 2023-10-10
AWS: Deny non-SSL transport and define bucket ownership

ameukam:
Related:
  - https://github.com/kubernetes/registry.k8s.io/issues/172

Attach a policy ensuring non-TLS connections are denied. Ensure bucket ownership for each bucket.
Ensure account policy for public buckets are disabled. Change logic to generate bucket name.

### DIFF
--- a/infra/aws/terraform/modules/registry-k8s-io-s3-bucket/main.tf
+++ b/infra/aws/terraform/modules/registry-k8s-io-s3-bucket/main.tf
@@ -16,6 +16,8 @@ limitations under the License.
 
 data "aws_region" "current" {}
 
+data "aws_caller_identity" "current" {}
+
 locals {
   prefix      = "k8s-infra"
   bucket_name = format("%v-registry-k8s-io-%s", local.prefix, data.aws_region.current.name)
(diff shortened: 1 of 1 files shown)


ameukam:
cc @bryantbiggs @dims

infra/aws/terraform/modules/registry-k8s-io-s3-bucket/main.tf
```diff
@@ -25,11 +27,25 @@ module "s3_bucket" {
   source  = "terraform-aws-modules/s3-bucket/aws"
   version = "3.15.1"
 
-  bucket = base64sha256(local.bucket_name)
-  acl    = "public-read"
+  bucket = replace(lower(base64sha256(local.bucket_name)), "/[^a-zA-Z0-9-_]/", "")
```
ameukam, line 30: One reason of doing this is to ensure we respect bucket naming rules: https://docs.aws.amazon.com/AmazonS3/latest/userguide/bucketnamingrules.html

ameukam: I completely forgot about his input. 🤦🏾 

We don't have detailed requirements for naming. The idea was to completely obfuscate the bucket name. I think we can revisit the `bucket_name`. I don't have a strong opinion about this.

bryantbiggs: I think its ok either way - if you go ahead with what you have currently, I would just recommend adding some descriptive tags so that you can connect the buckets back to the repo/use case, etc., since the name will be obfuscated

ameukam: I'll add tags as a follow-up PR so we can have a dedicated conversation.

PR #5964, opened 2023-10-11
Migrate EKS Prow build cluster to Bottlerocket

xmudrii:
We have stability issues in the EKS Prow build cluster since 2023-06-26 as explained in #5473. We have been working together with the AWS support and the EKS service team for the past few weeks and they strongly recommended migrating to a Linux distro maintained by AWS (AL2 or Bottlerocket). AL2 has way too old kernel version and doesn't work well with our CI jobs (mainly kind-based jobs). Bottlerocket comes with a much newer kernel (6.x) which makes it a good candidate to temporary replace Ubuntu.

Bottlerocket works a bit different compared to Ubuntu or AL2, for example:

- There's no option to use userdata, instead you must build a container image that's going to execute the bootstrap script
- The bootstrap logic is slightly different because volume mounts are handled in a different way compared to Ubuntu
- We can't use `sysctl` inside the bootstrap script because of difference in volume mounts (we have a DaemonSet to handle that instead)

Other than that, Bottlerocket works fine and I managed to get some tests passing. I'll continue running tests in the upcoming days.

Bottlerocket is supposed to temporarily replace Ubuntu for a few days. We're mostly like going to try to run it over the weekend when traffic is low. If it turns out to be stable, we can extend the testing time or consider switching permanently.

This change is already rolled out on canary.

For testing purposes, I temporarily created a public ECR registry in the Prow prod account where we can store the bootstrap container image. Because of that, I allowed `ecr-public:*` for the `EKSInfraAdmin` role so that we can push the image to the registry.

/assign @ameukam @dims @pkprzekwas 
/hold
We're yet to rollout this change to prod


pkprzekwas:
Looks good 👍🏻

xmudrii:
The cluster-autoscaler pod was stuck in CrashLoopBackoff with the following message:

```
failed to create containerd task: failed to create shim task: OCI runtime create failed: runc create failed: unable to start container process: error during container init: error mounting "/etc/ssl/certs/ca-certificates.crt" to rootfs at "/etc/ssl/certs/ca-certificates.crt": mount /etc/ssl/certs/ca-certificates.crt:/etc/ssl/certs/ca-certificates.crt (via /proc/self/fd/6), flags: 0x5001, data: context="system_u:object_r:data_t:s0:c463,c479": not a directory: unknown
```

The latest commit fixes this issue and it's supposed to be compatible with both Ubuntu and Bottlerocket. That said, I need another review for this PR (@ameukam @pkprzekwas)

xmudrii:
Production rollout progress thread: https://kubernetes.slack.com/archives/CCK68P2Q2/p1697443043000379

xmudrii:
Production rollout is completed successfully
/hold cancel

xmudrii:
@pkprzekwas @ameukam Can you please LGTM the PR again, I had to update the active node group on prod and I committed a helper script that I use for listing pods by node selector

PR #5966, opened 2023-10-12
AWS: update monthly threshold

ameukam:
We are now spending half of the monthly budget.


ameukam:
cc @dims 
We probably need to revisit the scalability tests.

xmudrii (approved):

PR #5990, opened 2023-10-18
Grant access to gcr.io/k8s-staging-autoscaling to new people working …

rsgowman:
…on workload autoscaling

As per https://github.com/kubernetes/autoscaler/blob/master/vertical-pod-autoscaler/RELEASE.md#permissions

This is to enable us to release new versions of the addon-resizer.

kwiesmueller is a k8s org member; the others (incl me) are not. IIUC, this change should allow any of us to test, but only kwiesmueller to create new images.


mwielgus (approved):
/lgtm
/approve

mwielgus (approved):
/lgtm

PR #6014, opened 2023-10-23
Fastly: fix cache config for cdn.dl.k8s.io

ameukam:
Potential fix to ensure objects are cached at edge for only requests responding 200 from the origin. We have a short ttl(2 min) for 404 responses from the origin but this is defined after the object is cached at edge.


ameukam:
cc @dims @BenTheElder

ameukam:
Deployed:

```bash
fastly service-version list --service-name cdn.dl.k8s.io -q | grep 'true'
36      true    2023-10-23 19:49
```

PR #6024, opened 2023-10-26
Bumping Gateway API admission-server to v1.0.0-rc2

youngnick:


robscott (approved):
Thanks @youngnick!

/lgtm

PR #6063, opened 2023-11-09
Promote smb and nfs csi driver images

andyzhangx:


andyzhangx:
/assign @msau42 @xing-yang 
could you approve? thanks.

PR #6082, opened 2023-11-15
Fastly: Create a service for cdn.dl-sandbox.k8s.dev

ameukam:
Set up a new VCL service to investigate VCL service configuration using
a private gcs bucket as an origin.


ameukam:
cc @dims

PR #6122, opened 2023-11-27
Add a GCP service account for dl.k8s.io

ameukam:
Add a service account that will be used by Fastly to do RO actions against the new bucket releases.
Also create a HMAC key that will be used for authentification. Fastly at this time don't support workload identity.


ameukam:
/hold

cc @dims @upodroid @BenTheElder

upodroid (approved):
LGTM

PR #6124, opened 2023-11-28
chore: add myself to k8s-infra-ii-coop

BobyMCbobs:
adds myself to the _k8s-infra-ii-coop_ group.

/cc @hh @dims


hh:
+1

BobyMCbobs:
@hh @dims, thank you!

PR #6167, opened 2023-12-07
Grant access to k8s-io-packages app to SIG Release Leads and xmudrii

xmudrii:
It appears that we can't grant access to namespaces by using existing groups. This PRs allows SIG Release Leads and @xmudrii (me) to access the k8s-io-packages application in the `aaa` cluster.

cc @kubernetes/sig-release-leads 

/assign @upodroid @ameukam


upodroid (approved):
LGTM

PR #6148, opened 2023-12-05
kops-ci: Use storage-based instances.

ameukam:
Use a storage-based instance (NVMe local disk of 474 GB) for the EKS cluster.

PR #6168, opened 2023-12-07
kops-ci: Use container image for boostrap

ameukam:
Follow-up of:
  - https://github.com/kubernetes/k8s.io/pull/6148

Use a custom image for bootstrap since Bottlerocket only TOML format for bootstrap

### DIFF
--- a/infra/aws/terraform/kops-infra-ci/resources/bootstrap/Dockerfile
+++ b/infra/aws/terraform/kops-infra-ci/resources/bootstrap/Dockerfile
@@ -1,0 +1,30 @@
+# Copyright 2023 The Kubernetes Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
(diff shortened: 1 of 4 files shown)


infra/aws/terraform/kops-infra-ci/resources/bootstrap/Dockerfile
```diff
@@ -0,0 +1,30 @@
...
+FROM alpine@sha256:34871e7290500828b39e22294660bee86d966bc0017544e848dd9a255cdf59e0
+
+RUN apk update \
+    && apk add --no-cache \
+        bash \
+        findutils \
+        util-linux \
+        xfsprogs
+
+ADD https://raw.githubusercontent.com/kubernetes/k8s.io/a8be6bcc88c61dc55ce0f6b888e7f4e6c0a5293d/infra/aws/terraform/prow-build-cluster/bootstrap_bottlerocket/node_bootstrap.sh /bootstrap.sh
```
dims, line 26: this this make it harder to update bootstrap.sh in the future? (update the script in one PR and then fix up the SHA here in the next one?)

ameukam: Assuming we are doing a lot of changes for this. We are not gonna change the instance type for a while (specially in we use 1-year) and the script is pretty stable. 
Happy to use the ref of a branch if you want.

bryantbiggs: what about https://github.com/kubernetes-sigs/sig-storage-local-static-provisioner/blob/master/docs/getting-started.md#option-5-eks-experimental instead?

ameukam:
Ref:
 - https://github.com/kubernetes/k8s.io/issues/5127

PR #6170, opened 2023-12-07
Add kubecost deployment to the prow clusters

koksay:
Adds Kubecost deployment to flux.

> Because the helm values for prod and canary are different, the ConfigMaps are created manually and added them under `{resource_dir}/kubecost/*-values.yaml` files.


infra/aws/terraform/prow-build-cluster/hack/flux-update.bash
```diff
@@ -65,6 +65,20 @@ flux create hr node-termination-handler \
...
+    --interval=${sync_interval} \
+    --export >> ${resources_dir}/flux-system/flux-source-helm-kubecost-chart.yaml
+
+boilerplate > ${resources_dir}/kubecost/flux-hr-kubecost.yaml
+flux create helmrelease kubecost --chart cost-analyzer \
+  --source HelmRepository/kubecost \
+  --chart-version 1.107.1 \
+  --namespace kubecost \
+  --values-from=ConfigMap/kubecost-values \
+  --export >> ${resources_dir}/kubecost/flux-hr-kubecost.yaml
```
xmudrii, line 80: Some suggestions to make the command consistent with others

```suggestion
flux create hr kubecost \
    --source=HelmRepository/kubecost.flux-system \
    --namespace=kubecost \
    --chart=cost-analyzer \
    --chart-version=1.107.1 \
    --values-from=ConfigMap/kubecost-values \
    --interval=${sync_interval} \
    --export >> ${resources_dir}/kubecost/flux-hr-kubecost.yaml
```

xmudrii: Maybe we should check if we can explicitly specify the namespace for `kubecost-values` using `.<namespace>` as for `--source`

infra/aws/terraform/prow-build-cluster/resources/kubecost/prow-build-cluster-values.yaml
```diff
@@ -0,0 +1,51 @@
+## save this to values.yaml file and run
+## k create configmap kubecost-helm-values --from-file=values.yaml -n kubecost
+## on the prow-build-cluster
```
xmudrii, line 3: Can we make this part of the `flux-update.bash` script?

infra/aws/terraform/prow-build-cluster/resources/kubecost/prow-build-cluster-values.yaml
```diff
@@ -0,0 +1,51 @@
...
+  federator:
+    enabled: true
+    primaryClusterID: prow-build-cluster
+kubecostModel:
+  containerStatsEnabled: true
+  cloudCost:
+    enabled: true # Set to true to enable CloudCost view that gives you visibility of your Cloud provider resources cost
+  etlCloudAsset: false # Set etlCloudAsset to false when cloudCost.enabled=true
+  federatedStorageConfigSecret: kubecost-object-store
+serviceAccount: # this example uses AWS IRSA, which creates a service account with rights to the s3 bucket. If using keys+secrets in the federated-store, set create: true
```
xmudrii, line 22:
```suggestion
serviceAccount: # this config uses AWS IRSA, which creates a service account with rights to the s3 bucket. If using keys+secrets in the federated-store, set create: true
```

xmudrii (changes requested):
@koksay Thank you for taking care of this! Some minor nits, otherwise LGTM!

infra/aws/terraform/prow-build-cluster/hack/flux-update.bash
koksay: Flux documentation does not mention about the namespace:

```bash
--values-from strings   a Kubernetes object reference that contains the values.yaml data key in the format '<kind>/<name>', where kind must be one of: (Secret,ConfigMap)

```

infra/aws/terraform/prow-build-cluster/resources/kubecost/prow-build-cluster-values.yaml
koksay: The problem here is that the values are different for canary and prod clusters. Otherwise, I could have just included them in the helmreleease file as well.

I can do it with a command like `kubectl cluster-info` (which reveals like `Kubernetes control plane is running at https://xxx.gr7.us-east-2.eks.amazonaws.com`) and use it with an if statement to create it. Wdyt?

xmudrii: We have `PROW_ENV` variable in Makefile that can be `canary` or `prod`. We use that variable mainly for Terraform, but we can re-use it for other purposes as well.

What do you think about:

- Modifying this script to fail very early (at the beginning) if `PROW_ENV` variable is unset
- Modifying Makefile and `flux-update` target to pass `PROW_ENV` to the script (replace `./hack/flux-update.bash` with something like `PROW_ENV=$(PROW_ENV) ./hack/flux-update.bash`)
- Modifying this script to act based on the value of `PROW_ENV`

infra/aws/terraform/prow-build-cluster/hack/flux-update.bash
xmudrii: Okay, then we're all good here 👍

infra/aws/terraform/prow-build-cluster/resources/kubecost/prow-build-cluster-values.yaml
koksay: Ok, that makes sense. I will implement it and update the PR accordingly.

koksay: I changed it to use values file (instead of ConfigMap), and change the filename as `{canary,prod}-cluster-values.yaml`

infra/aws/terraform/prow-build-cluster/resources/kubecost/prow-build-cluster-values.yaml
koksay: updated

xmudrii (approved):
/lgtm
/approve

PR #6184, opened 2023-12-13
Configure logs in httpRequest format - legacy

upodroid:
I configured nginx to parse the logs into a [httpRequest](https://cloud.google.com/logging/docs/reference/v2/rest/v2/LogEntry#HttpRequest) for easy analysis.

We can apply the same to the production redirector as well.

/cc @BenTheElder @ameukam @xmudrii


xmudrii (approved):
Thank you for taking care of this!
/lgtm
/approve

PR #6189, opened 2023-12-14
configure AWS SSO and grant it on the boskos account

upodroid:
We are ready to use AWS SSO. All I'm missing is some okta terraform

1. Amended the org-account module to always grant Admin on every account to sig-k8s-infra-leads group and readonly access to a read only group
2. sso.tf in the management layer now creates the Permissions sets, I created 3 important roles, AdministratorAccess, ReadOnly and BillingReadOnly for the management
3. I created a couple of groups manually in Okta and assigned them permissions.

The management account is a bit complicated so some applies were done with `-target` flag. 

/hold

/cc @ameukam @dims @xmudrii @BenTheElder @jeefy


ameukam:
/lgtm

Can we add instructions on how to configure the cli for SSO authn ?

upodroid:
> Can we add instructions on how to configure the cli for SSO authn ?

I'll put it in a followup PR

/hold cancel

PR #6199, opened 2023-12-16
Update OWNERS_ALIASES

aojea:
add aojea to sig-net leads

Ref: https://github.com/kubernetes/community/issues/7650


thockin (commented):
Thanks!

/lgtm
/approve

PR #6201, opened 2023-12-17
Configure DNS to allow Okta to send email via kubernetes.io

upodroid:
Okta allows you to use a custom email for outbound communications.

I have configured it to send emails as `okta@kubernetes.io` and I need to configure DKIM, SPF and domain validation before enabling this feature.

https://developer.okta.com/docs/guides/custom-url-domain/main/#about-custom-email-addresses

On that note, we are missing gmail DKIM config

/cc @ameukam @dims @BenTheElder


upodroid:
Verification is successful. All future emails from Okta will come from okta@kubernetes.io

PR #6202, opened 2023-12-18
Promote all new gcp-compute-persistent-disk-csi-driver releases

ialidzhikov:
You can double-check the digests in https://console.cloud.google.com/gcr/images/k8s-staging-cloud-provider-gcp/global/gcp-compute-persistent-disk-csi-driver.

Fixes https://github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/issues/1397


msau42:
/lgtm
/approve

Thanks for taking care of this. I plan to copy over the script we use for csi release promotions to make this easier in the future.

PR #6214, opened 2023-12-21
Use script to generate gcp promotion images

msau42:


registry.k8s.io/images/k8s-staging-cloud-provider-gcp/generate.sh
```diff
@@ -0,0 +1,39 @@
...
+    cloud-controller-manager
+)
+
+for image in "${images[@]}"; do
+    echo "- name: ${image}"
+    echo "  dmap:"
+    gcloud container images list-tags \
+        "${repo}/$image" \
+        --format="get(digest, tags)" \
+        --sort-by="tags" \
```
mattcary, line 36: too bad we can't sort by version number in the tags, the lexicographic ordering is annoying.

msau42: With some bash foo it could probably be resorted but I'll leave that as  future improvement :-)

Issue #6242, opened 2024-01-04
Add a `go.k8s.io` redirect for contact information


MadhavJivrajani:
There have been many instances where contact information of a particular SIG or just avenues of communication in the community need to be conveyed either to new contributors or just having the link handy helps.

Having said that, we could probably add a redirect to the existing `go.k8s.io` redirect to handle this:

https://github.com/kubernetes/k8s.io/blob/a2717cb3e792e43fb039b77f690cbdedbaaff8be/apps/k8s-io/configmap-nginx.yaml#L258

### Step 1: Add in the redirects

The following redirects would help:
1. `go.k8s.io/contact` goes to https://github.com/kubernetes/community/tree/master/communication 
2. `go.k8s.io/contact/<sig|wg|committee-name>` goes to https://github.com/kubernetes/community/tree/master/group-name#contact
  i. Example: `go.k8s.io/contact/sig-contributor-experience` would go to https://github.com/kubernetes/community/tree/master/sig-contributor-experience#contact

### Step 2: Update README.md

Update the README [here](https://github.com/kubernetes/k8s.io/tree/a2717cb3e792e43fb039b77f690cbdedbaaff8be/apps/k8s-io#redirections) to mention the redirects.

### Step 3: Add in tests for the redirects

Subsequently, add in tests for this new redirect here: 

https://github.com/kubernetes/k8s.io/blob/a2717cb3e792e43fb039b77f690cbdedbaaff8be/apps/k8s-io/test.py#L172

/sig contributor-experience
/good-first-issue

SD-13:
Hey @MadhavJivrajani I just tried `go.k8s.io/contact` but it is giving 404 Not found error even with a sig name. It is also not in the redirects list.

MadhavJivrajani:
@SD-13 yes, changes still need to be made in order for it to be a valid redirect. Please see the steps I've outlined in the issue.

@haripriya9647 feel free to ask questions here in case you need to as you work on this!

haripriya9647:
Sure @MadhavJivrajani

haripriya9647:
1. "https://go.k8s.io/pr-dashboard" 
2. "https://go.k8s.io/test-health"
3. "https://go.k8s.io/logo"
  link are broken

nitishfy:
Hi, Is this issue still open to work on? If yes, I'd like to take this up.

haripriya9647:
Yes Nitish , i have some issues you can take this up..

[...]

PR #6247, opened 2024-01-04
Redirects

haripriya9647:
#6242

### DIFF
--- a/apps/k8s-io/README.md
+++ b/apps/k8s-io/README.md
@@ -57,6 +57,8 @@ Redirections
 - https://go.k8s.io/test-history
 - https://go.k8s.io/triage
 - https://go.k8s.io/logo
+- https://go.k8s.io/contact
+- https://go.k8s.io/contact/sig-contributor-experience
 
 
 NOTE: please see configmap-nginx.yaml for rewrite rules.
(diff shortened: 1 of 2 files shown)


haripriya9647:
@MadhavJivrajani Do you want me to add links for all the committees or just for contributor experience

MadhavJivrajani (changes requested):
Thanks for working on this @haripriya9647!

I think there is some confusion as to what needs to be done. Let me try and clarify:
1. We need the redirect to be generic. When I said `sig-contributor-experience`, I only meant it as an example of what the URL that someone would type might look like.
2. You could hard code all groups as redirects, but new groups are added and removed quite frequently in the community, so the redirect list would then be quite stale.
3. You also have not yet added in what needs to be added: the actual redirect in the config file I had linked in the original issue, see the configurations listed here:

https://github.com/kubernetes/k8s.io/blob/5918433899a23de861ef3ff2ce95a5d8d97438d7/apps/k8s-io/configmap-nginx.yaml#L257-L284

4. You will need to add in 2 more entries under the `go.k8s.io` configuration:
```
rewrite ^/contact$          https://github.com/kubernetes/community/tree/master/communication/ redirect;
rewrite ^/contact/([^/]*)/?$ https://github.com/kubernetes/community/tree/master/$1#contact redirect;
```
Note that the above redirects are similar to the redirect configuation for `go.k8s.io/owners`, subsequently, the test cases you write also should be similar to the `/owners` redirect.

5. Update the README accordingly as well.

MadhavJivrajani:
Please also edit the PR description and title to convey what you're working on :)

haripriya9647:
Hello, @MadhavJivrajani. My apologies I had some issues with my laptop so I couldn't add my commits. 
I will finish it by today.

MadhavJivrajani:
Thanks @haripriya9647!

MadhavJivrajani:
/close 
Ref: https://github.com/kubernetes/k8s.io/issues/6242#issuecomment-1943934006

PR #6248, opened 2024-01-05
kOps CI: Deploy cluster-autoscaler

ameukam:
- Deploy Cluster Autoscaler matching the version of the EKS cluster.
- Deploy Metrics Server


ameukam:
cc @hakman

PR #6251, opened 2024-01-06
Sink logs from prod redirector instance

upodroid:
Applying #6184 to our production redirect instance. Also, we are sinking apt/yum URLs to bq for analysis.

@xmudrii


xmudrii (approved):
/lgtm
/approve

Issue #5168, opened 2023-04-25
eks-prow-build-cluster: Use Karpenter instead of cluster-autoscaler


xmudrii:
cluster-autoscaler doesn't support mixing different instance types. This is a problem because `r5ad.4xlarge` is way too expensive ($1.048/hr) and we might want to combine bigger and smaller instances (e.g. not all tests need such large instances).

We also might want to consider mixing on-demand and spot instances for additional savings, which Karpenter would allow us to do.

/priority important-soon

xmudrii:
@sbdtu5498 Are you still interested to work on this issue?

xmudrii:
/unassign @sbdtu5498 
/assign @xmudrii
(@koksay will eventually work on this, I'll reassign it to him later)
/milestone v1.30

xmudrii:
This is done! 🎉 
/close

PR #6257, opened 2024-01-09
eks-prow-build: introduce stable node group

xmudrii:
Some components are not taking often node rotations well:

- Flux: a lot of alerts coming into `#k8s-infra-alerts` upon evicting/rescheduling Flux pods
- KubeCost: it's watching pods and their resource usage all the time, as well as, has integrated Prometheus
- Our monitoring stack: same as KubeCost

To mitigate this issue, I created a stable node group, that's not autoscaled, with three nodes. This node group is tainted and it's only going to run these components (and potentially some other components that might benefit from this).

This PR is the first part, there'll be another PR to change manifests for these components. Terraform changes are already applied and the node group has been successfully created.

### DIFF
--- a/infra/aws/terraform/prow-build-cluster/eks.tf
+++ b/infra/aws/terraform/prow-build-cluster/eks.tf
@@ -106,5 +106,6 @@ module "eks" {
   eks_managed_node_groups = {
     build-blue  = local.node_group_build_blue
     build-green = local.node_group_build_green
+    stable      = local.node_group_stable
   }
 }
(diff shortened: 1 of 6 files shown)


upodroid (approved):
LGTM

alert fatigue :D

sftim:
Cool. Does this lay the ground for using spot launch more for everything else?

xmudrii:
> Does this lay the ground for using spot launch more for everything else?

In a way, but we likely need a more sophisticated cluster-autoscaling solution (e.g. [Karpenter](https://github.com/kubernetes/k8s.io/issues/5168))

PR #6270, opened 2024-01-12
Add jdzikowski to scalability oncall

jdzikowskigke:


marseel:
Looks like you are not kubernetes org member?

mborsz:
/ok-to-test

Jarek is not a kubernetes org member, but he is a part of oncall rotation.
Maybe after few oncall shifts/resolved bugs we can request kubernetes org membership with scalability oncall as a contribution?

WDYT, @marseel ?

marseel:
@mborsz 

Sure, makes sense.
/lgtm
/approve

PR #6296, opened 2024-01-18
prow-build-trusted: switch to newer storage class for ghproxy

ameukam:
The PD CSI driver have been abled since GKE 1.25 and the new storage class introduced after an auto-upgrade operation to 1.25

This is done to address an issue reported in
https://github.com/kubernetes/test-infra/issues/31618 where the ghproxy was down due to the PV not deleted after the upgrade of the cluster to GKE 1.26


ameukam:
The PVC and the PV were manually deleted.

PR #6299, opened 2024-01-19
Add dgrisonnet into metrics-server release OWNERS

CatherineF-dev:


dgrisonnet:
I updated the list in https://github.com/kubernetes/k8s.io/pull/6305.

/close

PR #6306, opened 2024-01-22
eks-prow-build: switch some components to the stable node group

xmudrii:
We recently introduced a stable node group in the EKS Prow build cluster to host components that might be sensitive to often interruptions caused by autoscaling. This PR moves some components to that node group, namely:

- cluster-autoscaler
- metrics-server
- boskos
- node-problem-detector

Other components, including Flux, will be moved in a follow up PR.

Additionally, the AWS module has been updated to 5.20 because it was required by Terraform/some submodule.

cc @upodroid @dims

### DIFF
--- a/infra/aws/terraform/prow-build-cluster/.terraform.lock.hcl
+++ b/infra/aws/terraform/prow-build-cluster/.terraform.lock.hcl
@@ -2,148 +2,148 @@
 # Manual edits may be lost in future updates.
 
 provider "registry.terraform.io/hashicorp/aws" {
-  version     = "5.19.0"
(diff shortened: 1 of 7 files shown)


infra/aws/terraform/prow-build-cluster/resources/node-problem-detector/node-problem-detector.yaml
```diff
@@ -37,6 +37,8 @@ spec:
                     operator: In
                     values:
                       - linux
+      nodeSelector:
```
upodroid, line 40: Is this deployment missing tolerations?

xmudrii: @upodroid Great catch, I dropped changes for this file. PTAL again

upodroid (approved):
Thanks

PR #6310, opened 2024-01-22
Remove k8s-gsm-tools from image promo

ameukam:
Project archived: https://github.com/kubernetes/org/issues/3432

We never did an image promo for this project.
The GCP project still exists and we will be deleted manually.


ameukam:
cc @BenTheElder @dims

PR #6313, opened 2024-01-23
Promote gcp-compute-persistent-disk-csi-driver v1.13.0

kon-angelo:
Promote new release for `v1.13.0` of gcp-compute-persistent-disk-csi-driver


mattcary:
/ok-to-test
/lgtm
/approve

Thanks!

Issue #6303, opened 2024-01-22
AWS: Pod got deleted unexpectedly


ameukam:
We have reports of prowjobs deleted on EKS cluster `eks-prow-build-cluster` with 

```
Job execution failed: Pod got deleted unexpectedly
```
  - https://github.com/kubernetes/kubernetes/issues/122822
  - https://github.com/kubernetes-sigs/cluster-api/issues/9901
  - https://github.com/kubernetes/test-infra/issues/31459
  
  
/kind bug
/area infra/aws
/priority important-soon
/milestone v1.30

ameukam:
/assign @xmudrii
cc @upodroid @dims

xmudrii:
A support ticket has been created with AWS to investigate this issue

xmudrii:
We have a discussion on Slack about potential root cause https://kubernetes.slack.com/archives/CCK68P2Q2/p1705919163947889

xmudrii:
We received a response from the AWS support and this is the most important bit:

> With regards to the behaviour you have observed, I have investigated into the nodes provided and can confirm that all except one of these nodes were terminated due to the AZRebalance feature as you have suspected. Additionally, there are about 1700 rebalancing activities for this Auto Scaling group over the past 6 weeks (6 weeks is the maximum history limit for Auto Scaling activity history) which matches the behaviour you are seeing in your environment.

They recommended us the following and it matches what we discussed with @tzneal yesterday:

> As this can result in an imbalance spread of instances between AZs, a possible long-term approach would be to utilize multiple node groups, where each node group is scoped to a single availability zone. This would allow cluster autoscaler to scale and balance nodes between multiple node groups, effectively balancing nodes between AZs. Note that "--balance-similar-node-groups" must be enabled on the cluster autoscaler for this feature
>
> The above approach is also recommended should you be utilizing EBS volumes for PVCs as EBS volumes are scoped to a single availability zone.

Out of 7 instances that I provided to the AWS support, 6 of them were removed by the AZRebalance feature and one was removed by cluster-autoscaler. We'll:

- switch to a dedicated node group per AZ
- continue monitoring if cluster-autoscaler is mistakenly removing nodes

xmudrii:
The proposed mitigation has been rolled out to the production cluster. I propose leaving this issue open for 7 days to monitor if the issue is gone. We can use Prow's Deck for monitoring: https://prow.k8s.io/?state=error&cluster=eks-prow-build-cluster

[...]

PR #6320, opened 2024-01-24
eks-prow-build-cluster: Use dedicated Managed Node Groups (MNGs) per Availability Zone (AZ)

xmudrii:
xref https://github.com/kubernetes/k8s.io/issues/6303 (see for more details and context)

As per recommendations received from the AWS support and @tzneal, we're replacing blue and green node groups with node groups per AZ. In other words, old node groups had instances in all three AZs. Now, we have a dedicated node group for each AZ. This is a short-term solution to fix the stability issues that we're facing.

This has been successfully rolled out to canary, I'll do prod rollout once this PR is reviewed and merged.

Notes:

- We plan to switch to Karpenter long-term (https://github.com/kubernetes/k8s.io/issues/5168)
- We didn't want to suspend AZRebalacing process because it's not natively supported by EKS, so we opted in for dedicated node groups as that solution can be fully-automated using Terraform
- We have dedicated Terraform objects and variables for each Node Group. I explicitly didn't want to use `count` because that would make rolling out upgrades too complicated

Follow-ups:

- Remove blue/green Terraform variables and objects
- Update docs to reflect changes in the upgrade procedure

/assign @upodroid @ameukam @dims


upodroid (approved):
LGTM

upodroid:
cancel the hold when you are ready to merge
/hold

xmudrii:
Let's get this merged and I'll create a new PR for follow-ups
/hold cancel

infra/aws/terraform/prow-build-cluster/node_group_us-east-2a.tf
```diff
@@ -0,0 +1,101 @@
...
+
+locals {
+  node_group_build_us_east_2a = {
+    name            = "build-us-east-2a"
+    description     = "EKS managed node group in US East 2a"
+    use_name_prefix = true
+
+    cluster_version = var.node_group_version_us_east_2a
+
+    taints = var.node_taints_build
```
bryantbiggs, line 25: FYI - you can use [`eks_managed_node_group_defaults`](https://github.com/terraform-aws-modules/terraform-aws-eks/blob/2cb1fac31b0fc2dd6a236b0c0678df75819c5a3b/examples/eks_managed_node_group/main.tf#L84)  to set the default values you wish to use across all of the managed nodegroups, and then any specific settings unique to that nodegroup can be set in the nodegroup definition. You can also set the defaults and still override the value in the nodegroup definition as well

So roughly speaking, something like the following might help cut down on the copy+pasta across nodegroup definitions:

```hcl
  eks_managed_node_group_defaults = {
    use_name_prefix = true
    
    taints = var.node_taints_build
    labels = var.node_labels_build
    ... anything else that you want common across the nodegroups
  }
  
  
```

xmudrii: That's neat, thanks for pointing it out! I'll keep it in mind and see if we can make use of it the next time we do some rollout

PR #6309, opened 2024-01-22
registry.k8s.io: Add europe-west3 and europe-west10 - Part 1

ameukam:
Add regions europe-west3 and europe-west10 to the promotion process.

PR #6328, opened 2024-01-25
registry.k8s.io: Update known GCP artifact registry prod repositories

ameukam:
Update Test manifest with europe-west3 and europe-west10

### DIFF
--- a/registry.k8s.io/manifests/manifest_test.go
+++ b/registry.k8s.io/manifests/manifest_test.go
@@ -55,14 +55,16 @@ type Manifest struct {
 // TODO: If you add any production registries you will have to update this list 🤷
 var ProdRegistries []string = []string{
 	"asia-east1-docker.pkg.dev/k8s-artifacts-prod/images/",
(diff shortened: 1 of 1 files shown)


ameukam:
cc @upodroid @dims @BenTheElder

ameukam:
`pull-k8sio-verify` is expected to fail. Adding both regions in https://github.com/kubernetes/k8s.io/pull/6309

xmudrii (approved):
/lgtm
/approve

yussufsh:
Just want to report that `pull-k8sio-verify` jobs are failing because of this change in all new PRs.

Apologies if I am wrong but on debugging I found the length of registries [here](https://github.com/kubernetes/k8s.io/blob/main/registry.k8s.io/manifests/k8s-staging-addon-manager/promoter-manifest.yaml) does not match total 22 production registries.

yussufsh:
I see now the PRs to add the new registries are in the open state. I'll wait and check back.

PR #6357, opened 2024-02-02
Add verify-conformance to apps/

BobyMCbobs:
apart of apisnoop donation to sig-architecture

depends on: https://github.com/kubernetes/test-infra/pull/31804, https://github.com/kubernetes/test-infra/pull/31777

ticket: https://github.com/kubernetes/org/issues/4760 (might need moving to k8s.io?)

# TODO
- [ ] secrets for GitHub App or token for bot user

---

/assign
/hold

### DIFF
--- a/apps/verify-conformance/OWNERS
+++ b/apps/verify-conformance/OWNERS
@@ -1,0 +1,10 @@
+# See the OWNERS docs at https://go.k8s.io/owners
+
+approvers:
+  - BobyMCbobs
+  - hh
+  - heyste
+  - dims
+
+labels:
+  - sig/architecture
--- a/apps/verify-conformance/README.md
+++ b/apps/verify-conformance/README.md
@@ -1,0 +1,17 @@
+# verify-conformance
+
+a Prow+GitHub bot for verifying Kubernetes conformance product submissions.
+
+# Setup and configuration
+
+Configure the targetted repos in [configmap.yaml](./configmap.yaml) and the deployment in [deployment.yaml](./deployment.yaml).
+
+# Deployment
+
+- Have [access](https://github.com/kubernetes/k8s.io/blob/main/running-in-community-clusters.md) to the GKE cluster `aaa`.
+
+- From the `apps/verify-conformance` directory run:
+
+```console
+./deploy.sh
+```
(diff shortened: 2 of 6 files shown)


apps/verify-conformance-release/deployment.yaml
```diff
@@ -0,0 +1,68 @@
+---
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: verify-conformance-release
+  namespace: prow
```
ameukam, line 6: Add a new dedicated namespace: https://github.com/kubernetes/k8s.io/blob/fa4da59ac6a42788c2d4bee59916de80f7fc37ca/infra/gcp/bash/namespaces/ensure-namespaces.sh#L145-L161

BobyMCbobs: @ameukam, thank you for taking a look!
As the bot interacts with PRs over at cncf/k8s-conformance, will the secrets be available to be authenticated as prow.k8s.io in a new namespace?

https://github.com/kubernetes/k8s.io/pull/6357/files/0b38794494a108fac67e0a8cf78cb969217eddd2#diff-03d8e004cec916a342fd5c6664a9ed4483a62b3ba1804a77854de00d3e94f9ebR56-R65

ameukam: We don't currently have a way to authenticate to `prow.k8s.io` in the `aaa` and we don't have access to the GKE cluster running `prow.k8s.io`.

BobyMCbobs: @ameukam, OK this is good to know.

Would think that coordinating with community github-management to create an OAuth app to then provide the credentials to the bot with is a suitable alternative solution?

ameukam: The historical approach of this is to request a token for one of the bots managed by github management team. Not sure how they feel about an OAuth app. Feel free to ask them.

BobyMCbobs: I think the historical approach would work just fine, I'll reach out. Thank you!

BobyMCbobs:
Renamed app from _verify-conformance-release_ to just _verify-conformance_, adjusted the deployment too.

BobyMCbobs:
Unsure how to handle the webhook and HMAC side of it. I'm aware of [sigs.k8s.io/prow/cmd/hmac](https://sigs.k8s.io/prow/cmd/hmac) but unsure whether it can be used here.

PR #6361, opened 2024-02-02
[Prow cluster] Flux version update from v2.0.0-rc.2 to v2.2.2

koksay:
- Flux version is upgraded to the latest v2.2.2
- Some deprecated v1beta2 APIs are updated.

When the PR is merged, Flux will upgrade the components as described [here](https://fluxcd.io/flux/installation/upgrade/#upgrade-with-git)

I also checked the [release notes](https://github.com/fluxcd/flux2/releases), and the removed fields don't seem to affect us.


xmudrii (approved):
/ok-to-test
/lgtm
/approve

PR #6379, opened 2024-02-05
[Prow cluster] Add flux kustomization for Kubecost

koksay:
Kustomization added for Kubecost, and values files are renamed not to have .yaml extension.


infra/aws/terraform/prow-build-cluster/resources/flux-system/flux-ks-kubecost.yaml
```diff
@@ -0,0 +1,29 @@
+# Copyright 2023 The Kubernetes Authors.
```
xmudrii, line 1: Nit: we need to figure out how to generate new manifests with 2024, but let's leave it for a follow up

xmudrii (approved):
/ok-to-test
/lgtm
/approve

PR #6393, opened 2024-02-07
Promote cluster-api-ipam-provider-in-cluster v0.1.0-rc.0

schrej:
cc @srm09 @rikatz


schrej:
/hold
I'll change the image name to "cluster-api-ipam-in-cluster-controller" to be more in line with other providers.

PR #6403, opened 2024-02-08
GCP/prow-build: Enable Cloud Key Management Service (KMS) API

pwschuurman:
Follow of:

 * https://github.com/kubernetes/test-infra/pull/31863

Some E2E tests requires the service to be enabled. Failure on https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/sigs.k8s.io_gcp-compute-persistent-disk-csi-driver/1375/pull-gcp-compute-persistent-disk-csi-driver-e2e/1704443546796298240/

Presubmit E2E tests are currently failing with:

```
rpc error: code = PermissionDenied desc = Cloud Key Management Service (KMS) API has not been used in project 773781448124 before or it is disabled. Enable it by visiting https://console.developers.google.com/apis/api/cloudkms.googleapis.com/overview?project=773781448124 then retry. If you enabled this API recently, wait a few minutes for the action to propagate to our systems and retry.
      error details: name = ErrorInfo reason = SERVICE_DISABLED domain = googleapis.com metadata = map[consumer:projects/773781448124 service:cloudkms.googleapis.com]
      error details: name = Help desc = Google developers console API activation url = https://console.developers.google.com/apis/api/cloudkms.googleapis.com/overview?project=773781448124
```

### DIFF
--- a/infra/gcp/terraform/modules/gke-project/main.tf
+++ b/infra/gcp/terraform/modules/gke-project/main.tf
@@ -23,6 +23,7 @@ locals {
   project_services = [
     "bigquery.googleapis.com",
     "cloudbuild.googleapis.com",
+    "cloudkms.googleapis.com",
     "cloudresourcemanager.googleapis.com",
     "compute.googleapis.com",
     "container.googleapis.com",


Bharadwajshivam28:
> /assign @ameukam
> 
> /cc @Sneha-at @Bharadwajshivam28

Hey @pwschuurman What i understand is that there is some permission issue or the API is not enable..

pwschuurman:
> Hey @pwschuurman What i understand is that there is some permission issue or the API is not enabled..

@Bharadwajshivam28 Correct, the Cloud KMS API is not enabled for the new `k8s-infra-prow-build` project. The PDCSI e2e tests require this API to be enabled, as they test interactions with this API.

Based on audit IAM permissions, I believe this is the only fix we need (default Service Account has editor permissions on the boskos rental projects), so this _should_ allow the PDCSI e2e tests to pass, but we shall see once these terraform changes apply to project.

PR #6423, opened 2024-02-14
[EKS Prow Cluster] Move metrics-server and cluster-autoscaler to Flux

koksay:
Currently, metrics-server, and cluster-autoscaler are deployed via terraform, and this PR moves the installation to flux.

In addition:
- External-secrets are moved to the stable nodes group.
- Flux controllers are moved to the stable nodes group (via kustomization)


infra/aws/terraform/prow-build-cluster/kubernetes.tf
```diff
@@ -14,36 +14,6 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-module "cluster_autoscaler" {
```
xmudrii, line 17: Can you also remove sources for those modules?

infra/aws/terraform/prow-build-cluster/resources/flux-system/kustomization.yaml
```diff
@@ -0,0 +1,27 @@
+# As described here: https://fluxcd.io/flux/installation/configuration/boostrap-customization/
+---
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+resources:
+- gotk-components.yaml
```
xmudrii, line 6: How is this file being applied?

koksay: This is the way to kustomize flux components: https://fluxcd.io/flux/installation/configuration/boostrap-customization/

xmudrii: I thought about how is `gotk-components.yaml` being applied by Flux. It turns out:
- we have a Kustomization resource: https://github.com/kubernetes/k8s.io/blob/main/infra/aws/terraform/prow-build-cluster/resources/flux-system/flux-ks-flux-system.yaml
- `kustomization.yaml` is automatically recognized by Flux when reconciling a Kustomization resource

> Maybe we need to list other files here in the resources, checking it out

I don't think that should be needed, because `kustomization.yaml` is not a Flux resources, it's a core kustomize resource. I _think_ that Flux is still going to apply all resources in the directory (because there's no `ignore` directive), but only apply given kustomizations to files listed in `kustomization.yaml`

xmudrii (approved):
/lgtm
/approve

PR #6446, opened 2024-02-20
Add Danil-Grigorev to CAPI Operator OWNERS

Fedosin:
Reference: https://github.com/kubernetes-sigs/cluster-api-operator/blob/main/OWNERS_ALIASES#L17


Danil-Grigorev (commented):
/lgtm

PR #6450, opened 2024-02-21
eks-prow-build-cluster: Update kustomization file for flux

koksay:
To fix the issue created with #6423


xmudrii (approved):
/lgtm
/approve

PR #6500, opened 2024-02-29
Promote dns 1.23.0 images

liangyuanpeng:
Promote dns 1.23.0 images.

Fixes: https://github.com/kubernetes/dns/issues/624

Related:
- https://github.com/kubernetes/kubernetes/pull/123310
- https://github.com/kubernetes/kubernetes/pull/123590

Quickly wrote a test to check if there is a digest mismatch issue with the current PR. Fortunately, there were no errors.  
Here is the test thing: Get the digest with tag, and then check the digest is it the digest from this PR.

```golang
func TestImages(t *testing.T) {
	newManifestList, err := image.NewManifestListFromFile(filepath.Join(imgfilepath))
	if err != nil {
		panic(err)
	}
	reg, err := remote.NewRegistry("gcr.io")
	if err != nil {
		panic(err)
	}
	total := 0
	checkDigestFailedTags := []string{}
	for _, imageData := range *newManifestList {
		src, err := reg.Repository(context.TODO(), "k8s-staging-dns/"+imageData.Name)
		if err != nil {
			panic(err)
		}
		for digest, tags := range imageData.DMap {
			for _, tag := range tags {
				if tag != "1.23.0" {
					continue
				}
				log.Println("checking image for:", "gcr.io/k8s-staging-dns/"+imageData.Name+":"+tag)
				total++
				dst := memory.New()
				desc, err := oras.Copy(context.TODO(), src, tag, dst, tag, oras.DefaultCopyOptions)
				if err != nil {
					panic(err)
				}
				if desc.Digest.String() != digest {
					checkDigestFailedTags = append(checkDigestFailedTags, "gcr.io/k8s-staging-dns/"+imageData.Name+":"+tag)
				}
			}
		}
	}
	log.Printf("check total:%d, failed:%d \n", total, len(checkDigestFailedTags))

}
```

logs:

```shell
...
2024/03/07 07:57:07 checking image for: gcr.io/k8s-staging-dns/k8s-dns-node-cache-arm64:1.23.0
2024/03/07 07:57:08 checking image for: gcr.io/k8s-staging-dns/k8s-dns-node-cache-ppc64le:1.23.0
2024/03/07 07:57:08 checking image for: gcr.io/k8s-staging-dns/k8s-dns-node-cache-s390x:1.23.0
2024/03/07 07:57:09 check total:24, failed:0 
--- PASS: TestImages (28.83s)
PASS
```


monotek (approved):

liangyuanpeng:
Quickly wrote a test to check if there is a digest mismatch issue with the current PR. Fortunately, there were no errors.  
Here is the test thing: Get the digest with tag, and then check the digest is it the digest from this PR.

```golang
func TestImages(t *testing.T) {
	newManifestList, err := image.NewManifestListFromFile(filepath.Join(imgfilepath))
	if err != nil {
		panic(err)
	}
	reg, err := remote.NewRegistry("gcr.io")
	if err != nil {
		panic(err)
	}
	total := 0
	checkDigestFailedTags := []string{}
	for _, imageData := range *newManifestList {
		src, err := reg.Repository(context.TODO(), "k8s-staging-dns/"+imageData.Name)
		if err != nil {
			panic(err)
		}
		for digest, tags := range imageData.DMap {
			for _, tag := range tags {
				if tag != "1.23.0" {
					continue
				}
				log.Println("checking image for:", "gcr.io/k8s-staging-dns/"+imageData.Name+":"+tag)
				total++
				dst := memory.New()
				desc, err := oras.Copy(context.TODO(), src, tag, dst, tag, oras.DefaultCopyOptions)
				if err != nil {
					panic(err)
				}
				if desc.Digest.String() != digest {
					checkDigestFailedTags = append(checkDigestFailedTags, "gcr.io/k8s-staging-dns/"+imageData.Name+":"+tag)
				}
			}
		}
	}
	log.Printf("check total:%d, failed:%d \n", total, len(checkDigestFailedTags))

}
```

logs:


```shell
...
2024/03/07 07:57:07 checking image for: gcr.io/k8s-staging-dns/k8s-dns-node-cache-arm64:1.23.0
2024/03/07 07:57:08 checking image for: gcr.io/k8s-staging-dns/k8s-dns-node-cache-ppc64le:1.23.0
2024/03/07 07:57:08 checking image for: gcr.io/k8s-staging-dns/k8s-dns-node-cache-s390x:1.23.0
2024/03/07 07:57:09 check total:24, failed:0 
--- PASS: TestImages (28.83s)
PASS
```

aojea:
/lgtm
/approve

Thanks for verifying it

ameukam:
@aojea +1 for make this as a presubmit.

liangyuanpeng:
> should we have this kind of test as a presubmit here? it will have to be a more involved just to check only the changed images though

Actually, Today I really wanted to implement a general go test, but I didn't think of a better implementation to only detect changed content.

> NOTE: This is a bad release because the container image and tar files do not match. And we could not fix it because registry.k8s.io do not allow tag mutation (https://github.com/kubernetes/k8s.io/pull/6522). Please use v0.8.17 instead, which is tagged to the same commit as v0.8.16.

I can work for it, avoid something like that would be great.

aojea:
> Actually, Today I really wanted to implement a general go test, but I didn't think of a better implementation to only detect changed content.

@pohly IIRC you already solved this problem, can you point us to that code?

pohly:
I did? Sorry, doesn't ring a bell.

Or do you mean the helper script that we are using in SIG Storage? That's here: https://github.com/kubernetes/k8s.io/blob/main/k8s.gcr.io/images/k8s-staging-sig-storage/generate.sh

brecht82:
@bowei @MrHohn can You review that PR? A large corporation is waiting for this version ;)

MrHohn (approved):
/approve

PR #6468, opened 2024-02-26
k8s-infra-prow-build: Add monitoring

koksay:
Adds monitoring components for the `k8s-infra-prow-build` cluster. 
Mainly copied from `eks-prow-build-cluster` [configuration](https://github.com/kubernetes/k8s.io/tree/main/infra/aws/terraform/prow-build-cluster/resources/monitoring)

Please note that, the `alertmanager-slack-token.yam` file has invalid data for this cluster. It will be updated later when the external secrets are in place.

PR #6504, opened 2024-03-01
k8s-infra-prow-build: Install kubecost

koksay:
Create `kubecost` namespsace and install Kubecost with manifests generated via `helm template`


koksay:
/hold
it seems that #6468 will be reverted; there is no point in deploying Kubecost w/o that one.

koksay:
/hold
postpone this after kubecon and k8s 1.30 release

xmudrii:
We're yet to consider how we are going to proceed with this one
/lifecycle frozen

PR #6472, opened 2024-02-27
Setup image release for leaderworkerset

liurupeng:
This is repo for the image release: https://github.com/kubernetes-sigs/lws

PR #6513, opened 2024-03-04
Add lws to registry.gcr.io

kerthcet:
Add lws to registry.gcr.io as a new sub-project, see https://github.com/kubernetes-sigs/lws/pull/30#discussion_r1510336189

/cc @ahg-g


kerthcet:
seems we have one for https://github.com/kubernetes/k8s.io/pull/6472 ...
/close then

PR #6521, opened 2024-03-05
k8s-infra-prow-build: remove node-exporter installation

koksay:
It seems that `node-exporter` daemonset caused the issue with cluster autoscaler. Removing it to see if this will help:

https://github.com/kubernetes/test-infra/issues/32157#issuecomment-1978086967

#6468


xmudrii (approved):
/lgtm
/approve

PR #6525, opened 2024-03-05
Switch to C3 machine types for prow build cluster

upodroid:
/cc @ameukam @BenTheElder @xmudrii 

We should get better performance out of these instances.

C3 instance types are a bit tricky. Our options are:
- c3d-highmem-8-lssd which provides 8 cores, 64GB of ram and a single local SSDs
- c3-standard-8-lssd which provides 8 cores, 32GB of ram and a two local SSDs

I'm in favour offering 7 cores and 28GB to all our prowjobs across the board. This setting guarantees a dedicated node for testing. The remaining core belongs to us to run our stuff on it

### DIFF
--- a/infra/gcp/terraform/k8s-infra-prow-build/main.tf
+++ b/infra/gcp/terraform/k8s-infra-prow-build/main.tf
@@ -83,7 +83,7 @@ module "prow_build_cluster" {
 # Why use UBUNTU_CONTAINERD for image_type?
 # - ipv6 jobs need an ipv6 stack; COS lacks one, so use UBUNTU
 # - k8s-prow-builds/prow cluster uses _CONTAINERD variant, keep parity
-module "prow_build_nodepool_n1_highmem_8_localssd" {
+module "prow_build_nodepool_c3d_highmem_8_lssd" {
   source       = "../modules/gke-nodepool"
   project_name = module.project.project_id
   cluster_name = module.prow_build_cluster.cluster.name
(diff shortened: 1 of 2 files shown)


BenTheElder:
> I also used a custom type to insert 2 extra cores for all the extra daemonset we need to run.

We really shouldn't need 2 cores of daemonsets ...?
/hold

We should be able to tighten this up, discarding 20% of the node as waste is excessive.

BenTheElder:
Maybe we can import from the existing metrics collection?

upodroid:
According to the calculations [here](https://github.com/kubernetes/test-infra/issues/32157#issuecomment-1979284058):
We could adopt `c3-standard-8-lssd` right now and disable calico. I'm not sure why we need a network policy enabled on the build cluster.

BenTheElder:
> [...] and disable calico. I'm not sure why we need a network policy enabled on the build cluster. The capacity being used by calico would be available for node-exporter

That sounds good to me (have not reviewed machine types yet). I don't recall an discussion about enabling network policy and I don't think we need it, 400m CPU is quite a bit back to fit other daemonsets cc @ameukam

BenTheElder:
Just checked: k8s-prow-builds in google.com (`cluster: default`) has calico network policy disabled.

upodroid:
One more question, are we ok with nodes having 32gb memory instead of 64gb?

BenTheElder:
`c3-standard-8-lssd` has lower memory / core (32gb for 8 cores) versus n1-highmem-8 (52 gb for 8 cores).

I'm not sure if that's a problem for any of our workloads or not. I think we might have a few that are a bit lopsided in memory>cpu requests currently

upodroid:
Apply plan:
1. apply the 1st commit using terraform
2. Cordon and Drain the old nodepool in the console
3. Apply the 2nd commit

BenTheElder:
I think we should probably query the jobs to see if any have 30+ GB or a larger memory : core ratio than c3, if so, we can do n2?

ameukam:
> > [...] and disable calico. I'm not sure why we need a network policy enabled on the build cluster. The capacity being used by calico would be available for node-exporter
> 
> That sounds good to me (have not reviewed machine types yet). I don't recall an discussion about enabling network policy and I don't think we need it, 400m CPU is quite a bit back to fit other daemonsets cc @ameukam

Network Policy was enabled by default at the time the cluster was created. We never took the opportunity to disable it. +1 for the disablement.

upodroid:
Instead of tweaking the prowjobs, let go with N2 instances(Ice Lake platform) and keep the same machine-type and merge that change this weekend.

PR #6543, opened 2024-03-07
disable calico in prow build clusters

upodroid:


upodroid:
/hold cancel
I applied the change in both clusters

PR #6557, opened 2024-03-12
Source Slack webhook token for k8s-infra-alerts from GCP SM

xmudrii:
We copied the Slack webhook for the `#k8s-infra-alerts` channel to the GCP Secrets Manager so that we can use it in both k8s-infra-prow-build and eks-prow-build clusters without having them in two places (GCP SM and AWS SM). This PR changes ESO objects to reference the GCP SM and I'll remove the secret from the AWS SM after this is reconciled.

/assign @dims @upodroid @ameukam


xmudrii:
The AWS SM secret has been scheduled for deletion in 7 days.

PR #6563, opened 2024-03-12
registry.k8s.io: Switch to Cloudfront for Chile

ameukam:
Related to:
  - https://github.com/kubernetes/registry.k8s.io/issues/194

Ensure we serve blobs from Cloudfront instead of S3.


ameukam:
/hold
cc @BenTheElder @dims

PR #6657, opened 2024-04-05
Add restart script to codesearch for daily restart

SohamChakraborty:
Resolves [2182](https://github.com/kubernetes/k8s.io/issues/2182)

Inspired from https://github.com/kubernetes/k8s.io/pull/3679

Adds a restart.sh script that will be run daily by the periodic-k8sio-deploy-app-codesearch prow job.
This will ensure we restart codesearch daily and update search indexes to include any new git changes.
Rolling upgrade is done along with checking the status for a timeout window of 30 minutes.


apps/codesearch/restart.sh
```diff
@@ -0,0 +1,51 @@
+#!/usr/bin/env bash
+
+# Copyright 2021 The Kubernetes Authors.
```
ameukam, line 3:
```suggestion
# Copyright 2024 The Kubernetes Authors.
```

PR #6810, opened 2024-05-14
Update groups.yaml to add content team shadows

nitishfy:
Add content-Team  shadows in the summit-team group


mfahlandt (commented):
/lgtm

PR #6832, opened 2024-05-22
add RBAC for verify-conformance

BobyMCbobs:
add permissions

apart of #6357, https://github.com/kubernetes/org/issues/4706


BobyMCbobs:
This change is no longer needed since verify-conformance is now running in GitHub Actions.
https://github.com/cncf/k8s-conformance/actions/workflows/verify-conformance.yml

PR #6839, opened 2024-05-23
promote pause:3.10

BenTheElder:
with build from https://github.com/kubernetes/kubernetes/pull/125105

digest is visible at the end of the logs from https://prow.k8s.io/view/gs/kubernetes-jenkins/logs/post-kubernetes-push-image-pause/1793743356572995584


BenTheElder:
this is promoted via https://github.com/kubernetes/k8s.io/pull/6840

PR #6853, opened 2024-05-30
add terraform code to deploy patch release notification service

cpanato:
this PR adds the terraform code to deploy the service we are building in https://github.com/kubernetes/release/pull/3627

/assign @xmudrii @saschagrunert @ameukam 

xref: https://github.com/kubernetes/release/pull/3627

### DIFF
--- a/infra/aws/terraform/sig-release/patch-release-notification/README.md
+++ b/infra/aws/terraform/sig-release/patch-release-notification/README.md
@@ -1,0 +1,32 @@
+# patch-release-notification service
+
+This terraform code deploys the code to notify the K8s community about the cherry pick deadline for the patch releases.
+
+The patch-release-notification code can be found in `cmd/patch-release-notification`
+
+Right now, the terraform is applied manually by the release managers that have access to the AWS account for the SIG-release.
+
+# Deploy
+
+To deploy will require to have both repositories cloned:
+
+- https://github.com/kubernetes/release/
+- https://github.com/kubernetes/k8s.io
+
+from https://github.com/kubernetes/k8s.io
+
+```
+# loging to the AWS ECR
+$ aws ecr get-login-password --region us-west-2 | docker login --username AWS --password-stdin 433650573627.dkr.ecr.us-west-2.amazonaws.com
+
+$ cd k8s.io/infra/aws/terraform/sig-release/patch-release-notification
+
+$ terraform init
+
+$ terraform plan -out=plan.out
+
+$ terraform apply "plan.out"
+```
+
+_note_: you will need to configure your AWS credentials before.
+_note2_: this is setup to run in the AWS SIG-Release account.
--- a/infra/aws/terraform/sig-release/patch-release-notification/variables.tf
+++ b/infra/aws/terraform/sig-release/patch-release-notification/variables.tf
@@ -1,0 +1,51 @@
+/*
+Copyright 2024 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+variable "region" {
+  description = "The AWS region to deploy the resources"
+  type        = string
+}
+
+variable "email_identity" {
+  description = "The email address or domain to verify"
+  type        = string
+}
+
+variable "to_email" {
+  description = "The email address to send the notification"
+  type        = string
+}
+
+variable "no_mock" {
+  description = "if will send the message to dev@kubernetes.io or just internal"
+  type        = bool
+}
+
+variable "days_to_alert" {
+  description = "when to send the notification"
+  type        = number
+}
+
+variable "schedule_path" {
+  description = "path where we can find the schedule.yaml"
+  type        = string
+}
+
+variable "repository" {
+  description = "The ECR repository to use for the image"
+  type        = string
+  default     = ""
+}
--- a/infra/aws/terraform/sig-release/patch-release-notification/versions.tf
+++ b/infra/aws/terraform/sig-release/patch-release-notification/versions.tf
@@ -1,0 +1,36 @@
+/*
+Copyright 2024 The Kubernetes Authors.
+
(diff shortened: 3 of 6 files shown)


infra/aws/terraform/sig-release/patch-release-notification/README.md
```diff
@@ -0,0 +1,32 @@
...
+Right now, the terraform is applied manually by the release managers that have access to the AWS account for the SIG-release.
+
+# Deploy
+
+To deploy will require to have both repositories cloned:
+
+- https://github.com/kubernetes/release/
+- https://github.com/kubernetes/k8s.io
+
+from https://github.com/kubernetes/k8s.io
```
xmudrii, line 16: I'm not a big fan of doing this because I think it's going to be very hard to automate the Terraform execution in the future. @upodroid Maybe you can help here, I'm thinking out loud, but can we setup SA with IRSA in EKS build cluster that would allow pushing to this registry, and then we run a ProwJob that's going to use that SA to build the image and push it to the registry?

cpanato: the main idea here is to use terraform to build the image when is needed and deploy the lambda, then everytime we run that and if there are changes in the code it will generate a new image

cpanato: but it is not hard to automate, in the automation we clone both repos, we do this in several places already for other things

upodroid:
Try to follow the approach that we do for archeio(registry.k8s.io). We build an image, publish it and then update the terraform code to use that revision.

xmudrii:
> but there is any reason to do that? why not build on the fly? we also reduce a few steps and overhead

One thing that I'm worried about is that depending on the CI/CD solution/approach we go with for Terraform, it might be hard/impossible to also clone k/release and run the build process. Building also requires some more resources than just applying Terraform configs, so there's that too. This is kind of wild guess, but it's better to avoid such problems if we can.

cpanato:
> > but there is any reason to do that? why not build on the fly? we also reduce a few steps and overhead
> 
> One thing that I'm worried about is that depending on the CI/CD solution/approach we go with for Terraform, it might be hard/impossible to also clone k/release and run the build process. Building also requires some more resources than just applying Terraform configs, so there's that too. This is kind of wild guess, but it's better to avoid such problems if we can.

I would disagree with that, but I will work on refactoring in the way k8s infra likes more.

infra/aws/terraform/sig-release/patch-release-notification/variables.tf
xmudrii, line 1: We should also commit `terraform.tfvars` (if no sensitive information) or add default values to this file.

infra/aws/terraform/sig-release/patch-release-notification/versions.tf
```diff
@@ -0,0 +1,36 @@
...
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+terraform {
+  backend "s3" {
+    bucket = "tf-state-sig-release"
```
xmudrii, line 19: Is this bucket manually created?

infra/aws/terraform/sig-release/patch-release-notification/versions.tf
```diff
@@ -0,0 +1,36 @@
...
+    key    = "cherry-pick-notification"
+    region = "us-west-2"
+  }
+
+  required_version = "1.8.0"
```
xmudrii, line 24: Can we relax the version if we don't have any dependency on 1.8.0?

infra/aws/terraform/sig-release/patch-release-notification/versions.tf
cpanato: yes, we need to bootstrap some how :)

infra/aws/terraform/sig-release/patch-release-notification/versions.tf
cpanato: sure

infra/aws/terraform/sig-release/patch-release-notification/variables.tf
cpanato: I will commit, but as we discussed here, I will need to refactor not to build the image here and use a prebuilt image that will be done in another job.

but, I need to know how to push the image to the AWS account and more specifically to the aws account we are interested in, how that can be done? cc @upodroid

xmudrii: @cpanato I would try the following approach if it's okay with @upodroid:

- Create a new IAM role that can push to the said registry
- Create a new ServiceAccount inside the EKS Prow build cluster
- Bind SA and IAM role via IRSA
- Create a ProwJob that utilizes the said SA and make it build and push the image

There might some additional steps between 1 and 2 in terms of configuring the AWS account, but we have done that before.

xmudrii: If y'all are okay with that, I can take care of setting up everything because I think I have access to all the resources/accounts that we need for this.

cpanato: we will build it using cloudbuild? or can i use `ko` (https://github.com/ko-build/ko) to do the image/push build?

infra/aws/terraform/sig-release/patch-release-notification/versions.tf
xmudrii: What I have been doing before for both SIG Release and SIG K8s Infra is using CloudFormation: https://github.com/kubernetes/k8s.io/blob/main/infra/aws/cloudformation/iam/cdn.packages.k8s.io/cloudformation.yaml

Maybe we can take the same approach here if that's okay for y'all. It still requires manual bootstrapping (i.e. running that CloudFormation script), but all the resources are created automatically.

infra/aws/terraform/sig-release/patch-release-notification/variables.tf
xmudrii: You should be able to use `ko`, let me take a look.

infra/aws/terraform/sig-release/patch-release-notification/versions.tf
cpanato: Yikes! cloudformation, i thought that did not exist anymore LOL

but sounds good to me, i can add that, and delete the bucket that I've create before.
I will add that as well in this PR
thanks!

xmudrii:
@cpanato Can I help with anything to get this moving forward?

PR #6854, opened 2024-05-30
[JobSet] Promote JobSet v0.5.2 to prod

danielvegamyhre:
Promote JobSet v0.5.2 to prod


danielvegamyhre:
@ahg-g  I added a hold since it turns out relaxing the validation alone is not sufficient to meet Kueue's requirements, see https://github.com/kubernetes-sigs/jobset/issues/586

danielvegamyhre:
Confirmed new image tagged v0.5.2 works via manual testing of updating a suspended JobSet's pod template labels, then resuming the JobSet and checking the Job and Pod both had the label.

PR #6857, opened 2024-05-31
Kueue: Add v0.7.0 image

tenzen-y:
I added the kueue image for the v0.7.0.


alculquicondor:
/hold
we need a new tag

PR #6869, opened 2024-06-05
Allow anyone to post messages to kueue-alerts, under moderation

alculquicondor:
Since we haven't received any alerts from testgrid so far, I don't think the settings were right.

I'm opening the group for moderation for now. We can allowlist some emails if somehow this group becomes noisy.


tenzen-y (commented):
/lgtm

PR #6921, opened 2024-06-27
GCP: ensure default compute SA can use KMS keys

pwschuurman:
Ensure the default compute SA can use KMS keys. This is required for the GCP PDCSI tests to call the GCE `instances.insert` API, as VMs are created with the Compute Engine default service account.

The GCP PDCSI e2e test prowjobs have failed after migrating to use the k8s-infra-prow-build cluster: https://github.com/kubernetes/test-infra/pull/32809

### DIFF
--- a/infra/gcp/bash/prow/ensure-e2e-projects.sh
+++ b/infra/gcp/bash/prow/ensure-e2e-projects.sh
@@ -55,6 +55,9 @@ function ensure_e2e_project() {
 
     ensure_project "${prj}"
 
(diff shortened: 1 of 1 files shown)


BenTheElder (commented):
/lgtm
/approve
/hold

BenTheElder (commented):
/lgtm

BenTheElder:
This is running.

BenTheElder:
It should be applied to all projects now.

PR #6923, opened 2024-06-27
GCP: ensure default engine service agent SA can use KMS keys

pwschuurman:
Ensure the default compute engine service agent SA can use KMS keys. Compute engine has a special service agent account that needs access to cloudkms. This is required for the GCP PDCSI tests to call the GCE `instances.insert` API, as VMs are created with the Compute Engine default service account. See https://github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/pull/1762#issuecomment-2195525892

The GCP PDCSI e2e test prowjobs have failed after migrating to use the k8s-infra-prow-build cluster: https://github.com/kubernetes/test-infra/pull/32809

### DIFF
--- a/infra/gcp/bash/prow/ensure-e2e-projects.sh
+++ b/infra/gcp/bash/prow/ensure-e2e-projects.sh
@@ -92,6 +92,12 @@ function ensure_e2e_project() {
       "serviceAccount:${project_number}-compute@developer.gserviceaccount.com" \
       "roles/cloudkms.cryptoKeyEncrypterDecrypter"
 
+    # Ensure GCP Default Compute Engine Service Agent Account can manage KMS
+    # keys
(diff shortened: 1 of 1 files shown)


BenTheElder (commented):
/lgtm
/approve

BenTheElder:
This is deploying ... will comment again when it's done for the main pool (happens first) and/or when it's totally done.

BenTheElder:
Main project pool should be good go to now, we can try this in the driver repo now.

(left this unattended and it had a connection issue, I'd invest in built-in retries but I don't think we're likely to be tweaking this often in the future)

Running again to apply to the remaining projects in the other pools.

BenTheElder:
Running again for the full set.

BenTheElder:
This is deployed to all projects.

PR #6950, opened 2024-07-06
Kops - Create S3 bucket for hosting CI results

rifelpet:
Barring a better solution for interacting with both a GCS bucket and the Kops AWS account in one prow job (see https://github.com/kubernetes/kops/issues/16637#issuecomment-2199004031), I'm experimenting with storing Kops' version markers and CI artifacts on S3 instead of GCS, that way the prow jobs only need to interact with one cloud provider.

I will test this with https://github.com/kubernetes/kops/pull/16659 and https://github.com/kubernetes/test-infra/pull/32918

/cc @upodroid @hakman

### DIFF
--- a/infra/aws/terraform/kops-infra-ci/main.tf
+++ b/infra/aws/terraform/kops-infra-ci/main.tf
@@ -76,3 +76,34 @@ resource "aws_s3_bucket_acl" "kops_oidc_store" {
   bucket   = aws_s3_bucket.kops_oidc_store.id
   acl      = "public-read"
 }
+
+## Used by kOps for hosting CI build artifacts and version markers
+resource "aws_s3_bucket" "kops_ci_results" {
+  provider = aws.kops-infra-ci
+  bucket   = "k8s-infra-kops-ci-results"
+  tags     = var.tags
+}
+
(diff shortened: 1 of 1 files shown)


upodroid:
Instead of doing this, I can fix the rbac issues

rifelpet:
Ok that works for me too. kubetest2-kops today uses the k8s-staging-kops bucket you introduced here:

The new `pull-kops-e2e-k8s-aws-calico-k8s-infra` job that runs on the `k8s-infra-kops-prow-build` prow cluster reports this error:
 
`AccessDeniedException: 403 prow-build@k8s-infra-prow-build.iam.gserviceaccount.com does not have storage.objects.create access to the Google Cloud Storage object. Permission 'storage.objects.create' denied on resource (or it may not exist).`

If we can fix that RBAC then we should be able to migrate all presubmit jobs.

If we reuse the k8s-staging-kops bucket for version markers, we just need to update the version markers in our prow jobs to use the new bucket:

https://github.com/kubernetes/test-infra/blob/ff39426505bbbfea5b4a70e9213ffd070bdb5db4/config/jobs/kubernetes/kops/build_jobs.py#L755

Let me know when you've fixed the RBAC and I can continue with this migration work.

PR #6958, opened 2024-07-09
fastly: bump terraform provider

ameukam:
Bump the provider to the latest version so we can define Fastly alerts.


ameukam:
cc @BenTheElder @dims

BenTheElder (commented):
/lgtm
/approve
thanks!

PR #6968, opened 2024-07-10
update slack-moderator-words to latest build, pin digest

BenTheElder:
TODO: these should all be served from registry.k8s.io AND pinned by digest ... running from staging via a tag is NOT secure

focusing on this deployment because https://github.com/kubernetes-sigs/slack-infra/issues/56 ... we're doing a LOT of unnecessary logging still, this is an easy win hopefully.

/sig contributor-experience


ameukam:
Deployed: https://prow.k8s.io/view/gs/kubernetes-jenkins/logs/post-k8sio-deploy-app-slack-infra/1811281764493037568

PR #6969, opened 2024-07-11
[WIP] oci-proxy: block most irrelevant requests at the edge with a simple path based rule

BenTheElder:
xref: https://github.com/kubernetes/registry.k8s.io/issues/286

I checked the logs for:
httpRequest.requestUrl!~"^https://registry.k8s.io/(?:$)|(?:v2/)"

We get a lot of requests that are malformed or not registry API or other spam, this won't catch everything but it will exclude a LOT.

404 at the WAF should be cheaper than forwarding it to cloud run.

If we ship this, we can probably drop the other WAF rules (except rate limiting) since nearly all of the other requests will already be blocked anyhow.

### DIFF
--- a/infra/gcp/terraform/modules/oci-proxy/cloud-armor.tf
+++ b/infra/gcp/terraform/modules/oci-proxy/cloud-armor.tf
@@ -21,6 +21,50 @@ resource "google_compute_security_policy" "cloud-armor" {
   project = var.project_id
   name    = "security-policy-oci-proxy"
 
+  # apply rate limits
+  rule {
+    action      = "throttle"
+    description = "Default rule, throttle traffic"
+    priority    = "2147483647"
+
+    match {
+      config {
+        src_ip_ranges = ["*"]
+      }
+      versioned_expr = "SRC_IPS_V1"
+    }
+
+    rate_limit_options {
+      conform_action = "allow"
+      exceed_action  = "deny(429)"
+
+      enforce_on_key = "IP"
+      # This is comparable to the GCR limits from k8s.gcr.io
+      rate_limit_threshold {
+        count        = 5000
+        interval_sec = 60
+      }
+    }
+
+    preview = false
+  }
+
+  // block all requests with obviously invalid paths at the edge
+  // we support "/", "/privacy", and "/v2/.*" API
+
+  rule {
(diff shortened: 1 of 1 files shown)


infra/gcp/terraform/modules/oci-proxy/cloud-armor.tf
```diff
@@ -21,6 +21,50 @@ resource "google_compute_security_policy" "cloud-armor" {
...
+
+  // block all requests with obviously invalid paths at the edge
+  // we support "/", "/privacy", and "/v2/.*" API
+
+  rule {
+    action   = "deny(404)"
+    priority = "2147483646"
+    match {
+      expr {
+        expression = "!request.path.match('(?:^/$)|(?:^/privacy$)|(?:^/v2/)')"
```
BenTheElder, line 60: ... untested, not sure if there's a good way to test this, and while the language is documented it's a bit sparse about things like regex support ...

but that's why we have the staging endpoint :-)

will look further tomorrow maybe

BenTheElder: I'm assuming the regex match API is RE2 compatible given it's a GCP product, but the docs don't specify that part.

ameukam: > The Google Cloud Armor custom rules language is a subset of the [Common Expression Language (CEL)](https://github.com/google/cel-spec).

https://cloud.google.com/armor/docs/rules-language-reference#allow_or_deny_traffic_for_a_request_uri_that_matches_a_regular_expression

And
> Regular expressions follow the [RE2 syntax](https://github.com/google/re2/wiki/Syntax).
https://github.com/google/cel-spec/blob/master/doc/langdef.md#regular-expressions

It should be fine ?

BenTheElder: let's try this in staging then

BenTheElder:
NOTE: this will be auto-deployed to staging and we'll iterate from there. This will NOT be deployed to registry.k8s.io (production) yet.

BenTheElder:
This appears to be WAI, filing a follow-up.

PR #6973, opened 2024-07-11
promote latest slack-infra images

BenTheElder:
TODO: get references updated after these are promoted, also likely get new builds and do it again ...

first part of https://github.com/kubernetes-sigs/slack-infra/issues/66


k8s.gcr.io/images/k8s-staging-slack-infra/images.yaml
```diff
@@ -1,15 +1,20 @@
 - name: slack-event-log
```
BenTheElder, line 1: gah, we should really delete these old promoter manifests, will get this moved over to the current one in registry.k8s.io/

BenTheElder:
NOTE: it's been so long since any were built, so I had to assume the currently available digests are good.

In the future it should be strongly preferred to obtain the digests from the cloud build logs following a secure automated build occurring, and promote those.

PR #6974, opened 2024-07-11
Update desired size for prow-build-cluster nodes

dims:
Discussion in slack:
https://kubernetes.slack.com/archives/CCK68P2Q2/p1720637247581549

### DIFF
--- a/infra/aws/terraform/prow-build-cluster/terraform.prod.tfvars
+++ b/infra/aws/terraform/prow-build-cluster/terraform.prod.tfvars
@@ -48,15 +48,15 @@ node_instance_types_stable     = ["r5ad.2xlarge"]
 
 node_min_size_us_east_2a     = 7
 node_max_size_us_east_2a     = 25
-node_desired_size_us_east_2a = 7
+node_desired_size_us_east_2a = 15
 
 node_min_size_us_east_2b     = 7
 node_max_size_us_east_2b     = 25
-node_desired_size_us_east_2b = 7
+node_desired_size_us_east_2b = 15
 
 node_min_size_us_east_2c     = 7
 node_max_size_us_east_2c     = 25
-node_desired_size_us_east_2c = 7
+node_desired_size_us_east_2c = 15
 
 node_desired_size_stable = 3
 


infra/aws/terraform/prow-build-cluster/terraform.prod.tfvars
```diff
@@ -48,15 +48,15 @@ node_instance_types_stable     = ["r5ad.2xlarge"]
...
 
 node_min_size_us_east_2b     = 7
 node_max_size_us_east_2b     = 25
-node_desired_size_us_east_2b = 7
+node_desired_size_us_east_2b = 15
 
 node_min_size_us_east_2c     = 7
 node_max_size_us_east_2c     = 25
-node_desired_size_us_east_2c = 7
+node_desired_size_us_east_2c = 15
```
BenTheElder, line 59: does this mean we're just forcing a larger minimum and bypassing scale-up issues that way?

dims: @BenTheElder one time thing for now as we did turn on a lot more jobs recently. (note this is not the minimum, it's the desired size)

dims: i am reflecting the change i made in the UI here.

BenTheElder: OK, I'm used to CA on GKE with min/max.

I looked at the docs and it reads like this will be clobbered by autoscaling, and would only impact the initial size??

https://docs.aws.amazon.com/eks/latest/APIReference/API_NodegroupScalingConfig.html#API_NodegroupScalingConfig_Contents

> If you use the Kubernetes [Cluster Autoscaler](https://github.com/kubernetes/autoscaler#kubernetes-autoscaler), you shouldn't change the desiredSize value directly, as this can cause the Cluster Autoscaler to suddenly scale up or scale down.

(are we?)

dims: you are right. they came back down.

image

dims: let me close this PR and we can continue on slack

BenTheElder: ACK, thanks for working on this 🙏

PR #6975, opened 2024-07-11
simplify WAF rules

BenTheElder:
follow-up to https://github.com/kubernetes/k8s.io/pull/6969

intend to deploy to staging

### DIFF
--- a/infra/gcp/terraform/modules/oci-proxy/cloud-armor.tf
+++ b/infra/gcp/terraform/modules/oci-proxy/cloud-armor.tf
@@ -23,9 +23,9 @@ resource "google_compute_security_policy" "cloud-armor" {
 
   # apply rate limits
   rule {
-    action      = "throttle"
-    description = "Default rule, throttle traffic"
-    # apply rate limit first (rules are applied sequentially by priority)
+    action      = "rate_based_ban"
+    description = "Limit excessive usage"
+    # apply rate limits first (rules are applied sequentially by priority)
     # https://cloud.google.com/armor/docs/security-policy-overview#eval-order
     priority = "0"
 
@@ -39,20 +39,29 @@ resource "google_compute_security_policy" "cloud-armor" {
     rate_limit_options {
       conform_action = "allow"
       exceed_action  = "deny(429)"
-
       enforce_on_key = "IP"
-      # This is comparable to the GCR limits from k8s.gcr.io
+      # TODO: revisit these values
+      # above this threshold we serve 429, currently ~83/sec in a 1 minute window
       rate_limit_threshold {
+        # NOTE: count cannot exceed 10,000
+        # https://cloud.google.com/armor/docs/rate-limiting-overview
         count        = 5000
         interval_sec = 60
       }
+      # if the user continues to exceed the rate limit, temp ban
+      # otherwise users may ignore transient 429 and keep running right at the limit
+      # clients that respect the 429 and backoff will not hit this
+      # (or better yet, https://github.com/kubernetes/registry.k8s.io/blob/main/docs/mirroring/README.md)
+      ban_threshold {
+        count        = 10000
+        interval_sec = 120
+      }
+      ban_duration_sec = 1800
     }
-
-    preview = false
   }
 
   // block all requests with obviously invalid paths at the edge
-  // we support "/", "/privacy", and "/v2/.*" API
+  // we support "/", "/privacy", and "/v2/.*" API, GET or HEAD
 
   rule {
     action = "deny(404)"
@@ -60,7 +69,7 @@ resource "google_compute_security_policy" "cloud-armor" {
     priority = "1"
     match {
       expr {
-        expression = "!request.path.match('(?:^/$)|(?:^/privacy$)|(?:^/v2/)')"
+        expression = "!request.path.matches('(?:^/$)|(?:^/privacy$)|(?:^/v2/)')"
       }
     }
   }
--- a/infra/gcp/terraform/modules/oci-proxy/cloud-armor.tf
+++ b/infra/gcp/terraform/modules/oci-proxy/cloud-armor.tf
@@ -25,7 +25,9 @@ resource "google_compute_security_policy" "cloud-armor" {
   rule {
     action      = "throttle"
     description = "Default rule, throttle traffic"
-    priority    = "2147483647"
+    # apply rate limit first (rules are applied sequentially by priority)
+    # https://cloud.google.com/armor/docs/security-policy-overview#eval-order
+    priority = "0"
 
     match {
       config {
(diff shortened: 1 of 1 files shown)


BenTheElder:
this is easier to read commit-by-commit

infra/gcp/terraform/modules/oci-proxy/cloud-armor.tf
```diff
@@ -37,162 +39,39 @@ resource "google_compute_security_policy" "cloud-armor" {
...
-      expr {
-        expression = "evaluatePreconfiguredWaf('php-v33-stable', {'sensitivity': 3})"
-      }
-    }
-    description = "PHP"
-
-    preview = false
-  }
+  // block all requests with obviously invalid paths at the edge
+  // we support "/", "/privacy", and "/v2/.*" API, GET or HEAD
```
BenTheElder, line 64: we aren't blocking unknown methods here because we can't return 405, only 403, 404, or 502

BenTheElder: (we do block them at oci-proxy, but by then we've had to serve this with cloud run, still we don't redirect them to AR)

kwilczynski:
@BenTheElder, is there any visibility into what rules block how much traffic? Like a dashboard showing a count blocked or a percentage of blocked requests? Would be nice to be able to tell which ones are the most effective ones... the rules, that is.

For example, Cloudflare has a nice breakdown in their WAF, so I wonder if Cloud Armor has something like that. I don't have it [Cloud Armor] deployed anywhere, so I am unable to check things at the moment.

BenTheElder:
> @BenTheElder, is there any visibility into what rules block how much traffic? Like a dashboard showing a count blocked or a percentage of blocked requests? Would be nice to be able to tell which ones are the most effective ones... the rules, that is.

Yes, but there are some complications ...

SIG K8s Infra Tech Leads (including myself) have access to pretty detailed structured logs for debugging purposes, we can get quite a lot of detail from querying what rules were applied etc., aside from metrics panels etc.

BUT ... cloud armor applies rules sequentially in the specified priority order, so it's obviously biased towards earlier rules, and it's hard to determine the overlap because the rest of these basically just say "please apply apply this [OWASP CoreRuleSet](https://github.com/coreruleset/coreruleset) standard set of rules by name" (https://cloud.google.com/armor/docs/waf-rules), so we'd have to go dig through the OWASP CRS definitions.

That said ... I'm using the structured logs to inform the new rules.

Almost all requests currently are either:
a) not blocked
b) an invalid api call / path
c) rate limited
d) rate limited AND an invalid api call / path (e.g. we're currently rate limiting some clients that hit a made-up "/.well-known/app-capabilities" path (https://github.com/quay/container-security-operator/issues/141#issuecomment-2221782520 ?), which is impressive given we currently permit 5000 queries / minute per IP ...)

I'm looking at log query results for e.g.:
```
jsonPayload.enforcedSecurityPolicy.configuredAction!="THROTTLE"
httpRequest.requestUrl!~"://registry.k8s.io/v2/.*"
```
```
jsonPayload.enforcedSecurityPolicy.configuredAction!="THROTTLE"
httpRequest.requestUrl=~"://registry.k8s.io/v2/.*"
```



These new rules will cover most of b) c) d) by just bluntly 404-ing invalid paths at the edge, which isn't a behavior change in terms of which requests are accepted (https://github.com/kubernetes/registry.k8s.io/blob/main/cmd/archeio/docs/request-handling.md), we just shift most of it to the edge and drop the OWASP rules in favor of rejecting anything with an invalid path.

I'm planning to iterate on refining the regex to only permit valid, standard, supported API calls (plus the two information endpoints `/` and `/privacy`) once this is proven out in staging, but this blocks most of them already.

Unfortunately we have a lot of remaining spam that are technically valid looking API calls related to images we don't host by pull-through cache/mirrors that are misconfigured or poorly behaved and trying to pull garbage images (like `/v2/dockerfile/tags/list?` AKA an image named "dockerfile" 🙄 ) or what are clearly commercial internal images, or various non-standard APIs that are under `/v2/`.

We'll want to block more of that, but the OWASP rules generally weren't blocking those calls anyhow (and if they are, they risked false positives for valid image API calls).

The results are not public for privacy reasons, but community members that contribute and build trust can be granted access, for now we've limited it to the tech leads for the infra group. Most of our infra is a bit more open access to community members, but in this case we're concerned about the details in the ephemeral debug / loadbalancer logs.

BenTheElder:
This change also partially addresses that we do have some clients constantly hitting the rate limit, by extending the rate limiting with a temp ban if they ignore the 429 response and keep hitting the registry at `2*rate_limit_period` with `2*rate_limit_count` (rate limit: 5000/60s, temp_ban: 10000/120s), we'll probably need to iterate on that. I'm hoping these users will notice when there's a 30m 429 block instead of hitting the limit inside 60s and then continuing to keep hitting the loadbalancer all throughout anyhow and getting more API calls in the next minute.


We might want to put longer rate limit windows, but that will also likely imply tighter limits, because 10000 is the max configurable count for a limit of any time period. 

That said, most clients should be fine with a lower limit, we picked the current limit because it was easy to implement and comparable-ish to what was GCR's limit at the time of 50000/600s (it's currently 60000/600s).

infra/gcp/terraform/modules/oci-proxy/cloud-armor.tf
```diff
@@ -37,162 +39,39 @@ resource "google_compute_security_policy" "cloud-armor" {
...
   rule {
-    action   = "deny(403)"
-    priority = "1010"
+    action = "deny(404)"
+    # apply this broad 404 for unexpected paths second
+    priority = "1"
     match {
       expr {
-        expression = "evaluatePreconfiguredExpr('cve-canary')"
+        expression = "!request.path.matches('(?:^/$)|(?:^/privacy$)|(?:^/v2/)')"
```
BenTheElder, line 72: This would block a bit more:

```suggestion
        expression = "!request.path.matches('(?:^/$)|(?:^/privacy$)|(?:^/v2$)|(?:^/v2/$)|(?:^/v2/.+/(:?blobs|manifests|tags)/.+')"
```

Need to double check if there's any other calls. Also it would solve https://github.com/kubernetes/registry.k8s.io/issues/287 and similar such issues. If we add support for new APIs we can just add them to the regex.

BenTheElder:
/hold cancel

will deploy this and then iterate on the path match once I see the rest working

infra/gcp/terraform/modules/oci-proxy/cloud-armor.tf
```diff
@@ -23,9 +23,11 @@ resource "google_compute_security_policy" "cloud-armor" {
 
   # apply rate limits
   rule {
-    action      = "throttle"
-    description = "Default rule, throttle traffic"
```
BenTheElder, line 27: You _must_ have a default rule with this specific priority (The highest value / last applied rule), not doing so is invalid.

We still want throttling to be applied first, however, so garbage requests are factored into throttling.

BenTheElder: It failed to deploy for this reason:
https://prow.k8s.io/view/gs/kubernetes-jenkins/logs/post-k8s-deploy-registry/1811794309389750272

> googleapi: Error 400: Invalid value for field 'priority': '2147483647'. Cannot remove the default rule., invalid

https://stackoverflow.com/a/60401055

BenTheElder: https://github.com/kubernetes/k8s.io/pull/6979

infra/gcp/terraform/modules/oci-proxy/cloud-armor.tf
BenTheElder: https://github.com/kubernetes/k8s.io/pull/6979

PR #6976, opened 2024-07-12
Tweak scale-down-utilization-threshold to reduce pod scheduling issues

dims:
Trying to see if this would help with the situation we are seeing:
https://kubernetes.slack.com/archives/CCK68P2Q2/p1720637247581549


xmudrii:
The issue that we encountered has been related to something else (a node stuck in the ASG). Given that, I don't think we need this PR at the moment, but if we merge it, please check with @koksay if we need to change anything in our Karpenter implementation to reflect this

dims:
closing it. thanks @xmudrii

PR #6979, opened 2024-07-12
oci-proxy: add default rule, improve path regex

BenTheElder:
follow-up to https://github.com/kubernetes/k8s.io/pull/6975

PR #6980, opened 2024-07-12
implement regex without capture groups, which are not permitted

BenTheElder:
https://github.com/kubernetes/k8s.io/pull/6979#issuecomment-2225945822

.... not too bad, hopefully there isn't some other undocumented restriction >.>


BenTheElder:
I had to manually remove and delete the apparmor policy then retrigger the deployment, it got stuck on trying to convert the default rule from throttle => allow as far as I can tell. Seems to be a bug in the tf module.

We will want to do that quickly when we move to prod, but prod deploys are also faster since there's no image push.

https://prow.k8s.io/view/gs/kubernetes-jenkins/logs/post-k8s-deploy-registry/1811829196788338688

successful deployment, now on to testing ..

PR #6982, opened 2024-07-12
Add a AWS cli command for getting counts of instances in node groups

dims:
CLI is a bit easier to run with a `watch` instead of refreshing the web page.


koksay:
/lgtm
/approve

CLI makes more sense, thanks @dims

PR #6984, opened 2024-07-12
/tags/list

BenTheElder:
... it was even in the comment, just not the regex

`curl -L https://registry.k8s.io/v2/pause/tags/list`
versus
`curl -L https://registry-sandbox.k8s.io/v2/pause/tags/list`


jeremyrickard (approved):
/lgtm

BenTheElder:
https://github.com/opencontainers/distribution-spec/blob/main/spec.md#endpoints

PR #7009, opened 2024-07-18
delete k8s.gcr.io directory

BenTheElder:
we can look in git history if we really still want to see these, otherwise they're just confusing for OWNERS updates and image promotion PRs, which should be done to registry.k8s.io/ instead for a year+ now easily.


BenTheElder:
/kind cleanup
/sig k8s-infra

will manually bypass the blocked-paths and then drop that from prow config once this is lgtm+approve

BenTheElder:
ah, we had a test script still testing it ...

PR #7027, opened 2024-07-18
Azure: Storage accounts for Terraform states

ameukam:
Create a dedicated resource group and required storage accounts to use Azure as a remote backend.

### DIFF
--- a/infra/azure/bash/ensure-resource-groups.sh
+++ b/infra/azure/bash/ensure-resource-groups.sh
@@ -1,0 +1,97 @@
+#!/usr/bin/env bash
+
+# Copyright 2024 The Kubernetes Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# This script creates a Azure resource group, storage accounts and storage needed
+# to store Terraform states.
+
+set -o errexit
+set -o nounset
+set -o pipefail
+
+# shellcheck disable=SC2034
+SCRIPT_DIR=$(dirname "${BASH_SOURCE[0]}")
+
+if ! [[ -x "$(command -v jq)" ]]; then
+  echo "Please install jq. https://stedolan.github.io/jq/download/.  Aborting."
+  exit 1
+fi
+
+if ! [[ -x "$(command -v az)" ]]; then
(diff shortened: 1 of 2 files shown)


ameukam:
cc @upodroid @BenTheElder

infra/azure/bash/ensure-resource-groups.sh
```diff
@@ -0,0 +1,95 @@
...
+fi
+
+RESOURCE_GROUP="k8s-infra-tf-states-rg"
+RESOURCE_GROUP_LOCATION=$(az group show --name $RESOURCE_GROUP --output tsv --query 'location')
+
+# storage accounts rules: https://learn.microsoft.com/en-us/azure/azure-resource-manager/management/resource-name-rules#microsoftstorage
+#TODO define a role assignement with Azure Entra Groups
+readonly TERRAFORM_STATE_BUCKET_ENTRIES=(
+  k8sinfratfstatesub
+  k8sinfratfstateprowbuild
```
upodroid, line 44: A single bucket is fine, given at users will be able to access all the buckets in the sub.

upodroid: You apply RBAC to the resource groups instead of the subscriptions.

ameukam: so one bucket where we keep tf changes impacting the subscription and  only accessible for the sub owners would be fine ?

infra/azure/bash/ensure-resource-groups.sh
```diff
@@ -0,0 +1,95 @@
...
+    echo "resource group $RESOURCE_GROUP already exists"
+  fi
+
+  echo "Retrieving locks for resource group: $RESOURCE_GROUP"
+  LOCKS=$(az lock list -g $RESOURCE_GROUP --query '[].{name:name}' --output json)
+  echo "$LOCKS"
+
+  if [[ $LOCKS == "[]" ]]; then
+    echo "Creating lock for resource group: $RESOURCE_GROUP"
+    az lock create --name "DO_NOT_DELETE" --lock-type CanNotDelete --resource-group "$RESOURCE_GROUP"
```
ritikaguptams, line 90: Lets do a tag here of DO_NOT_DELETE instead of a lock? I think that should suffice for the cleanup app to skip this resource group.

ameukam: Beside the cleanup app, we need to ensure rgs are not accidentally deleted.

ritikaguptams: Yes alright makes sense, then I think we should let the lock be and also add a tag

ameukam:
@hakman I think we want a separate rg for kops e2e tests ?

hakman:
> @hakman I think we want a separate rg for kops e2e tests ?

sure, that would be cool. thanks!

ameukam:
@upodroid @ritikaguptams PTAL.

infra/azure/bash/ensure-resource-groups.sh
ameukam: Tags added.

upodroid (approved):
LGTM

PR #7044, opened 2024-07-22
Allow members to see membership

SergeyKanzhelev:
This will allow group members to invite the group to meetings


SergeyKanzhelev:
Without this permission scheduling meetings by members is not possible. And some workstream meetings needs to be scheduled by the group members

ArangoGutierrez (approved):
/lgtm

PR #7046, opened 2024-07-22
update prod registry to latest

BenTheElder:
/hold

NOTE: we don't have automated deployments here yet (somewhat by design, this is sensitive and a human should be on hand to rollback when updating it, we DO automatically deploy to staging which is not end-user facing) so this is for visibility, I'm going to deploy it and then rollback if necessary.

### DIFF
--- a/infra/gcp/terraform/modules/oci-proxy/variables.tf
+++ b/infra/gcp/terraform/modules/oci-proxy/variables.tf
@@ -17,8 +17,8 @@ limitations under the License.
 // this one is overridden to the latest image in staging typically
 variable "digest" {
   type = string
-  # v0.3.0-4-g1003b13
-  default  = "sha256:deaae834dbf9247e61bf322d28c758f6a47440d700f734d5f6260957fd05cbf4"
+  # v0.4.0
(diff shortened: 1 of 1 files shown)


BenTheElder:
`terraform plan` looks as expected
- image update
- WAF rewrite

applying

BenTheElder:
Apply complete.

BenTheElder:
I'm looking at:
```
severity=WARNING
jsonPayload.enforcedSecurityPolicy.configuredAction="RATE_BASED_BAN"
jsonPayload.enforcedSecurityPolicy.rateLimitAction.outcome="RATE_LIMIT_THRESHOLD_EXCEED"
```

```
severity=WARNING
jsonPayload.enforcedSecurityPolicy.configuredAction="RATE_BASED_BAN"
jsonPayload.enforcedSecurityPolicy.rateLimitAction.outcome="BAN_THRESHOLD_EXCEED"
```

To check on rate limiting.

So far we have none of the latter, so I don't think that part is working as intended, but at least the rate limit is still doing its job and everything looks fine otherwise. (Also this isn't super surprising, the two limits are too similar, but now that the updated rule types is in place we can iterate on the thresholds, and I was prioritizing the other changes this time)

BenTheElder:
So far no sign of issues.

PR #7055, opened 2024-07-23
Increase max capacity for eks-prow-build-cluster to 102 nodes

xmudrii:
The maximum capacity for the EKS Prow build cluster has been increased from 75 to 102 (34 nodes per the AZ), as we have had maxed out the capacity.

xref https://kubernetes.slack.com/archives/CCK68P2Q2/p1721745534898059

/assign @dims @ameukam @upodroid

### DIFF
--- a/infra/aws/terraform/prow-build-cluster/terraform.prod.tfvars
+++ b/infra/aws/terraform/prow-build-cluster/terraform.prod.tfvars
@@ -47,15 +47,15 @@ node_instance_types_us_east_2c = ["r5ad.4xlarge"]
 node_instance_types_stable     = ["r5ad.2xlarge"]
 
 node_min_size_us_east_2a     = 7
-node_max_size_us_east_2a     = 25
+node_max_size_us_east_2a     = 34
 node_desired_size_us_east_2a = 7
 
 node_min_size_us_east_2b     = 7
-node_max_size_us_east_2b     = 25
+node_max_size_us_east_2b     = 34
 node_desired_size_us_east_2b = 7
 
 node_min_size_us_east_2c     = 7
-node_max_size_us_east_2c     = 25
+node_max_size_us_east_2c     = 34
 node_desired_size_us_east_2c = 7
 
 node_desired_size_stable = 3


dims:
/approve
/lgtm

/hold

xmudrii:
It's already applied
/hold cancel

BenTheElder:
I think we should this increase a lot, like 2x from before, especially if we've been hitting the previous limit, unless we think it's going to send us over budget (seems pretty unlikely and we monitor spend) or we're going to hit some scaling issues with this cluster (seems unlikely at 100s of nodes)

The default cluster is 150 x n1-highmem-8 last I checked, so decommissioning that is definitely expected to increase the community cluster size. We're nearing completion there but still have ~100s of jobs, and we probably want to shift a bit more over between the community clusters for cost engineering.

xmudrii:
@BenTheElder We're are also very closely monitoring the status of the cluster and we can react very quickly if we see that it's not enough. This cluster is using the `r5ad.4xlarge` instance type which has 16 vCPUs and 128 GB RAM, which is double compared to `n1-highmem-8`. With that, 100 nodes in the EKS cluster are corresponding to ~200 nodes in the GKE cluster, so if we did fine with 150 nodes in the GKE cluster, I think we should be fine with 100 nodes in the EKS cluster. I'd say let's monitor this and we can make further adjustments in the coming days if needed.

BenTheElder:
OK, I just think we shouldn't _have_ to monitor the cluster closely and react quickly, we can let the autoscaler do that, and the limit is artificial as long as it's within budget and not causing cluster components to fall over under the load.

xmudrii:
I agree with that point, but we anyways have to monitor it closely because the autoscaler is not very stable at the moment and tends to get stuck. If it gets stuck on a high number of nodes, it can burn the credits very quickly, so we have to be aware of that. This will be mitigated by migrating to Karpenter.

BenTheElder:
>  If it gets stuck on a high number of nodes, it can burn the credits very quickly, so we have to be aware of that.

This is true even with the current limit since we expect to not use much overnight typically, but that's also why we monitor spend.

PR #6896, opened 2024-06-14
[EKS Prow cluster] Karpenter configuration with nodepool and ec2nodeclass

koksay:
This PR is a follow-up for #6895 . It will add `nodepool` and `ec2nodeclass` objects

Make sure to update the `.spec.role` in the `ec2nodeclass.yaml` file.

PR #6895, opened 2024-06-14
[EKS Prow Cluster] Add karpenter terraform module for eks-prow-build cluster

koksay:
- Add karpenter module, which creates related roles, sqs queue and events
- Update EKSAdmin role to be able to create above resources
- Add manually added SSO role to terraform

Make sure to update the `serviceAccount.annotations` field in the `infra/aws/terraform/prow-build-cluster/resources/karpenter/flux-hr-karpenter.yaml` file (also in `infra/aws/terraform/prow-build-cluster/resources/karpenter/prod-cluster-values`):

```bash
# You may use this command to get the arn after terraform apply:
terraform state show module.karpenter.aws_iam_role.controller\[0\] | grep " arn "
```

There will be a follow-up PR to add nodepool and nodeclass configration.

PR #7036, opened 2024-07-19
kops: Bump terraform providers and modules for the build cluster.

ameukam:

PR #7063, opened 2024-07-24
eks-prow-build-cluster: Deploy the Karpenter module

xmudrii:
This PR deploys the Karpenter module to the EKS Prow build cluster to be used as a replacement for the cluster-autoscaler. Karpenter is better tailored to AWS, doesn't use ASGs that were causing us problems before, and provides many more configuration options that we might want to utilize in the future.

This PR:

- Adds a new [Terraform module](https://registry.terraform.io/modules/terraform-aws-modules/eks/aws/latest/submodules/karpenter) to deploy Karpenter
  - The Terraform module includes all the needed resources such as the IAM roles, SQS queues, events, and more
- Update the appropriate IAM policies to allow deploying this module

This PR supersedes #6895. The most of the work in this PR has been done by @koksay, but given that I can't push to his branch and that he's on a vacation, I decided to create a new PR that addresses comments left by @bryantbiggs and me.

~This PR depends on #7062, so please hold off with reviews until that PR is not merged and this one rebased. I also want to do some additional manual testing in the meanwhile.~

Once we merge this PR, we need to merge #6896 in order to enable Karpenter.


xmudrii:
We still need to figure out the migration plan, will be discussed on Slack
/hold

infra/aws/terraform/prow-build-cluster/karpenter.tf
```diff
@@ -18,11 +18,15 @@ module "karpenter" {
...
+  cluster_name = module.eks.cluster_name
+
   create_access_entry             = true
   create_pod_identity_association = true
 
-  namespace = "karpenter"
+  node_iam_role_use_name_prefix = false
+  node_iam_role_name            = "karpenter-nodes"
+
+  queue_name = "karpenter-queue"
```
bryantbiggs, line 29: will there be more than one karpenter controller per account, per region (thinking about potential naming conflicts). Same for the role name. Injection the cluster name usually helps resolve this

xmudrii: No, I don't think so. There are two different accounts and clusters, and I wanted to standardize the naming. These two clusters have different names, so that's why I wanted to avoid injecting the cluster name.

infra/aws/terraform/prow-build-cluster/locals.tf
```diff
@@ -39,6 +39,7 @@ locals {
   auto_scaling_tags = {
     "k8s.io/cluster-autoscaler/${var.cluster_name}" = "owned"
     "k8s.io/cluster-autoscaler/enabled"             = true
```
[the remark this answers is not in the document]
bryantbiggs: yes please - EKS MNG already adds these, not sure if they are still causing conflicts from what @ameukam saw previously https://github.com/kubernetes/k8s.io/pull/7036#issuecomment-2246231339

infra/aws/terraform/prow-build-cluster/locals.tf
```diff
@@ -39,6 +39,7 @@ locals {
...
+    "karpenter.sh/discovery"                        = var.cluster_name
```
bryantbiggs, line 42: I don't think we want this here - we want this on the security group that Karpenter is going to use (i.e. - `node_security_group_tags`) and the subnets it will use for discovery. In terms of the security group though, this tag should only be on one security group

infra/aws/terraform/prow-build-cluster/resources/karpenter/nodepool-default.yaml
```diff
@@ -0,0 +1,47 @@
...
+        kind: EC2NodeClass
+        name: default
+
+      requirements: # r5ad.4xlarge amd64 non-spot
+        - key: topology.kubernetes.io/zone
+          operator: In
+          values: ["us-east-2a", "us-east-2b", "us-east-2c"]
+        - key: kubernetes.io/arch
+          operator: In
+          values: ["amd64"]
```
bryantbiggs, line 19: do we need to be as restrictive on some of these? if folks want to run arm, why not let them?

likewise below for the `instance-*`, why not let Karpenter decide the best fit?

infra/aws/terraform/prow-build-cluster/vpc.tf
```diff
@@ -65,6 +65,7 @@ module "vpc" {
   public_subnet_tags = {
     "kubernetes.io/role/elb"                    = 1
     "kubernetes.io/cluster/${var.cluster_name}" = "owned"
+    "karpenter.sh/discovery"                    = var.cluster_name
```
bryantbiggs, line 68: I think we want to put this in the private subnets, no? do we have public facing instances?

infra/aws/terraform/prow-build-cluster/resources/karpenter/nodepool-default.yaml
ameukam: We could have a dedicated nodepool with arm64 instances ?

xmudrii: At the moment yes, we don't want to change the behavior now, this needs a bit more discussion with SIG Testing.

infra/aws/terraform/prow-build-cluster/vpc.tf
xmudrii: I don't think so, but I could be mistaken. We want our instances to have public IP addresses, the only way we managed to do this is by putting them in public subnets. This is required because jobs were running into rate limits (e.g. from GitHub) when behind NAT, so we had to rule it out.

infra/aws/terraform/prow-build-cluster/resources/karpenter/nodepool-default.yaml
bryantbiggs: I guess what I am saying is why block or restrict things that require users to come ask us to unblock. If a group wanted to test arm64 instances, it's seems like we should just let them do that. And on the instance types/sizes - Karpenter works better (delivers on its benefits) when it's not restricted. The workloads define their requirements and Karpenter supplies compute to satisfy those requirements

infra/aws/terraform/prow-build-cluster/vpc.tf
bryantbiggs: ah ok - yes I think I remember seeing/hearing that

infra/aws/terraform/prow-build-cluster/resources/karpenter/nodepool-default.yaml
xmudrii: I'm +1 with this, but the plan is that we try to replicate the cluster-autoscaler behavior in the very beginning, and then we start to add additional functionalities after the initial deployment.

koksay:
/lgtm
Great work @xmudrii !

xmudrii:
The Terraform changes are applied to the production:
```
Apply complete! Resources: 20 added, 4 changed, 0 destroyed.
```
/hold cancel

infra/aws/terraform/prow-build-cluster/locals.tf
xmudrii: Thanks, fixed!

infra/aws/terraform/prow-build-cluster/locals.tf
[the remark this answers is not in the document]
xmudrii: This will be applied in a follow up PR

PR #7067, opened 2024-07-25
eks-prow-build-cluster: Automatically force Flux on stable nodes

xmudrii:
This PR automates enforcing Flux GitOps Toolkit (GOTK) components onto the stable worker nodes in the EKS Prow build cluster. Until now, we had to manually change the `gotk-components.yaml` manifest to add the needed `nodeSelector` and `tolerations` properties, which is very error prone.

We initially tried to do this in #6423, however that caused a cluster-wide outage on 2024-02-21 by removing all resources managed by Flux. The reason for this incident is that the provided `kustomization.yaml` file only specified the `gotk-components.yaml` file, but there were many other files in that folder. That made Flux to only apply the `gotk-components.yaml` file, while pruning resources defined in other files from the cluster. This incident has been documented as the part of the postmortem: https://docs.google.com/document/d/1PMgbClhYwIls8NdEw2jE80OqHmnD-Ty7TermtUe4MTE/edit?usp=sharing

This PR takes a bit different approach that is much safer. Instead of adding `kustomization.yaml` to the `flux-system` folder which is often changed (new files are often added, which would require updating `kustomization.yaml`), the `gotk-components.yaml` file has been moved to a new folder (`resources/flux`) that only contains that file and the needed `kustomization.yaml`. With that, the `kustomization.yaml` file is simple, risks are very minimum, while keeping it easy to maintain Flux as long as we use it.

This PR includes some additional changes to the documentation and scripts. Mainly, the `hack/flux-update.bash` script will not update Flux by default by any longer, instead `make flux-update-with-gotk` must be used. This script is also used when adding new resources to be managed by Flux and we want to make Flux updates more coordinated (and not by accident).

Initially, this PR requires taking some manual steps to ensure that Flux is not going to get pruned one this PR is merged. We need to suspend the `flux-system` kustomization, then apply the `flux` kustomization, and then resume the `flux-system` kustomization. In other words:

```bash
flux suspend kustomization flux-system

# Merge this PR

kubectl apply -f resources/flux-system/flux-ks-flux.yaml

# Wait for the flux kustomization to get fully-reconciled

flux resume kustomization flux-system
```

This is based on the instructions from https://github.com/fluxcd/flux2/discussions/2318

This change has been already successfully deployed on the Canary cluster.

/hold

/assign @upodroid @ameukam @dims


dims:
i mainly looked at `GitOps.md` which looks good to me :) The other changes resonate as well, but i did not look very deep.

/approve
/lgtm

xmudrii:
Successfully reconciled :shipit:

PR #7068, opened 2024-07-25
eks-prow-build-cluster: Switch the cluster to the API authentication

xmudrii:
This PR switches the EKS Prow build cluster to the EKS Auth API authentication mode (currently it's EKS Auth API and the `aws-auth` ConfigMap). It also removes the `aws-auth` ConfigMap.

It's important to note that this change is irreversible, once we apply it, there's no way to go back to the combined (EKS Auth API and the `aws-auth` ConfigMap) authentication mode. This will be the only authentication option with the new major version of the EKS Terraform module.

/assign @dims @upodroid @ameukam 

/hold
We'll apply this change after a testing period (after 2024-07-29)


xmudrii:
Applied:
```
Apply complete! Resources: 1 added, 6 changed, 2 destroyed.
```
/hold cancel
@dims @ameukam needs another LGTM because of the needed fix

PR #7070, opened 2024-07-25
tweak rate limiting following log analysis

BenTheElder:
1. shorten rate limit window with same approximate rate limit rounded down to the nearest integer-per-second
2. increase temp-ban duration to maximum (1h)
3. update and clarify comments

### DIFF
--- a/infra/gcp/terraform/modules/oci-proxy/cloud-armor.tf
+++ b/infra/gcp/terraform/modules/oci-proxy/cloud-armor.tf
@@ -42,22 +42,28 @@ resource "google_compute_security_policy" "cloud-armor" {
       exceed_action  = "deny(429)"
       enforce_on_key = "IP"
       # TODO: revisit these values
-      # above this threshold we serve 429, currently ~83/sec in a 1 minute window
+      # above this threshold we serve 429, currently 83/sec in a 30s window
       rate_limit_threshold {
         # NOTE: count cannot exceed 10,000
         # https://cloud.google.com/armor/docs/rate-limiting-overview
-        count        = 5000
-        interval_sec = 60
+        # This is set to be like the ban limit, for a shorter time period
+        # 83/second for 30s versus 83.33/second for two minutes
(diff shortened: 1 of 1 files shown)


dims:
/approve
/lgtm
/hold

( for additional approvals or until you are ready to land this @BenTheElder )

BenTheElder:
/hold cancel

first we'll watch this deploy to staging, which also gives some more time for any objection raising
(also, we can roll back any production changes if there are objections later, I think this should be fine, it's a relatively small tweak)

BenTheElder:
auto-deploy to staging: https://prow.k8s.io/view/gs/kubernetes-jenkins/logs/post-k8s-deploy-registry/1816575907268136960

BenTheElder:
working in staging, will roll out to prod

BenTheElder:
done

BenTheElder:
So far we've not seen anyone hit the ban threshold (which is NOT any higher after this change) since tweaking the rate limit threshold (maybe we're getting clients to back off more now? too early to say)

We are rate limiting 8 IPs in the past hour:


| ASN | Number of IPs |
|--|--|
| AMAZON-02 | 1 |
| ENTEL CHILE S.A., CL | 1 |
| GOOGLE-CLOUD-PLATFORM | 5 |
| MICROSOFT-CORP-MSN-AS-BLOCK | 1 |

And temp-banning none since the change.

Will have to check back later, 1h is a small sample period for the registry and we previously only temp-banned a subset of clients (that's expected, the two tier limit is intended to first throttle short bursts to maintain the rate limit, and then temp ban sustained excessive usage ignoring the throttling).

BenTheElder:
In the past 12 hours We have 16 IPs hitting the temp-ban threshold and 64 hitting the rate limit threshold.

I didn't stop to track down the ASNs but at a glance it looks similar to before.

This seems like a good result, by far the majority of clients that wind up getting rate limited only hit a temporary rate limit, and still less than 20 clients actually get temp-banned for persisting past the 30s burst limit.

PR #7077, opened 2024-07-27
Switch alpine images to new source.

michelle192837:
The test-infra alpine image is now being published to a new location (k8s-staging-test-infra) in addition to the old one (k8s-prow). Use the images in the new location rather than the old one.

Ref https://github.com/kubernetes/test-infra/issues/32432


michelle192837:
Thanks!

/hold cancel

PR #7084, opened 2024-07-29
eks-prow-build-cluster: Increase the Karpenter capacity to 150 nodes

xmudrii:
We deployed Karpenter with the limit of 50 nodes, now we're increasing it to 150 nodes to match the limits that we use on GKE.

Also, the CPU limits value has been quoted as recommended by Karpenter: https://karpenter.sh/docs/concepts/nodepools/#speclimits

> CPU limits are described with a DecimalSI value. Note that the Kubernetes API will coerce this into a string, so we recommend against using integers to avoid GitOps skew.

/assign @dims @BenTheElder @koksay


dims:
/approve
/lgtm
/hold

PR #7103, opened 2024-07-30
Store credentials for Fastly service account

ameukam:
In order to define and access a private bucket as origin for the Fastly service defined for dl.k8s.io, we store the credentials required in GCP secrets.
Fastly don't support workload identity so the credentials will be accessible through [data sources](https://developer.hashicorp.com/terraform/language/data-sources) outside of this configuration.


ameukam:
/hold

cc @dims @BenTheElder

BenTheElder:
thank you!

PR #7116, opened 2024-08-01
Fastly: Fix authentication with HMAC for dl-sandbox

ameukam:
Update VCL service to pull objects from a private GCS bucket

### DIFF
--- a/infra/fastly/terraform/dl-sandbox.k8s.dev/vcl/binaries.vcl
+++ b/infra/fastly/terraform/dl-sandbox.k8s.dev/vcl/binaries.vcl
@@ -53,9 +53,8 @@ sub vcl_fetch {
 
   # TODO: Drop this when the origin(GCS bucket) is owned by the community
   # See: https://github.com/kubernetes/k8s.io/issues/2396
-  if (beresp.status == 200 && req.url.path ~ "^/release/") {
-    set beresp.http.Cache-Control = "private, no-store"; # Don't cache in the browser
-    set beresp.ttl = 30d;
+  if (beresp.status == 206 && req.url.path ~ "^/release/") {
+    set beresp.ttl = 1y;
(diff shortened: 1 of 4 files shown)


ameukam:
cc @dims @BenTheElder 

You should be able to test with:

```bash
::1     localhost ip6-localhost ip6-loopback
ff02::1 ip6-allnodes
ff02::2 ip6-allrouters
151.101.0.204 cdn.dl-sandbox.k8s.dev
2a04:4e42::204 cdn.dl-sandbox.k8s.dev
```

infra/fastly/terraform/dl-sandbox.k8s.dev/vcl/binaries.vcl
```diff
@@ -53,9 +53,8 @@ sub vcl_fetch {
 
   # TODO: Drop this when the origin(GCS bucket) is owned by the community
   # See: https://github.com/kubernetes/k8s.io/issues/2396
-  if (beresp.status == 200 && req.url.path ~ "^/release/") {
-    set beresp.http.Cache-Control = "private, no-store"; # Don't cache in the browser
-    set beresp.ttl = 30d;
+  if (beresp.status == 206 && req.url.path ~ "^/release/") {
```
BenTheElder, line 56: should we allow more status codes?

ameukam: 🤔  I don't think we want to cache other responses from the origin for the migration. With a private bucket, we only get `206` from the origin based on some investigation with Fastly Support.

BenTheElder: OK, It seems like we might want to cache 2XX (at least 200??) just in case but we can revisit later

PR #7118, opened 2024-08-02
aws/kops: Setup AWS CloudWatch Observability agent

ameukam:
Temporarily install observability agent to collect metrics.


ameukam:
I would like to collect some metrics (IOPS on local disk, memory...) to help change the instance type.

cc @hakman 

FYI @bryantbiggs

PR #7120, opened 2024-08-03
Use new location for Prow images instead of old.

michelle192837:
Prow images have been published to a new location for a while now (us-docker.pkg.dev/k8s-infra-prow/images/). These should be the same images as published to the old location, so update images to pull from the new location instead.

Ref https://github.com/kubernetes/test-infra/issues/32432

### DIFF
--- a/hack/autobump-config.yaml
+++ b/hack/autobump-config.yaml
@@ -26,12 +26,12 @@ prefixes:
     summarise: false
     consistentImages: true
   - name: "ghproxy"
-    prefix: "gcr.io/k8s-prow/ghproxy"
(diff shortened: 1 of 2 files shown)


xmudrii (approved):
/lgtm
/approve

upodroid:
I defined one here and it doesn't work.
https://github.com/kubernetes/test-infra/blob/1e6d39ad85f8a3e9e69e6fc3858eca01fdbf1023/config/prow/autobump-config/prow-job-autobump-config.yaml#L38

michelle192837:
/hold cancel

Thanks folks!

PR #7123, opened 2024-08-05
Deploy arm64 nodepool on the GKE Build Cluster

upodroid:
/cc @ameukam @BenTheElder 

According to the docs, arm64 nodepools are automatically tainted by GCP to block amd64 workloads running on there. Jobs have to either use nodeselectors or taint the pod to tolerate arm64 nodes if the image is multiarchitecture.

https://cloud.google.com/kubernetes-engine/docs/how-to/prepare-arm-workloads-for-deployment#overview

Fixes:


ameukam:
/lgtm
/hold

In case, other folx want to shime in.

upodroid:
This has been deployed. Need another lgtm

PR #7134, opened 2024-08-06
Mark some variables non-nullable, part 2

sftim:
Make some inputs non-nullable. Nullable variables can be [expensive](https://www.infoq.com/presentations/Null-References-The-Billion-Dollar-Mistake-Tony-Hoare/).

This change strengthens additional validation on values.

### DIFF
--- a/infra/aws/terraform/prow-build-cluster/variables.tf
+++ b/infra/aws/terraform/prow-build-cluster/variables.tf
@@ -16,12 +16,14 @@ limitations under the License.
 
 variable "eks_cluster_admins" {
   type        = list(string)
+  nullable    = false
   description = "List of maintainers that have administrator access to the cluster."
   default     = []
 }
 
 variable "eks_cluster_viewers" {
   type        = list(string)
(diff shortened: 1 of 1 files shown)


infra/aws/terraform/prow-build-cluster/variables.tf
```diff
@@ -71,111 +74,133 @@ variable "vpc_intra_subnet" {
...
+  nullable    = false
 }
 
 variable "node_group_version_us_east_2a" {
   type        = string
   description = "Kubernetes version of the EKS-managed node group in US East 2a"
+  nullable    = false
 }
 
 variable "node_group_version_us_east_2b" {
```
sftim, line 98: We could refactor this to a map by AZ, instead.

xmudrii: We will drop all `node_group_*` fields in the coming days.

xmudrii (approved):
/lgtm
/approve

PR #7140, opened 2024-08-08
Add kubernetes.io managed groups for all remaining SIG/WGs

mrbobbytables:
This PR creates managed kubernetes.io groups for all remaining sigs/wgs that do not currently have them. Migrating to managed groups will resolve several issues we've had with  public googlegroups (e.g. mailing list limits on calendar invites or leads losing control of mailing lists), and enable us to use a common shared drive.

To complete this, I am adding myself to each new list to complete configuration (some settings can't currently be set via our automation). Once complete, I will remove myself from all the groups.

It also cleans up the general leads list and references each individual sig/wgs leads list so it won't have to be updated in more than one place. There are a few remaining emails in there that I'm waiting for acks on, but all in all its a significant improvement.

/hold

PR #7142, opened 2024-08-08
Add `rexagod` to `leads`

rexagod:
👋 Hello, it took me a while to figure out why I wasn't receiving any invites, CFPs, etc., but I believe this patch should address that.


rexagod:
Not needed after the refactor done in https://github.com/kubernetes/k8s.io/pull/7140.

Priyankasaggu11929:
> Not needed after the refactor done in #7140.

yea, that should address the mailing lists permissions :)

PR #7075, opened 2024-07-26
Remove resources for k8s-infra-prow (k8s-infra-prow.k8s.io)

ameukam:
Remove the resources required to run https://k8s-infra-prow.k8s.io.

PR #7145, opened 2024-08-08
gcsweb: Drop deleted bucket

ameukam:
Stop serve deleted bucket in
https://github.com/kubernetes/k8s.io/pull/7075.


ameukam:
/hold
cc @dims @BenTheElder

PR #7168, opened 2024-08-14
re-scope azure storage blob contributor role assignment to subscription

nojnhuh:
This change increases the scope of the "Storage Blob Data Contributor" role assignment to the `az-cli-prow` Service Principal from only the `k8sprowstoragecomm` Storage Account to the entire subscription in order to also support the `--auth-mode login` flag for `az storage` commands against the ephemeral `capz-wi-*` Storage Accounts created by CAPZ for Workload Identity.

### DIFF
--- a/infra/azure/terraform/capz/main.tf
+++ b/infra/azure/terraform/capz/main.tf
@@ -118,7 +118,6 @@ module "role_assignments" {
   source                   = "./role-assignments"
   resource_group_name      = var.resource_group_name
   container_registry_scope = module.container_registry.container_registry_id
-  storage_account_scope    = azurerm_storage_account.k8sprowstorage.id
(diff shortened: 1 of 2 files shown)


ritikaguptams:
@nojnhuh Could you also remove this the storage_account_scope variable from being passed down from the main? https://github.com/kubernetes/k8s.io/blob/69da28271896d35365b225ef331087e07c240616/infra/azure/terraform/capz/main.tf#L121

nojnhuh:
> @nojnhuh Could you also remove this the storage_account_scope variable from being passed down from the main?
> 
> https://github.com/kubernetes/k8s.io/blob/69da28271896d35365b225ef331087e07c240616/infra/azure/terraform/capz/main.tf#L121

Done, thanks!

PR #7198, opened 2024-08-20
Fastly: Ensure specific file types are not cached.

ameukam:
Ensure HTML and JSON are not cached at the edge and are directly
served.
We are also specific about the sub-paths for the different version
markers.
This will required to purge some objects from the cache.

### DIFF
--- a/infra/fastly/terraform/dl-sandbox.k8s.dev/vcl/binaries.vcl
+++ b/infra/fastly/terraform/dl-sandbox.k8s.dev/vcl/binaries.vcl
@@ -44,8 +44,15 @@ sub vcl_fetch {
     set beresp.stale_while_revalidate = 60s; // 1 minute
   }
 
-  //Ensure version markers are not cached
-  if (req.url.ext == "txt") {
+  # Ensure version markers are not cached at the edge
+  if (req.url.path ~ "^/release/(latest|stable)(-\d+(\.\d+))?\.txt\z") {
+    set beresp.cacheable = false;
+    set beresp.ttl = 0s;
+    return (pass);
+  }
+
+  # Ensure HTML and JSON files are not cached at the edge
+  if (req.url.ext ~ "(html|json)\z") {
     set beresp.cacheable = false;
     set beresp.ttl = 0s;
     return (pass);
--- a/infra/fastly/terraform/dl.k8s.io/vcl/binaries.vcl
+++ b/infra/fastly/terraform/dl.k8s.io/vcl/binaries.vcl
@@ -44,8 +44,15 @@ sub vcl_fetch {
     set beresp.stale_while_revalidate = 60s; # 1 minute
   }
 
-  # Ensure version markers are not cached
-  if (req.url.ext == "txt") {
+  # Ensure version markers are not cached at the edge
+  if (req.url.path ~ "^/release/(latest|stable)(-\d+(\.\d+))?\.txt\z") {
+    set beresp.cacheable = false;
+    set beresp.ttl = 0s;
+    return (pass);
+  }
+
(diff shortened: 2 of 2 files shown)


ameukam:
cc @upodroid @BenTheElder

infra/fastly/terraform/dl-sandbox.k8s.dev/vcl/binaries.vcl
```diff
@@ -44,8 +44,8 @@ sub vcl_fetch {
     set beresp.stale_while_revalidate = 60s; // 1 minute
   }
 
-  //Ensure version markers are not cached
-  if (req.url.ext == "txt") {
+  # Ensure HTML, JSON and TXT files are not cached at the edge
+  if (req.url.ext ~ "(html|json|txt)$") {
```
BenTheElder, line 48: ummm are we mutating these??

BenTheElder: are there some specific paths we should exclude instead?

ameukam: The version markers (latest|stable) are mutated.

exclude specific paths or subpaths for those kind of files is not worth it IMHO. They are too small to cache at the edge.

BenTheElder: Well, today, but we may trip over this assumption in the future ... someday we will add a huge json object to releases

(e.g. OCI image individual blobs)

I would strongly prefer to except the marker paths if we can instead.
We can add more exceptions if we need, but we don't want to over-except and miss something important.

ameukam: Yeah. We may encounter it in the future but it's unlikely to happen in the short term (EOY ?).

BenTheElder: The problem is who else will even know and expect this, and then later we figure out the issue by way of the budget exploding.

BenTheElder: Versus it will be very clear if we're caching something we didn't want cached, and that's a quick fix.

ameukam: I tried to be specific about the url paths for the version markers. PTAL.

infra/fastly/terraform/dl-sandbox.k8s.dev/vcl/binaries.vcl
```diff
@@ -44,8 +44,15 @@ sub vcl_fetch {
...
-  if (req.url.ext == "txt") {
+  # Ensure version markers are not cached at the edge
+  if (req.url.path ~ "^/release/(latest|stable)(-\d+(\.\d+))?\.txt\z") {
+    set beresp.cacheable = false;
+    set beresp.ttl = 0s;
+    return (pass);
+  }
+
+  # Ensure HTML and JSON files are not cached at the edge
+  if (req.url.ext ~ "(html|json)\z") {
```
BenTheElder, line 55: FYI @kubernetes/release-engineering, something to keep in mind, if you see artifacts that should or shouldn't be cached (e.g. if releases start including html/json files) we need to refine these rules and deploy them (not automated yet)

BenTheElder (commented):
/lgtm
/approve

ameukam:
This is deployed.

ameukam:
`cdn.dl.k8s.io/release/release-notes-index.json` is purged from the cache.

PR #7211, opened 2024-08-22
Enable autosync for prow control plane

upodroid:
The migration is complete and we want enable auto sync on for prow


xmudrii (approved):
/lgtm
/approve
/hold
unhold when ready

PR #7224, opened 2024-08-27
publish scheduler-plugins v0.29.8

Huang-Wei:


Huang-Wei:
@ingvagabund could you /lgtm? Once it's merged, `registry.k8s.io/scheduler-plugins/kube-scheduler:v0.29.8` should be live and /bin/kube-scheduler is expected to be the default entrypoint.

PR #7225, opened 2024-08-27
Promote the first half of CSI sidecar releases

jsafrane:
Promote CSI sidecars that were already tagged and released.


registry.k8s.io/images/k8s-staging-sig-storage/images.yaml
```diff
@@ -46,6 +47,7 @@
     "sha256:ee4458b1b0b9612de81f2607b4214c6838fa333b133c845da0995fc40fb1936e": [ "v0.9.0" ]
 - name: csi-node-driver-registrar
   dmap:
+    "sha256:f80c426eda32c648c7d7698511a186b8b11e029b2d62c5f2aa0989e35df1cb92": [ "v1.12.0" ]
```
jsafrane, line 50: This was an accident, I created tag v1.12.0 instead of v2.12.0 and the robots were too quick

jsafrane:
cc @msau42 @xing-yang

PR #7231, opened 2024-08-28
Start autobumping Community Prow

upodroid:
1. refConfigFile isn't required, have a look at https://github.com/kubernetes-sigs/prow/blob/main/cmd/generic-autobumper/main.go#L158
2. We want to bump any images for prow/boskos in the kubernetes folder
3. I cleaned up some directories that don't hold prow deployments

/cc @BenTheElder @ameukam


xmudrii (approved):
/lgtm
/approve

PR #7218, opened 2024-08-24
Manage the gke build cluster via argocd - part one

upodroid:
This PR configures argocd to manage the test-pods namespace of the gke build cluster except the secrets. We need to migrate the secrets to ESO from KES and resolve https://github.com/kubernetes/k8s.io/issues/3123 in a separate PR.

PR #7233, opened 2024-08-29
fix boskos deployment typos

upodroid:
Fixing typos from #7218 
/cc @dims


dims:
not just typos :) looks like we are moving the crd to another file intentionally right?

/approve
/lgtm
/hold

kubernetes/gke-prow-build/prow/crds/boskos.yaml
```diff
@@ -0,0 +1,139 @@
+apiVersion: apiextensions.k8s.io/v1
```
upodroid, line 1: ArgoCD is complaining about empty crds folder. I thought it might ignore an empty folder.
/hold cancel

upodroid: The prow CRD is too big to be applied without server-side apply. context: https://github.com/kubernetes/k8s.io/pull/7141/commits/4d3858b0ecf494c32956461dc69cbe5d48be9c5f

PR #7241, opened 2024-08-30
fix disk size for arm nodepools

upodroid:
https://kubernetes.slack.com/archives/CCK68P2Q2/p1724975407113939

etcd robustness CI jobs were failing, logs indicated `The node was low on resource: ephemeral-storage` was the cause, so adding higher local 500GB ssd boot disk for the arm nodepools in more regions as well.


dims:
/approve
/lgtm

i added a couple of sentences of context, please check if that looks ok @upodroid

upodroid:
Description looks good, thanks

PR #7260, opened 2024-09-04
[WIP] Fastly: Drop cookies and http headers for dl-sandbox

ameukam:
Drop all cookies and only all specific HTTP headers for any client request.

### DIFF
--- a/infra/fastly/terraform/dl-sandbox.k8s.dev/vcl/binaries.vcl
+++ b/infra/fastly/terraform/dl-sandbox.k8s.dev/vcl/binaries.vcl
@@ -18,6 +18,13 @@ sub vcl_recv {
   if (req.method != "HEAD" && req.method != "GET" && req.method != "FASTLYPURGE") {
     return(pass);
   }
+
+  # Drop all the headers except the list providers
+  header.filter_except(req,"Fastly-Debug");
+
+  # Remove all cookies from the request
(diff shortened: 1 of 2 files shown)


ameukam:
cc @BenTheElder 

I need figure out how to filter/drop the parameters passed by the client.

infra/fastly/terraform/dl-sandbox.k8s.dev/vcl/binaries.vcl
```diff
@@ -18,6 +18,13 @@ sub vcl_recv {
   if (req.method != "HEAD" && req.method != "GET" && req.method != "FASTLYPURGE") {
     return(pass);
   }
+
+  # Drop all the headers except the list providers
+  header.filter_except(req, "Accept", "User-Agent", "Accept-Encoding");
```
BenTheElder, line 23: I don't think we actually want to send these to the backend, I think we want to drop all of them?

It would be nice if the CDN side respected gzip request but it shouldn't care about User-Agent, and we shouldn't need to pass any of it to GCS.

ameukam:
@BenTheElder I did some experiment and this is somehow breaks caching. I need to investigate deeper but I need to first ensure we have logging for the VCL service which is currently not set up.

ameukam:
Will revisit this later.
/close

PR #7271, opened 2024-09-07
Define owners for the prow configuration.

ameukam:


ameukam:
cc @dims @upodroid @BenTheElder 
FYI @michelle192837

PR #7206, opened 2024-08-21
Use the new prow endpoint

upodroid:

PR #7302, opened 2024-09-16
delete nginx ingress and add some dns records for services on the utility cluster

upodroid:
1. nginx ingress is too buggy to deploy something like this that I need to expose ArgoCD https://mykubert.com/blog/ollama-kubernetes-deployment-cost-effective-and-secure/#Security_Setup_with_OAuth2_Proxy_and_Keycloak
2. One of the reasons why we need to use a custom ingress instead of gce ingress is to support multi cluster observability for Thanos https://wrossmann.medium.com/putting-thanos-grpc-endpoints-behind-a-k8s-ingress-a5e21c7d0d85, as well as the complex proxy for argo.

I'm also adding DNS records for services in the utility cluster.

/cc @ameukam @BenTheElder

### DIFF
--- a/dns/zone-configs/k8s.io._0_base.yaml
+++ b/dns/zone-configs/k8s.io._0_base.yaml
@@ -255,6 +255,8 @@ pr-test:
 cli.triage:
   type: A
   value: 34.117.106.163
+
+### prow.k8s.io related records, running in the k8s-infra-prow GCP project
 prow:
   - type: A
     ttl: 600 # this has needed to change in the past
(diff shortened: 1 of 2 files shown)


dns/zone-configs/k8s.io._0_base.yaml
```diff
@@ -269,9 +271,28 @@ hooks.prow:
   - type: AAAA
     ttl: 600 # this has needed to change in the past
     value: "2600:1901:0:b465::"
+# prow-certificates in k8s-infra-prow project
+_acme-challenge.prow:
```
upodroid, line 275: Deleted unintentionally in https://github.com/kubernetes/k8s.io/pull/7206

PR #7330, opened 2024-09-23
Formatting Terraform code. No code changes.(Minor)

vishnuswmech:


vishnuswmech:
If possible, we can have a Terraform Format checks on PR stage itself. It is a standard practice.

vishnuswmech:
@BenTheElder Thanks for pointing out.

While we haven't faced issues before, consistent formatting enhances better readability, which makes the code easier to review and understand.

And it will be more helpful for the new contributors to understand the code quickly.

Kindly share your thoughts on this.

ameukam:
better readability is the benefit of this PR and it doesn't affect the state of the resources. IMHO, we should start with some non-blocking HCL checks.

/lgtm
/approve

PR #7342, opened 2024-09-25
gcp/redirect: Bump nginx tag

ameukam:
Changelog: https://nginx.org/en/CHANGES-1.26

### DIFF
--- a/apps/k8s-io/deployment.yaml
+++ b/apps/k8s-io/deployment.yaml
@@ -34,7 +34,7 @@ spec:
           # Map all keys to files.
       containers:
       - name: nginx
-        image: nginx:1.25-alpine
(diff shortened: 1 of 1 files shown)


ameukam:
cc @BenTheElder @dims @upodroid

apps/k8s-io/deployment.yaml
```diff
@@ -34,7 +34,7 @@ spec:
           # Map all keys to files.
       containers:
       - name: nginx
-        image: nginx:1.25-alpine
+        image: nginx:1.26-alpine
```
BenTheElder, line 37: can we please at least pin the digest?

this is a bit terrifying in the critical path to e.g. binary downloads

ameukam: Updated. PTAL.

BenTheElder:
/lgtm
/approve
thanks!

PR #7344, opened 2024-09-26
[eks-prow-build] Upgrade karpenter to v0.37.3

koksay:
This PR upgrades Karpenter to latest minor version before the v1.0 upgrade as suggested on the [migration doc](https://karpenter.sh/docs/upgrading/v1-migration/)

Following steps with a separate PR:

- Upgrade `karpenter` and `karpenter-crd` helmreleases to v1.0.2
- At this point, there will be reconciliation errors due to some changes on the IAM side. We need to add `enable_v1_permissions = true` to `karpenter.tf` and run terraform.


xmudrii (approved):
/lgtm
/approve

PR #7351, opened 2024-09-29
publish cloud-provider-kind images

aojea:
```
crane digest gcr.io/k8s-staging-kind/cloud-provider-kind:v20240929-43cbeb6
sha256:cacb2da53a0d3b363bf53f11c9e0f714d7605471130a43ae7d1e161da0c3cbdb
```

Ref: https://github.com/kubernetes-sigs/cloud-provider-kind/issues/115

/assign @BenTheElder

### DIFF
--- a/registry.k8s.io/images/k8s-staging-kind/OWNERS
+++ b/registry.k8s.io/images/k8s-staging-kind/OWNERS
@@ -1,0 +1,6 @@
+# See the OWNERS file documentation:
+# https://github.com/kubernetes/kubernetes/blob/master/docs/devel/owners.md
+
+approvers:
+- aojea
+- bentheelder
--- a/registry.k8s.io/images/k8s-staging-kind/images.yaml
+++ b/registry.k8s.io/images/k8s-staging-kind/images.yaml
@@ -1,0 +1,3 @@
+- name: cloud-provider-kind
+  dmap:
+    "sha256:cacb2da53a0d3b363bf53f11c9e0f714d7605471130a43ae7d1e161da0c3cbdb": ["v0.4.0"]


aojea:
@ameukam what is the public url of this image?

registry.k8s.io/images/k8s-staging-kind/OWNERS
```diff
@@ -0,0 +1,6 @@
+# See the OWNERS file documentation:
+# https://github.com/kubernetes/kubernetes/blob/master/docs/devel/owners.md
```
BenTheElder, line 2: cloud provider kind is a distinct subproject and should be published via it's own staging?

BenTheElder: too late now I guess, but we should be careful about this part in the future, with immutable images we're stuck with naming and we have arbitrary nesting so we could do distinct paths for each.

BenTheElder: ... not too late

BenTheElder:
> @ameukam what is the public url of this image?

nowhere. k8s-staging-kind isn't setup to publish yet, there's two sets of configs.

let's fix that by creating a k8s-staging-cloud-provider-kind and stick these there.

BenTheElder:
ideally they should wind up `registry.k8s.io/cloud-provider-kind/..` and we'll leave the `registry.k8s.io/kind` namespace for the main project for now

also let's make sure the staging GCB project etc is isolated. we should not encourage subprojects to reuse these. adding more projects is cheap and increases isolation.

https://github.com/kubernetes/k8s.io/tree/main/registry.k8s.io#creating-image-promoter-manifests

aojea:
> ideally they should wind up `registry.k8s.io/cloud-provider-kind/..` and we'll leave the `registry.k8s.io/kind` namespace for the main project for now

IIRC we decided to use kind because cloud-provider-kind was too long and failed

BenTheElder:
> IIRC we decided to use kind because cloud-provider-kind was too long and failed

🤨 

Too long where?

aojea:
> > IIRC we decided to use kind because cloud-provider-kind was too long and failed
> 
> 🤨
> 
> Too long where?

the size of the name bucket :)

PR #7353, opened 2024-09-30
Set mod level to none for leads lists

mrbobbytables:
Many leads are missing notifications (including cal invites) because their own lists are moderated and they miss the mod notification. This sets the general mod level to none. If it becomes an issue we can revert.

(want to do this before we send out the updated cal invite for the monthly chair/tl meeting)


ameukam:
/lgtm
/approve
/hold

Remove the hold when ready.

mrbobbytables:
Thanks @ameukam !
/hold cancel

PR #7355, opened 2024-09-30
chore: CAPA group membership changes

richardcase:
Updates to the Google groups for recenet changes to the CAPA maintainers.


groups/sig-cluster-lifecycle/groups.yaml
```diff
@@ -1,5 +1,4 @@
 groups:
-
```
fabriziopandini, line 2: I know this should not make a difference, but from a quick look it seems that all groups files have those empty lines, can we leave them?

richardcase: Ahhhh, format on save 😢  Yes no problem, sorry about that.

richardcase: Fixed the erroneous formatting.

vincepri (approved):

PR #7358, opened 2024-10-01
gcp/bash: drop GCR repository creation

ameukam:
Since May 15, new GCR repos are hosted on Artifact Registry and GCR repos can't no longer be created.
This is to ensure to can handle special cases for existing GCR repos.


ameukam:

upodroid (approved):
LGTM
/hold cancel when you are ready to merge

PR #7433, opened 2024-10-19
azure/bash: Fix location variable

ameukam:
Fix incorrect variable and drop incorrect flag for the creation of the storage containers.


ameukam:
cc @hakman

PR #7461, opened 2024-10-28
Add terraform for Azure community image gallery

mboersma:
Adds terraform needed to recreate the basic Azure community gallery resources used to host Cluster API Provider for Azure reference images.

The managed images, SIG image versions, SIG image definitions, and staging gallery are all transient and fungible resources that will be recreated if needed by the publishing pipeline, so they do not have to be handled by terraform.

See also:
- kubernetes-sigs/image-builder#1578
- kubernetes-sigs/cluster-api-provider-azure#5167

### DIFF
--- a/infra/azure/terraform/capz/cluster-api-gallery/main.tf
+++ b/infra/azure/terraform/capz/cluster-api-gallery/main.tf
@@ -1,0 +1,70 @@
+/*
+Copyright 2024 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+variable "resource_group_name" {
+  type = string
+}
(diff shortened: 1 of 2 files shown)


mboersma:
/cc @jsturtevant @nojnhuh @ritikaguptams

I'm not entirely sure what I'm doing here, so I used the `aztfexport` tool against the existing resource group and massaged the output into something that looks like our existing terraform structure. Please let me know what should change.

infra/azure/terraform/capz/cluster-api-gallery/main.tf
```diff
@@ -0,0 +1,70 @@
...
+
+# Create the shared image gallery with community permissions
+resource "azurerm_shared_image_gallery" "community_gallery" {
+  description         = "Shared image gallery for Cluster API Provider Azure"
+  location            = var.location
+  name                = "community_gallery"
+  resource_group_name = "cluster-api-gallery"
+  tags = {
+    creationTimestamp = "2024-10-24T00:00:00Z"
+    jobName           = "image-builder-sig-ubuntu-2404"
```
nojnhuh, line 43: Is this tag meaningful? Seems odd to have one for a specific image build.

nojnhuh (approved):
/lgtm

infra/azure/terraform/capz/cluster-api-gallery/main.tf
mboersma: No, it really isn't informative. I think we just propagate the same `${TAGS}` to several resources, and that job is the one that lazily created the gallery. I'll delete it after I wait to see if there's more feedback on this PR.

PR #7472, opened 2024-10-30
Azure: Setup Prow build cluster

ameukam:
Setup an AKS cluster where we can run ProwJobs similarly to the GKE
cluster `k8s-infra-prow-build`.

Signed-off-by: Arnaud Meukam <ameukam@gmail.com>

PR #7499, opened 2024-11-07
Azure: update tag for resource group post cluster creation

ameukam:
Follow-up of:
  - https://github.com/kubernetes/k8s.io/pull/7472

https://github.com/Azure/rg-cleanup/ is deployed at
the subscription level to delete resource groups older than 3 days to
clean resources left post prowjobs execution. This is also affect the resource
groups auto-created by AKS clusters.
`rg-cleanup` will not clean up resource groups with specific tags
(e.g.`DO-NOT-DELETE`)
Tag is added to ensure `rg-cleanup` do remove the resource groups of the
build cluster.

### DIFF
--- a/infra/azure/terraform/k8s-infra-prow-build/aks.tf
+++ b/infra/azure/terraform/k8s-infra-prow-build/aks.tf
@@ -121,3 +121,13 @@ module "prow_build" {
 
   depends_on = [module.prow_network]
 }
+
+# Prevent resource group deletion
+resource "null_resource" "prow_nodepool_rg_tag" {
+
+  provisioner "local-exec" {
+    command = "az group update --resource-group ${module.prow_build.node_resource_group} --tags DO-NOT-DELETE=true"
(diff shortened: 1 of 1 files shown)


ameukam:
cc @dims @BenTheElder @upodroid

ameukam:
> Won't this conflict with [`main`/infra/azure/terraform/k8s-infra-prow-build/main.tf#L38-L45](https://github.com/kubernetes/k8s.io/blob/main/infra/azure/terraform/k8s-infra-prow-build/main.tf#L38-L45) ?
> 
> Should we just add this tag to the common tags, or a union of common tags plus this one?

> Should we just add this tag to the common tags

the resource group is automatically created by the AKS API and I couldn't find a way to propagate this tag using the terraform provider (and currently the az cli don't allow this). I'm currently doing this as a hack until the AKS team figure out how to handle exception in their cleanup.

Issue #7447, opened 2024-10-23
stable-1.txt & latest-1.txt are cached at the CDN edge


ameukam:
https://cdn.dl.k8s.io/release/stable-1.txt value is currently cached by our CDN provider.

```bash
http -F https://cdn.dl.k8s.io/release/stable-1.txt "fastly-debug:1"
HTTP/1.1 200 OK
Accept-Ranges: bytes
Age: 78
Cache-Control: private, no-store
Connection: keep-alive
Content-Length: 7
Content-Type: text/plain
Date: Wed, 23 Oct 2024 11:57:27 GMT
ETag: "070f49ef6152e49d62961ceb15a53df8"
Fastly-Debug-Digest: 7c429746fb142e022f052417aaaf72beea474417a040ce31e5f8d72ee6f11a8e
Fastly-Debug-Path: (D cache-par-lfpg1960051-PAR 1729684647) (D cache-chi-kigq8000048-CHI 1729684641)
Fastly-Debug-TTL: (H cache-par-lfpg1960051-PAR - - 78) (H cache-chi-kigq8000048-CHI 31535927.882 31536000.000 72)
Last-Modified: Wed, 23 Oct 2024 04:42:28 GMT
Server: UploadServer
Surrogate-Key: 081F2D302F85E1D217FE19018C9D3C7DE8D71E6F28C31055FD622432FF711E05
Via: 1.1 varnish, 1.1 varnish
X-Cache: HIT, HIT
X-Cache-Hits: 68, 1
X-Served-By: cache-chi-kigq8000048-CHI, cache-par-lfpg1960051-PAR
access-control-allow-origin: *

v1.31.2
```

TODO: provide more details

ameukam:
Fixed and validated by the patch releases of Nov '24.

/close

PR #7518, opened 2024-11-14
Fastly: fix the cache rule for the version markers.

ameukam:
Related:
  - https://github.com/kubernetes/k8s.io/issues/7447

We do not capture the latest-1.txt and stable-1.txt due to the regex
rule define for all the version markers.

### DIFF
--- a/infra/fastly/terraform/dl-sandbox.k8s.dev/vcl/binaries.vcl
+++ b/infra/fastly/terraform/dl-sandbox.k8s.dev/vcl/binaries.vcl
@@ -45,7 +45,7 @@ sub vcl_fetch {
   }
 
   # Ensure version markers are not cached at the edge
-  if (req.url.path ~ "^/release/(latest|stable)(-\d+(\.\d+))?\.txt\z") {
+  if (req.url.path ~ "^/release/(latest|stable)([^/]*)\.txt\z") {
     set beresp.cacheable = false;
     set beresp.ttl = 0s;
     return (pass);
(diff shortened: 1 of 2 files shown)


ameukam:
/hold

Not exactly sure about this regex rule.

how to test it:
```bash
echo "151.101.0.204 cdn.dl-sandbox.k8s.dev" > /etc/hosts
curl -vvv -H "fastly-debug:1" http://cdn.dl-sandbox.k8s.dev/release/latest-1.txt
```

A `MISS` is expected for this version marker

cc @BenTheElder

infra/fastly/terraform/dl-sandbox.k8s.dev/vcl/binaries.vcl
```diff
@@ -45,7 +45,7 @@ sub vcl_fetch {
   }
 
   # Ensure version markers are not cached at the edge
-  if (req.url.path ~ "^/release/(latest|stable)(-\d+(\.\d+))?\.txt\z") {
+  if (req.url.path ~ "^/release/(latest|stable)(-\d+(\.\d+)?)?\.txt\z") {
```
BenTheElder, line 48: Why don't we do:
```suggestion
  if (req.url.path ~ "^/release/(latest|stable)([^/]*)\.txt\z") {
```

So any top level release file that starts with latest or stable, ends with .txt

ameukam: @BenTheElder SGTM. PTAL.

BenTheElder (commented):
/lgtm
/approve
:shipit:

ameukam:
Deployed

PR #7523, opened 2024-11-14
Add a redirect for bluesky community list

dims:
I have created a bluesky list with members of all our github orgs:
https://bsky.app/profile/did:plc:kfztyuziv2i44b5kpecth77y/lists/3lau2wjkn3g2s

Currently the list is under my bluesky id, I have a script that updates the bluesky list every 6 hours. It picks up the github id(s) listed in various org.yaml(s), searches GH profiles to see if there is a bluesky URL and then adds them auto-magically to that list above. the code is here:
https://github.com/dims/gh-bsky-sync/

Until we find a good spot for this list and being able to maintain this list, we need a redirect we can give to folks, So the URL remains the same even when we move the list to a better home.

proposing `https://go.k8s.io/bsky` as the redirect for the same!

### DIFF
--- a/apps/k8s-io/configmap-nginx.yaml
+++ b/apps/k8s-io/configmap-nginx.yaml
@@ -291,6 +291,7 @@ data:
         location / {
           rewrite ^/api-review$      https://github.com/kubernetes/community/blob/master/sig-architecture/api-review-process.md redirect;
           rewrite ^/bot-commands$    https://prow.k8s.io/command-help redirect;
+          rewrite ^/bsky$            https://bsky.app/profile/did:plc:kfztyuziv2i44b5kpecth77y/lists/3lau2wjkn3g2s redirect;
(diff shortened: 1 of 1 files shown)


sftim:
We can have https://k8s.io/bsky too if we want

mfahlandt:
In hindsight of the whole .io story, i would rather move it towards bsky.kubernetes.dev or smth like this

dims:
@mfahlandt there are 20 of those redirects if you see the yaml file all starting with `go.k8s.io/xyz`, this would be just another one. it would be good to deal with all of them in the same fashion in another PR if/when needed. the `.io` changes are years away and whatever we do with others we would do with this one too.

if that's not acceptable, i am happy to drop this PR. i really don't mean to deal with deal with the larger picture in this one.

sftim:
LGTM

Don't take https://github.com/kubernetes/k8s.io/pull/7523#issuecomment-2477507414 as blocking a merge.

PR #7532, opened 2024-11-18
updating release signal shadows

drewhagen:
This seems to have been lost in the scramble somewhere

@gracenng @fsmunoz @katcosgrove


fsmunoz:
@drewhagen seems that it now requires a rebase first

PR #7536, opened 2024-11-18
Add self to sig-instrumentation leads

richabanker:


ameukam:
@richabanker Do you mind update the SIG leads list in this PR https://github.com/kubernetes/k8s.io/blob/main/groups/sig-instrumentation/groups.yaml#L22-L30 ?

richabanker:
> @richabanker Do you mind update the SIG leads list in this PR https://github.com/kubernetes/k8s.io/blob/main/groups/sig-instrumentation/groups.yaml#L22-L30 ?

Done, thanks for pointing that out!

PR #7538, opened 2024-11-19
Update OWNERS files

ameukam:
Update OWNERS to use aliases and drop fellow contributors.

### DIFF
--- a/apps/OWNERS
+++ b/apps/OWNERS
@@ -1,10 +1,7 @@
 # See the OWNERS docs at https://go.k8s.io/owners
 
 approvers:
-- ameukam
-- dims
-- spiffxp
-- thockin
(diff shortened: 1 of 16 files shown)


ameukam:
cc @dims

dims:
/approve
/lgtm

/hold

ameukam:
FYI @BenTheElder @upodroid

BenTheElder:
I will again note that the problem with this is it loses the history of which packages you actually reviewed and approved, makes emeritus impossible, and for a repo like this you'd just add the leads to the root owners anyhow

BenTheElder:
(But maybe also explicitly list some or all of them in the subdirs they're actively reviewing with root owners as a fallback, we want to assign leaf reviewers / approvers)

PR #7543, opened 2024-11-20
add atlantis DNS record

upodroid:
Required for #7520

### DIFF
--- a/dns/zone-configs/k8s.io._0_base.yaml
+++ b/dns/zone-configs/k8s.io._0_base.yaml
@@ -274,6 +274,11 @@ oauth2-proxy:
     value: 34.66.218.218
   - type: AAAA
     value: "2600:1900:4000:627b:8000::"
+atlantis:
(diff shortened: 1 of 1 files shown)


dns/zone-configs/k8s.io._0_base.yaml
```diff
@@ -274,6 +274,11 @@ oauth2-proxy:
     value: 34.66.218.218
   - type: AAAA
     value: "2600:1900:4000:627b:8000::"
+atlantis:
+  - type: A
```
BenTheElder, line 278: This is in cluster AAA?

BenTheElder: Ack- Sorry I meant to come back and wound up on the kubebuilder thread through 😴 (not the one in the channel, direct message with me and Fede)

upodroid:
Thanks

/hold cancel

PR #7550, opened 2024-11-21
GCP: Enable Artifact Registry for the Boskos projects

ameukam:
Ensure the service is enabled for all the Boskos projects.
It's currently not consistent across the entire list.

Also auto-linting of the bash script.

### DIFF
--- a/infra/gcp/bash/prow/ensure-e2e-projects.sh
+++ b/infra/gcp/bash/prow/ensure-e2e-projects.sh
@@ -25,11 +25,11 @@ SCRIPT_DIR=$(dirname "${BASH_SOURCE[0]}")
 . "${SCRIPT_DIR}/../lib.sh"
 
 function usage() {
-    echo "usage: $0 [repo...]" > /dev/stderr
-    echo "example:" > /dev/stderr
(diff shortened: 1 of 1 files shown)


ameukam:
/hold

BenTheElder:
uhhhh maybe we shouldn't do that?

BenTheElder:
Do we have retention policies for these? What are we pushing to boskos projects? Which users are doing this?

I could've sworn we only used GCS tarballs for kube-up and kops?

ameukam:
> Do we have retention policies for these?

We probably want to set this on AR repos if e2e tests continue to produce those images.

ameukam:
I think we at least need a container registry for `cloud-controller-manager`. No exactly sure which tests need this

ameukam:
@BenTheElder Friendly ping. :)

BenTheElder:
OK, so we'll need to patch the tests to push to AR instead of GCR next, and then we need to see if we can make boskos clean these or otherwise enable a retention policy in every project.

BenTheElder (commented):
/lgtm
/approve

PR #7572, opened 2024-11-27
Prow: Add an AKS cluster

ameukam:
Follow-up of:
  - https://github.com/kubernetes/k8s.io/pull/7472

Include the AKS cluster as a Prow build cluster.
It's possible we may need to recreate this cluster.


ameukam:
cc @dims @BenTheElder

PR #7606, opened 2024-12-12
Add Kubermatic to k8s-infra-staging-apisnoop project

koksay:
Access to `k8s-infra-staging-apisnoop` project to work on apisnoop and conformance tests.

Fyi, @xmudrii @mfahlandt @jeefy 

/cc @BobyMCbobs

### DIFF
--- a/groups/sig-architecture/groups.yaml
+++ b/groups/sig-architecture/groups.yaml
@@ -68,6 +68,9 @@ groups:
       - hh@ii.coop
       - jbelamaric@google.com
       - zz@ii.coop
+      - koray@kubermatic.com
+      - mario@kubermatic.com
+      - marko@kubermatic.com
 
   #
   # k8s-infra gcs write access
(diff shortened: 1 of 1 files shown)


BobyMCbobs:
Thank you @koksay, these are the perms we need.

Seeing how I can help below
/lgtm
/approve

groups/sig-architecture/groups.yaml
```diff
@@ -68,6 +68,9 @@ groups:
       - hh@ii.coop
       - jbelamaric@google.com
       - zz@ii.coop
+      - koray@kubermatic.com
+      - mario@kubermatic.com
+      - marko@kubermatic.com
```
xmudrii, line 73: Can you please use `mudrinic.mare@gmail.com` as I use it for everything else in the upstream projects?

koksay: right, maybe I should use my gmail one as well for the same reason. @mfahlandt do you have any preference?

PR #7621, opened 2024-12-17
images: Bump the k8s-infra image components

ameukam:


ameukam:
cc @xmudrii @upodroid @dims

PR #7627, opened 2024-12-18
Promote kOps 1.31.0-beta.1 binary artifacts

justinsb:


ameukam:
🤔 may we should also promote the cni plugins ?

justinsb:
In this case though, @hakman already did this - thank you 👍 

/close

PR #7661, opened 2025-01-07
remove Jason Detiberus from capi ilsts

cprivitere:
Remove Jason Detiberus's emails from the various CAPI related email lists here. 

Jason retired from CAPI a couple years ago. Also, the @equinix.com address is no longer valid.

### DIFF
--- a/groups/sig-cluster-lifecycle/groups.yaml
+++ b/groups/sig-cluster-lifecycle/groups.yaml
@@ -59,8 +59,6 @@ groups:
       ReconcileMembers: "true"
     members:
       - naadir@randomvariable.co.uk
-      - detiber@gmail.com
-      - jdetiberus@equinix.com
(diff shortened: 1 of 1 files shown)


fabriziopandini:
cc @detiber 

No strong opinions from me, but we usually let folks to self remove from those lists (might be they still want to receive those emails)

blind3dd:
Perhaps it's worth defining a process that would automatically create those kind of PRs after coming up with dedicated logic, maybe emailing that person first to make sure they're still interested as they may come back at certain point, and if there is no answer for more than xyz, than perhaps this kind of PRs should be automatically open for review.

cprivitere (commented):
Yeah, not trying to kick @detiber off any lists, was just tidying up while I was in these files. Happy to cancel the PR.

PR #7669, opened 2025-01-12
Ingress NGINX: Promote Test Runner v20250112-a188f4eb/v20250112-01b7af21.

Gacko:
/cc @strongjz @tao12345666333 @cpanato

I'm also not super proud of this. Basically the reason for promoting two images with similar names and only the hash being different is as follows:

The tag we set in `images/test-runner/TAG` in the Ingress NGINX project is not being used for the image tag. Instead we're generating one. This wouldn't be an issue if we would just use the same Test Runner image built once across all branches. Then we would just trigger a build on `main` and use the same result everywhere.

Sadly reality is different: Currently there's an NGINX base image v2.0.0 in use on `main` since we bumped NGINX to v1.27.1 there. On `release-1.12` it's still NGINX v1.25.3 and the NGINX base image is tagged as v1.2.0. The Test Runner is built on top of the NGINX image.

Due to the fact that this is a quite big difference, I'm rather safe than sorry and therefore promote two images. Difference between `release-1.12` and `release-1.11` is small to not existing when it comes to the Test Runner, so we can safely use the same image for both of them.


strongjz (approved):
/lgtm

PR #7695, opened 2025-01-19
Bump CPU and Mem for codesearch pods

dims:
Add missing CPU limit, codesearch/hound is very sensitive and needs CPU for serving requests. Needs more memory as well.


dims:
(already trying it out in `aaa` cluster)

dims:
Running experiment from https://github.com/kubernetes/k8s.io/issues/2182#issuecomment-2515795607 bumping CPU/Mem definitely helps!

image
image

PR #7704, opened 2025-01-21
Update groups.yaml to add jmfran@google.com

jm-franc:
Adding myself to get access to gcr.io/k8s-staging-autoscaling to create a VPA release, per https://github.com/kubernetes/autoscaler/blob/master/vertical-pod-autoscaler/RELEASE.md#permissions


raywainman:
/approve

(I think this should work now?)

gjtempleton:
Interesting, wonder why it's not picking up the update to the OWNERS file...
/approve

PR #7718, opened 2025-01-23
Promote cs-canary to prod cs.k8s.io

dims:
Fixes https://github.com/kubernetes/k8s.io/issues/2182

With the recent changes in https://github.com/kubernetes/k8s.io/pull/7707 and https://github.com/kubernetes/k8s.io/pull/7695 the performance is on par with the baremetal host we were running on equinix. Let's switch over the default cs.k8s.io to our own infra so we can turn down the equinix host. Thanks for all the support all these years to Equinix/Packet folks!

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,7 +3,7 @@
 Kubernetes project infrastructure, managed by the kubernetes community via [sig-k8s-infra]
 
 - `apps`: community-managed apps that run on the community-managed `aaa` cluster
-  - `codesearch`: instance of [codesearch] at <https://cs-canary.k8s.io> - owned by [sig-k8s-infra]
+  - `codesearch`: instance of [codesearch] at <https://cs.k8s.io> - owned by [sig-k8s-infra]
   - `elekto`: instance of [elekto] at <https://elections.k8s.io> - owned by Elections officers (on behalf of [sig-contributor-experience])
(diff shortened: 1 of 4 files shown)


BenTheElder:
what about https://github.com/kubernetes/k8s.io/pull/7707#pullrequestreview-2568452627 ?

dims:
@BenTheElder seemed unrelated. I probably have to check if we create the NAPs in terraform as well ( not that familiar with TF )

```
 Error: Error when reading or editing SQL Database Instance "terraform-20210921031444719700000001": googleapi: Error 403: Cloud SQL Admin API has not been used in project 270672962129 before or it is disabled. Enable it by visiting https://console.developers.google.com/apis/api/sqladmin.googleapis.com/overview?project=270672962129 then retry. If you enabled this API recently, wait a few minutes for the action to propagate to our systems and retry.
│
```

dims:
@BenTheElder let's please land this today if possible, i'll kick the tires over the weekend to make sure it works ok

dims:
thanks @BenTheElder 

/hold cancel
/skip

vielmetti:
Thanks @dims !

dims:
@vielmetti give me a few days to clean up things on that box :) then it's all yours!

dims:
@vielmetti please feel free to take back / de-provision the server. thanks!

PR #7729, opened 2025-01-27
WIP: Move staging repo away from gcr.io

xing-yang:
As stated here https://cloud.google.com/artifact-registry/docs/transition/transition-from-gcr, gcr Container Registry will be shut down after March 18, 2025. We need to move our staging away from gcr.io.

Images on staging repo can be viewed here after we moved over (https://github.com/kubernetes-csi/csi-release-tools/pull/271):
https://console.cloud.google.com/artifacts/docker/k8s-staging-sig-storage

### DIFF
--- a/registry.k8s.io/images/k8s-staging-sig-storage/generate.sh
+++ b/registry.k8s.io/images/k8s-staging-sig-storage/generate.sh
@@ -18,10 +18,10 @@ set -o errexit
 set -o nounset
 set -o pipefail
 
-readonly repo="gcr.io/k8s-staging-sig-storage"
+readonly repo="us-central1-docker.pkg.dev/k8s-staging-sig-storage"
 readonly tag_filter="tags~^v\d+\.\d+\.\d+\$"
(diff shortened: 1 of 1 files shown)


registry.k8s.io/images/k8s-staging-sig-storage/generate.sh
```diff
@@ -18,10 +18,10 @@ set -o errexit
 set -o nounset
 set -o pipefail
 
-readonly repo="gcr.io/k8s-staging-sig-storage"
+readonly repo="us-central1-docker.pkg.dev/k8s-staging-sig-storage"
```
jsafrane, line 21: This gives me 404 today (`gcloud container images list-tags us-central1-docker.pkg.dev/k8s-staging-sig-storage/csi-attacher`).

How can we confirm it's the right address?

xing-yang: We have not pushed any images to the new staging repo yet, so I think you will get a 404 if you are looking specifically for csi-attacher images.

PR #7761, opened 2025-02-05
Added the csi-snapshot-metadata image.

carlbraganza:


carlbraganza:
@xing-yang I did not modify the `images.yaml` file. If I were to regenerate it, besides adding csi-snapshot-metadata, it would re-sort csi-snapshotter and delete a bunch of images from smbplugin.

Should I regenerate the `images.yaml` file or manually add the csi-snapshot-metadata entry to it?

PR #7776, opened 2025-02-10
Add terraform for capz monitoring cluster

willie-yao:
Adds terraform needed to recreate the capz monitoring cluster used to monitor/export metrics from CAPZ e2e tests.

### DIFF
--- a/infra/azure/terraform/capz/capz-monitoring/main.tf
+++ b/infra/azure/terraform/capz/capz-monitoring/main.tf
@@ -1,0 +1,79 @@
+/*
+Copyright 2024 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+variable "resource_group_name" {
+  type = string
+}
+
+variable "location" {
+  type = string
+}
+
+variable "subscription_id" {
+  type = string
+}
+
+# Create the "capz-monitoring" resource group
+resource "azurerm_resource_group" "capz-monitoring" {
+  location = var.location
+  name     = var.resource_group_name
+  tags = {
+    DO-NOT-DELETE     = "contact capz"
+    creationTimestamp = timestamp()
+  }
+}
+
+resource "azurerm_user_assigned_identity" "capz_monitoring_user_identity" {
+  name                = "capz-monitoring-user-identity"
+  location            = azurerm_resource_group.capz-monitoring.location
+  resource_group_name = azurerm_resource_group.capz-monitoring.name
+}
+
+resource "azurerm_role_assignment" "monitoring_reader" {
+  principal_id         = azurerm_user_assigned_identity.capz_monitoring_user_identity.principal_id
+  role_definition_name = "Monitoring Reader"
+  scope                = "/subscriptions/${var.subscription_id}"
+  depends_on = [ azurerm_user_assigned_identity.capz_monitoring_user_identity ]
+}
+
+resource "azurerm_kubernetes_cluster" "capz-monitoring" {
+  dns_prefix            = var.resource_group_name
+  location              = var.location
+  name                  = var.resource_group_name
+  resource_group_name   = var.resource_group_name
+  tags = {
+    DO-NOT-DELETE     = "contact capz"
+    creationTimestamp = timestamp()
+  }
+  depends_on = [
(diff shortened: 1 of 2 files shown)


willie-yao:
/hold for reviews
/cc @jsturtevant @nojnhuh @ritikaguptams @mboersma 

My first go with terraform stuff. I just modeled the file off of my existing capz-monitoring cluster. Is it okay to hard-code the name and location since I don't think it's currently set to the default? If not, I can re-create the cluster in the correct location

mboersma (approved):
/lgtm

> Is it okay to hard-code the name and location

I'm not sure actually, I followed the `var.location` convention when I made changes, but I'm not a terraform expert. I doubt it's a problem in practice to use `eastus`, but maybe @jsturtevant knows?

infra/azure/terraform/capz/capz-monitoring/main.tf
```diff
@@ -0,0 +1,89 @@
...
+resource "azurerm_resource_group" "capz-monitoring" {
+  location = var.location
+  name     = var.resource_group_name
+  tags = {
+    DO-NOT-DELETE     = "contact capz"
+    creationTimestamp = "2024-10-24T00:00:00Z"
+  }
+}
+
+resource "azurerm_resource_group" "MC_capz-monitoring_capz-monitoring_eastus" {
```
jsturtevant, line 39: I am surprised that this resource group needs to be captured in the terraform.  This is a RG that is managed by AKS and shouldn't be managed by terraform AFAIK.

https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs/resources/kubernetes_cluster#node_resource_group-1

Says confirms that for me, in that we could provide a name but otherwise don't need to specify it.

jsturtevant:
> My first go with terraform stuff. I just modeled the file off of my existing capz-monitoring cluster. Is it okay to hard-code the name and location since I don't think it's currently set to the default? If not, I can re-create the cluster in the correct location

I think you've done the right thing with the location/name. Those are variables, you can specify them with variables to use them with pre-created clusters.  The part you don't need is the `node_resource_group` which shouldn't be managed by Terraform, if I understand the docs correct: https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs/resources/kubernetes_cluster

infra/azure/terraform/capz/capz-monitoring/main.tf
willie-yao: Ah that makes sense. I included it because I noticed that this resource group has been getting deleted on the CNCF sub periodically while I'm testing even though I have the do not delete tag set on it. Does this need to be in terraform if I don't want it to be deleted, or is that being caused by something else?

jsturtevant: I believe that if you put the tags do-not-delete on the AKS resource group and AKS cluster, it will carry those through when it creates managed RG.

willie-yao: Hmm okay, next time it gets deleted I might need some help figuring out what is causing it. Thanks!

infra/azure/terraform/capz/capz-monitoring/main.tf
```diff
@@ -0,0 +1,79 @@
...
+  type = string
+}
+
+# Create the "capz-monitoring" resource group
+resource "azurerm_resource_group" "capz-monitoring" {
+  location = var.location
+  name     = var.resource_group_name
+  tags = {
+    DO-NOT-DELETE     = "contact capz"
+    creationTimestamp = "2024-10-24T00:00:00Z"
```
jsturtevant, line 35: maybe we could use `timestamp()` here (https://developer.hashicorp.com/terraform/language/functions/timestamp). rg-cleanup uses rf3339 so `timestamp()` should work here https://github.com/Azure/rg-cleanup/blob/1f19a91e03d7c78f0ce1634f1e022a6e8c2713d3/main.go#L31

infra/azure/terraform/capz/capz-monitoring/main.tf
```diff
@@ -0,0 +1,79 @@
...
+}
+
+resource "azurerm_kubernetes_cluster" "capz-monitoring" {
+  dns_prefix            = var.resource_group_name
+  location              = var.location
+  name                  = var.resource_group_name
+  resource_group_name   = var.resource_group_name
+  tags = {
+    DO-NOT-DELETE     = "contact capz"
+    creationTimestamp = "2024-10-24T00:00:00Z"
```
jsturtevant, line 59: same here for `timestamp`

jsturtevant (commented):
LGTM, minor nit on timestamp but otherwise looks good

mboersma (approved):
/lgtm

jsturtevant (approved):

marosset (approved):
LGTM

willie-yao:
/hold
whoops forgot to squash

willie-yao:
Can I get one more approval here? Sorry I had to squash @jsturtevant @mboersma @marosset

mboersma (approved):
/lgtm
/approve

PR #7781, opened 2025-02-12
Add mbianchidev to release editors

mbianchidev:
Update groups.yaml to add myself to release editors.

CC @jimangel (as my release management "sponsor")


jimangel:
/lgtm
@kubernetes/release-managers if we could get an approval here, this should allow @mbianchidev to lead the alpha.2 cut tomorrow with me shadowing.

PR #7792, opened 2025-02-13
update sig-storage image list and promotion script

BlaineEXE:
The COSI project has 2 images that need to be promoted to the k8s.io registry for its latest release.

COSI staged images previously had multiple tags including the semver tag. The sig-storage `generate.sh` script doesn't output multi-tagged images correctly, so COSI modified its release tagging to use only the semver tag for staged release images. In order to filter out existing staged COSI images that have multiple tags, introduce an inverted grep to the script in this commit.

Run the script to generate an updated `images.yaml`.

### DIFF
--- a/registry.k8s.io/images/k8s-staging-sig-storage/generate.sh
+++ b/registry.k8s.io/images/k8s-staging-sig-storage/generate.sh
@@ -61,5 +61,8 @@ for image in "${images[@]}"; do
         --format="get(digest, tags)" \
         --sort-by="tags" \
         --filter="${filter}" \
+        | grep --invert-match ';' \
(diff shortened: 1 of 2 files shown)


BlaineEXE:
In theory, I could have instead modified the script to support multiple tags, but the change would a bit more involved. It seems safer to make this small change, rather than change the image promotion tooling in ways that could have effects I don't know how to anticipate.

For comparison, I did find examples of multi-tagged images in an `images.yaml` here https://github.com/kubernetes/k8s.io/blob/1bd588ff62e241d030ec07fc4e0b47d4048c6a28/registry.k8s.io/images/k8s-staging-artifact-promoter/images.yaml#L17-L18

PR #7797, opened 2025-02-15
redirect: more short links for k8s infra

ameukam:
more replacements of bit.ly


ameukam:

PR #7801, opened 2025-02-18
fix bad promotion, tag is already there

cpanato:
follow up of #7799 

/assign @xmudrii @saschagrunert 


xmudrii (approved):
/lgtm
/approve

PR #7818, opened 2025-02-21
Add Chart Image approval to Kueue maintainers

tenzen-y:
Based on this discussion, I added the chart directory approval permissions to Kueue maintainers.

https://kubernetes.slack.com/archives/CCK68P2Q2/p1739983650931409?thread_ts=1739869280.030129&cid=CCK68P2Q2

Related to https://github.com/kubernetes-sigs/kueue/issues/4143

### DIFF
--- a/registry.k8s.io/images/charts/OWNERS
+++ b/registry.k8s.io/images/charts/OWNERS
@@ -1,0 +1,6 @@
+# See the OWNERS file documentation:
+#  https://github.com/kubernetes/community/blob/master/contributors/devel/owners.md
+
+approvers:
+- tenzen-y
+- mimowo


mimowo:
/lgtm
Thanks!

registry.k8s.io/images/charts/OWNERS
```diff
@@ -0,0 +1,6 @@
+# See the OWNERS file documentation:
+#  https://github.com/kubernetes/community/blob/master/contributors/devel/owners.md
+
+approvers:
+- tenzen-y
+- mimowo
```
tenzen-y, line 6: @upodroid If we need to add sig labels, please let me know, thanks!

https://github.com/kubernetes/k8s.io/blob/5b37813bd4f4e3e7a675dc26ffcceb4f75691b41/registry.k8s.io/images/k8s-staging-kueue/OWNERS#L17-L18

upodroid (approved):
We'll add other chart maintainers in the future as needed

tenzen-y:
> We'll add other chart maintainers in the future as needed

That sounds reasonable.
Thanks all!
/hold cancel

aojea:
> We'll add other chart maintainers in the future as needed

cc: @BenTheElder , we may need to find a different model

BenTheElder:
Seems like we should do registry.k8s.io/$project/charts/ ...?

We want to be able to delegate adding new no-conflict artifacts, scoping them to a directory solves this.

tenzen-y:
> Seems like we should do registry.k8s.io/$project/charts/ ...?
> 
> We want to be able to delegate adding new no-conflict artifacts, scoping them to a directory solves this.

@BenTheElder Hi, thank you for checking this. Based on this discussion (https://docs.google.com/document/d/1xa2me_O0N5sHMD_np9ztr2O9GL3uX1yKBgDcApDqV_8/edit?resourcekey=0-JHI0r4CeitKdIEJII7c8eg&tab=t.0#heading=h.kuew5wcageoq) with sig k8s-infra, we added charts to the current directory.

If you or other infra-related sig members have different recommendations, please let me know, thanks!
cc: @mimowo

tenzen-y:
> > If you or other infra-related sig members have different recommendations, please let me know, thanks!
> 
> cc @kubernetes/sig-k8s-infra-leads
> 
> I really think we should give each subproject their own charts directory, the same as we do with images.
> 
> Having a single shared charts directory / publishing location has the following downsides:
> 
> * poor security isolation
>   
>   * anyone able to publish a chart can publish any of the charts (presumably we will have more projects soon)
> * poor signal/noise ratio
>   
>   * git commits will be for all of these projects, right now you can just follow the commits for the project specific subdir for images
> * potential for naming conflicts
>   
>   * This is impossible if they're published to different root paths

That sounds reasonable. @upodroid, could you confirm @BenTheElder 's opinion and that there are no drawbacks?
If you do not have any objections, I think we can migrate this to @BenTheElder 's solution.

PR #7849, opened 2025-03-06
Adding image sha for v0.2.0 RC

kfswain:

PR #7850, opened 2025-03-06
Adding SHA for Gateway Inference Extension v0.2.0-rc.1

robscott:
Replaces #7849 because CLA doesn't like Kellen today.

/cc @ahg-g @danehans @kfswain


kfswain:
> CLA doesn't like Kellen today.

Can't say I blame it tbh.

Thanks for savin the day Rob!
/lgtm
/approve

PR #7852, opened 2025-03-06
Update groups.yaml for sig scalability

mengqiy:
Per https://github.com/kubernetes/community/pull/8343 and https://github.com/kubernetes/community/issues/8370

### DIFF
--- a/groups/sig-scalability/groups.yaml
+++ b/groups/sig-scalability/groups.yaml
@@ -63,6 +63,7 @@ groups:
       SIG scalability general discussion group
     owners:
       - marcel.zieba@isovalent.com
+      - mengqiyu1991@gmail.com
       - shyam123.jvs95@gmail.com
       - wojtekt@google.com
     settings:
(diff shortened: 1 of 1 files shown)


groups/sig-scalability/groups.yaml
```diff
@@ -116,11 +117,13 @@ groups:
       - siarkowicz@google.com
       - polynomial@redhat.com
       # sig-scalability leads
-      - jeedigv@amazon.com
-      - mmatejczyk@google.com
-      - wojtekt@google.com
       - dordel@google.com
       - marcel.zieba@isovalent.com
+      - mengqiyu1991@gmail.com
```
shyamjvs, line 122: Looks like everywhere else you used corp email. Use the same here for consistency?

mengqiy: My GH handle is used in other PRs in https://github.com/kubernetes/community/issues/8370#event-2136003055
I guess you may confuse my GH handle with my corp email. There's one letter difference.
I think I'm using my gmail in k8s community consistently. And both my gmail and corp email have been associated with my GH account. I don't think there will be any issue.

PR #7862, opened 2025-03-11
add ppc64le cluster to prow

upodroid:
/cc @Rajalakshmi-Girish @ameukam @BenTheElder

PR #7904, opened 2025-03-20
Add ClusterSecretStore for IBM provider's Secrets Manager

Rajalakshmi-Girish:
This ClusterSecretStore is to fetch the secrets from Secrets Manager in IBM Cloud provider.
This would be deployed on IBM's ppc64le build cluster added by https://github.com/kubernetes/k8s.io/pull/7862

### DIFF
--- a/kubernetes/ibm-ppc64le/prow/external-secrets.yaml
+++ b/kubernetes/ibm-ppc64le/prow/external-secrets.yaml
@@ -1,0 +1,36 @@
+---
+apiVersion: external-secrets.io/v1beta1
+kind: ExternalSecret
+metadata:
+  name: k8s-infra-ibmcloud-iam
+  namespace: test-pods
+spec:
+  refreshInterval: 60m
+  secretStoreRef:
+    name: secretstore-ibm-k8s
+    kind: ClusterSecretStore
+  target:
+    name: k8s-infra-ibmcloud-iam
+    creationPolicy: Owner
+  data:
+  - secretKey: key
+    remoteRef:
+      key: iam_credentials/c4c5e90a-408d-69de-38b2-0d56d58d29db
+---
+apiVersion: external-secrets.io/v1beta1
+kind: ExternalSecret
+metadata:
+  name: k8s-infra-ssh-key
+  namespace: test-pods
+spec:
+  refreshInterval: 60m
+  secretStoreRef:
+    name: secretstore-ibm-k8s
+    kind: ClusterSecretStore
+  target:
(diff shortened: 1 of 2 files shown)


kubernetes/ibm-ppc64le/prow/secretstore-ibm-k8s.yaml
```diff
@@ -0,0 +1,14 @@
...
+spec:
+  provider:
+    ibm:
+      serviceUrl: "https://3297fd32-6322-45e2-af3f-00b1a5af3565.us-south.secrets-manager.appdomain.cloud"
+      auth:
+        secretRef:
+          secretApiKeySecretRef:
+            name: ibm-sm-apikey
+            key: API_KEY
+            namespace: test-pods
```
mkumatag, line 14: is this a requirement to have this secret in the test-pods? if not lets avoid this.

mkumatag, line 14: is this a requirement to have this secret in the test-pods? if not lets avoid this.

mkumatag (changes requested):
hope will have external secrets also created soon to pull the keys from the secret manager.

kubernetes/ibm-ppc64le/prow/secretstore-ibm-k8s.yaml
upodroid: Can we add that secret to the external-secrets namespace instead?

Like this

kubernetes/ibm-ppc64le/prow/external-secrets.yaml
```diff
@@ -0,0 +1,36 @@
+---
+apiVersion: external-secrets.io/v1beta1
+kind: ExternalSecret
+metadata:
+  name: ibmcloud-iam
```
mkumatag, line 5: can we have more meaningful name based on who is consuming this?

mkumatag (changes requested):

mkumatag:
@upodroid post merging this change, I don't see the clustersecretstores or external secret, so do we need to deploy them manually or some automation will take care of this?

upodroid:
Automation will take care of it but there something was wrong, fixed in 7914

mkumatag:
> Automation will take care of it but there something was wrong, fixed in 7914

ah, okay. How to avoid such errors? do we have instructions or guidelines how to add these new resources and is there way to run dry-run commands to validate the specs?

PR #7992, opened 2025-04-15
SIG Autoscaling - Mailing list update with new SIG leads

gjtempleton:
Ref https://github.com/kubernetes/community/issues/8368

* Also alphabetically sorted email addresses in infra-staging group

### DIFF
--- a/groups/sig-autoscaling/groups.yaml
+++ b/groups/sig-autoscaling/groups.yaml
@@ -12,8 +12,10 @@ groups:
     description: |-
       SIG autoscaling leads
     owners:
-      - maciekpytel@google.com
       - guyjtempleton@googlemail.com
+      - jack.francis@microsoft.com
+      - jtuznik@google.com
+      - rwainman@google.com
     settings:
       AllowWebPosting: "true"
       ReconcileMembers: "true"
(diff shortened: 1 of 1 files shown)


gjtempleton:
/hold
@towca could you just give the current membership of the infra group a look over and check whether we should prune any more of the existing emails there?

groups/sig-autoscaling/groups.yaml
```diff
@@ -79,14 +83,14 @@ groups:
...
       - bskiba@google.com
-      - maciekpytel@google.com
-      - ksiedlecki@google.com
-      - jbartosik@google.com
-      - jtuznik@google.com
+      - bwroblewski@google.com
       - danielmk@google.com
       - guyjtempleton@googlemail.com
+      - jack.francis@microsoft.com
+      - jbartosik@google.com
```
towca, line 90: We can remove these:
* jbartosik@google.com
* kgolab@google.com

AFAIK, the rest of google.com addresses here are people still working on Cluster Autoscaler or VPA at least occasionally.

@raywainman to confirm the VPA side.

towca:
Done, added a comment!

Otherwise /lgtm

groups/sig-autoscaling/groups.yaml
raywainman: confirmed for VPA folks

PR #8029, opened 2025-04-23
GCP: bump terraform providers

ameukam:
Bump the GCP provider so we can use external terraform modules requiring the 6.x provider

### DIFF
--- a/infra/gcp/terraform/k8s-infra-prow-build-trusted/00-provider.tf
+++ b/infra/gcp/terraform/k8s-infra-prow-build-trusted/00-provider.tf
@@ -30,11 +30,11 @@ terraform {
   required_providers {
     google = {
       source  = "hashicorp/google"
-      version = "~> 5.45.0"
+      version = "~> 6.31.0"
     }
     google-beta = {
       source  = "hashicorp/google-beta"
-      version = "~> 5.45.0"
+      version = "~> 6.31.0"
     }
   }
 }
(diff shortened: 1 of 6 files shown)


BenTheElder:
I assume these are still not BSL?

Can we let the merge backlog from code thaw land before any potentially disruptive infra changes?

BenTheElder:
(also assuming they're compatible with our old terraform copy ...)

ameukam:
> I assume these are still not BSL?
> 
> Can we let the merge backlog from code thaw land before any potentially disruptive infra changes?

This is not disruptive because it's not triggering any change to the infrastructure. And we potentially only reconcile what's was already changed by the provider.

ameukam:
> I assume these are still not BSL?

https://github.com/hashicorp/terraform-provider-google/blob/main/LICENSE

BenTheElder:
> This is not disruptive because it's not triggering any change to the infrastructure. And we potentially only reconcile what's was already changed by the provider.

It's triggering at least one change (which we hope is not disruptive), but I also don't see the rush, the merge queue should be cleared by tomorrow.

dims:
/lgtm
/hold

for merge queue to clear (and then we can remove hold)

BenTheElder:
we have around one more batch merge to go https://prow.k8s.io/tide

ameukam:
atlantis apply

PR #8057, opened 2025-04-30
Promote vpa 1.3.1 images

raywainman:
Promote latest VPA images as part of https://github.com/kubernetes/autoscaler/issues/8073.


voelzmo:
/lgtm
Thanks!

PR #8059, opened 2025-04-30
fix issue search shortcuts

BenTheElder:
the current https://go.k8s.io/good-first-issue and https://go.k8s.io/help-wanted don't work, we need to switch github endpoints

PR #8081, opened 2025-05-08
Update the search filter for help-wanted and good-first-issue

shotarok:
Fixes https://github.com/kubernetes/community/issues/8452.

https://go.k8s.io/help-wanted and https://go.k8s.io/good-first-issue don't show any issues at this moment because the search filters include multiple org keywords without OR. We can see the issues by adding OR. Please take a look at the screenshot on the issue.


shotarok:
I just found the other fix PR: https://github.com/kubernetes/k8s.io/pull/8059, so I’ll close mine.

PR #8109, opened 2025-05-15
Move from bit.ly URLs to rel.k8s.io for the release team

Vyom-Yadav:
Bit ly has started adding ads between redirects, so we're moving away from bit ly. 

Context: https://kubernetes.slack.com/archives/C2C40FMNF/p1739462854147559

/cc @BenTheElder @katcosgrove @fsmunoz @neoaggelos

### DIFF
--- a/apps/k8s-io/configmap-nginx.yaml
+++ b/apps/k8s-io/configmap-nginx.yaml
@@ -345,9 +345,11 @@ data:
         server_name releases.k8s.io rel.k8s.io releases.kubernetes.io rel.kubernetes.io;
         listen 80;
 
-        location / {
-          rewrite ^/$                  https://github.com/kubernetes/kubernetes/releases redirect;
-          rewrite ^/([^/]*)(/.*)?$     https://github.com/kubernetes/kubernetes/tree/$1$2 redirect;
+          rewrite ^/$                               https://github.com/kubernetes/kubernetes/releases redirect;
+          rewrite ^/k8s-release-cal                 https://calendar.google.com/calendar/embed?src=kipmnllvl17vl9m98jen6ujcrs%40group.calendar.google.com redirect;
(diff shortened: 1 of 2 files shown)


apps/k8s-io/configmap-nginx.yaml
```diff
@@ -346,8 +346,25 @@ data:
...
+          # Special case for versions up to v1.33 - redirect to bit.ly
+          if ($uri ~ "^/v([0-9])([0-9]+)/([a-zA-Z]+)$") {
+            set $major $1;
+            set $minor $2;
+            set $remaining $3;
+
+            # For versions <=1.33, redirect to bit.ly
+            if ($major = "1" && $minor <= "33") {
+              return 301 https://bit.ly/k8s$major$minor-$remaining;
+            }
```
Vyom-Yadav, line 358: Thinking about this again, this could allow for something similar to subdomain takeover or unknown redirect.

`rel.k8s.io/v123/steal-your-passwords` would redirect externally to `https://bit.ly/k8s123-steal-your-passwords` which isn't safe imo. 

For maintaining backwards compatibility, we can have a generic links.md in sig-release which would have the generic URLs for previous release.

BenTheElder: We should also constrain the lower bound if we go this route.

I think it might be best to directly include them here without the bit.ly link in the middle?

Vyom-Yadav: Yeah, having bit ly remains doesn't seem like a great solution. I can write a short script which can auto populate links.md for the older releases. 

Updated this to do a redirect to sig-release only.

Vyom-Yadav:
ping @BenTheElder I addressed the review comment above.

BenTheElder:
@Vyom-Yadav sent me a nudge in slack, sorry I let this fall off my radar. 

/lgtm
/approve

BenTheElder:
I think this still has to be manually deployed.

PR #8131, opened 2025-05-26
rollout the c4 instances more widely and complete the cgroups v2 migration for GKE build cluster

### LINKED ISSUE #5276: Migrate k8s-infra-prow-builds to cgroups v2
We'll probably want to coordinate with test-infra-oncall to do this for the existing default cluster as well.

https://cloud.google.com/kubernetes-engine/docs/how-to/node-system-config#cgroup-mode-options

cgroups v2 is the future and eventually when kubernetes etc stops supporting v1 then we will need this to run things like KIND and local-up-cluster.sh

upodroid:
Fixes #5276 

/hold


upodroid:
atlantis plan

upodroid:
atlantis apply

upodroid:
/hold cancel

old nodepool has been scaled down to 0

PR #8156, opened 2025-06-03
fix arm64 nodepool for kind

upodroid:
/cc @ameukam @aojea @BenTheElder 

The kernel bug appeared in the kind jobs on arm64 nodes so I changed their os too.

c4d is in preview so I'm forcing the nodepool to always run atleast 10 nodes of this type per zone. Looking at the nodepool size metrics, it has consistently stayed above 20 per zone for weeks. The autoscaler seems to be preferring c4 pool over c4d pool, which isn't ideal.

image


upodroid:
atlantis plan

upodroid:
atlantis plan

BenTheElder:
thanks

PR #8169, opened 2025-06-06
Add neoaggelos to release editors

neoaggelos:
### Summary

add myself to release editors, cc @mbianchidev as my release management sponsor

### DIFF
--- a/groups/sig-release/groups.yaml
+++ b/groups/sig-release/groups.yaml
@@ -52,6 +52,7 @@ groups:
       - pal.nabarun95@gmail.com
       - saschagrunert@gmail.com
       - admin@mb-consulting.dev
+      - neoaggelos@gmail.com
 
   - email-id: k8s-infra-release-viewers@kubernetes.io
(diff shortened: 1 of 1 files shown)


groups/sig-release/groups.yaml
```diff
@@ -52,6 +52,7 @@ groups:
       - pal.nabarun95@gmail.com
       - saschagrunert@gmail.com
       - admin@mb-consulting.dev
+      - neoaggelos@gmail.com
```
Verolop: Please add your email address in the correct alphabetical position, just above `pal.nabarun95@gmail.com` to keep the list sorted. While you're at it, would you mind rearranging the list in alphabetical order so it's easy for folks to identify the pattern in the future?
This is to maintain the style of the rest of the lists and docs.
Thanks!

neoaggelos: updated!

Verolop (commented):
/lgtm
/approve

PR #8171, opened 2025-06-08
openstack v1.33 images

zetaab:


zetaab:
@kayrus can you review

PR #8178, opened 2025-06-10
Kueue: Sync Oweners with the latest ones

tenzen-y:
The current owners are a bit of outdated.

/assign @mimowo

### DIFF
--- a/registry.k8s.io/images/k8s-staging-kueue/OWNERS
+++ b/registry.k8s.io/images/k8s-staging-kueue/OWNERS
@@ -6,9 +6,7 @@ approvers:
 - mimowo
 
 reviewers:
-- denkensk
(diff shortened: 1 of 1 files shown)


tenzen-y:
> The following users are mentioned in OWNERS file(s) but are untrusted for the following reasons. One way to make the user trusted is to add them as [members](https://git.k8s.io/community/community-membership.md#member) of the kubernetes org. You can then trigger verification by writing `/verify-owners` in a comment.
> 
> * mbobrovskyi
>   
>   * User is not a member of the org. Satisfy at least one of these conditions to make the user trusted.
> * pbundyra
>   
>   * User is not a member of the org. Satisfy at least one of these conditions to make the user trusted.

Updated.

PR #8204, opened 2025-06-18
IBMCLOUD: add tags to powervs workspaces

Prajyot-Parab:
- IBMCLOUD: add tags to powervs workspaces

### DIFF
--- a/infra/ibmcloud/terraform/k8s-power-conformance/modules/pvs_workspace/variables.tf
+++ b/infra/ibmcloud/terraform/k8s-power-conformance/modules/pvs_workspace/variables.tf
@@ -18,3 +18,4 @@ variable "datacenter" {}
 variable "pi_workspace_name" {}
 variable "resource_group_id" {}
 variable "image_name" {}
+variable "tags" { default = [] }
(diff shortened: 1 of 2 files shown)


infra/ibmcloud/terraform/k8s-power-conformance/pvs.tf
```diff
@@ -73,6 +77,7 @@ module "powervs_workspace_syd05" {
   pi_workspace_name = "k8s-boskos-powervs-syd05"
   resource_group_id = module.resource_group.k8s_rg_id
   image_name        = var.image_name
+  tags              = local.workspace_tags
```
mkumatag: I'm unsure whether we actually need to track these temporary tags as part of the source code. Instead, I'd prefer to manage them offline and roll back when we're ready to reconsider those workspaces for scheduling.

mkumatag (changes requested):

infra/ibmcloud/terraform/k8s-power-conformance/pvs.tf
Prajyot-Parab: Sure @mkumatag, I will update the PR to drop the specific tag attachment code bits + will keep only the tag feature support code.

Prajyot-Parab: Done.

mkumatag (approved):
/lgtm

PR #8222, opened 2025-06-23
add nirrozenbaum as owner of gateway api inference extension

nirrozenbaum:
(instead of jeffwan)


kfswain:
I think we need @Jeffwan for approval, I'll try tho

/approve

PR #8229, opened 2025-06-24
switch to hyperdisk-balanced for ghproxy in prow-build-trusted

BenTheElder:
we need to use a disk type compatible with the nodes

see previously 4702c302431380cdab598c996970e6e214dd1d7e


michelle192837 (approved):
/hold

Unhold at will

BenTheElder:
after https://github.com/kubernetes/k8s.io/pull/8231 this synced and there's a green ghproxy deployment

PR #8234, opened 2025-06-25
gcve: remove nsx-gateway

chrischdi:
The case for GCVE got resolved.

> I would like to share an update regarding the issue we noticed that  there is an issue with Cloud NAT port allocation. After reviewing the situation, we identified that the problem stemmed from the port allocation settings. To resolve this, we enabled dynamic port allocation on the Cloud NAT and set the minimum and maximum ports per VM to 32,768 and 65,536, respectively.

This was part of GCVE which is not accessible as a user and got solved by the support team.

Note: terraform is already applied / destroyed (for nsx-gateway)


sbueringer:
Thank you very much!

/lgtm
/approve

infra/gcp/terraform/k8s-infra-gcp-gcve/README.md
```diff
@@ -50,15 +50,6 @@ resource pools and other vSphere components required when running tests. See:
 
 # Working around network issues
```
sbueringer, line 51: @chrischdi Could it be that the Gateway VM is still on the picture? (GVCE.draw.io.png) (should be GCVE btw :D)

PR #8247, opened 2025-06-26
override kubectl image repository

Prajyot-Parab:
- override kubectl image repository to use k8s vended image [ref](https://explore.ggcr.dev/?image=registry.k8s.io%2Fkubectl:v1.32.3), by default it uses `bitnami/kubectl` image which lacks mutli-arch support [ref](https://explore.ggcr.dev/?image=bitnami%2Fkubectl%3A1.32.3)


kubernetes/ibm-ppc64le/helm/kyverno.yaml
```diff
@@ -0,0 +1,8 @@
+---
+webhooksCleanup:
+  image:
+    repository: registry.k8s.io/kubectl
```
mkumatag, line 4: you need to override the tag as well to point to the right image

kubernetes/ibm-ppc64le/helm/kyverno.yaml
```diff
@@ -0,0 +1,8 @@
...
+
+policyReportsCleanup:
+  image:
+    repository: registry.k8s.io/kubectl
```
mkumatag, line 9: same as above

Prajyot-Parab: fixed.

kubernetes/ibm-ppc64le/helm/kyverno.yaml
Prajyot-Parab: fixed.

PR #8314, opened 2025-07-22
promote etcd 3.6.2-0

pacoxu:
promote after https://github.com/kubernetes/kubernetes/pull/132864


ahrtr:
FYI. we are going to release etcd v3.6.3 soon (today), so as to catch up with Kubernetes 1.34.0.

PR #8325, opened 2025-07-23
prow-build: use hyperdisk-balanced disks for prometheus

xmudrii:
The `pd-balanced` disks that we have been using for Prometheus in the GKE Prow build cluster are not available with the C4 instances that we migrated to: https://cloud.google.com/compute/docs/machine-resource#machine_type_comparison

This PR addresses this by migrating to the Hyperdisk Balanced disks that are supported on the C4 instances. This change has been already reconciled manually.

/assign @BenTheElder


xmudrii:
This is not a terraform change
/override "atlantis/plan"

BenTheElder (commented):
/lgtm
/approve
thanks!

PR #8327, opened 2025-07-25
increase KETTLE memory limit and add CPU request

BenTheElder:
these are based on inspecting actual utilization of the workload

x-ref: https://github.com/kubernetes/test-infra/issues/35173

/hold

### DIFF
--- a/apps/kettle/deployment.yaml
+++ b/apps/kettle/deployment.yaml
@@ -26,9 +26,11 @@ spec:
           value: kubernetes-public/kubernetes-ci-logs-updates/k8s-infra-kettle
         resources:
           requests:
-            memory: 4Gi
+            cpu: 2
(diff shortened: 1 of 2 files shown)


BenTheElder:
this can't be deployed with the way we have the autoscaling setup, and I think also the configuration of aaa is out of sync with the terraform.

BenTheElder:
@dims I can't find the codesearch nodepool in the terraform locally, but one of those two nodes is sitting unused, and we're at autoscaling limits for memory (which I haven't touched yet as I try to figure out if it's safe to touch the terraform)

BenTheElder:
Ahh that's because it's a NAP pool

BenTheElder:
atlantis plan

BenTheElder:
this is deployed

PR #8332, opened 2025-07-27
connect the s390x cluster to argocd

upodroid:
Part of https://github.com/kubernetes/k8s.io/issues/8082

/hold

@jeefy Did IBM handover this account correctly yet?


jeefy:
We do have accounts on here and access to handle/observe billing so should be g2g!

PR #8350, opened 2025-07-30
groups: drop setting for a staging group

ameukam:
Allow publishing is not required for Google groups using for staging repos.


ameukam:
Also attempting to debug why the group is not reconciled.

cc @xmudrii @upodroid

upodroid (approved):

PR #8352, opened 2025-07-30
groups: fix reconciliation for SIG UI groups

ameukam:
Follow-up of:
  - https://github.com/kubernetes/k8s.io/pull/8350

Look like we are using the wrong key.


ameukam:
cc @xmudrii @upodroid

upodroid (approved):

PR #8367, opened 2025-08-06
Make v$version also work for lws

kerthcet:
fix: https://github.com/kubernetes-sigs/lws/issues/609

I can not change the helm chart values now, let's make the v or non-v image tag always work. I'm always confusion here ...


kerthcet:
@yankay would you mind to tag a LGTM here considering your timezone.

yankay:
OK @kerthcet 
/lgtm

PR #8412, opened 2025-08-17
add the s390x cluster to prow

upodroid:
Part of #8082 

@brahads


kubernetes/gke-prow/prow/gce-ssd-retain_storageclass.yaml
upodroid, line 1: We are using C4 nodes so pd-ssd is no longer available.

kubernetes/gke-prow/prow/trusted_serviceaccounts.yaml
upodroid, line 1: These were copied over when we migrated prow to this new cluster. These service accounts are not used today.

PR #8506, opened 2025-09-13
update sig network projects to latest release

aojea:
Hash in https://prow.k8s.io/view/gs/kubernetes-ci-logs/logs/post-kube-network-policies-image/1966892555559768064

The nat64 was obtained directly from the registry


aojea:
/assign @dims @BenTheElder @upodroid @ameukam 

if some weekend person is around :)

PR #8517, opened 2025-09-16
Image promotion for build-image v0.8.2

dims:
Image promotion for build-image v0.8.2
This is an automated PR generated from `kpromo`
```
kpromo pr --fork dims --interactive --project build-image --tag v0.8.2
```

/hold


cpanato:
I just saw now why take 2?

We are not promoting the same tag rigth?

dims:
@cpanato kpromo missed the entries in this PR last time around.

no, we are not promoting the same tag again (i believe we have safe guards to prevent that from happening as well)

cpanato:
yes we have, but it is a pain to clean up

thanks for the clarification

PR #8523, opened 2025-09-17
Revert "Update boskos to v20250908-5e4ab72, prow to v20250903-f49c6158b and ghproxy as needed"

BenTheElder:
Reverts kubernetes/k8s.io#8467

fixes https://github.com/kubernetes/kubernetes/issues/134115 until boskos can be rolled forward safely

cc @dims @kubernetes/sig-k8s-infra-leads


stmcginnis:
Looks like a clean revert.

/lgtm

PR #8564, opened 2025-09-25
Update SIG-Autoscaling TL to adrianmoisey

raywainman:
Updates SIG Autoscaling TL from myself to @adrianmoisey per [this announcement](https://groups.google.com/g/kubernetes-sig-autoscaling/c/VH_Dj1nWCUw/m/ulj2jMCuAQAJ).

See https://github.com/kubernetes/community/issues/8629.

### DIFF
--- a/registry.k8s.io/images/k8s-staging-autoscaling/OWNERS
+++ b/registry.k8s.io/images/k8s-staging-autoscaling/OWNERS
@@ -1,6 +1,7 @@
 # See the OWNERS docs at https://go.k8s.io/owners
 
 approvers:
+- adrianmoisey
 - bskiba
 - maciekpytel
 - krzysied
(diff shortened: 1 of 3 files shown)


registry.k8s.io/images/k8s-staging-autoscaling/OWNERS
```diff
@@ -1,6 +1,7 @@
 # See the OWNERS docs at https://go.k8s.io/owners
 
 approvers:
+- adrianmoisey
```
adrianmoisey, line 4: Do you need to remove yourself from this list? On line 15

raywainman: Was going to leave myself here given there are other folks here but it looks like it is mostly Googlers + SIG leads so makes sense to remove myself.

Done.

jackfrancis (approved):
/lgtm
/approve

PR #8585, opened 2025-10-01
Update Google groups for 1.35 release docs shadows

Urvashi0109:
Update Google groups for 1.35 release docs shadows


dipesh-rawat (commented):
/lgtm

PR #8594, opened 2025-10-03
Rotate SIG K8s leads

ameukam:
Related to:
  - https://github.com/kubernetes/community/issues/8633
  - https://github.com/kubernetes/community/issues/8634

### DIFF
--- a/groups/sig-testing/groups.yaml
+++ b/groups/sig-testing/groups.yaml
@@ -174,9 +174,7 @@ groups:
     - k8s-infra-release-viewers@kubernetes.io
     # members
     - alarcj137@gmail.com
-    - ameukam@gmail.com
     - antonio.ojea.garcia@gmail.com
-    - bartek@smykla.com
     - eddiezane@gmail.com
     - gmccloskey@google.com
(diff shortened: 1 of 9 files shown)


groups/sig-testing/groups.yaml
BenTheElder, line 166: we should probably add sig-testing leads alias

BenTheElder: we should drop this, we're way past depending on google oncall, amerai should still be included but we can do that via the testing leads alias

BenTheElder (commented):
there's a lot going on here, but on first pass only minor comments that are not blockers

groups/sig-testing/groups.yaml
ameukam: Done

ameukam: Done

BenTheElder:
https://github.com/kubernetes/k8s.io/compare/d2322c5a41254e8d7864cd541b8414963ae4d14a..5330319e6047b02e866da2d97cea6ec768f71fae LGTM

would like additional eyes on the overall diff

PR #8600, opened 2025-10-05
add single process oom kill kubeletconfig to GKE nodes

upodroid:
This PR is busy but it does the following

- use the gke project factory module instead of our custom module
- The current GKE cluster is route based so we can't use the gke module we [use](https://github.com/kubernetes/k8s.io/blob/main/infra/gcp/terraform/k8s-infra-prow/gke.tf) in k8s-infra-prow project.
- The single_process_oom_kill field hasn't made it to the GKE module [yet](https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/tree/main/modules/gke-node-pool). Once it does, I'll mark the PR ready for review.

/cc @BenTheElder @ameukam

### DIFF
--- a/infra/gcp/terraform/k8s-infra-prow-build/provider.tf
+++ b/infra/gcp/terraform/k8s-infra-prow-build/provider.tf
@@ -23,18 +23,18 @@ This file defines:
 terraform {
 
   backend "gcs" {
-    bucket = "k8s-infra-tf-prow-clusters"
-    prefix = "k8s-infra-prow-build/prow-build" // $project_name/$cluster_name
+    bucket = "k8s-infra-terraform"
+    prefix = "k8s-infra-prow-build"
   }
 
   required_providers {
     google = {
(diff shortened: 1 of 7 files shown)


infra/gcp/terraform/k8s-infra-prow-build/provider.tf
```diff
@@ -23,18 +23,18 @@ This file defines:
 terraform {
 
   backend "gcs" {
-    bucket = "k8s-infra-tf-prow-clusters"
-    prefix = "k8s-infra-prow-build/prow-build" // $project_name/$cluster_name
+    bucket = "k8s-infra-terraform"
```
upodroid, line 26: This is a state bucket in the k8s-infra-seed project instead of the kubernetes-public

BenTheElder:
Not super familiar with the terraform project but we might have to file an issue about the kubelet option

PR #8612, opened 2025-10-07
create the scale-cache-puller in the prow-build project

upodroid:
Fixes: https://github.com/kubernetes/kubernetes/issues/134429

TIL; you can send base64 encoded service account keys to AR. https://cloud.google.com/artifact-registry/docs/docker/authentication#json-key

/cc @BenTheElder 

I applied the changes already and testing in https://github.com/kubernetes/kubernetes/pull/134461


BenTheElder (commented):
/lgtm
/approve

PR #8633, opened 2025-10-10
Add bowei to OWNERS_ALIASES

bowei:
Ref: https://github.com/kubernetes/community/issues/8608

### DIFF
--- a/OWNERS_ALIASES
+++ b/OWNERS_ALIASES
@@ -72,6 +72,7 @@ aliases:
     - skitt
   sig-network-leads:
     - aojea
+    - bowei
     - danwinship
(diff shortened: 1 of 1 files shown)


OWNERS_ALIASES
```diff
@@ -72,6 +72,7 @@ aliases:
     - skitt
   sig-network-leads:
     - aojea
+    - bowei
```
hakman, line 75: Probably you also shoud remove `thockin`.

danwinship: He stepped down as chair but is still a tech lead, so he's still a sig-network-lead.

PR #8660, opened 2025-10-18
drop storage migrator registry

upodroid:
Part of #8663

As discussed in https://kubernetes.slack.com/archives/C0EG7JC6T/p1760724978153999

/cc @enj @michaelasp


enj:
/lgtm
/hold
For ack from @kubernetes/sig-api-machinery-leads

upodroid:
If there are no objections, I'll cancel the hold in 2 days

PR #8661, opened 2025-10-18
drop unused sig-auth staging registry

upodroid:
Part of #8663

@ligitt @dims 

There are no active github repositories pushing jobs to this staging registry for several years.(last prod promo was in 2022)

https://github.com/search?q=repo%3Akubernetes%2Ftest-infra%20k8s-staging-sig-auth&type=code

I found this on Slack https://kubernetes.slack.com/archives/CHGFYJVAN/p1708975545162689


liggitt:
I don't really object to dropping this, but we should clean up references from things like https://github.com/kubernetes/kubernetes/blob/3eeb838aba6d5b311dcfb82f79bd55adfbd45f99/staging/src/k8s.io/pod-security-admission/webhook/manifests/50-deployment.yaml#L31 first

PR #8662, opened 2025-10-18
drop kubebuilder staging registry

upodroid:
Part of #8663

kubebuilder is not publishing any images from its staging registry so we are removing it from the image promoter. Previously published images will still be available.

kubebuilder's staging repository was created to copy over `kube-rbac-proxy` in #6690 to serve them from our registry and this function is complete now. The kubebuilder project no longer uses this component anymore.


upodroid:
bump

/cc @camilamacedo86

camilamacedo86:
Hi @upodroid,

Thanks for looking into this.

Just to confirm: we are removing the configuration, but the images will still be published there. Is that correct?

If so, then yes. LGTM

upodroid:
The current images will still be available

Issue #8663, opened 2025-10-18
ci-k8sio-image-promo is broken


upodroid:
This job is currently broken with errors like these:

```
time="07:46:58.968" level=fatal msg="run `cip run`: promote images: filtering edges: filtering promotion edges: reading registries: getting tag list: GET https://gcr.io/v2/token?scope=repository%3Ak8s-staging-kubebuilder%2Fkube-rbac-proxy%3Apull&service=gcr.io: unexpected status code 412 Precondition Failed: Container Registry is deprecated and shutting down, please use the auto migration tool to migrate to Artifact Registry (gcloud artifacts docker upgrade migrate --projects='k8s-staging-kubebuilder'). For more details see: https://cloud.google.com/artifact-registry/docs/transition/auto-migrate-gcr-ar" diff=14ms
```

A small number of staging registries weren't migrated to AR as they have not been active. I'm going to remove them from the image promoter to speed it up and resolve this issue. If these projects do need to promote new images, we can recreate their registries.

upodroid:
This has been fixed now

https://prow.k8s.io/job-history/gs/kubernetes-ci-logs/logs/ci-k8sio-image-promo

PR #8664, opened 2025-10-18
dop unused k8s-staging-windows-svc-proxy

upodroid:
Part of https://github.com/kubernetes/k8s.io/issues/8663

We never promoted an image from windows-service-proxy project, so I'm removing their staging registry.

/cc @ameukam @hakman


ameukam:
cc @marosset @zylxjtu

PR #800, opened 2020-04-23
Update audit as of 2020-04-23

spiffxp:
Supersedes https://github.com/kubernetes/k8s.io/pull/781

PR #8672, opened 2025-10-20
enable nested virt & set single process oom kill

upodroid:
More immediate version of https://github.com/kubernetes/k8s.io/pull/8600
Fixes: #800

Before this change is applied:
- Merge https://github.com/kubernetes/test-infra/pull/35741
- Create the new nodepool, taint the old ones and drop them

### DIFF
--- a/infra/gcp/terraform/k8s-infra-prow-build/main.tf
+++ b/infra/gcp/terraform/k8s-infra-prow-build/main.tf
@@ -60,14 +60,6 @@ resource "google_project_iam_member" "k8s_infra_prow_viewers" {
   member  = "group:k8s-infra-prow-viewers@kubernetes.io"
 }
 
-// Allow prow-deployer service account in k8s-infra-prow-build-trusted to deploy
-// to the cluster defined in here
-resource "google_project_iam_member" "prow_deployer_for_prow_build" {
-  project = module.project.project_id
-  role    = "roles/container.admin"
-  member  = "serviceAccount:prow-deployer@k8s-infra-prow-build-trusted.iam.gserviceaccount.com"
-}
-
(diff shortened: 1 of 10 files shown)


infra/gcp/terraform/k8s-infra-prow-build/main.tf
```diff
@@ -90,14 +90,15 @@ module "prow_build_nodepool_c4_highmem_8_localssd" {
...
-  disk_type       = "hyperdisk-balanced"
-  service_account = module.prow_build_cluster.cluster_node_sa.email
+  name                         = "pool6"
+  initial_count                = 1
+  min_count                    = 1
+  max_count                    = 80
+  machine_type                 = "c4-highmem-8-lssd"
+  disk_size_gb                 = 100
+  disk_type                    = "hyperdisk-balanced"
+  enable_nested_virtualization = true
```
BenTheElder, line 93: What's the use case? Nested virt can have poor performance, create noise for neighbors, and we need to make sure the VMs are not leaked ...

upodroid: Here is an example, trying to enable workflows like these in prow. 
https://github.com/kubernetes/kops/blob/master/.github/workflows/e2e.yml

https://github.com/kubernetes/kops/actions/runs/18663330367/job/53208805245

BenTheElder: ACK, in kind we also use actions for this, which is mostly ~fine.

upodroid:
/hold cancel

Successfully applied

upodroid:
need another lgtm to merge this

PR #8687, opened 2025-10-23
configure the workflow server-side instead of inrepo

upodroid:
The workflow changes I made in #8684 have no effect as atlantis by default doesn't allow repositories to specify custom workloads. We'll retain this secure behaviour and put the workflow definition server-side

### DIFF
--- a/.atlantis.yaml
+++ b/.atlantis.yaml
@@ -10,23 +10,3 @@ projects:
     branch: /main/
     dir: infra/aws/terraform/artifacts.k8s.io
     workflow: aws
-# For AWS, atlantis needs to assume a specific role in each account except the management account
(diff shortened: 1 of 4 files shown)


ameukam:
Can we ignore specific paths ?
I think it's possible with `autodiscover.ignore_paths` 
cc @GenPage

upodroid:
For AWS, we can disable autodiscover at `infra/aws/terraform` and manually add projects.

ameukam:
And also ignore in this PR:
`infra/fastly/`
`infra/ibmcloud/` 
?

upodroid:
I'll follow it up in a separate PR, fastly can work easily

PR #8694, opened 2025-10-24
restrict valid github repo name in nginx config

BenTheElder:
https://stackoverflow.com/a/59082561

### DIFF
--- a/apps/k8s-io/configmap-nginx.yaml
+++ b/apps/k8s-io/configmap-nginx.yaml
@@ -27,7 +27,7 @@ data:
           return 200 'ok';
         }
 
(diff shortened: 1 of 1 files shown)


thockin:
LGTM

ameukam:
/lgtm
/hold
Will rollout out on Monday.

ameukam:
Failure of postsubmit:
https://prow.k8s.io/view/gs/kubernetes-ci-logs/logs/post-k8sio-deploy-app-k8s-io/1982839624044646400

thockin:
Failure is because the old regex matched `""` and the new one requires 1 character, and that rule is used to do the "main site" redirect.  Changing it to `*` instead of `+` fixes tests and canary passes.

Is there any reason to demand at least 1 character?

PR #8708, opened 2025-10-29
Ingress NGINX: Promote other images.

Gacko:
/cc @strongjz @cpanato @tao12345666333


Gacko:
@strongjz @cpanato @tao12345666333: Anyone? 🙂 I'd like to get the releases out in the next few days.

PR #8723, opened 2025-11-03
IBMCLOUD: Update terraform configurations

Prajyot-Parab:
- Update build cluster configuration
- Add new PowerVS workspace in lon06 for conformance

### DIFF
--- a/infra/ibmcloud/terraform/k8s-power-build-cluster/variables.tf
+++ b/infra/ibmcloud/terraform/k8s-power-build-cluster/variables.tf
@@ -146,7 +146,7 @@ variable "bastion" {
   type = object({ memory = string, processors = string })
   default = {
     memory     = "8"
-    processors = "0.5"
+    processors = "0.25"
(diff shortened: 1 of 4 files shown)


infra/ibmcloud/terraform/k8s-power-build-cluster/variables.tf
```diff
@@ -164,9 +164,9 @@ variable "control_plane" {
 
 variable "compute" {
   default = {
-    count      = 2
```
mkumatag, line 167: why only 2?

Prajyot-Parab: changed to 5 (this line is removal of 2 below is change to 5)

mkumatag: but we had more than 2 worker nodes earlier as far as I remember.

PR #8732, opened 2025-11-05
bump cloud provider gcp images

upodroid:
@mattcary @sunnylovestiramisu @aojea @BenTheElder 

Please promote these images promptly in the future as kops and cluster api for GCP depend on these images.


upodroid:
btw, the generate.sh script needs to be removed, it has no purpose as you should be promoting every time you cut a tag and reading the digest from prow build job

BenTheElder:
I don't own this directory or the source repo(s) for the images.

I think @justinsb might know more about what's next for these.

PR #8761, opened 2025-11-18
configure kops-infra-ci to use atlantis

iftachk:


infra/aws/terraform/kops-infra-ci/locals.tf
```diff
@@ -16,10 +16,6 @@ limitations under the License.
 
 
 locals {
-  kops-infra-ci-name       = "kops-infra-ci"
-  kops-infra-ci-index      = index(data.aws_organizations_organization.current.accounts.*.name, local.kops-infra-ci-name)
```
GenPage, line 20: Are we sure these values aren't referenced anywhere else? I only see the removal of account id local.

iftachk: @GenPage 
I did confirm that `kops-infra-ci-index` and `kops-infra-ci-name` are not referenced anywhere else

PR #8762, opened 2025-11-18
Add WG Checkpoint Restore mailing lists

rst0git:
Reference: https://github.com/kubernetes/community/pull/8508

### DIFF
--- a/OWNERS_ALIASES
+++ b/OWNERS_ALIASES
@@ -176,6 +176,11 @@ aliases:
     - intUnderflow
     - kwilczynski
     - rthallisey
+  wg-checkpoint-restore-leads:
+    - adrianreber
+    - haircommander
+    - rst0git
+    - viktoriaas
   wg-policy-leads:
     - JimBugwadia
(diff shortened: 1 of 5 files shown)


haircommander:
do you have examples of having multiple emails per person for one of these lists? it doesn't seem typical to me and I'm not sure if there's some policy around it

/lgtm

rst0git:
@haircommander I've updated the patch to use one email per person.

BenTheElder:
> do you have examples of having multiple emails per person for one of these lists? it doesn't seem typical to me and I'm not sure if there's some policy around it

Yes this is fine, for example multiple steering members (including myself) have a personal email included in the steering lists for reasons like employer google docs restrictions

I'm not aware of any policy prohibiting it.

At some point we might want to drive leads type lists from the central source of truth in kubernetes/community sigs.yaml, but I don't know anyone planning to work on this and if/when that happens there will be other cases like this.

BenTheElder (commented):
/lgtm
/approve

PR #8776, opened 2025-11-20
Add inference-perf mailing list

jjk-g:
Adds k8s-infra-staging-inference-perf@kubernetes.io for inference-perf artifacts in registry.k8s.io.

https://github.com/kubernetes-sigs/inference-perf


jjk-g:
cc @SachinVarghese

PR #8778, opened 2025-11-20
Update resource group and secret references for s390x

sudharshanibm:
Added build resource group outputs and data source, updated IAM access group policy to use project resource group, and changed TTL for secret rotator. Updated secret keys in helm and prow manifests to reference updated keys


infra/ibmcloud/terraform/k8s-s390x-conformance/modules/secrets_manager/secret_manager.tf
```diff
@@ -64,7 +64,7 @@ resource "ibm_sm_iam_credentials_secret" "secret_rotator" {
   }
 
   //The time-to-live (TTL) or lease duration of generated secret 86400seconds = 24hrs
```
Prajyot-Parab:
```suggestion
  //The time-to-live (TTL) or lease duration of generated secret 91800seconds = 25.5hrs
```

PR #8782, opened 2025-11-20
Temporary change sig-contribex-leads to member in zoom-moderators

xmudrii:
The groups job is failing to reconcile the newly added `zoom-moderators` Google Group with the following error:

```
2025/11/19 16:27:34 unable to add sig-contribex-leads@kubernetes.io to "zoom-moderators@kubernetes.io" as MANAGER: googleapi: Error 400: Invalid Input: memberKey, invalid
```

I don't have much idea why is this the case, but some random thought that I have is that maybe you first have to add the user and then make it a manager. Or that groups can't be managers. But let's first try the former option and see if it helps.

### DIFF
--- a/groups/sig-contributor-experience/groups.yaml
+++ b/groups/sig-contributor-experience/groups.yaml
@@ -291,8 +291,7 @@ groups:
       ReconcileMembers: "true"
     owners:
       - contributors@kubernetes.io
-    managers:
-      - sig-contribex-leads@kubernetes.io
     members:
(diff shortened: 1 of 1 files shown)


chris-short:
@xmudrii I had the same thoughts.

/lgtm

BenTheElder:
passed :-)

https://prow.k8s.io/view/gs/kubernetes-ci-logs/logs/post-k8sio-groups/1991567863290269696

we could probably add a check for this

PR #8783, opened 2025-11-20
Change sig-contribex-leads back to manager in zoom-moderators

xmudrii:
Technically a revert of #8782 

Let's see if we can make it a manager now that it is part of the group. Otherwise, we'll revert it again.

### DIFF
--- a/groups/sig-contributor-experience/groups.yaml
+++ b/groups/sig-contributor-experience/groups.yaml
@@ -291,7 +291,8 @@ groups:
       ReconcileMembers: "true"
     owners:
       - contributors@kubernetes.io
+    managers:
+      - sig-contribex-leads@kubernetes.io
     members:
       - chadmcrowell@gmail.com
       - chris@chrisshort.net
-      - sig-contribex-leads@kubernetes.io


groups/sig-contributor-experience/groups.yaml
```diff
@@ -291,7 +291,8 @@ groups:
       ReconcileMembers: "true"
     owners:
       - contributors@kubernetes.io
+    managers:
+      - sig-contribex-leads@kubernetes.io
```
BenTheElder, line 295: should we continue listing it as part of the group?
did we compare to any similar groups?

xmudrii: From what I saw, managers are not listed as members and my understanding of the source code is that it's not needed. It'll reconcile it as a member with the manager role. I think it was only about that you can't add user directly/for the first time as a manager. But there's only one way to find that out. :)

BenTheElder: hmm ok, requiring two PRs for this feels bad, maybe our tool needs to internally handle the two steps.

otherwise maybe explicitly listing as both ...

BenTheElder:
/lgtm
let's try it

xmudrii:
This didn't work. It appears that other Google Group cannot be managers or owners.

PR #8784, opened 2025-11-20
Revert changing sig-contribex-leads back to manager in zoom-moderators

xmudrii:
This is a revert of #8783 because it appears that other Google Groups cannot be managers/owners (we get the same error as in #8782).


Priyankasaggu11929:
Thanks @xmudrii, @BenTheElder for taking care of it. :pray:

PR #8786, opened 2025-11-21
Add bowei@google.com as an owner in groups.yaml

aojea:
Added bowei@google.com as an owner to both sig-network leads and general discussion groups.


aojea:
/assign @bowei 
ping @kubernetes/sig-network-leads

PR #8807, opened 2025-11-26
CAPO v0.12.6

lentzi90:
```console
$ docker run --rm gcr.io/k8s-staging-capi-openstack/capi-openstack-controller@sha256:72266b71c4832c3a7ca4b880ae84a8efa08d36f8482485dd4c24132e083101e5 --version
v0.12.6
```


lentzi90:
Hmm we should get you two Kubernetes membership also. You will need it for these image promotions. Since you already have kubernetes-sigs, there isn't really any other requirement to get it anyway.

PR #8857, opened 2025-12-11
add DNS entry for node-readiness-controller.sigs.k8s.io

Priyankasaggu11929:
ref: https://github.com/kubernetes/org/issues/6015


xmudrii (approved):
/lgtm
/approve

PR #8858, opened 2025-12-11
Promote echo-basic to be used on docs and tests

rikatz:
As part of creating a document to demonstrate a quickstart on Gateway API, we have identified that the best image to use as a backend is our echo-basic image, used on conformance tests.

To avoid pointing people to the staging bucket/registry, I would like to promote this image to production, making it clear that the image IS NOT PRODUCTION READY (aka it can have old libraries, we will NOT keep bumping and promoting it for every go.mod change), it SHOULD NOT be deployed on a production cluster.


rikatz:
/hold

Until both Nick and Rob are on agreement with it

robscott:
Thanks @rikatz!

/lgtm
/approve

rikatz:
/hold cancel
Thanks Rob, I think we can move with it

PR #8865, opened 2025-12-13
Fix XSS in go-get handling

thockin:
Followup to #8694 - fix the tests.

Don't REQUIRE at least 1 character (which broke tests).

Retain trailing / on URLs if specified, but with simpler regex handling.

@ameukam @BenTheElder @cji

### DIFF
--- a/apps/k8s-io/configmap-nginx.yaml
+++ b/apps/k8s-io/configmap-nginx.yaml
@@ -27,7 +27,8 @@ data:
           return 200 'ok';
         }
 
-        location ~ ^/(?<repo>[A-Za-z0-9\-_\.]+)(?<subpath>/.*)?$ {
+        # NOTE: subpath is not currently used.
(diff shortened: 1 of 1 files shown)


BenTheElder:
> Don't REQUIRE at least 1 character (which broke tests).

I want to say requiring a character in this particular match was necessary but it's been a while, maybe we should be matching the no-characters case seperately? (and that case shouldn't allow query params etc?)

thockin:
I didn't want to go refactoring all the config, so I opted to make it as similar in function as possible.  I agree that it's a little weird to have the same location handling all of those cases...

thockin:
Do we have this on auto-push or does someone need to manyually run the deploy script?

BenTheElder:
Auto-push:
https://prow.k8s.io/view/gs/kubernetes-ci-logs/logs/post-k8sio-deploy-app-k8s-io/1999918180268511232

PR #8878, opened 2025-12-16
CAPO: Sync owners from repo

lentzi90:
For reference, the owners and aliases in CAPO:
- https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/main/OWNERS
- https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/main/OWNERS_ALIASES

### DIFF
--- a/groups/sig-cluster-lifecycle/groups.yaml
+++ b/groups/sig-cluster-lifecycle/groups.yaml
@@ -222,6 +222,8 @@ groups:
     members:
       - emacchi@redhat.com
       - lennart.jern@est.tech
+      - m.andre@redhat.com
+      - stephenfin@redhat.com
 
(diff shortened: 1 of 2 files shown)


lentzi90:
@EmilienM I see your handle is all lower case in CAPO. It was working fine so I guess this is case insensitive. Let me know if I should change it!

lentzi90:
Oh now I noticed that it says the file is invalid :open_mouth: 
Let's see if I can figure it out
/hold

lentzi90:
The issue is that @smoshiur1237 isn't in the kubernetes org yet. Fixed by https://github.com/kubernetes/org/issues/6034

lentzi90:
/hold cancel
We are good to go :tada:

smoshiur1237 (commented):
/lgtm

PR #8896, opened 2025-12-19
k8s-infra-staging-sig-security@: Remove nested groups, use direct group membership.

tabbysable:
It seems that google cloud does not support nested groups in the way we were trying to use them. Let's list out the group memberships directly, to see if this makes the project appear in the users' google cloud console.


tabbysable:
@mtardy

PR #8924, opened 2026-01-06
drop testgrid-proxy and testgrid-data dns record

upodroid:
/hold

We don't need the proxy anymore as we can access the bucket directly.

`testgrid-data.k8s.io` has been replaced with `testgrid-api.prow.k8s.io`


upodroid:
/hold cancel

The api can access the testgrid bucket directly

PR #8925, opened 2026-01-07
Promoting images for IGW v1.3.0 RC 1 cut

kfswain:
https://pantheon.corp.google.com/artifacts/docker/k8s-staging-images/us-central1/gateway-api-inference-extension?e=13802955&mods=logs_tg_prod


registry.k8s.io/images/k8s-staging-gateway-api-inference-extension/images.yaml
```diff
@@ -102,6 +105,7 @@
     "sha256:80e9821bc3b1aff4788121487d48b536720817c9e2186949d6875e35903dc28d": ["v1.2.0-rc.1"]
     "sha256:2400789434d31f0e4fc692e1b2b83d6cbc68da7fc450be3c19e11af8f05333f8": ["v1.2.0"]
     "sha256:ba95718c5ecc202eb9d7e40d6fde3fb93414451d311f9bff6500bdc9b8d8f86f": ["v1.2.1"]
+    "sha256:52b6617b69416dfe7b739511de9bec7850f03fb925b18d4f9cb096d9d0088459": ["v1.3.0-rc.1"]
 - name: lora-syncer
```
nirrozenbaum, line 109: @kfswain what about the lora-syncer?
as long as it's still part of IGW, can we update the images?

kfswain: This is part of the deprecation of the syncer we have talked about. I'll include in the patch notes that we will no longer be publishing new images, and eventually removing the syncer from the codebase. But keep the old images.

The syncer has been the same for 6+ months, so continuing to update the versions doesn't seem necessary, especially since we are deprecating

PR #8933, opened 2026-01-09
promote gvisor image for minikube

medyagh:
got the SHA from https://prow.k8s.io/view/gs/kubernetes-ci-logs/logs/post-minikube-gvisor-addon-image/2009434597758078976

closes https://github.com/kubernetes/minikube/issues/22404

### DIFF
--- a/registry.k8s.io/images/k8s-staging-minikube/images.yaml
+++ b/registry.k8s.io/images/k8s-staging-minikube/images.yaml
@@ -3,4 +3,4 @@
     "sha256:57f19c6a0b6a78f180f0ed65b7d548602839f2a5379f1febf3bd7055e729f629": ["v1", "v2", "v11"]
 - name: gvisor
   dmap:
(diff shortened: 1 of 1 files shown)


upodroid:
hey, please promote a correct semver tag and `latest` is not allowed

You won't be able to move tags later

/hold

medyagh:
> hey, please promote a correct semver tag and `latest` is not allowed
> 
> You won't be able to move tags later
> 
> /hold

thanks I removed the latest, btw I am curious if we later wanna change v1 to be a different iamge, we wont be able to do that ?

the bootcamp image tutorial might need to be changed in the future for v1 ,v2 to be different tutorials

upodroid:
That's right, tags can never be moved, which is why we ask maintainers to promote vX.Y.Z instead of vX or latest

medyagh:
> That's right, tags can never be moved, which is why we ask maintainers to promote vX.Y.Z instead of vX or latest

alright thanks @upodroid  moved changed it to semver format. can you please unhold

upodroid (approved):
/hold cancel

PR #8954, opened 2026-01-16
Fix ibmcloud-secret-rotator CLI to apply --region before instance

sudharshanibm:
This change fixes an issue where the ibmcloud-secret-rotator CronJob was attempting to contact the us-south Secrets Manager endpoint, even though the Secrets Manager instance is provisioned in eu-de, causing repeated DNS resolution failures.


upodroid (approved):

mkumatag (approved):
/lgtm

PR #8972, opened 2026-01-20
fix argocd-differ

upodroid:
So our argocd-diff-preview hasn't been working correctly for a few reasons:
- The bundled argocd in the docker image doesn't allow kustomize with helm so I had to fix that with a modified argocd install
- `Generated 0 applications from 9 ApplicationSets for branch: main (branch: main)` this happened because we generate applications from appsets that require the clusters to be added, which is why we apply the clusters.yaml and the differ works correctly.
- The IBM clusters pull their credentials from Secret Manager so I substituted them with dummy cluster credentials but same labels and name.

### DIFF
--- a/kubernetes/gke-utility/argocd/clusters-diff.yaml
+++ b/kubernetes/gke-utility/argocd/clusters-diff.yaml
@@ -1,0 +1,51 @@
+# This file is used by the argocd-diff github actions workflow, never add it kustomization.yaml
+# The IBM clusters are loaded via ExternalSecrets which the ephemeral argocd-diff-preview cluster
+# cannot access. Therefore we create the same clusters with server set to localhost:8080
+# and random username/password so it appears as a valid cluster and we can generate the appsets
+apiVersion: v1
+kind: Secret
+metadata:
+  name: ibm-ppc64le
+  labels:
+    name: ibm-ppc64le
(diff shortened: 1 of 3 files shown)


kubernetes/gke-utility/argocd/clusters-diff.yaml
```diff
@@ -0,0 +1,54 @@
...
+    clusterType: prow
+    environment: prod
+    prowNamespace: test-pods
+    cloud: ibm
+type: Opaque
+stringData:
+  name: ibm-ppc64le
+  server: http://localhost:8080
+  config: |
+    {
```
hakman:
```suggestion
    {
```

kubernetes/gke-utility/argocd/clusters-diff.yaml
```diff
@@ -0,0 +1,54 @@
...
+    clusterType: prow
+    environment: prod
+    prowNamespace: test-pods
+    cloud: ibm
+type: Opaque
+stringData:
+  name: ibm-s390x
+  server: http://localhost:8080
+  config: |
+    {
```
hakman:
```suggestion
    {
```

upodroid: fixed

PR #8992, opened 2026-01-22
Image promotion for gcp-compute-persistent-disk-csi-driver v1.23.x

kon-angelo:
Updates the generate script with the following changes:
- cloud-controller-manager first in list (to generate the least amount of changes per the current `images.yaml` format)
- automatically replace the `images.yaml` file (because it is unintuitive that it only prints the new images)

Also run the script and updated the images so that the new releases of the CSI provider like `v1.23.x` are available for consumption. 

- Removed manually some v999 images that appeared (see latest commit vs first) as they seemed out of place.


kon-angelo:
cc: @mattcary (since you reviewed previous similar PRs in the past) and I intend to apologise if you should not be invoked in future reviews 😅

ialidzhikov (approved):
/lgtm

registry.k8s.io/images/k8s-staging-cloud-provider-gcp/images.yaml
```diff
@@ -173,7 +181,7 @@
     "sha256:56674ef544860014ddbeac390d22e36dcf63a0dff93d949523031c316f681cf1": ["v1.9.13"]
     "sha256:1238f4e6e084b2de2f369bb227e86357393e02cfc4839d1a6234f4ecedf4d028": ["v1.9.14"]
     "sha256:e371e47ebc3873553d5de31b980b27955d0d37922b5da415a811b9288678e792": ["v1.9.15"]
-    "sha256:e023510d585ce95cb30a658995583513bad6e77528d3cc4b6b51c4b9c902f69d": ["v1.9.16"]
+    "sha256:37912f6535fe17f8c24d6217e1caf30505b9f5a6136cfceee214766368ede1ef": ["v1.9.16"]
```
msau42: I don't know why the hash changed but image tags are supposed to immutable so we may need to revert this line to the original.

kon-angelo: Ok, @msau42 I will revert these. 

Should something like a filter list be considered (in a separate PR) ? Just to complete the script automation so to say

kon-angelo: Done with latest commit. Please have a look again :)

msau42: thanks! yeah a filter list would be great

PR #9009, opened 2026-01-24
Add more redirect links support for rel.k8s.io

rytswd:
**Before**:

- ❌ `rel.k8s.io/v136` (no config in place)
- ❌ `rel.k8s.io/v1.36`
- ✅ `rel.k8s.io/v136/releasemtg`
- ❌ `rel.k8s.io/v1.36/releasemtg`

**After**:

- ✅ `rel.k8s.io/v136` (defaults to directory view on GitHub, which shows README)
- ✅ `rel.k8s.io/v1.36`
- ✅ `rel.k8s.io/v136/releasemtg`
- ✅ `rel.k8s.io/v1.36/releasemtg`

I believe the initial implementation of `v136` support matches how we used to generate links on bitly, but there is no strict requirements on that with NGINX config, and thus extending the support.

### DIFF
--- a/apps/k8s-io/README.md
+++ b/apps/k8s-io/README.md
@@ -78,7 +78,10 @@ NOTE: please see k8s.io/k8s.io/configmap-nginx.yaml for `server` definitions
 ### Version-specific Redirects
 
 Example:
+- https://rel.k8s.io/vXYY
+- https://rel.k8s.io/vX.YY
 - https://rel.k8s.io/vXYY/releasemtg
+- https://rel.k8s.io/vX.YY/releasemtg
 - https://rel.k8s.io/vXYY/retro
 - https://rel.k8s.io/v1XYY/contacts (Note: Access is restricted through Google Authorization)
 


rytswd:
CC @dipesh-rawat @Vyom-Yadav

apps/k8s-io/README.md
```diff
@@ -78,7 +78,10 @@ NOTE: please see k8s.io/k8s.io/configmap-nginx.yaml for `server` definitions
 ### Version-specific Redirects
 
 Example:
+- https://rel.k8s.io/vXYY
```
jenshu, line 81: do we still need to support vXYY for backwards compatibility?
i'm wondering whether we can just move it all to `vX.YY` format

rytswd: I agree we should just use v1.xx notation, but it won't harm to have this extra mapping (and we've got a lot this format used in various docs already)

dipesh-rawat: For now, I think it is fine to extend this while keeping the existing one. We can revisit it later and remove it if it is no longer needed. This could also be a good retro discussion item to gather everyone’s thoughts.

PR #9067, opened 2026-02-06
Update owners and staging repo ACL for dra-driver-cpu

pravk03:
Owners updated based on https://github.com/kubernetes-sigs/dra-driver-cpu/blob/main/OWNERS

### DIFF
--- a/registry.k8s.io/images/k8s-staging-dra-driver-cpu/OWNERS
+++ b/registry.k8s.io/images/k8s-staging-dra-driver-cpu/OWNERS
@@ -4,6 +4,9 @@ approvers:
 - johnbelamaric
 - klueska
 - pohly
+- ffromani
+- pravk03
+- sig-node-leads
 labels:
 - wg/device-management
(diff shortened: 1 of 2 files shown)


haircommander:
should this still be draft?

pravk03:
Thanks @haircommander. Un-drafted.

registry.k8s.io/images/k8s-staging-dra-driver-cpu/OWNERS
```diff
@@ -4,6 +4,8 @@ approvers:
 - johnbelamaric
 - klueska
 - pohly
```
SergeyKanzhelev, line 6: it is a good practice to add SIG Node leads as approvers to everything that has sig-node label. If you can do it in this PR, it will be great. Thanks!

pravk03: Thanks. Added.

PR #9113, opened 2026-02-13
update gcp cloud provider staging OWNERS

zhang-xuebin:
Add back previous owners in storage team so they can work on image promotion for CSI.


aojea:
Storage folks publish they're artifacts in this file, the name of the folder made us think it was only for the cloud controller manager and we remove them during last clean up.
Bring them back so they can manage the promotion of their artifacts as they always did

/lgtm 
/approve

PR #9126, opened 2026-02-16
Promoting IGW images for v1.3.1 rc1

kfswain:
https://pantheon.corp.google.com/artifacts/docker/k8s-staging-images/us-central1/gateway-api-inference-extension?e=13802955&invt=AcGaWQ&mods=logs_tg_prod


kfswain:
> should we add images also for standalone chart?

Not in a patch release IMO, I think the next minor release, yes.

PR #9132, opened 2026-02-18
Use k8s-admin user and disable root SSH

sudharshanibm:
- Disable root SSH access on newly created VPC–VSI instances
- Introduce and use a dedicated non-root user (k8s-admin) for all automation
- Grant required privileges via sudo (wheel group), instead of direct root login
- Remove any dependency on `ansible_become_user: root`


kishen-v:
/unhold 
LGTM

Rajalakshmi-Girish (commented):
/lgtm

PR #9163, opened 2026-02-27
promote gatewayapi echo image for v1.5.0

rikatz:
Promote echo-image used for Gateway API 1.5.0 conformance


rikatz:
/hold

let's just decide if we want this tag name, or simply v1.5.0

robscott:
Agreed in Slack to go with this name, thanks @rikatz!

/lgtm
/approve
/hold cancel

PR #9323, opened 2026-04-09
Update WG Checkpoint Restore mailing lists

rst0git:
As a follow up on the update in https://github.com/kubernetes/community/pull/8927, this pull request adds @andreyvelich to the owners for `wg-checkpoint-restore-leads` and `wg-checkpoint-restore`


rst0git:
@cblecker Thank you for your help! This is a follow-up pull request for https://github.com/kubernetes/community/pull/8927.
Would it be possible to review it as well?

PR #9336, opened 2026-04-14
add guicassolato@gmail.com as signet group owner

guicassolato:
- Adds guicassolato@gmail.com to both sig-network leads and general discussion group owners.
- Removes sutt@redhat.com from both sig-network leads and general discussion group owners.


guicassolato:
/assign @guicassolato
ping @kubernetes/sig-network-leads

PR #9356, opened 2026-04-16
feat: Promote dranet v1.2.0 from staging to prod

gauravkghildiyal:
/assign @BenTheElder


fmuyassarov:
Hi @BenTheElder 
We had a little hiccup in the chart releasing process.
Basically, there was a wrong path indicated to push the helm charts during the release process and it landed to `gcr.io/k8s-staging-networking/charts/dranet/dranet:1.2.0` instead of `gcr.io/k8s-staging-networking/charts/dranet:1.2.0`.

Is there any chance, we could do manual copy exceptionally so that we could promote charts too instead of waiting for the next release?
```
crane copy \
   gcr.io/k8s-staging-networking/charts/dranet/dranet:1.2.0 \
   gcr.io/k8s-staging-networking/charts/dranet:1.2.0
```

fmuyassarov:
> Hi @BenTheElder We had a little hiccup in the chart releasing process. Basically, there was a wrong path indicated to push the helm charts during the release process and it landed to `gcr.io/k8s-staging-networking/charts/dranet/dranet:1.2.0` instead of `gcr.io/k8s-staging-networking/charts/dranet:1.2.0`.
> 
> Is there any chance, we could do manual copy exceptionally so that we could promote charts too instead of waiting for the next release?
> 
> ```
> crane copy \
>    gcr.io/k8s-staging-networking/charts/dranet/dranet:1.2.0 \
>    gcr.io/k8s-staging-networking/charts/dranet:1.2.0
> ```

alternatively, there might be an option to re-trigger the Prow job to re-run the cloudbuild, but it is the same situation that we don't have access for that.

fmuyassarov:
I want to note that we probably shouldn’t let the charts block the promotion of the container image. If it’s possible to copy the image, we can open a follow-up PR to promote the charts. Otherwise, we will handle the chart promotion in the next release.

fmuyassarov:
If we’re unable to fix the path right now manually, we can still reference the staging (with incorrect path)version of the chart image. This means users who want to try DRANET with Helm aren’t blocked. The only caveat is that the chart resides in the staging repository, which has a 90-day retention policy, which should not be a problem really until the next release (I hope).

```sh
$ helm install dranet oci://gcr.io/k8s-staging-networking/charts/dranet/dranet --version 1.2.0

Pulled: gcr.io/k8s-staging-networking/charts/dranet/dranet:1.2.0
Digest: sha256:3ae7d18d2a25e699c90ea0a50b492790f427265ae3073cbb09da12abda1cb193
NAME: dranet
LAST DEPLOYED: Fri Apr 17 01:16:03 2026
NAMESPACE: default
STATUS: deployed
REVISION: 1
TEST SUITE: None
```

gauravkghildiyal:
> we can still reference the staging (with incorrect path)version of the chart image.

@fmuyassarov That sounds reasonable to me. 

/unhold

PR #9112, opened 2026-02-12
serve k8s-release-dev and k8s-staging-kops via Fastly

upodroid:
Commenting inline, but verified this change successfully.

# CI Markers
```
The MISS is expected as it's fetched from the bucket
 mahamed  Mac  ~  Desktop  Git  1  $  curl -H "Fastly-Debug: 1" https://dl.k8s.dev/ci/latest.txt -i
HTTP/2 200
content-type: text/plain
cache-control: private, max-age=0, no-transform
last-modified: Thu, 12 Feb 2026 15:37:54 GMT
etag: "0b56f27cbf612e34d9ee122990f8d7f4"
accept-ranges: bytes
server: UploadServer
surrogate-key: 6208FBA01F9D15A877FDA8525CE954669E3639BFA499258D8FA2EDA5CF058917
date: Thu, 12 Feb 2026 17:22:25 GMT
via: 1.1 varnish
fastly-debug-path: (D cache-mrs10558-MRS 1770916945)
fastly-debug-ttl: (M cache-mrs10558-MRS - - -)
fastly-debug-digest: 4facbaed5b7a2367bed8a043030db74703f8e846de68a39aa6ca9f69c41c2b2a
x-served-by: cache-mrs10558-MRS
x-cache: MISS
x-cache-hits: 0
access-control-allow-origin: *
alt-svc: h3=":443";ma=86400,h3-29=":443";ma=86400,h3-27=":443";ma=86400
content-length: 34

v1.36.0-alpha.1.139+150247a304f2cb
```

# Non txt files are cached
```
 mahamed  Mac  ~  Desktop  Git  1  $  curl -H "Fastly-Debug: 1" https://dl.k8s.dev/ci/v1.36.0-alpha.1.111+311071d3007c42/bin/linux/amd64/kubectl -i
HTTP/2 200
content-type: application/octet-stream
last-modified: Wed, 11 Feb 2026 19:26:58 GMT
etag: "68e2eb456ef5209104cc2143b0031fcf"
server: UploadServer
surrogate-key: 3FC097F271723711FA956B564A57599B7F36BD3A86315346C5F1BD028A8E4907
via: 1.1 varnish, 1.1 varnish
cache-control: public, max-age=86400
accept-ranges: bytes
age: 29589
date: Thu, 12 Feb 2026 17:23:00 GMT
[...]

PR #9390, opened 2026-04-25
fix broken redirector service

### LINKED ISSUE #9389: SSL Certificate Expired For Many k8s.io Subdomains (including pkgs.k8s.io, git.k8s.io, k8s.io, ...)
### Description

I hope i am at the right place to open this issue. Otherwise please let me know where to open it if it is known.

The SSL certificate for `pkgs.k8s.io` seems to have expired today (2026-04-25), causing `apt update` to fail when using the Kubernetes apt repository.

### Error Message

```
Err:8 https://pkgs.k8s.io/core:/stable:/v1.35/deb  InRelease
  Certificate verification failed: The certificate is NOT trusted. The certificate chain uses expired certificate. Could not handshake: Error in the certificate verification. [IP: 34.107.204.206 443]
```

### Steps to Reproduce

1. Have any Kubernetes apt repository (e.g. `pkgs.k8s.io/core:/stable:/v1.35/deb`) configured on a Debian/Ubuntu system
2. Run `sudo apt update`
3. Observe certificate verification failure

### Expected Behavior

`apt update` completes successfully with a valid SSL certificate.

### Actual Behavior

`apt update` fails due to an expired certificate on `pkgs.k8s.io`. I could see in my browser that the certificate recently expired:

Image

### Impact

From my understanding all users relying on `pkgs.k8s.io` for Kubernetes package installation or updates are affected and cannot pull package metadata until the certificate is renewed.

Thanks for looking into this!

Kind Regards
Vasili

upodroid:
In #9112, I forgot to remove dl.k8s.io and dl.kubernetes.io from the ManagedCertificate and it broke cert-renewal.
Fixes: https://github.com/kubernetes/k8s.io/issues/9389

This has been fixed, but we have a few changes we'll apply:
- Serve pkgs.k8s.io entirely from AWS Cloudfront, it will be much faster and eliminates an unnecessary hop.
- Migrate this service to use Gateway API + Managed Certificates with wildcards.
- Revisit the list of redirects and eliminate what we don't need. Also, we don't need every k8s.io domain be served at kubernetes.io. I'm not sure why we did, but we should cut it down.

### DIFF
--- a/apps/k8s-io/test.py
+++ b/apps/k8s-io/test.py
@@ -218,14 +218,10 @@ class RedirTest(HTTPTestCase):
     def test_yum(self):
         for base in ('yum.k8s.io', 'yum.kubernetes.io'):
             self.assert_temp_redirect(base, 'https://kubernetes.io/blog/2023/08/31/legacy-package-repository-deprecation/')
-            self.assert_temp_redirect(base + '/$id',
-                'https://packages.cloud.google.com/yum/$id', id=rand_num())
 
     def test_apt(self):
         for base in ('apt.k8s.io', 'apt.kubernetes.io'):
             self.assert_temp_redirect(base, 'https://kubernetes.io/blog/2023/08/31/legacy-package-repository-deprecation/')
-            self.assert_temp_redirect(base + '/$id',
-                'https://packages.cloud.google.com/apt/$id', id=rand_num())
 
     def test_packages(self):
         for base in ('packages.k8s.io', 'packages.kubernetes.io', 'pkgs.k8s.io', 'pkgs.kubernetes.io'):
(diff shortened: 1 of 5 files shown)


apps/k8s-io/test.py
```diff
@@ -218,14 +218,10 @@ def test_go(self):
     def test_yum(self):
         for base in ('yum.k8s.io', 'yum.kubernetes.io'):
             self.assert_temp_redirect(base, 'https://kubernetes.io/blog/2023/08/31/legacy-package-repository-deprecation/')
-            self.assert_temp_redirect(base + '/$id',
```
upodroid, line 221: Google yanked these repos in 2024; we should serve the 404s directly instead of sending them to Google.

dns/zone-configs/kubernetes.io._0_base.yaml
```diff
@@ -90,9 +90,6 @@ apt:
 yum:
   type: CNAME
   value: yum.k8s.io.
-dl:
```
upodroid, line 93: dl.kubernetes.io was never advertised to endusers anywhere

GenPage (approved):
/lgtm

PR #9347, opened 2026-04-16
add cluster-inventory-api staging registry

kahirokunn:
We would like to host images for [cluster-inventory-api](https://github.com/kubernetes-sigs/cluster-inventory-api) in registry.k8s.io, see [kubernetes-sigs/cluster-inventory-api#34](https://github.com/kubernetes-sigs/cluster-inventory-api/issues/34).

Specifically, we plan to publish the official credential plugins of cluster-inventory-api ([`secretreader`](https://github.com/kubernetes-sigs/cluster-inventory-api/tree/main/plugins/secretreader/cmd/plugin) and [`kubeconfig-secretreader`](https://github.com/kubernetes-sigs/cluster-inventory-api/tree/main/plugins/kubeconfig-secretreader/cmd/plugin)) as OCI artifacts so that users can consume them via [Image Volume](https://kubernetes.io/docs/tasks/configure-pod-container/image-volumes/).

/sig multicluster

Signed-off-by: kahirokunn <okinakahiro@gmail.com>

PR #9402, opened 2026-04-28
Fix sig-multicluster groups config

kahirokunn:
Fix the `groups/sig-multicluster/groups.yaml` config so it is actually reconciled by the Google Groups automation.

This PR also shortens the new staging access group from `k8s-infra-staging-cluster-inventory-api@kubernetes.io` to `k8s-infra-staging-cluster-inv-api@kubernetes.io`, because staging access groups must keep the project-name suffix within the 18 character limit.

### DIFF
--- a/groups/restrictions.yaml
+++ b/groups/restrictions.yaml
@@ -154,7 +154,7 @@ restrictions:
       - "^sig-instrumentation-leads@kubernetes.io$"
   - path: "sig-multicluster/groups.yaml"
     allowedGroups:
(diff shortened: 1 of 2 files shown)


kahirokunn:
@ameukam Could you please take a look when you have a chance?

This should fix the missing Google Group issue that caused the Atlantis apply failure in #9347. The previous groups PR used `teams` as the top-level key, but the reconciler only consumes `groups`, so the new group was not actually created.

PR #9403, opened 2026-04-28
kubernetes/gke-utility: expose kops-discovery via HTTPRoute

ameukam:
Add HTTPRoute for discovery.kops.k8s.io attached to the existing istio-ingressgateway in the utility cluster.


ameukam:
cc @hakman @upodroid @justinsb

PR #9417, opened 2026-04-29
Update kops-discovery probes to use HTTPS scheme

ameukam:
This replaces the tcpSocket probes with httpGet probes using the HTTPS scheme. fix the error saw in the discovery pod:

```bash
http: TLS handshake error from 10.250.227.1:49662: EOF
```


ameukam:
cc @hakman @justinsb

PR #9419, opened 2026-04-30
Change memory default request from 1Gi to 128Mi for ibm-ppc64le prow components

kishen-v:
The boskos-* pods consume far less memory than the requests, and the peak usage is seen to be at 40Mi, hence 128Mi will just be sufficient

### DIFF
--- a/kubernetes/ibm-ppc64le/prow/limit-range.yaml
+++ b/kubernetes/ibm-ppc64le/prow/limit-range.yaml
@@ -15,5 +15,5 @@ metadata:
 spec:
   limits:
   - defaultRequest:
-      memory: 1Gi
+      memory: 128Mi
     type: Container


kubernetes/ibm-ppc64le/prow/limit-range.yaml
```diff
@@ -15,5 +15,5 @@ metadata:
 spec:
   limits:
   - defaultRequest:
-      memory: 1Gi
+      memory: 64Mi
```
mkumatag: can we bump to 128Mi for a safer side..

mkumatag (changes requested):

kubernetes/ibm-ppc64le/prow/limit-range.yaml
kishen-v: Addressed, thanks @mkumatag

mkumatag (approved):
/lgtm

PR #9420, opened 2026-04-30
Increase api request and limits for testgrid-api deployment

### LINKED ISSUE #84: Increase resources allocated to the API deployment
The API deployment used by K8s is defined in https://github.com/GoogleCloudPlatform/testgrid/tree/main/cluster (specifically https://github.com/GoogleCloudPlatform/testgrid/blob/main/cluster/canary/api.yaml and https://github.com/GoogleCloudPlatform/testgrid/blob/main/cluster/prod/api.yaml).

The CPU/memory used for the API is pretty spiky (especially because it's basically only used for the TestGrid UI at the moment, so it only acts when there's a request made), but it's pretty low at the moment (0.5 CPU requested, 2GB memory requested). I think we should try increasing the overall resources used and see if that helps the response time for using the UI (since it can take like 1-3s to respond to a given request).

geewynn:
Increase resource request and limit to testgrid-api deployment.

Fixes: https://github.com/kubernetes-sigs/testgrid/issues/84

### DIFF
--- a/kubernetes/gke-prow/prow/testgrid.yaml
+++ b/kubernetes/gke-prow/prow/testgrid.yaml
@@ -29,7 +29,7 @@ spec:
         - name: http
           containerPort: 8080
         resources:
-          requests: 
+          requests:
             cpu: 1
             memory: 2Gi
           limits:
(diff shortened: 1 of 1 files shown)


upodroid (approved):
FYI @michelle192837

PR #9428, opened 2026-05-03
add karpenter config for arm64 nodepool

geewynn:
This PR does the following:
- Adds a new Karpenter NodePool to support arm64 (Graviton) build nodes in the EKS prow-build-cluster
- Enables Kyverno on this EKS cluster by adding a cloud: eks generator to the existing ApplicationSet, plus an empty values file matching the IBM/AKS pattern.
- Adds a taint to the arm64 NodePool (kubernetes.io/arch=arm64:NoSchedule) so amd64-default workloads don't accidentally land there.
- Adds a Kyverno ClusterPolicy that injects the matching toleration when a pod has nodeSelector: kubernetes.io/arch: arm64.

### DIFF
--- a/kubernetes/apps/kyverno.yaml
+++ b/kubernetes/apps/kyverno.yaml
@@ -16,6 +16,10 @@ spec:
               operator: In
               values:
                 - aks-prow-build
+    - clusters:
+        selector:
+          matchLabels:
+            cloud: eks
   template:
     metadata:
       name: "kyverno-{{ .name }}"
--- a/kubernetes/eks-prow-build/kube-system/nodepool.yaml
+++ b/kubernetes/eks-prow-build/kube-system/nodepool.yaml
@@ -45,3 +45,56 @@ spec:
     expireAfter: 720h
     budgets:
       - nodes: "20%" # Allowing rotating 20% of instances at any given time
+
+---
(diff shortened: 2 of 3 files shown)


xmudrii:
Is there any issue or request for doing this? What would be the use case for these nodes?

geewynn:
> Is there any issue or request for doing this? What would be the use case for these nodes?

yes, I got the request to do this from Mahamed cc @upodroid

upodroid:
Thank you for making this change.

Does EKS automatically taint arm64 nodes and mutate the pods when they have the arch selector?

For example, GKE does this:
- All arm64 nodes are tainted with `kubernetes.io/arch=arm64:NoSchedule`
- If you use the nodeSelector  `kubernetes.io/arch: arm64`, it will mutate the pod and add the taints.

If EKS doesn't do this natively, we can do it using kyverno.

geewynn:
no, EKS doesn't do this natively. Unlike GKE, there's no automatic taint on Graviton nodes and no admission controller that mutates pods to add tolerations based on the kubernetes.io/arch selector.

I'll push an update with the taint on the arm64 NodePool + a Kyverno policy to inject the toleration from the selector.

upodroid (approved):
LGTM

thank you

upodroid (approved):

PR #9459, opened 2026-05-08
Removed weave.works addresses

micahhausler:
Weaveworks is gone, and their domain `weave.works` now points to a Thai gambling site.


fabriziopandini (commented):
/lgtm
/approve

PR #9460, opened 2026-05-08
Removed weave.works addresses

micahhausler:
Weaveworks is gone, and their domain `weave.works` now points to a Thai gambling site.

### DIFF
--- a/groups/sig-network/groups.yaml
+++ b/groups/sig-network/groups.yaml
@@ -263,7 +263,6 @@ groups:
     members:
       - k8s-infra-release-editors@kubernetes.io
       - brucema19901024@gmail.com
-      - bryan@weave.works
       - cdc@redhat.com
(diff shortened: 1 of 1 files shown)


groups/sig-network/groups.yaml
danwinship, line 271: Are we even still using this list?

cdc@redhat.com should be casey.callendrello@isovalent.com now (@squeed)
dcbw@redhat.com and mcambria@redhat.com are also gone, with no Kubernetes-relevant forwarding address. None of the others except thockin@google.com are current members of the sig-network mailing list...

micahhausler: I'd like to just merge this as is if that's ok so we reduce untrusted domain access to K8s resources

danwinship:
ok

/lgtm
/approve

PR #9476, opened 2026-05-13
harden the shared staging image project

upodroid:

k8s-staging-images is our shared staging registry project. Our current approach uses a single service account, which allows any job to write to any registry in the project.

This new approach makes the following changes:
- every repository gets its own service account in Kubernetes and GCP.
- Cloud Build jobs now must specify a service account or it won't work at all. We'll activate this policy by enforcing this [constraint](https://docs.cloud.google.com/build/docs/cloud-build-service-account-updates).
- This dedicated service account can only write to its own registry and will fail if it tries to write to anywhere else
- `gcb-builder@k8s-infra-prow-build-trusted.iam.gserviceaccount.com` will eventually be deprecated from this project
- Sharing a single scratch bucket is not an issue because it gets hashed by GCB and used immediately.
- I'll amend the tests in the k/test-infra to require a kube service account to be set.

image

N/A

/cc @BenTheElder @kubernetes/sig-k8s-infra-leads


ameukam:
+1

hakman:
+1

ameukam:
we should probably update a doc to include the need of a k8s service account

upodroid:
I will after I test the change in registry.k8s.io GCB job first, I have a PR in k/t-i that needs to be merged

BenTheElder:
Thank you!

upodroid:
This is ready to merged, I had to tweak the registry.k8s.io sandbox as we do deploys via Cloud Build. 

The image is successful but the job is red because of the deployment changes this PR will solve.

https://prow.k8s.io/job-history/gs/kubernetes-ci-logs/logs/post-registry-push-images

infra/gcp/terraform/k8s-infra-oci-proxy/provider.tf
```diff
@@ -22,8 +22,7 @@ This file defines:
 
 terraform {
   backend "gcs" {
-    bucket = "k8s-infra-tf-oci-proxy"
-    prefix = "sanbox"
+    bucket = "k8s-infra-oci-proxy"
```
upodroid, line 25: I created this bucket within the staging project so the principal isn't granted permissions on the prod bucket.

ameukam:
atlantis unlock

PR #9498, opened 2026-05-17
monitoring: include DinD child containers in prow:job usage rules

dims:
For preset-dind-enabled jobs like ci-kubernetes-build the actual work (go compile via kube-cross) runs in containers dockerd spawns inside the pod, not in the `test` container. Switch the usage rules to the pod cgroup (container="") so those siblings are counted.

The resource_requests / resource_limits rules continue to filter on container="test" — those are per-container values.

Applies the same change to the AWS prow-build-cluster mirror.


upodroid (approved):
Thanks, I forgot about this

PR #9507, opened 2026-05-18
Promote Gateway API echoserver 1.6.0-dev.2

rikatz:

Promote Gateway API echo server v1.6.0-dev.2, that contains UDP server used on UDP conformance tests

- [X] I have verified the digest with [gcrane](https://github.com/google/go-containerregistry/blob/main/cmd/gcrane/README.md) and added it as a comment to the PR or in the body.

- [X] I'm not promoting a tag that resolves to latest or moving tags as these are not supported

- [X] registry.k8s.io is an immutable registry and I'll need to cut a new release if the digest is wrong.


rikatz:
/close
replaced by https://github.com/kubernetes/k8s.io/pull/9508/changes

Issue #8351, opened 2025-07-30
Rebuild the monitoring stack


ameukam:
The monitoring stack is currently spread between 2 cloud providers with standalone deployments which make maintainers navigate between different visualizations platorms (monitoring-(eks|gke).prow.k8s.io) for the metrics produced by the CI.

We historically used to have [monitoring.prow.k8s.io](https://monitoring.prow.k8s.io) for this but at the time we only had GCP. we now have extended the infra to multiple cloud providers.

We should attempt to centralize all those metrics from all our CI clusters.

ameukam:
IMHO, this should a subproject under this SIG.

/help
/sig k8s-infra
/lifecycle frozen
(I don't know how long it's will take to do this)

elieserr:
I would be happy to help with this if possible

elieserr:
I see we have in place the deployment for GKE in [here](https://github.com/kubernetes/k8s.io/tree/cb7060183a7b80f8e811f5cb7b22c2671757eb89/infra/gcp/terraform/k8s-infra-prow-build/prow-build/resources/monitoring) and for EKS config is mostly [here](https://github.com/kubernetes/k8s.io/tree/cb7060183a7b80f8e811f5cb7b22c2671757eb89/infra/aws/terraform/prow-build-cluster/resources/monitoring), I can spent a min looking at what we have and try to see what the next steps would be

In case you have an opinion on what next steps look like, please let me know

ameukam:
Have an understanding of what we have is a good start. I don't have a strong opinion what are the next steps but I'm happy to talk about it.

elieserr:
I think i have a better perspective of this, I was able to track what we have in https://monitoring-eks.prow.k8s.io/ and https://monitoring-gke.prow.k8s.io/ and where everything is created. 

What I am now missing is where the config for https://monitoring.prow.k8s.io/ is? i see its consuming metrics from `thanos-query-frontend.monitoring` but i wasn't able to get where this was set up, maybe the case that the main grafana is a clickops instance? or maybe im missing something

also please correct me if im wrong but all this monitoring changes has been applied manually right?

elieserr:
A couple of things come to mind thinking of how to move forward, we could

1. Use the existing grafana instance served in `monitoring.prow.k8s.io` and tune it to be able to consume the data from the EKS cluster, this will require

    - exposing the EKS prometheus instance outside of the cluster

[...]

PR #9524, opened 2026-05-25
create gcs buckets for mimir long term storage

geewynn:
This PR creates 3 GCS buckets for mimir long term storage. This is part of #8351.

cc @ameukam @upodroid 

### DIFF
--- a/infra/gcp/terraform/k8s-infra-prow/buckets.tf
+++ b/infra/gcp/terraform/k8s-infra-prow/buckets.tf
@@ -202,44 +202,11 @@ module "prow_security_bucket" {
 }
 
 
-module "mimir_blocks_bucket" {
+module "mimir_bucket" {
   source  = "terraform-google-modules/cloud-storage/google//modules/simple_bucket"
   version = "~> 11.1"
 
-  name       = "k8s-infra-prow-mimir-blocks"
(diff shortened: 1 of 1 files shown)


infra/gcp/terraform/k8s-infra-prow/buckets.tf
```diff
@@ -200,3 +200,53 @@ module "prow_security_bucket" {
...
+
+  iam_members = [
+    {
+      role   = "roles/storage.objectUser"
+      member = "principal://iam.googleapis.com/projects/16065310909/locations/global/workloadIdentityPools/k8s-infra-prow.svc.id.goog/subject/ns/mimir/sa/mimir"
+    },
+  ]
+}
+
+module "mimir_alertmanager_bucket" {
```
ameukam: why do we need this ?

geewynn: Mimir's alertmanager needs a backend storage for state persistence and HA. If an Alertmanager replica can't sync state from other replicas on startup, it falls back to what's persisted in object storage to avoid losing alert state entirely.

upodroid:
Can we use a single bucket with storage prefixes? Thanks

ameukam:
atlantis plan

PR #9525, opened 2026-05-25
groups: resolve relative restrictions-path from the config directory

immanuwell:
**What this PR does / why we need it**

`groups/config.yaml` says `restrictions-path` is relative to the config file. In practice we resolved it from the current working dir, so the reconciler could look for the wrong `restrictions.yaml` if you run it from somewhere else. Kinda sneaky.

This makes `restrictions-path` follow the same rule as `groups-path`, and adds a regression test. Same area as #2401.

**Special notes for your reviewer**

Repro before this patch:

1. Keep `config.yaml` in one dir, start from another cwd.
2. Set `groups-path: .` and `restrictions-path: restrictions.yaml`.
3. `groups-path` resolves from the config dir, but `restrictions-path` resolves from cwd, so it points at the wrong file.

`go test ./... -run TestConfigLoadPaths` fails before this patch, passes with it.

### DIFF
--- a/groups/reconcile.go
+++ b/groups/reconcile.go
@@ -397,6 +397,8 @@ func (c *Config) Load(configFilePath string, confirmChanges bool) error {
 
 	if c.RestrictionsPath == "" {
 		c.RestrictionsPath = filepath.Join(filepath.Dir(configFilePath), defaultRestrictionsFile)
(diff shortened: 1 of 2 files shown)


immanuwell:
@kaslin @MadhavJivrajani please take a look

PR #9533, opened 2026-05-26
tests: verify that images being promoted are multi-arch

justinsb:
We have an allowlist that allows well-known architecture suffixes (-amd64, -arm etc),
and a second allowlist of specific images.

We also allow helm charts to not be multiarch.

---

Created in collaboration with AI.

### DIFF
--- a/registry.k8s.io/manifests/multiarch_test.go
+++ b/registry.k8s.io/manifests/multiarch_test.go
@@ -1,0 +1,391 @@
+/*
+Copyright 2026 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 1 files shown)


registry.k8s.io/manifests/multiarch_test.go
```diff
@@ -0,0 +1,391 @@
...
+}
+
+// allowedImageSuffixes automatically allowlists an image that ends with the specified string e.g. kubectl-amd64
+var allowedImageSuffixes = []string{
+	"-amd64",
+	"-arm64",
+	"-ppc64le",
+	"-s390x",
+	"-arm",
+	"-386",
```
ameukam, line 60: I think we no longer support this ?

PR #9547, opened 2026-05-29
Move boskos pods to prefer being scheduled on infra nodes

kishen-v:
This PR allows the boskos pods to be deployed to the dedicated `infra` nodes, which host other critical deployments tied to the `ppc64le` build cluster. This frees up the compute nodes from hosting the build cluster infrastructure related pods.


mkumatag (approved):
/lgtm

PR #9559, opened 2026-06-02
Create multi-zone Boskos resources for parallel s390x jobs

sudharshanibm:
## Summary

This PR expands the boskos resource pool from one VPC to three zone-specific VPCs.

Each s390x  job can lease a separate boskos VPC resource, allowing multiple s390x conformance jobs to run concurrently

## Changes

- Create VPC and subnet resources in:
  - `eu-de-1`
  - `eu-de-2`
  - `eu-de-3`

## Expected Result

Up to three s390x conformance jobs can lease independent VPC resources and run in parallel, subject to available Boskos resources.

## Testing

- Verified the Terraform configuration changes.
- Confirmed that the Boskos ConfigMap contains all three zone-specific VPC resource names.

### DIFF
--- a/infra/ibmcloud/terraform/k8s-s390x-conformance/README.md
+++ b/infra/ibmcloud/terraform/k8s-s390x-conformance/README.md
@@ -57,3 +57,15 @@ Terraform will display a plan of the actions it will take, and you'll be prompte
 ```
 terraform output
 ```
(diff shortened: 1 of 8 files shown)


Prajyot-Parab (commented):
/lgtm

PR #9562, opened 2026-06-03
setup mimir on gke utility cluster

geewynn:
Part of #8351 

This PR:
 - Adds Mimir deployment using argocd application set referecing gke utility cluster. 
 - Adds Mimir helm values file based on mimir small helm value example. [mimir-small.yaml](https://github.com/grafana/mimir/blob/main/operations/helm/charts/mimir-distributed/small.yaml) 
 - Adds Gateway-API HTTPRoute for `mimir.prow.k8s.io` with path-based routing to distributor, query-frontend, alertmanager, and ruler.
 - Adds mimir.prow.k8s.io A and AAAA records.
 - Adds mimir.prow.k8s.io to the oauth2-proxy AuthorizationPolicy.
 - Adds htpasswd-based basic auth for mimir endpoints.

cc @ameukam @upodroid 

### DIFF
--- a/kubernetes/gke-utility/helm/mimir.yaml
+++ b/kubernetes/gke-utility/helm/mimir.yaml
@@ -1,0 +1,194 @@
+serviceAccount:
+  create: true
+  name: mimir
+
+nginx:
+  enabled: false
+
+minio:
+  enabled: false
+
+kafka:
+  enabled: false
+
+gateway:
+  enabled: true
+  replicas: 2
+  resources:
+    requests:
+      cpu: 100m
+      memory: 256Mi
+    limits:
+      memory: 512Mi
+  nginx:
+    basicAuth:
+      enabled: true
+      existingSecret: mimir-credentials
+
+mimir:
+  structuredConfig:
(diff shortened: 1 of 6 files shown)


kubernetes/gke-utility/helm/mimir.yaml
```diff
@@ -0,0 +1,194 @@
...
+  affinity:
+    podAntiAffinity:
+      requiredDuringSchedulingIgnoredDuringExecution:
+        - labelSelector:
+            matchExpressions:
+              - key: app.kubernetes.io/component
+                operator: In
+                values:
+                  - ingester
+          topologyKey: "kubernetes.io/hostname"
```
ameukam: Do we really need this ? I'm fine having those running in the same node.

geewynn: No, I initially added that for HA but mimir can run without it.

ameukam: I think we can start without it and patch it later if needed.

PR #9584, opened 2026-06-10
setup kps for prow consolidated monitoring

geewynn:
- This PR adds an empty Grafana + Prometheus install on gke-utility for consolidated prow monitoring. 
- Prometheus remote-writes to the in-cluster Mimir gateway. 
- Removed monitoring.prow.k8s.io from oauth2-proxy auth-policy for public access.
part of #8351 

cc @ameukam

### DIFF
--- a/kubernetes/gke-utility/helm/monitoring.yaml
+++ b/kubernetes/gke-utility/helm/monitoring.yaml
@@ -35,7 +35,7 @@ kube-state-metrics:
     - deployments=[app.kubernetes.io/name,app.kubernetes.io/component,app.kubernetes.io/instance]
 
 nodeExporter:
-  enabled: true
+  enabled: false
 
 prometheus:
   prometheusSpec:
--- a/kubernetes/gke-utility/helm/monitoring.yaml
+++ b/kubernetes/gke-utility/helm/monitoring.yaml
@@ -1,0 +1,85 @@
+grafana:
+  enabled: true
+  defaultDashboardsEnabled: false
+  admin:
+    existingSecret: grafana-credentials
+    userKey: username
+    passwordKey: password
+  service:
+    type: ClusterIP
+  sidecar:
+    dashboards:
+      enabled: true
+    datasources:
+      enabled: true
+  persistence:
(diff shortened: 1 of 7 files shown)


kubernetes/gke-utility/helm/monitoring.yaml
```diff
@@ -0,0 +1,85 @@
...
+  defaultDashboardsEnabled: false
+  admin:
+    existingSecret: grafana-credentials
+    userKey: username
+    passwordKey: password
+  service:
+    type: ClusterIP
+  sidecar:
+    dashboards:
+      enabled: true
```
ameukam, line 12: why do we need this ?

kubernetes/gke-utility/helm/monitoring.yaml
```diff
@@ -0,0 +1,85 @@
...
+            key: password
+        headers:
+          X-Scope-OrgID: prow
+
+coreDns:
+  enabled: false
+kubeControllerManager:
+  enabled: false
+kubeScheduler:
+  enabled: false
```
ameukam, line 85: are the other components (API Server, Etcd, etc...) disabled by default ?

kubernetes/gke-utility/helm/monitoring.yaml
geewynn: We need this so grafana can dynamically discover and load dashboard ConfigMaps from the cluster.

kubernetes/gke-utility/helm/monitoring.yaml
geewynn: No, they're not. 

Confirmed against the chart's values.yaml at v80.10.0, only kubeDns is disbaled by default. Everything else (kubeApiServer, kubeEtcd, kubeProxy, kubelet, coreDns, kubeControllerManager, kubeScheduler) is enabled by default.

ameukam: Let's disable all of them. I don't think we need them.

PR #9587, opened 2026-06-10
Update groups.yaml

rreimi:
Add reimi@google.com to scalability-oncall

**What this PR does / why we need it**: Onboarding reimi to KCP dev on call

**Special notes for your reviewer**: N/A

### DIFF
--- a/groups/sig-scalability/groups.yaml
+++ b/groups/sig-scalability/groups.yaml
@@ -107,6 +107,7 @@ groups:
       ReconcileMembers: "true"
     members:
       # scalability-oncall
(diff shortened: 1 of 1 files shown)


groups/sig-scalability/groups.yaml
```diff
@@ -107,6 +107,7 @@ groups:
       ReconcileMembers: "true"
     members:
       # scalability-oncall
+      - reimi@google.com
```
wojtek-t: Can you please put in alphabetic order?

I know it's not fully sorted, but lets at least not mess it up further :)

PR #9618, opened 2026-06-18
Add davanum@gmail.com to k8s-infra-prow-viewers

dims:
Adding myself to the `k8s-infra-prow-viewers` group for view access to resources used by prow.k8s.io (build clusters and e2e projects). This is a self-service, non-privileged read-only group per the comment in the file.

/sig testing


BenTheElder (approved):
/lgtm
/approve

PR #9625, opened 2026-06-20
configure prow to work with a github app and set sane resource limits

### LINKED ISSUE #6449: Personal Access Token to GitHub App Migration for prow.k8s.io
We really need to move from a PAT to a github app, other major instances (e.g. openshift) are using a github app, which should dramatically improve API quota (scales with the size of the org installed to) vs personal access token (fixed per-user quota).

Prow seems to be running into issues related to this again, we should really just migrate.

I'm not sure who will have the bandwidth, but we'll need to coordinate across SIGs, I'm hoping to start the conversation here.

cc @kubernetes/sig-testing-leads @kubernetes/sig-k8s-infra-leads @kubernetes/owners 

/sig testing

upodroid:
I temporarily disabled ArgoCD autosync and applied the changes manually, I'm observing the performance of the new resource limits before merging.

Fixes: https://github.com/kubernetes/org/issues/6449

All the components except the cherrypicker are using the GitHub app.

The secret in GCP SM looks like this:

```json
{
  "app_id": "REDACTED",
  "app_client_id": "REDACTED",
  "app_private_key": "REDACTED",
  "oauth_config": "client_id: REDACTED\nclient_secret: REDACTED\nredirect_url: https://prow.k8s.io/github-login/redirect\nfinal_redirect_url: https://prow.k8s.io/pr\n"
}
```

### DIFF
--- a/dns/zone-configs/k8s.io._0_base.yaml
+++ b/dns/zone-configs/k8s.io._0_base.yaml
@@ -244,13 +244,6 @@ prow:
   - type: AAAA
     ttl: 600 # this has needed to change in the past
     value: "2600:1901:0:b465::"
-hooks.prow:
-  - type: A
-    ttl: 600 # this has needed to change in the past
-    value: 34.128.150.99
-  - type: AAAA
-    ttl: 600 # this has needed to change in the past
-    value: "2600:1901:0:b465::"
 testgrid.prow:
   - type: A
     value: 34.128.150.99
--- a/hack/autobump-config.yaml
+++ b/hack/autobump-config.yaml
@@ -19,10 +19,6 @@ prefixes:
     repo: "https://github.com/kubernetes-sigs/boskos"
     summarise: false
     consistentImages: true
-  - name: "ghproxy"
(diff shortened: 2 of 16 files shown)


dns/zone-configs/k8s.io._0_base.yaml
```diff
@@ -244,13 +244,6 @@ prow:
   - type: AAAA
     ttl: 600 # this has needed to change in the past
     value: "2600:1901:0:b465::"
-hooks.prow:
```
upodroid, line 247: I moved the webhooks to the app now and cleaned up the webhooks from all the organizations:

- kubernetes
- kubernetes-sigs
- kubernetes-client
- kubernetes-csi
- etcd-io

In the future, adding the app to a repo/org will automatically capture the events via the webhook configured in the app.

also, this was a temp transition domain. The webhook endpoint is prow.k8s.io/hook

alvaroaleman:
While this is a great change and will help with token usage, I highly doubt this will fix https://github.com/kubernetes/test-infra/issues/37286 so maybe edit the PR body?

upodroid:
Done

PR #9670, opened 2026-07-01
fix karpenter and use newer bottlerocket features

upodroid:
karpenter was deployed incorrectly from #9503 

I also fixed a few things:
- bootstrap is no longer needed, bottlerocket automatically formats and makes the nvme disk available as an ephemeral disk
- upgraded to the latest Intel VMs with local ssds. We moved forward 3 generations and got faster disks.
- simplified the nodepool requirements
- Applied `single_process_oom_kill` kubelet config to report oom pods correctly. Applied previously to GKE at #8672


upodroid:
atlantis plan

upodroid:
/override "atlantis/plan"
/override "atlantis/plan: infra/aws/terraform/prow-build-cluster/default"

PR #9674, opened 2026-07-02
Enable image registry for Gateway API conformance images

rikatz:

Gateway API conformance images has a new home at k-sigs/gateway-api-conformance-images.

This PR adds the required manifests to bootstrap the new AR and set the right permissions on Gateway API group

### DIFF
--- a/groups/sig-network/groups.yaml
+++ b/groups/sig-network/groups.yaml
@@ -134,9 +134,8 @@ groups:
     members:
       - antonio.ojea.garcia@gmail.com
       - bowei@google.com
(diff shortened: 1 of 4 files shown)


hakman (approved):

rikatz:
it seems it gets on some infinite loop trying to pull manifests and fails

hakman:
@rikatz I think https://github.com/kubernetes/test-infra/pull/37383 should fix the presubmit.

rikatz:
nice, thank you!

PR #9703, opened 2026-07-10
add docs for rotating cluster credentials

upodroid:
I added docs to show how we rotate credentials for IBM build clusters that don't use Workload Identity. I also added minimal readmes to every cluster + argocd status badges.


xmudrii (approved):
/lgtm
/approve

PR #9708, opened 2026-07-11
Replace wdc06 DC with eu-de-1 region in boskos on ibm ppc64le

Rajalakshmi-Girish:
Removing old workspaces in wdc04 zone and adding eu-de-1 back.


mkumatag (approved):
/lgtm

PR #9712, opened 2026-07-13
Promote coredns to v1.14.6

yashsingh74:
Coredns v1.14.6 release is available - https://github.com/coredns/coredns/releases/tag/v1.14.6

```
❯ gcrane digest coredns/coredns:1.14.6
sha256:900f9c109f7a33545d3c811516e8376df9019147b750f5ce3e254468769176ea
```


yashsingh74:
cc: @dims @stp-ip @pacoxu

pacoxu:
/lgtm
BTW, should we promote the v1.14.5 image as well?

stp-ip:
Pushed v1.14.6 into staging and confirmed the correct digest.

V1.14.5 has an issue on the coredns side as it doesn't have a tagged multi arch image and therefore currently we can't promote.

PR #9728, opened 2026-07-16
Update images for csi hostpath driver and snapshot metadata

xing-yang:


registry.k8s.io/images/k8s-staging-sig-storage/images.yaml
```diff
@@ -369,6 +371,7 @@
     "sha256:054a2d495119b1380c5cdb6f1ed03580ecb766a262e2b33772951d98ab72872b": [ "v1.20.0-windows-hp" ]
     "sha256:e5b148629c4b22554b60bc428195c15c992ee3fc24a45aff9873edd0a401a9e8": [ "v1.20.1" ]
     "sha256:d66147764fd5936aad08581ca8f7c0e7054cf550971cc956f39d61b11c79f0ca": [ "v1.20.1-windows-hp" ]
+    "sha256:aaf7eb300a7c1a8895e7b55302964c390e25e1002307d657f848d9a8e5995189": [ "v1.20.2" ]
```
xing-yang, line 374: @andyzhangx should I update this?

xing-yang: I don't see an image for `v1.20.2-windows-hp`.  Should they always be promoted together?

andyzhangx: `v1.20.2-windows-hp` is missing since there is image build issue is this smb csi driver, that's the reason why I cut a new v1.20.3 release quickly.

andyzhangx: L374 change is ok

andyzhangx (approved):
/lgtm

Issue #8904, opened 2025-12-23
Rebuild our Prow Build Clusters to support dualstack networking


upodroid:
Our Prow build clusters don't support IPv6, and IPv6 has been GA in Kubernetes for several years:

The GKE cluster running the prow control plane has IPv6 enabled already.

Clusters in scope:

```
k8s-infra-prow-build
k8s-infra-prow-build-trusted
eks-prow-build-cluster
k8s-infra-aks-admin (This cluster never worked)
```

Runbook:
1. Fix all the following outstanding issues:
    1. https://github.com/kubernetes/k8s.io/issues/3123
    1. https://github.com/kubernetes/k8s.io/issues/6442
2. Provision new Infrastructure
    1. New dual-stack VPCs for gke prow build clusters
    1. Modify the existing AWS VPC to support IPv6
    1. Deploy new GKE clusters like this: https://github.com/kubernetes/k8s.io/blob/main/infra/gcp/terraform/k8s-infra-prow/gke.tf
    1. Delete the old clusters
    1. Deploy the new clusters
3. Post cluster provisioning
     1. Update the endpoints of the clusters in ArgoCD, https://github.com/kubernetes/k8s.io/blob/main/kubernetes/gke-utility/argocd/clusters.yaml
     1. Update the endpoints of the clusters in Prow by modifying the kubeconfig secrets in GCP Secret Manager

We expect this operation to take 3 hours and ideally be executed on the weekend before the start of v1.36 release cycle.

PR #9737, opened 2026-07-20
fix(infra): add IPv6 egress firewall rule for Prow VPC network

marqc:
GCP VPC networks configured with external IPv6 subnets do not automatically permit outbound IPv6 egress traffic. Without an explicit egress firewall rule targeting ::/0, outbound IPv6 connections from Prow nodes and CI workloads time out.
This change adds an outbound firewall rule allowing all IPv6 egress traffic (destination ::/0) in the Prow VPC.
Ref: https://github.com/kubernetes/kubernetes/issues/140489

Adds an explicit outbound IPv6 firewall rule (`prow-allow-ipv6-egress` for destination `::/0`) to the Prow VPC network in `infra/gcp/terraform/k8s-infra-prow/vpc.tf`. GCP VPC networks configured with external IPv6 subnets (`ipv6_access_type = "EXTERNAL"`) do not automatically permit outbound IPv6 egress traffic by default, causing outbound IPv6 connections from Prow nodes and CI job workloads to time out.

Tested and verified in a sandbox GCP project with fresh VPC network. Before adding the egress firewall rule, outbound IPv6 requests (`curl -6 https://ipv6.google.com`) timed out (`CURL_EXIT_CODE: 28`). After applying the firewall rule, outbound IPv6 HTTP/2 connections succeeded (`CURL_EXIT_CODE: 0`, `HTTP 200 OK`).

### DIFF
--- a/infra/gcp/terraform/k8s-infra-prow/vpc.tf
+++ b/infra/gcp/terraform/k8s-infra-prow/vpc.tf
@@ -67,3 +67,18 @@ module "nat" {
   router        = "prow-nat"
   name          = "prow-nat"
 }
+
+resource "google_compute_firewall" "allow_ipv6_egress" {
+  name        = "prow-allow-ipv6-egress"
+  project     = module.project.project_id
+  network     = module.vpc.network_name
+  direction   = "EGRESS"
+  priority    = 1000
+
+  allow {
+    protocol = "all"
+  }
+
+  destination_ranges = ["::/0"]
+}
+


aojea:
/assign @ameukam @upodroid  @BenTheElder 

the main problem with IPv6 is the security angle, if this is just egress traffic it looks safe ... the moment it allows coming in we enter in a security complexity when things are globally routable

as I commented in the referenced issue I have also my doubts about the useful of that test, kubernetes should not make any opinion about internet traffic, just north-south in general

upodroid:
This wouldn't even work, the cluster doesn't have IPv6 enabled.

We would need to rebuild the cluster and decided against that in #8904

You'll need to use the AKS build cluster if you need IPv6.

PR #9758, opened 2026-07-22
rightsize the eks prometheus instance

upodroid:
This should fix the broken eks broken instance


kubernetes/gke-prow/prow/hook.yaml
```diff
@@ -55,7 +55,7 @@ spec:
               memory: 3000Mi
             limits:
               cpu: 500m
-              memory: 1000Mi
+              memory: 6000Mi
```
upodroid, line 58: Fixing a typo from #9748

hakman (approved):

PR #9740, opened 2026-07-21
Fix: Slack inviter link

tico88612:
**What this PR does / why we need it**:

communityinviter.com has been deprecated; they have moved to inviter.co

**Special notes for your reviewer**:

> Community Inviter has been deprecated. Please continue with the latest experience at inviter.co.

PR #9766, opened 2026-07-23
slack: modify to a 302 Found redirect

samuelkarp:
301 Moved Permanently redirects are aggressively cached by browsers, which can make changing the target of the redirect more difficult if it ever experiences problems (such as the change from communityinviter.com to inviter.co done in https://github.com/kubernetes/k8s.io/pull/9740 and related discussion in https://kubernetes.slack.com/archives/C4M06S5HS/p1739578731492559). 302 Found redirects are followed every time.

LLM assistant disclosure: [Antigravity CLI](https://antigravity.google/product/antigravity-cli) was used to validate the changes in a GKE cluster.

N/A, no image promotion in this PR.

### DIFF
--- a/apps/k8s-io/configmap-nginx.yaml
+++ b/apps/k8s-io/configmap-nginx.yaml
@@ -361,7 +361,7 @@ data:
         listen 80;
 
         location / {
-          rewrite ^/.*$ https://inviter.co/kubernetes permanent;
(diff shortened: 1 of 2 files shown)


BenTheElder (commented):
/lgtm
/approve

PR #9768, opened 2026-07-25
Promote VPA 1.7.1

adrianmoisey:

Promotes VPA 1.7.1 from staging

I'm unsure what this is or what I should do. I added these using `./add-version.sh`, which I assume is sufficient

I get a 404 here


omerap12 (commented):
/lgtm

PR #9789, opened 2026-07-29
start using cdn.registry.k8s.io in registry-sandbox.registry.k8s.io

upodroid:
cdn.registry.k8s.io was provisioned in #8819, and I'm rolling it out on our sandbox instance.

The IAM change is to fix the broken prowjob https://prow.k8s.io/view/gs/kubernetes-ci-logs/logs/post-registry-push-images/2082519543736963072


upodroid:
bump

PR #9796, opened 2026-07-30
Promote provider-aws v1.32.4 / v1.33.1 / v1.36.1 release artifacts

hakman:
Image promotion for cloud-controller-manager v1.32.4, v1.33.1 and v1.36.1, plus file promotion for the v1.36.1 ecr-credential-provider binaries (resurrects #9735, checksums verified against the staged .sha256 files).

These releases were built and staged but never promoted. v1.36.1 is currently the only usable release of the 1.36 branch, since the v1.36.0 image build failed and was never staged (kubernetes/cloud-provider-aws#1450). v1.33.3, v1.34.1 and v1.35.1 are not included because their tagged builds failed and nothing exists in staging for them.

Note: file promotion manifests for v1.32.0 through v1.35.0 are merged, but nothing has landed in gs://k8s-artifacts-prod/binaries/cloud-provider-aws/ since v1.31.7, so the file promoter may need separate attention.

/cc @kmala @rifelpet


hakman:
Thanks @kmala! 🙂

PR #9801, opened 2026-08-02
fix: use Recreate strategy for Grafana on gke-utility

ameukam:
Grafana uses a ReadWriteOnce persistent volume. RollingUpdate can schedule the replacement pod on another node while the existing pod still holds the volume, leaving the rollout stuck until it exceeds its progress deadline. Recreate releases the volume before starting the replacement pod.


ameukam:
cc @geewynn @hakman

PR #9802, opened 2026-08-02
fix: avoid Grafana RWO rolling update deadlock

ameukam:
Use a zero-surge rolling update so the existing Grafana pod releases its ReadWriteOnce volume before Kubernetes creates the replacement. Keeping the strategy type as RollingUpdate also allows Argo server-side apply to update the existing Deployment without Kubernetes rejecting the retained rollingUpdate field.


ameukam:
cc @geewynn @hakman @upodroid

hakman (approved):

PR #9803, opened 2026-08-02
fix: eliminate monitoring Argo drift

ameukam:
Remove server-side diff from the monitoring ApplicationSet because it treats legacy field-manager ownership as live drift after the Argo upgrade.

Preserve Grafana's chart-generated admin password and checksum during syncs, and explicitly declare ExternalSecret and HTTPRoute defaults that Argo cannot infer from CRD schemas. This lets the monitoring applications converge without rotating credentials or restarting Grafana.

### DIFF
--- a/kubernetes/apps/monitoring.yaml
+++ b/kubernetes/apps/monitoring.yaml
@@ -18,8 +18,6 @@ spec:
   template:
     metadata:
       name: 'monitoring-{{ .name }}'
-      annotations:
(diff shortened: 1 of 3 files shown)


ameukam:
cc @geewynn @hakman @upodroid

ameukam:
> /hold, there is a better way to do this

@upodroid do you mind share a better approach to this ?

PR #9830, opened 2026-08-12
Promote gateway-api-inference-extension v1.6.0 lwepp image

nicolexin:
Promote gateway-api-inference-extension images for v1.6.0.

Source tag: v1.6.0
Release tracker: https://github.com/kubernetes-sigs/gateway-api-inference-extension/issues/3007


nirrozenbaum (approved):
/lgtm
/approve

PR #9843, opened 2026-08-18
deploy the registry.k8s.io CDN in Fastly

upodroid:
Serving registry.k8s.io blobs via Fastly will free up about 1M USD a year that we can use towards general compute.

I have been using CloudFront to shield the traffic for the past few days, so we have an idea of how much bandwidth we serve to non GCP/AWS users. We are looking at about 27 TB on a weekend day and 40 TB during a working day, which gives about 1.1PB of bandwidth per month. This is around 11% of our monthly bandwidth budget, and we currently use about 50% of our bandwidth budget.


upodroid:
atlantis plan

hakman (approved):

upodroid:
/override "atlantis/plan"
/override "atlantis/plan: infra/fastly/terraform/cdn/default"
/hold cancel

PR #9853, opened 2026-08-20
promote ExternalDNS v0.22.0

mloiseleur:

This PR promotes ExternalDNS v0.22.0 images, coming from this tag:

```
gcr.io/k8s-staging-external-dns/external-dns:v20260820-v0.22.0
```


Raffo (approved):
/approve
/lgtm

PR #9865, opened 2026-08-24
switch to using fastly cdn for registy.k8s.io

upodroid:
The service is deployed, and I manually created the acme record, so the cert has been created and it's ready to serve traffic.
/hold


BenTheElder:
might need to walk this back for a bit https://github.com/kubernetes/registry.k8s.io/issues/333

also noting that we are scheduled to release kubernetes in ~2d

PR #9918, opened 2026-09-10
Image promotion for sp-operator seccomp base profiles and test fixtures

saschagrunert:
Image promotion for sp-operator seccomp profiles, generated with `kpromo` and extended by hand for the base profiles.

Two kinds of artifact, both in the KEP-6061 runtime format, a single layer holding an OCI runtime-spec seccomp profile identified by `application/vnd.cncf.seccomp-profile.config.v1+json`:

- `base/runc` and `base/crun` are the base profiles the operator records against the container runtimes it supports. Container runtimes consume them for `type: OCI` seccomp profiles and the operator reads them for `oci://` base profiles.
- `seccomp-test-profiles` holds fixtures for the Kubernetes `e2e_node` tests: a profile that constrains an observable syscall, one more permissive than any node baseline, one a runtime has to reject, and one above the default size limit.

Only versioned tags are promoted. The staging registry also carries a `latest` tag for the base profiles, which is deliberately left out, since tags here cannot be repointed and a promoted `latest` would be frozen at whatever it pointed to first.

/cc @kubernetes/release-engineering

### DIFF
--- a/registry.k8s.io/images/k8s-staging-sp-operator/images.yaml
+++ b/registry.k8s.io/images/k8s-staging-sp-operator/images.yaml
@@ -1,3 +1,11 @@
+- name: seccomp-test-profiles
+  dmap:
+    "sha256:15a0050ff613b742d960e7bf4ed3df77c91adee1e1472d5aea57718884ad8261": ["baseprofile-crun"]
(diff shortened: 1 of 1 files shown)


saschagrunert:
@kubernetes/release-engineering @pjbgf PTAL

puerco (approved):
:shipit:

PR #9945, opened 2026-09-17
Let the sp-operator image push build run as gcb-image-builder

saschagrunert:
Let the security-profiles-operator image push build run as `gcb-image-builder@k8s-staging-sp-operator`, so it can sign images and attach provenance and SBOM attestations keylessly (kubernetes-sigs/security-profiles-operator#3442).

Every staging project already gets `gcb-image-builder` with token creator on itself from `ensure_staging_gcb`. This adds a special case for `k8s-staging-sp-operator` that:

- grants the account `roles/cloudbuild.builds.builder`, the same permissions as the default Cloud Build service account (logs bucket, image push)
- lets `gcb-builder@k8s-infra-prow-build-trusted` act as the account, so prow can submit builds with `--service-account`

Needs `infra/gcp/bash/ensure-staging-storage.sh sp-operator` to be run after merge. If `roles/cloudbuild.builds.builder` is broader than wanted, Artifact Registry write plus object access on `gs://k8s-staging-sp-operator-gcb` would be the narrower alternative. The matching test-infra job change follows once this is applied.

/area infra
/sig k8s-infra security


saschagrunert:
PTAL @kubernetes/sig-k8s-infra

upodroid:
/hold

I have an alternate, more secure approach.

saschagrunert:
Closing in favor of #9951, which moves the security-profiles-operator staging builds to `k8s-staging-images` with a dedicated service account.

Issue #9486, opened 2026-05-14
Start using dedicated Google Service Account for every staging registry.


upodroid:
For all projects using `us-central1-docker.pkg/k8s-staging-images/*` staging registries listed [here](https://github.com/kubernetes/k8s.io/blob/main/infra/gcp/terraform/k8s-staging-images/registries.tf#L19), you'll need to start using user-supplied service account that is dedicated to your GitHub repository.

Docs: https://docs.cloud.google.com/build/docs/securing-builds/configure-user-specified-service-accounts

in cloudbuild.yaml, you'll need to add this line:
```yaml
serviceAccount: projects/k8s-staging-images/serviceAccounts/[REGISTRY_NAME]-sa@k8s-staging-images.iam.gserviceaccount.com
```

/priority important-soon

PR #3854, opened 2022-06-16
Add staging service accounts to `roles/iam.serviceAccountTokenCreator`

saschagrunert:
The idea is to use the GCB service account to be able to sign images in
staging and promote them later into production. This way we can use
cloudbuild.yaml to specify something like:

```
env:
  GOOGLE_SERVICE_ACCOUNT_NAME=448637284062@cloudbuild.gserviceaccount.com
```

We have to obtain the project ID since it is created randomly for each
staging project.

Refers to https://github.com/kubernetes-sigs/security-profiles-operator/pull/984
Should fix the permission issue mentioned in https://github.com/kubernetes-sigs/security-profiles-operator/pull/984#issuecomment-1154820860

/assign @puerco @ameukam

PR #9951, opened 2026-09-17
create sp-operator staging registry

upodroid:
xref #9486 

We are hardening the staging registry GCP project by requiring new registries to use a dedicated service account.

Push the images to `us-central1-docker.pkg/k8s-staging-images/sp-operator` and use this GCP service account `sp-operator-sa@k8s-staging-images.gserviceaccount.com`

/cc @saschagrunert


upodroid:
atlantis apply

saschagrunert:
Thanks, moving to `k8s-staging-images` works for us. One question before I adapt the SPO build:

The staging build will sign images, profiles and the Helm chart as Sigstore bundles (kubernetes-sigs/security-profiles-operator#3448), using cosign's `GOOGLE_SERVICE_ACCOUNT_NAME` impersonation to get an identity token. That needs `roles/iam.serviceAccountTokenCreator` for `sp-operator-sa` on itself, which the old staging accounts got in #3854, but `registries.tf` doesn't grant it. Could we add that binding, or is there another way you intend builds in `k8s-staging-images` to get an identity token?

The 90 day cleanup is fine for us. The promoted SPO catalog currently references the staging bundle image, I'll change it to reference the promoted bundle before switching, so nothing production facing depends on staging.

PR #9962, opened 2026-09-18
Update cloud-provider-gcp

mattcary:

Update cloud provider gcp images, including recent PD CSI driver and  cloud controller manager.

- [X ] I have used the generate.sh script to create the update

- [ X] I'm promoting a multi-arch image that matches all the [supported](https://github.com/kubernetes/sig-release/tree/master/release-engineering/platforms) platforms of Kubernetes.

- [ X] I'm not promoting a tag that resolves to latest or moving tags as these are not supported

- [ X] registry.k8s.io is an immutable registry and I'll need to cut a new release if the digest is wrong.


mattcary:
Note there is some weirdness with the cloud controller manager, it looks like some hashes might get changed by the generate script. Not sure what's happening there, if the cloud controller manager maintainers could comment it would be great.

mattcary:
Ok, the generate script is broken for all the images, not just cloud controller manager.

It seems that some of the older tags in gcloud container don't match what were promoted. The generate script doesn't try to update the images list, it just recreates the whole thing :-/

I just updated the newer pd csi images we want to promote.

PR #9965, opened 2026-09-20
Add kahirokunn as sig UI lead

kahirokunn:
Related to https://github.com/kubernetes/community/issues/9164


groups/sig-ui/groups.yaml
shu-mutou, line 45: Is it not necessary to become a member?

shu-mutou: if you don't need to become a member of k8s-infra-staging-headlamp, LGTM

kahirokunn: Thanks for your review 🙏 
I was thinking of applying for the permissions additionally if a situation arises where I need them, but what do you think?

kahirokunn: I felt I should have the permission, so I added it. Thank you!

shu-mutou:
LGTM

PR #9967, opened 2026-09-21
Setup ibm-ppc64le monitoring to mimir

geewynn:

setup ibm-ppc64le monitoring to mimir

/cc @upodroid


kubernetes/ibm-ppc64le/kube-system/powervs-tier1-sc.yaml
```diff
@@ -0,0 +1,11 @@
+apiVersion: storage.k8s.io/v1
```
upodroid, line 1: Isn't there a default storage class in the cluster?

geewynn: Not as far as I can tell, I did not see any one in the powervs csi driver repo.

